PR #24, opened 2025-01-15
1043: Record event when failing an ECT

ghost:
ensure event is send after failing an ECT induction
also add additional service assertions for pass induction

ticket https://github.com/DFE-Digital/register-ects-project-board/issues/1043


peteryates (approved):
Nice!

PR #26, opened 2025-01-15
Add TRS refresh timestamp to teachers table

peteryates:
It'll be useful to know when the teacher record was last synced with TRS and we might even use it to ensure we don't leave any records too long between sync.


tonyheadford (approved):
Nice idea

PR #27, opened 2025-01-16
[#893]: Create PostgreSQL backups to Azure Storage for snapshot production database

ltello:
This PR adds github workflows and actions to:
- create a backup of the default database of any environment and store it in Azure Storage
- restore a backup from Azure Storage to any app default database
- add a snapshot database to the production environment.
- setup a nightly workflow to backup the production database, store it in Azure Storage and restore the snapshot database from it.

Currently, 2 options are included for the last task (restore the snapshot db nightly from production):

- a dedicated GitHub workflow that creates a dump of the main db in production and restores the snapshot db from it, all inside a single workflow. 
This is a straightforward approach that requires an extra workflow scheduled nightly to produce and store a second dump of the db in production if we want to keep daily copies for security reasons.

- a GitHub workflow that eventually calls the generic workflows for backup/restore of any environment db.
This approach have the advantage of enabling us to manually trigger (any time from the Github Actions UI) a call to restore the snapshot db. Also, the same daily dump used to restore the snapshot db is stored in Azure Storage and be the backup copy we want to keep on a regular basis. No need for a second dump to be run every day.

Also, we need to consider the space limitations of our account in Azure Storage if we are storing a dump per day or we want to overwrite them daily and keep just the most recent one.

Easier to review commit by commit.

### DIFF
--- a/config/terraform/application/variables.tf
+++ b/config/terraform/application/variables.tf
@@ -94,7 +94,7 @@ variable "postgres_flexible_server_sku" {
 }
 
 variable "postgres_snapshot_flexible_server_sku" {
-  default = "GP_Standard_D2ds_v4"
+  default = "B_Standard_B1ms"
 }
 
 variable "postgres_enable_high_availability" {


config/terraform/application/variables.tf
```diff
@@ -93,6 +93,10 @@ variable "postgres_flexible_server_sku" {
   default = "B_Standard_B1ms"
 }
 
+variable "postgres_snapshot_flexible_server_sku" {
+  default = "GP_Standard_D2ds_v4"
+}
```
peteryates, line 98: I think this might be a bit overpowered (and expensive) for the snapshot server, we should probably be using a burstable plan - and then upgrade it later if needed.

Can it be `B_Standard_B1ms` for now?

.github/workflows/backup-db.yml
```diff
@@ -0,0 +1,103 @@
...
+          echo "DB_SERVER=${AZURE_RESOURCE_PREFIX}-${SERVICE_SHORT}-${CONFIG_SHORT}-pg" >> $GITHUB_ENV
+          if [ "${BACKUP_FILE}" == "schedule" ]; then
+            BACKUP_FILE=${SERVICE_SHORT}_${CONFIG_SHORT}_${TODAY}
+          elif [ "${BACKUP_FILE}" == "default" ]; then
+            BACKUP_FILE=${SERVICE_SHORT}_${CONFIG_SHORT}_adhoc_${TODAY}
+          else
+            BACKUP_FILE=${BACKUP_FILE}
+          fi
+          echo "BACKUP_FILE=${BACKUP_FILE}" >> $GITHUB_ENV
+          echo "KEYVAULT_NAME=${AZURE_RESOURCE_PREFIX}-${SERVICE_SHORT}-${CONFIG_SHORT}-inf-kv" >> $GITHUB_ENV
```
peteryates, line 82: I know this is needed but it's _dense_.

config/terraform/application/variables.tf
ltello: Sure

.github/workflows/backup-db.yml
ltello: Ok, I have split it into 2 steps and added some comments to clarify a bit their purpose.

peteryates (approved):
Looks great!

PR #31, opened 2025-01-16
Update content on error pages

PriyaNahar2:
Review prohibited page as content is kept super generic to follow suit with school reg side. Do we need to be more specific?

### DIFF
--- a/app/views/appropriate_bodies/claim_an_ect/errors/prohibited_from_teaching.md.erb
+++ b/app/views/appropriate_bodies/claim_an_ect/errors/prohibited_from_teaching.md.erb
@@ -2,6 +2,6 @@
 title: You cannot register <%= PendingInductionSubmissions::Name.new(@pending_induction_submission).full_name %>
 ---
 
-Our records show that <%= PendingInductionSubmissions::Name.new(@pending_induction_submission).full_name %> cannot be claimed as an ECT.
+Our records show that <%= PendingInductionSubmissions::Name.new(@pending_induction_submission).full_name %> is prohibited from teaching. You can use the <a href="https://www.gov.uk/guidance/check-a-teachers-record" class="govuk-link">Check a teacher's record service</a> to get more information.
 
 [Find and claim another ECT](<%= new_ab_claim_an_ect_find_path %>)
--- a/app/views/appropriate_bodies/claim_an_ect/errors/prohibited_from_teaching.md.erb
+++ b/app/views/appropriate_bodies/claim_an_ect/errors/prohibited_from_teaching.md.erb
@@ -1,5 +1,7 @@
 ---
-title: <%= PendingInductionSubmissions::Name.new(@pending_induction_submission).full_name %> is prohibited from teaching
+title: You cannot register <%= PendingInductionSubmissions::Name.new(@pending_induction_submission).full_name %>
 ---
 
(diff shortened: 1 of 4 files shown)


app/views/appropriate_bodies/claim_an_ect/errors/prohibited_from_teaching.md.erb
peteryates, line 1: This page is so uninformative I can't imagine it's much use.

The URL path gives me more info than the page.

I think we should either:

* make the url path (ends in `/prohibited-from-teaching`) equally vague
* tell the user **why** they can't register the person so they don't think it's a problem with the app and try again

claire-hughez: It's worth saying that Policy and engagement have said no one has been registered before that has been prohibited from teaching. I'm not sure how important this page is, which is why we kept is as blank and simple as possible.

The other reason we had for this was because we didn't think the purpose of our service was to inform of prohibitions - that is what Check a teacher's record is for - and we were worried any data or security implications here.

peteryates: > Our records show that John Smith is prohibited from teaching. You can use the [Check a teacher's record](https://www.gov.uk/guidance/check-a-teachers-record) service to get more information.

PriyaNahar2: Hmm that makes sense - shall we keep it generic to stay in line with the school side for now and revisit it if needed? @emily-prudence-dfe - what do you think? Pete's also got a good suggestion

emily-prudence-dfe: Honestly it's so unlikely to be used, it doesn't matter too much. I'm happy to go with Pete's suggestion directing people towards CTR - it might not be in line with schools now, but I don't think we're at risk of overexposing since it's data any AB can see on any record now

joe-harrison-dfe:
@PriyaNahar2 will cover the small bug on the exemption error page, fixing content and changing 'Claim another induction' link to a button.

PriyaNahar2:
Added link as per @peteryates suggestion:
Before and after screenshots: 
| Before | After |
|  ------ | ------ |
| Screenshot 2025-01-22 at 13 38 51| Screenshot 2025-01-22 at 13 38 34|

Made exempt page more intuitive:
Before and after screenshots: 
| Before | After |
|  ------ | ------ |
|Screenshot 2025-01-22 at 13 40 43|Screenshot 2025-01-22 at 13 40 19| 

cc @joe-harrison-dfe @emily-prudence-dfe @peteryates

emily-prudence-dfe (approved):

app/views/appropriate_bodies/claim_an_ect/errors/prohibited_from_teaching.md.erb
claire-hughez: I think that makes sense - some school users (probably many) won't have access to Check a teacher's record, but I think nearly all AB users will

peteryates:
I've resolved the conflicting files, this is ready to go now @PriyaNahar2 🚀

PR #43, opened 2025-01-18
Make teacher search work on word prefixes

peteryates:
Now search terms will receive a wildcard suffix that means results will be returned if all of the provided words match a teacher's name fields.

This means we can find John Smith by searching for:

* John Smith
* john smith
* jo sm
* jo
* sm


JSJohal (approved):
Looks good!

PR #45, opened 2025-01-19
Hide the edit link for AB past induction periods

peteryates:
This was added for admins and incorrectly shown on the AB side too. Now it can be configured with the enable_edit argument which defaults to false.

| Before | After |
| ----- | ------ |
| Screenshot From 2025-01-19 17-03-06 | Screenshot From 2025-01-19 17-03-21


PriyaNahar2:
Looks good. Ready to merge @peteryates

PR #48, opened 2025-01-20
Register ECT/Mentors - Create TeacherEmail table to store emails

TobyRet:
### Ticket
https://github.com/DFE-Digital/register-ects-project-board/issues/1091 

### Note 
This 1 of 2 possible solutions up for discussion. [Please also see the other proposed solution](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/60).

### Changes
- Create TeacherEmail table to store `teacher_id` and `email`. The `ect_at_school_period` and `mentor_at_school_period` tables now reference this new table. 
- This allows us to ensure that the combination of `teacher_id` and `email` is unique, as well as ensuring that `emails` by themselves are unique (so that other Mentors and ECTs cannot take an existing address)
- Validate at both the model and db layers

### Validation summary ( `teacher_emails` table)

Feature | Allowed?
-- | --
Duplicate email for same teacher | No
Teacher can have more than one email | Yes
Duplicate email across all teachers | No
Blank (nil) emails | Yes

### DIFF
--- a/app/models/ect_at_school_period.rb
+++ b/app/models/ect_at_school_period.rb
@@ -11,6 +11,10 @@ class ECTAtSchoolPeriod < ApplicationRecord
   has_many :events
 
   # Validations
+  validates :email,
+            notify_email: true,
+            allow_nil: true
+
   validates :started_on,
             presence: true
 
--- a/app/models/mentor_at_school_period.rb
+++ b/app/models/mentor_at_school_period.rb
@@ -9,6 +9,10 @@ class MentorAtSchoolPeriod < ApplicationRecord
   has_many :events
 
   # Validations
+  validates :email,
+            notify_email: true,
+            allow_nil: true
+
   validates :started_on,
             presence: true
 
--- a/db/migrate/20250123161607_add_email_to_mentor_at_school_periods.rb
+++ b/db/migrate/20250123161607_add_email_to_mentor_at_school_periods.rb
@@ -1,0 +1,5 @@
+class AddEmailToMentorAtSchoolPeriods < ActiveRecord::Migration[8.0]
+  def change
+    add_column :mentor_at_school_periods, :email, :string
+  end
+end
--- a/db/migrate/20250123161652_add_email_to_ect_at_school_periods.rb
+++ b/db/migrate/20250123161652_add_email_to_ect_at_school_periods.rb
@@ -1,0 +1,5 @@
+class AddEmailToECTAtSchoolPeriods < ActiveRecord::Migration[8.0]
+  def change
+    add_column :ect_at_school_periods, :email, :string
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,8 +10,8 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[8.0].define(version: 2025_01_20_140655) do
-# These are extensions that must be enabled in order to support this database
+ActiveRecord::Schema[8.0].define(version: 2025_01_23_161652) do
+  # These are extensions that must be enabled in order to support this database
   enable_extension "pg_catalog.plpgsql"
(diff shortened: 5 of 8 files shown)


db/migrate/20250120094641_add_email_to_ect_at_school_periods.rb
```diff
@@ -0,0 +1,5 @@
+class AddEmailToECTAtSchoolPeriods < ActiveRecord::Migration[8.0]
+  def change
+    add_column :ect_at_school_periods, :email, :string
```
tonyheadford, line 3: Could we add a unique constraint in some form on these? We might need to add an emails table though and make these references to that table as I'm not sure having them in place would allow it?  Any thoughts?

db/migrate/20250120134632_add_unique_constraint_to_email.rb
```diff
@@ -0,0 +1,6 @@
+class AddUniqueConstraintToEmail < ActiveRecord::Migration[8.0]
+  def change
+    add_index :ect_at_school_periods, :email, unique: true, where: "email IS NOT NULL", name: "unique_email_on_ect_at_school_periods"
+    add_index :mentor_at_school_periods, :email, unique: true, where: "email IS NOT NULL", name: "unique_email_on_mentor_at_school_periods"
```
peteryates, line 4: I don't think we need the `where:` clause here, nulls aren't considered equal.

db/migrate/20250120094641_add_email_to_ect_at_school_periods.rb
TobyRet: Hey @tonyheadford  - as discussed have added constraints on the model and db

db/migrate/20250120134632_add_unique_constraint_to_email.rb
TobyRet: Depends possibly - the default behaviour of the unique index is to disregard null values when enforcing uniqueness but if we are potentially going to have _lots_ of emails with 'null' then adding the WHERE clause will make the index smaller. 

Happy to take you recommendation on this!

tonyheadford: Will this work for teachers:
1. using their personal email and they move school and want to use the same email
2. leave a school and then return at some point in the future and wish to keep their original school email

app/migration/builders/ect/school_periods.rb
```diff
@@ -12,12 +12,16 @@ def build
         success = true
         school_periods.each do |period|
           school = School.find_by!(urn: period.urn)
+          # Check where the email address may come from
+          teacher_email = TeacherEmail.find_or_create_by!(teacher_id: teacher.id, email: nil)
```
TobyRet, line 16: @tonyheadford - can the email be sourced or should we set to nil for now?

app/migration/builders/mentor/school_periods.rb
```diff
@@ -12,12 +12,16 @@ def build
         success = true
         school_periods.each do |period|
           school = School.find_by!(urn: period.urn)
+          # Check where the email address may come from
```
TobyRet, line 15: https://github.com/DFE-Digital/register-early-career-teachers-public/pull/48/files#r1923775532

db/migrate/20250120094641_add_email_to_ect_at_school_periods.rb
TobyRet: As discussed we are leaving constraints alone for now

db/migrate/20250120134632_add_unique_constraint_to_email.rb
TobyRet: As discussed we are leaving constraints alone for now

TobyRet:
Closing in favour of https://github.com/DFE-Digital/register-early-career-teachers-public/pull/48

PR #58, opened 2025-01-22
Restrict the registration of ECTs who are prohibited from teaching

leoapost:
This PR prevents ECTs who are prohibited from teaching from being registered in the service.

We don't anticipate schools attempting to register ECTs who are prohibited from teaching. However, we implemented this functionality as an added safeguard.

Since this involves sensitive safeguarding information, we display a more generic page stating that the ECT cannot be registered.

Changes
- Added new wizard step `CannotRegisterECTStep` to handle prohibited teacher cases
- Added view to display appropriate message when registration is prevented
- Added two TRS teachers with prohibited alerts to facilitate testing in non-prod envs


tonyheadford (approved):
LGTM

ltello:
@leo we are also adding a feature test for each unhappy path.

ltello (changes requested):
LGTM but please see my comment about a missing feature test.

PR #62, opened 2025-01-22
Fix bug where induction period not saved

edujackedu:
This happens becuase we were checking for the truthiness of each success step but calling `.create` on a invalid model returns the unsaved record which is truthy.

Now we call `#persisted?` at the end of the method and only call record_induction_period_event if the induction period was saved.

Additionally, the event job errored when deserialising non-persisted models that were rolled back if a transaction errored.

ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/1157


peteryates:
Sorry @edujackedu I merged #74 without realising it contained this PR 🤦🏽

PR #68, opened 2025-01-24
Add feature spec for blocking schools from registering prohibited ECTs

leoapost:
Follow-up PR to #58, adding the missing feature test.


peteryates (approved):
Looks good!

PR #71, opened 2025-01-24
1136: Remove terms associated with an IP when end date is removed 

ghost:
ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/1136


joe-harrison-dfe:
@edujackedu on that user you linked, when you remove the end date and press update, the number of terms still remains on the record so not sure that is working? When I try it on an historic period then it blocks via overlapping

PR #76, opened 2025-01-27
Restrict the registration of mentors who are prohibited from teaching

leoapost:
Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/972

This PR prevents mentors who are prohibited from teaching from being registered in the service.

We don't anticipate schools attempting to register mentors who are prohibited from teaching. However, we implemented this functionality as an added safeguard.

Since this involves sensitive safeguarding information, we display a more generic page stating that the mentor cannot be registered.


ltello (commented):
Look great.
It'd be good to include the new logic in find mentor step specs.

ltello (approved):

PR #79, opened 2025-01-27
1146: admin back link 

peterdavidhamilton:


app/views/admin/induction_periods/edit.html.erb
```diff
@@ -1,8 +1,11 @@
 <% page_data(
   title: "Edit induction period",
-  error: @induction_period.errors.present?
+  error: @induction_period.errors.present?,
+  backlink_href: admin_teacher_path(@induction_period.teacher)
 ) %>
 
+<%= content_for(:backlink_or_breadcrumb) { govuk_back_link(href: admin_teacher_path(@induction_period.teacher)) } %>
```
peteryates, line 7: @peterdavidhamilton - `content_for` is called [in the helper](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/helpers/application_helper.rb#L11-L13) already so I think this is unnecessary.

.tool-versions
```diff
@@ -1 +1,2 @@
 terraform 1.9.8
+nodejs 22.5.0
```
peterdavidhamilton, line 2: This could be removed if `.node-version` can use a SemVer `2.5.0`. ASDF complained otherwise.

peteryates (approved):
Nice!

PR #85, opened 2025-01-28
1136 Validate number of terms absence when nil end date

ghost:
When end date is cleared, number of terms must be empty.

ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/1136

### DIFF
--- a/app/models/pending_induction_submission.rb
+++ b/app/models/pending_induction_submission.rb
@@ -38,9 +38,15 @@ class PendingInductionSubmission < ApplicationRecord
             on: %i[release_ect record_outcome]
 
   validates :number_of_terms,
-            presence: { message: "Enter a number of terms" },
-            inclusion: { in: 0..16,
-                         message: "Terms must be between 0 and 16" },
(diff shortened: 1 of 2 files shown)


app/models/induction_period.rb
```diff
@@ -12,10 +12,15 @@ class InductionPeriod < ApplicationRecord
...
-            if: -> { finished_on.present? }
+            inclusion: {
+              in: 0..16, message: "Terms must be between 0 and 16", if: -> { finished_on.present? }
+            },
+            presence: {
+              message: "Enter a number of terms", if: -> { finished_on.present? }
+            },
+            absence: {
+              message: "Delete the number of terms if the induction has no end date", if: -> { finished_on.blank? }
+            }
```
peteryates, line 23: Should this be in `PendingInductionSubmission` too? Maybe they can be together in `SharedInductionPeriodValidation` (although the optional validation context might restrict that)

peteryates (approved):
Looks good

PR #91, opened 2025-01-29
Adding a manual workflow for deploying to an environment

tonyheadford:
The migration environment is no longer in the normal deploy workflow to prevent issues with deployments being triggered by merges during migration or while fixing/reviewing data.
In addition, the introduction of DfE Analytics will prevent manual deployments to envs as terraform apply in the console now requires gcloud logins and not all developers will have gcloud accounts or access.

This PR adds a workflow that enables deploying to an environment manually (this is virtually identical to a workflow added to ECFv1).

### DIFF
--- a/.github/workflows/deploy_migration.yml
+++ b/.github/workflows/deploy_migration.yml
@@ -1,0 +1,58 @@
+name: "Manual deploy image"
+
+on:
+  workflow_dispatch:
+    inputs:
+      environment:
+        description: Deployment environment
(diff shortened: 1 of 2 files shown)


tonyheadford:
@cwrw has found some issues that we'll need to work around and is requesting a change to the docker build script managed by the ops team. Agreed to leave this on hold until there's a way forward.

ltello (approved):
LGTM

Issue #65, opened 2025-01-23
Importing DQT data


peteryates:
This was copied from @joe-harrison-dfe's [original data quality issues doc](https://educationgovuk-my.sharepoint.com.mcas.ms/:w:/g/personal/joe_harrison_education_gov_uk/EWSM1D1uTgJMki_Ov414EssBcUM5_ZqEWBFsr60vPu_w_A?e=KpQrkf).

## 1. Bad Status:

* [x] Implement bad status logic

This issue is when a record has an open induction period alongside an overall induction status of pass, fail or exempt. 

* There are 18 records that have an open induction period with a status that indicates they shouldn’t which we will work with the ops team to manually check.  
* There are a further 40 which are down as in progress but with no induction period which we will also manually check.  
* 100 records have no induction start and an AB that is exempt in Wales so we recommend updating these to exempt. @bbelward agrees  
* There are just under 35,000 records that have no induction dates with an overall status  they shouldn’t have, so we suggest just updating those to ‘Required to Complete’RTC. 
  * Rationale – no induction period means they shouldn’t be in our database  
  * Worst case – someone corrects and says they have actually done induction, which we would have to correct anyway

> [!NOTE]  
> These are going to be investigated and fixed on the TRS side, we don't need to take them into account as part of the migration.

 ## 2. Induction Period start date after end date:

* [ ] Implement period start date after end date logic

This is an issue where an induction period has a start date that isn’t blank and is after the period end date. 

* There are 156 records that contain this issue looking at our data cut from Rob. 
* @easeynathan  has conducted some spot checks on these and initial recommendation is to not migrate any of these records over. 
* Live in TRS and can check if we get a query associated with this, odds are low. @bbelward  agrees

> [!NOTE]  
> These are going to be investigated and fixed on the TRS side, we don't need to take them into account as part of the migration.
  
## 3. Induction period before overall induction start: 

* [ ] Implement induction period before overall induction start logic

This is an issue where the overall induction start date is not blank and is after the period start and/or end date where populated. 

* There are 123 records that contain this issue from our data cut from Rob. 
* This is normally caused by a bug in the DQT that can reset an overall induction start date to a more recent induction period start. 
* Recommendations are to revert records that have had their start date automatically moved and use the earliest start date associated with the record. 
* This is a bug, overall start shouldn’t change  @bbelward agrees  

## 4. Future start or end date:

* [ ] Implement future start or end date logic

[...]

PR #92, opened 2025-01-29
Improve the AB data importer

peteryates:
This change makes the three importers (teacher, AB, IP) less independent and puts the `AppropriateBodies::Importers::Importer` class in charge.

This means the three importers can depend on each other, which means we can make the process *much faster*.

The raw teachers table has 1.8m records and parsing it into a Ruby CSV file is slow, and we only want ~50k of those records so it's a pointless exercise. Now we use the induction periods table to work out which teachers we need, and loop through the CSV file looking for the rows we want, before parsing.

It's not finished yet, but it imports more or less what we need in more or less the right way. The remaining tasks listed in #65 can be solved on the back of this.


lib/tasks/appropriate_body_import.rake
peteryates, line 1: It remains to be seen whether the rake task is useful, might not make the final cut!

PR #93, opened 2025-01-29
Replace teacher.trn with teacher.id in routes

ghost:
ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/1151


peteryates (approved):
Looks good, tested it in both the AB and admin interfaces.

PR #101, opened 2025-01-30
Bugfix: trying to claim an ECT that was released by a different AB with a date prior to their last IP

ghost:
ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/1134

### DIFF
--- a/app/controllers/appropriate_bodies/claim_an_ect/check_ect_controller.rb
+++ b/app/controllers/appropriate_bodies/claim_an_ect/check_ect_controller.rb
@@ -2,6 +2,7 @@ module AppropriateBodies
   module ClaimAnECT
     class CheckECTController < AppropriateBodiesController
       def edit
+        @current_appropriate_body = current_user.appropriate_body
         @pending_induction_submission = find_pending_induction_submission
         @teacher = Teacher.find_by(trn: @pending_induction_submission.trn)
       end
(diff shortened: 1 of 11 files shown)


app/models/pending_induction_submission.rb
```diff
@@ -74,11 +74,26 @@ class PendingInductionSubmission < ApplicationRecord
...
+
 private
 
   def start_date_after_qts_date
     return if trs_qts_awarded_on.blank?
 
     ensure_start_date_after_qts_date(trs_qts_awarded_on)
   end
+
+  def no_future_induction_periods
```
peterdavidhamilton, line 87: General question. How has the team decided on when to move things out, in this case, into validators?

app/models/pending_induction_submission.rb
```diff
@@ -74,11 +74,26 @@ class PendingInductionSubmission < ApplicationRecord
...
+    teacher = Teacher.find_by(trn:)
+
+    return if teacher.blank?
+
+    teacher.induction_periods.where.not(finished_on: nil).find_each do |period|
+      if started_on < period.finished_on
+        errors.add(:started_on, "Enter a start date after the last induction period finished (#{period.finished_on.to_fs(:govuk)})")
+        break
+      end
+    end
```
ltello, line 97:
```suggestion
    latest_date_of_induction = teacher.induction_periods.maximum(:finished_on)
    if started_on < latest_date_of_induction
        errors.add(:started_on, "Enter a start date after the last induction period finished (#{latest_date_of_induction.to_fs(:govuk)})")
    end
```

PR #102, opened 2025-01-30
[1059] - block school user from adding an ect as their own mentor

ltello:
[Issue](https://github.com/DFE-Digital/register-ects-project-board/issues/1059)

This story blocks school users from adding an ECT as the ECT's own mentor as this is against the policy intent. This is most likely to happen in error if a user enters the ECT's TRN instead of the mentor's TRN - this new page will block the journey in this case.

This PR:
- adds a new step to display an error page when the user enters the TRN of the ECT to mentor in FindMentorStep.
- allow registering a teacher a second time as a mentor if it was already an ECT or viceverse. Otherwise, raise an exception
- prevents ECTs from auto-assigning themselves as mentors if they are already registered as mentors.

Better review commit by commit.

Video: 
https://github.com/user-attachments/assets/3874d17c-b156-4e3a-a96f-69aafe2b09ef

### DIFF
--- a/app/forms/schools/assign_mentor_form.rb
+++ b/app/forms/schools/assign_mentor_form.rb
@@ -30,7 +30,7 @@ module Schools
     end
 
     def mentorship_authorized
-      errors.add(:mentor_id, "This teacher cannot be mentoring this ECT") if eligible_mentors.exclude?(mentor)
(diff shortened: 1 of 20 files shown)


peteryates (approved):
This looks good to me.

PR #103, opened 2025-01-30
Configure rack attack

peteryates:
Add and configure `rack-attack` to protect against abusive requests.

For this to work in production we need to enable the Redis cache. It's already provisioned and present for the production-like environments.

## Testing locally with [siege](https://openbenchmarking.org/test/pts/apache-siege&eval=26ae0a03bb5a412022a81331100103a2e4882950)

To test locally you need to enable the cache in dev.

```
bundle exec rails dev:cache
```

Then create a siege file, e.g. in `~/protected-paths`:

```
HOST=http://localhost:3000
$(HOST)/sign-in
$(HOST)/otp-sign-in
```

Now we can run it with `siege ~/protected-paths`. Hopefully things are quickly throttled, like this:

Screenshot From 2025-01-30 15-10-17

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -44,6 +44,7 @@ COPY . .
 # Precompile assets
 RUN RAILS_ENV=production SECRET_KEY_BASE=required-to-run-but-not-used \
     GOVUK_NOTIFY_API_KEY=TestKey \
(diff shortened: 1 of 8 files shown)


peteryates:
Note, this might not be necessary if we can do rate limiting in Azure front door.

PR #110, opened 2025-01-31
[893] - keep only the last 7 days of backups in Azure Storage

ltello:
- Nightly production db backups will create dump files named `cpdec2_pd_adhoc_day_of_week`. 
  This way every week the new backups will be overwriting those from the previous one keeping so only the last 7 days copies in Azure Storage.

- Display full path to the backup file in Azure Storage on Slack notifications.


peterdavidhamilton (approved):
LGTM

PR #116, opened 2025-02-03
Bugfix: Allow data from TRS API to be persisted

peterdavidhamilton:
Character limit prevented ECTs with a status of "RequiredToComplete" from being processed.

- ~Increase size of `PendingInductionSubmission` `trs_induction_status` character limit from `16` to `18`.~
- Remove `PendingInductionSubmission` `trs_induction_status` character limit.
- Add validation when registering.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[8.0].define(version: 2025_01_23_161652) do
+ActiveRecord::Schema[8.0].define(version: 2025_01_31_171406) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "pg_catalog.plpgsql"
   enable_extension "pg_trgm"
@@ -322,7 +322,7 @@ ActiveRecord::Schema[8.0].define(version: 2025_01_23_161652) do
     t.string "trs_first_name", limit: 80
     t.string "trs_last_name", limit: 80
     t.date "date_of_birth"
-    t.string "trs_induction_status", limit: 16
+    t.string "trs_induction_status", limit: 18
     t.enum "induction_programme", enum_type: "induction_programme"
     t.date "started_on"
     t.date "finished_on"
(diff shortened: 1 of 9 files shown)


tonyheadford (approved):
LGTM!

ltello (commented):
Maybe we can define `enum` in the model and db like we have done with other fields

db/schema.rb
```diff
@@ -322,7 +322,7 @@
     t.string "trs_first_name", limit: 80
     t.string "trs_last_name", limit: 80
     t.date "date_of_birth"
-    t.string "trs_induction_status", limit: 16
+    t.string "trs_induction_status", limit: 18
```
claire-hughez, line 325: Can we make this larger? 50? Unless there's a good reason not to. It feels like it might bite us again potentially.

peterdavidhamilton: Certainly could, that was my initial reaction. Waiting to hear what the devs think.

ltello: I don't see any problem if we increase the length to 50

peteryates:
I set this to `16`  way back before we had any TRS integration etc (ab4a64782834f4ecf7e913) under the assumption TRS would have sensible names.

`RequiredToComplete` doesn't really make sense, you're still `RequiredToComplete` if you're part way through. It should really be called `NotStarted`.

As we can't fix it and need to rely on this having been validated on TRS' end already, I'm happy to accept whatever comes to us. Our code won't know what to do with it if something new/unexpected comes along, but it shouldn't _break_.

If we were doing anything important with the status I'd be inclined to follow @ltello's suggestion of an enum, so we'd be notified immediately if something changed, but we just display it and prevent people in the finished states from being claimed.

db/schema.rb
peteryates, line 325:
```suggestion
    t.string "trs_induction_status", limit: nil
```

peterdavidhamilton: Cheers @peteryates I'll get this merged tomorrow 👍🏼

PR #122, opened 2025-02-04
Add dfe-analytics BigQuery integration

leoapost:
This PR sets up Dfe Analytics in our service to send events to BigQuery.

We only care about the following four tables for now with all PII hidden:
- teachers
- appropriate_bodies
- induction_periods
- induction_extensions

Changes:
  - Add `dfe-analytics` gem
  - Whitelist entities and attributes as specified by stakeholders. Hide all PII attributes
  - Send events in a new, low-priority queue
  - Enable and schedule the dfe-analytics `Entity Table Check` job to run every day at 00:30
  - Disabled by default

How to test:
1. Set `DFE_ANALYTICS_ENABLED=true` in your `.env`
2. Start the app and the jobs with `bin/dev`
3. Create one of the whitelisted entities (i.e. teachers)
4. Confirm the job to send the event is created
5. Verify PII are hidden

You may want to temporarily set `config.log_only = true` in `config/initializers/dfe_analytics.rb` and check the logs for the events.


config/analytics_blocklist.yml
```diff
@@ -0,0 +1,505 @@
...
+  - provider_partnership_id
+  - lead_provider_id
+  - delivery_partner_id
+  - user_id
+  - author_type
+  - author_id
+  - author_name
+  - author_email
+  - created_at
+  - updated_at
```
peteryates, line 300: Sorry, can we move this to `config/analytics.yml` - events will be useful for the data team, but `heading` and `body` to the `config/analytics_hidden_pii.yml`?

peteryates (approved):
Looks great!

peteryates (approved):

PR #129, opened 2025-02-05
fix bug: some args are not been passed correctly to to github actions

ltello:
- deploy-env-to-aks:  rename arg `inputs.current-commit-sha` to its right name `inputs.commit-sha`
 - deploy: `deploy-production` needs to depend on `docker` step. Otherwise, its output (`needs.docker.outputs.docker-image-tag`) is not accessible


peteryates (approved):
Looks sensible

PR #134, opened 2025-02-05
Replace 'Do it yourself' with 'School-based induction programme'

peteryates:
DIY is an internal term, we should use terms the users will understand in the UI. This was brought up earlier this morning in a UR session.


claire-hughez (approved):
HOT TO GO

PR #141, opened 2025-02-06
1134: Update date validations of IPs and Pending Induction Submissions

ghost:
- validate Pending Induction Submission cannot be created if ECT has a future IP
- validate IP started_on != finished_on 

fixes: 
- https://github.com/DFE-Digital/register-ects-project-board/issues/1135
- https://github.com/DFE-Digital/register-ects-project-board/issues/1134

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,8 @@
 .idea
 .vscode
 
+bin/konduit.sh
+
 documentation/**/*.html
 !documentation/site/**/*.html
 documentation/site/output
--- a/documentation/setup.md
+++ b/documentation/setup.md
@@ -41,7 +41,7 @@ This assumes your username is `joey`. You can check yours by running `whoami`
 
    Then, running `psql` should connect and give us a prompt like this:
 
-   ```
+   ```sh
    $ psql
 
    psql (16.3)
@@ -59,7 +59,7 @@ Once you have Ruby and NodeJS installed we're ready to install the application's
 
 We can do this in one step with `bin/setup`:
 
-```bash
+```sh
 $ bin/setup
 
 # == Installing dependencies ==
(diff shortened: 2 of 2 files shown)


spec/features/appropriate_bodies/claim_an_ect_spec.rb
```diff
@@ -12,7 +12,7 @@
       FactoryBot.create(:induction_period, :active, teacher:, appropriate_body: other_body)
     end
 
-    scenario 'Button is hidden when induction is completed' do
+    scenario 'Button is hidden when induction is active' do
```
peteryates, line 15:
```suggestion
    scenario 'Button is hidden when induction is ongoing' do
```

app/models/induction_period.rb
```diff
@@ -52,4 +53,10 @@ def start_date_after_qts_date
 
     ensure_start_date_after_qts_date(teacher.trs_qts_awarded_on)
   end
+
+  def started_on_different_from_finished_on
+    return if started_on != finished_on
+
+    errors.add(:finished_on, "End date must be different from start date")
+  end
```
peteryates, line 61: I think we need to talk this through with the team tomorrow - but to me it makes sense.

peteryates: Actually we kind of have this already in `Interval#period_dates_validation`.

There it's checking `if finished_on <= started_on` though. I think that covers this too 🤔

peteryates (approved):
Looks good but I think we need to chat through with the team first. Banning 0 duration induction periods is quite a big step.

emily-prudence-dfe:
I've tested both on review app and they look like they're working as expected to me @edujackedu

PR #144, opened 2025-02-06
[1159] - fix the url to the opt sign in page in the emails containing the otp code when user logs in

ltello:
Fix the url to the opt sign in page in the emails containing the otp code when user logs in


peteryates:
Sorry @ltello I think we already addressed this here #143

PR #148, opened 2025-02-06
Fix roadmap

claire-hughez:
I can never get bullets right in Markdown.

Particularly because they look right in code editor previews and then don't in prod.

It makes me sad.

### DIFF
--- a/documentation/site/content/product/roadmap/later.md
+++ b/documentation/site/content/product/roadmap/later.md
@@ -1,5 +1,7 @@
 ## Later 🔮
+
 We will:
+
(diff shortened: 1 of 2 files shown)


peteryates (approved):

peteryates:
100% not your fault. There is no single _markdown_, every implementation of it seems to differ a bit. The closest thing to a standard is [commonmark](https://commonmark.org/help/), but not everything follows it exactly.

Rule of thumb though, always leave an empty line between different kinds of content. Headings, paragraphs, lists, tables, quotes. You don't lose anything by doing it but the resulting page is more likely to render nicely.

PR #154, opened 2025-02-07
Save teacher email

peterdavidhamilton:


ltello (commented):
Looks good. 
I think we just need to add the email to the specs in `spec/wizards/schools/register_ect/ect_spec.rb#register!`

PR #166, opened 2025-02-10
Move TRS API variables to to app config

peteryates:
We don't really need to store the `TRS_BASE_URL` (now renamed to `TRS_API_BASE_URL`) in the keyvault so I've moved that to the terraform env-specific yaml files too and will delete from the keyvaults.

@gunndabad also kindly bumped the TRS API version to `20250203` so we'll have the latest functionality available without having to rely on the `Next` (beta) version, so that's the default for the deployed envs.


peterdavidhamilton (approved):
`TRS_API_VERSION: 'Next'` what does this do for us?

Early warning ⏰

peteryates:
It basically is the 'beta' version of the API, some experimental stuff lands there before it's in a properly released version.

PR #167, opened 2025-02-10
Import extensions

peteryates:
- **Log the import messages rather than puts them**
- **Move the actual importing to the import! method**
- **Split up the importer into sensible methods**
- **Add events metadata jsonb column**
- **Remove the unique index on la code and establishment no**
- **Build and save events**


lib/appropriate_bodies/importers/importer.rb
```diff
@@ -1,70 +1,115 @@
...
           induction_period_rows << ip
         end
       end
 
-      InductionPeriod.insert_all(induction_period_rows.map(&:to_record))
+      induction_period_ids = InductionPeriod.insert_all!(induction_period_rows.map(&:to_record), returning: [:id])
+      Rails.logger.info("Induction periods inserted: #{InductionPeriod.count}")
+
+      # FIXME: how do we set titles?
+      #        can do it by executing a single line of SQL after insert
```
peteryates, line 75: 🤦🏽 this is already done on line 95.

PR #174, opened 2025-02-11
Try and fix migration deployment

tonyheadford:
Manually deploying to migration is failing.
It appears to be related to analytics and the `Migration::` models built for connecting to ECTv1 tables.

DfE Analytics has a config option to exclude models so attempting to see if we can exclude the entire `Migration::` namespace of models to remedy the issue.


peteryates (approved):
Didn't realise it could do that!

PR #181, opened 2025-02-12
Make the record failing outcome button red

peteryates:
| Before | After |
| ------ | ------- |
| Screenshot From 2025-02-12 09-52-19 | Screenshot From 2025-02-12 09-52-27 |

Also, I added some view specs covering these pages.

- **Make the fail button red**
- **Add view specs for the pass/fail induction pages**


PriyaNahar2 (approved):
Looks good to me 👍

PR #183, opened 2025-02-12
1213 number of terms decimal validation

ghost:
Allow number of terms with maximum one decimal point
move number of terms validation to separate shared object
change max number of terms to 16 for all objects

### DIFF
--- a/.github/workflows/reseed-staging-db.yml
+++ b/.github/workflows/reseed-staging-db.yml
@@ -17,7 +17,7 @@ jobs:
       with:
         token: ${{ secrets.GITHUB_TOKEN }}
 
-    - uses: Azure/login@v1
+    - uses: Azure/login@v2
       with:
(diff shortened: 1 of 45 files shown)


app/validators/terms_range_validator.rb
```diff
@@ -2,22 +2,11 @@ class TermsRangeValidator < ActiveModel::EachValidator
...
     return if value.nil?
 
-    range = case record
-            when InductionExtension
-              0.1...16
-            else
-              0..16
-            end
+    range = record.class::VALID_NUMBER_OF_TERMS
+    min, max = range.values_at(:min, :max)
```
peteryates, line 6: Nice.

peteryates (commented):
Looks good, a definite improvement

peterdavidhamilton (approved):

PR #187, opened 2025-02-12
ECT service tidy up

peterdavidhamilton:


peteryates (commented):
Thanks @peterdavidhamilton. This is much cleaner and the query itself is dead simple.

I do think that it should be in a service class though, even if it's minimal - purely because it makes both it and the controller easier to test and extend.

spec/services/schools/home_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+                      started_on: 1.year.ago)
+    FactoryBot.create(:mentorship_period, :active,
+                      mentor: mentor_period,
+                      mentee: ect_period,
+                      started_on: 1.year.ago)
+  end
+
+  describe '#ects_and_mentors' do
+    it "returns the school's ECTs and mentors" do
+      expect(service.ects_and_mentors).to eq([ect_period])
```
peterdavidhamilton, line 35: @peteryates following your comments this is back in a service object. I'm not keen on the method name, I think it's misleading.

app/services/schools/home.rb
```diff
@@ -0,0 +1,16 @@
+module Schools
+  class Home
+    attr_accessor :school
+
+    def initialize(school:)
+      @school = school
+    end
+
+    def ects_and_mentors
```
ltello, line 9: maybe name it `ects_with_mentors`?

ltello (approved):
Looks good to me. 
Maybe I would add a mentorship period ended in the spec and check it is not included in  `.ects_with_mentors`

PR #190, opened 2025-02-12
Disable delete protection for review apps

vipin-dfe:
## Description
Review apps donot need delete protection and are deleted when PRs are merged.

## How to review
Closing the PR will trigger the delete review app workflow which should run successfully.


config/terraform/application/variables.tf
```diff
@@ -104,6 +104,10 @@ variable "postgres_enable_high_availability" {
 
 variable "enable_logit" { default = false }
 
+variable "gcp_table_deletion_protection" {
+  type = bool
```
saliceti, line 108: This requires a default value

vipin-dfe: done.

saliceti (approved):

PR #193, opened 2025-02-12
prepopulate attributes on each step of the mentor journey

ltello:


peterdavidhamilton:
image

peterdavidhamilton (commented):
It all looks sensible and populating the forms will make manual testing of my ticket much easier. My only concern is the rename using `ECF` as a prefix only because I'm getting to grips with `ECT` (but recently heard about `ECTE` as an abbreviation). `DateHash` could be a good alternative. It's giving Ronseal

peterdavidhamilton (approved):
In anticipation of the spec fix 👍🏼

PR #202, opened 2025-02-13
Change induction period overlap validation to target the offending date

peterdavidhamilton:
This is done in a way that can be rolled out to other period types.

### DIFF
--- a/app/models/induction_period.rb
+++ b/app/models/induction_period.rb
@@ -33,9 +33,21 @@ private
   end
 
   def teacher_distinct_period
-    overlapping_siblings = InductionPeriod.siblings_of(self).overlapping_with(self).exists?
+    return unless overlap?
 
-    errors.add(:base, "Induction periods cannot overlap") if overlapping_siblings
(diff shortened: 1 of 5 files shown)


app/models/induction_period.rb
```diff
@@ -42,9 +42,21 @@ def valid_date_order?
...
+    elsif siblings.any? { |s| s.range.include?(finished_on) }
+      errors.add(:finished_on, "End date cannot overlap another induction period")
+    end
+  end
+
+  def siblings
+    InductionPeriod.siblings_of(self)
+  end
+
+  def overlap?
```
peteryates, line 58:
```suggestion
  def overlaps_with_siblings?
```

peteryates (approved):
Looks great. Left one suggestion on the method name but it's very much a take it or leave it.

emily-prudence-dfe:
@peterdavidhamilton Finally had a proper play around with this and it looks really good and helpful, thank you for picking this up. 

Happy for this to be merged 😊

PR #204, opened 2025-02-13
[1248] - when a user signing in via DSI clicks sign out sign them out of dfe sign in too

ltello:
[Issue](https://github.com/DFE-Digital/register-ects-project-board/issues/1248)

- Compress, encrypt and store the id_token (from DSI in the sign in callback) into a cookie for later use in the sign out.
- Extend OmniAuth DSI provider to callback RECT to log users out.
- Configure session users to set their specific logout url
- Call DSI to close the session there as well as in the service when user clicks 'Sign out'


app/services/sessions/manager.rb
```diff
@@ -41,6 +44,12 @@ def check_authorisation!(session_user)
       fail(MissingAccessLevel)
     end
 
+    def encrypt_token(token)
+      secret_key = Rails.application.secret_key_base.byteslice(0, 32)
```
peteryates, line 48: I feel like we should set this as a a separate environment variable called `DFE_SIGN_IN_ID_TOKEN_KEY` or similar, which would make rotating easier should we have to.

peteryates (approved):
Looks great, thanks @ltello.

app/services/sessions/manager.rb
peteryates: Yeah. Was just thinking if for some reason we need to rotate it, it's easier than doing the Rails creds too

PR #214, opened 2025-02-14
Add appeal content to fail induction page

PriyaNahar2:
Bug ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/1234 

Added ECT appeal content and left existing content as it is because we have the red CTA now which indicates a fail quite obviously. 

Content does need cleaning up / refining post MVP but for now, it's good to go.

 Before and after screenshots: 
| Before | After |
|  ------ | ------ |
| Screenshot 2025-02-14 at 15 11 07 | Screenshot 2025-02-14 at 15 07 25 |


emily-prudence-dfe (approved):
Looks great, it's exactly the type of thing we needed just to trigger the user to stop and think. Thanks @PriyaNahar2!

peteryates (approved):

PR #215, opened 2025-02-14
Add Error Message to Exempt ECTs

ghost:
when exempt, ECTs shoudn't be claimable and a notice should be displayed to the user

https://github.com/DFE-Digital/register-ects-project-board/issues/1251

### DIFF
--- a/spec/components/appropriate_bodies/claim_ect_actions_component_spec.rb
+++ b/spec/components/appropriate_bodies/claim_ect_actions_component_spec.rb
@@ -21,73 +21,70 @@ RSpec.describe AppropriateBodies::ClaimECTActionsComponent, type: :component do
     )
   end
 
-  context "when teacher is registered with another appropriate body" do
-    let(:other_appropriate_body) { FactoryBot.create(:appropriate_body) }
+  context "with teacher" do
     let(:teacher) { FactoryBot.create(:teacher) }
-    let!(:induction_period) do
-      FactoryBot.create(
-        :induction_period,
-        :active,
-        teacher:,
-        appropriate_body: other_appropriate_body
-      )
-    end
-
-    it "renders inset text explaining teacher cannot be registered" do
-      render_inline(component)
-
-      expect(page).to have_text("You cannot register John Doe")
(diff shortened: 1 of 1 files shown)


spec/components/appropriate_bodies/claim_ect_actions_component_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+
+  subject(:component) do
+    described_class.new(
+      teacher:,
+      pending_induction_submission:,
+      current_appropriate_body:
+    )
+  end
+
+  context "when teacher is registered with another appropriate body" do
```
peterdavidhamilton, line 24: If you want you could nest this block and define `teacher` once.

```ruby
context 'with teacher' do 
   let(:teacher) { FactoryBot.create(:teacher) }
   context 'when registered ...' do
   end 
   context 'when exempt' do
   end
end
```

app/components/appropriate_bodies/claim_ect_actions_component.rb
```diff
@@ -0,0 +1,50 @@
...
+    def induction_status
+      ::Teachers::InductionStatus.new(
+        teacher:,
+        induction_periods: teacher&.induction_periods,
+        trs_induction_status: pending_induction_submission.trs_induction_status
+      )
+    end
+
+    def pending_induction_submission_full_name(pending_induction_submission)
+      "#{pending_induction_submission.trs_first_name} #{pending_induction_submission.trs_last_name}"
```
peterdavidhamilton, line 47: `PendingInductionSubmissions::Name` does this, we should probably leverage that here too?

peterdavidhamilton (changes requested):
Opportunity to use an existing service but LGTM. I included method naming and spec formatting suggestions for consideration.

peterdavidhamilton (approved):

emily-prudence-dfe:
@edujackedu good to go on this one.

I did spot as I was testing that the same issue is happening with closed inductions. We should merge this regardless and I'll create another ticket to tackle closed inductions.

PR #223, opened 2025-02-18
Add fallback phase banner text

peteryates:
We're missing this in production and there should be something there, or we may as well drop the banner altogether.


ltello (approved):

PriyaNahar2 (approved):
Looking good to me 👍

PR #225, opened 2025-02-18
[Migration] Allow existing Teacher to be updated by migration

tonyheadford:
If a `Teacher` record already exists the migrator generates a migration failure error to reflect this. However in the real environment the majority of `Teacher` records will already exist, brought in via the AB data.
This PR prevents the error being generated and also if the `full_name` attribute from ECF does not match the `trs_first_name` and `trs_last_name` it will set the `corrected_name` attribute.

### DIFF
--- a/app/migration/builders/teacher.rb
+++ b/app/migration/builders/teacher.rb
@@ -10,13 +10,34 @@ module Builders
     end
 
     def build
-      ::Teacher.create!(trn:, trs_first_name: parser.first_name, trs_last_name: parser.last_name, legacy_id:)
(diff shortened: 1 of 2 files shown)


peterdavidhamilton (approved):
Are there middle names in ECF1? I noticed the corrected name trait in the factory included one.

tonyheadford:
> Are there middle names in ECF1? I noticed the corrected name trait in the factory included one.

I'm sure there are, we only check the first name for a match with DQT in ECF1 so the schools can easily add middle name or multiple surname parts

PR #227, opened 2025-02-18
1116: Remove number of terms checks from form

ghost:
To remove the validation we need to convert the field to a text field with numeric input mode. I don't think this small change requires additional tests.

Screenshot 2025-02-18 at 13 55 21


app/views/admin/induction_periods/edit.html.erb
```diff
@@ -18,14 +18,11 @@
       <%= f.govuk_date_field :finished_on,
           legend: { text: "End date", size: "m" } %>
 
-      <%= f.govuk_number_field :number_of_terms,
+      <%= f.govuk_text_field :number_of_terms,
```
peterdavidhamilton, line 21: I was able to type in letters without a validation catching it which resulted in `0.0` being saved. I think this introduces the risk of collecting bad data. 

`4.r5` becomes `4.0` so typos could result in subtle deviations. Am I right in thinking this could ultimately impact payments?

peteryates: Yeah - we should be using `input type="text"` here. There's [some guidance in the design system on the topic here](https://design-system.service.gov.uk/components/text-input/#asking-for-decimal-numbers).

emily-prudence-dfe:
Update to validation looks good to me in terms of the style of validation message etc.

Happy for this to go ahead once dev comments are addressed - and also happy to discuss if we need (esp. if the issue raised here around letters is on AB facing service too)

peteryates (approved):
Looks good and definitely more in line with what's recommended in the design system docs.

PR #232, opened 2025-02-19
Add a custom error when DfE Sign-in org not found

peteryates:
This should provide us a little more clarity on what's going on when the org isn't found and we can't fall back to personas.

Currently in Sentry the error doesn't make it clear who is failing to log in.


ltello (approved):
Good!. We need to know more when a sign in fails.

PR #235, opened 2025-02-19
Add school user home page mentors tab

peterdavidhamilton:
image

Add a new tab for a School user's home page listing mentors

### DIFF
--- a/app/services/schools/home.rb
+++ b/app/services/schools/home.rb
@@ -12,5 +12,11 @@ module Schools
         .eager_load(:teacher, :school, mentors: :teacher)
         .merge(MentorshipPeriod.ongoing)
     end
+
+    def mentors_with_ects
+      MentorAtSchoolPeriod
+        .where(school:)
+        .eager_load(:teacher, :school)
+    end
   end
 end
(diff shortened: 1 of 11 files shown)


app/services/schools/home.rb
```diff
@@ -12,5 +12,11 @@ def ects_with_mentors
         .eager_load(:teacher, :school, mentors: :teacher)
         .merge(MentorshipPeriod.ongoing)
     end
+
+    def mentors_with_ects
```
tonyheadford, line 16: Why is this `mentors_with_ects` when that isn't the query - is this just a list of mentors at the school?

peterdavidhamilton: My understanding was that the cards we are iterating over will need their ECTs (mentees) included, but that is in the next ticket and out of scope right now. I was setting out the intention.

tonyheadford: OK, not all mentors will have ECTs all the time, is the intention not to show those do you know?

tonyheadford (approved):

app/services/schools/home.rb
peterdavidhamilton: There should be tickets to consider what to do if a mentor has no mentees or has more than could comfortably be shown. The initial design used an unordered list but I was imagining another status indicator too.

PR #236, opened 2025-02-19
Rename`legacy_id` type fields more appropriately

peteryates:
The name `legacy_id` isn't much help, `dqt_id` at least hints at the source of the value.

Refs DFE-Digital/register-ects-project-board#1199

**Note**, this follows on from #240 and should be rebased once it's merged. No point in updating all the importer scripts with this change only to delete them.

This now includes renaming `legacy_` fields used for ECF migration into much nicer things


leoapost (approved):
Looks good

PR #238, opened 2025-02-19
Remove redundant require from spec files

leoapost:
The rails_helper is configured in the .rspec to be loaded by default.


peteryates (approved):
Thanks Leo!

PR #239, opened 2025-02-19
make login error-checking more robust

ltello:


peteryates:
This looks much better.

Think it'd make sense to just convert the `users.email` field to [citext](https://edgeapi.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/PostgreSQLAdapter.html#constant-NATIVE_DATABASE_TYPES)?

peteryates (commented):
I think the robustness side of this PR is great but I'm less sure about auto-converting values to lowercase.

It feels confusing that we have an additional type we use for one `email` column but not others (perhaps this could be helped by naming it after what it does) and that we're introducing db check constraints to ensure conversions are happening.

I think this might just be clearer by setting it to this and leaving the search as it originally was:

```ruby
create_table "users" ... do |t|
  # ...
   t.citext "email", null: false
  # ...
end
```

ltello:
> I think the robustness side of this PR is great but I'm less sure about auto-converting values to lowercase.
> 
> It feels confusing that we have an additional type we use for one `email` column but not others (perhaps this could be helped by naming it after what it does) and that we're introducing db check constraints to ensure conversions are happening.
> 
> I think this might just be clearer by setting it to this and leaving the search as it originally was:
> 
> ```ruby
> create_table "users" ... do |t|
>   # ...
>    t.citext "email", null: false
>   # ...
> end
> ```

I think you are right and a `citext` field would go better than downcasing emails. 
And yes, for consistency I think we should apply it to all email fields we have.
I'll modify this PR.

peteryates (approved):
Looks great, definitely robust and easier to follow.

PR #242, opened 2025-02-20
Simplify `Teachers::Search`

peteryates:
This change reworks the teachers search class and makes it more generic, and removes its need to have knowledge about where it's being called for.

It also makes the critiera combinable, so we'll be able to support searching within an AB's ECTs by providing both a search term and an appropriate body.

It also will allow us to search by multiple appropriate bodies if needed.


spec/services/teachers/search_spec.rb
```diff
@@ -1,200 +1,107 @@
...
-        it 'initiates a full text search with the given search string' do
-          teacher = FactoryBot.create(:teacher, trs_first_name: 'Captain', trs_last_name: 'Scrummy')
-          FactoryBot.create(
-            :induction_period,
-            :active,
-            teacher:,
-            started_on: 1.month.ago.to_date,
-            induction_programme: 'fip'
-          )
+        let(:teacher1) { FactoryBot.create(:teacher, trs_first_name: 'Captain', trs_last_name: 'Scrummy') }
```
peterdavidhamilton, line 70: 😂

peterdavidhamilton (approved):

spec/services/teachers/search_spec.rb
peteryates: image

PR #244, opened 2025-02-20
Add before and after scopes to interval

peteryates:
We'll undoubtably need to filter all kinds of periods by whether they started before, on or after a certain date. This change adds some common scopes that means we can do it consistently without having to use snippets of SQL.

- **Add started before/on or after scopes to interval**
- **Use started_before scope instead of where clause**

### DIFF
--- a/app/models/concerns/interval.rb
+++ b/app/models/concerns/interval.rb
@@ -12,7 +12,10 @@ module Interval
     scope :finished, -> { where.not(finished_on: nil) }
     scope :earliest_first, -> { order(started_on: 'asc') }
     scope :latest_first, -> { order(started_on: 'desc') }
-
+    scope :started_before, ->(date) { where(arel_table[:started_on].lt(date)) }
+    scope :started_on_or_after, ->(date) { where(arel_table[:started_on].gteq(date)) }
+    scope :finished_before, ->(date) { where(arel_table[:finished_on].lt(date)) }
+    scope :finished_on_or_after, ->(date) { where(arel_table[:finished_on].gteq(date)) }
     scope :containing_period, ->(period) { where("range @> daterange(?, ?)", period.started_on, period.finished_on) }
   end
 
--- a/app/models/concerns/interval.rb
+++ b/app/models/concerns/interval.rb
@@ -12,10 +12,10 @@ module Interval
     scope :finished, -> { where.not(finished_on: nil) }
     scope :earliest_first, -> { order(started_on: 'asc') }
     scope :latest_first, -> { order(started_on: 'desc') }
-    scope :started_before, ->(date) { where(arel_table[:started_on].lt(date)) }
-    scope :started_on_or_after, ->(date) { where(arel_table[:started_on].gteq(date)) }
-    scope :finished_before, ->(date) { where(arel_table[:finished_on].lt(date)) }
-    scope :finished_on_or_after, ->(date) { where(arel_table[:finished_on].gteq(date)) }
+    scope :started_before, ->(date) { where(started_on: ...date) }
+    scope :started_on_or_after, ->(date) { where(started_on: date..) }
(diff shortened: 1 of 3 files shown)


app/services/admin/update_induction_period_service.rb
```diff
@@ -32,18 +30,19 @@ def validate_can_update!
...
 
-    def earliest_period?
-      !InductionPeriod.where(teacher:)
-        .where("started_on < ?", induction_period.started_on)
+    def ect_has_earlier_induction_periods?
+      InductionPeriod
+        .siblings_of(induction_period)
+        .started_before(induction_period.started_on)
         .exists?
     end
```
peteryates, line 47: @edujackedu - I refactored this to make it easier to follow - thoughts?

app/models/concerns/interval.rb
```diff
@@ -12,7 +12,10 @@ module Interval
     scope :finished, -> { where.not(finished_on: nil) }
     scope :earliest_first, -> { order(started_on: 'asc') }
     scope :latest_first, -> { order(started_on: 'desc') }
-
+    scope :started_before, ->(date) { where(arel_table[:started_on].lt(date)) }
+    scope :started_on_or_after, ->(date) { where(arel_table[:started_on].gteq(date)) }
+    scope :finished_before, ->(date) { where(arel_table[:finished_on].lt(date)) }
+    scope :finished_on_or_after, ->(date) { where(arel_table[:finished_on].gteq(date)) }
```
ltello, line 18: Isn't it more readable using endless ranges?

```suggestion
    scope :started_before, ->(date) { where(started_on: ...date) }
    scope :started_on_or_after, ->(date) { where(started_on: date..) }
    scope :finished_before, ->(date) { where(finished_on: ...date) }
    scope :finished_on_or_after, ->(date) { where(finished_on: date..) }
```

peteryates: Readable yes, but I'm not sure it's more _understandable_. `lt` and `gteq` are easier to understand in my opinion than `..` vs `...` - but the endless ranges are more idiomatic Rails so let's go with them.

peterdavidhamilton (approved):

app/models/concerns/interval.rb
```diff
@@ -12,7 +12,10 @@ module Interval
     scope :finished, -> { where.not(finished_on: nil) }
     scope :earliest_first, -> { order(started_on: 'asc') }
     scope :latest_first, -> { order(started_on: 'desc') }
-
+    scope :started_before, ->(date) { where(started_on: ...date) }
+    scope :started_on_or_after, ->(date) { where(started_on: date..) }
+    scope :finished_before, ->(date) { where(finished_on: ...date) }
+    scope :finished_on_or_after, ->(date) { where(finished_on: date..) }
     scope :containing_period, ->(period) { where("range @> daterange(?, ?)", period.started_on, period.finished_on) }
   end
```
ltello, line 20: would it make sense to extend this concern with this:
```
  scope :siblings_of, ->(instance) {  raise ImplementationError }

  def sibblings = self.class.sibblings_of(self)
  def predecessors = sibblings.started_before(started_on)
  def successors = sibblings.started_on_or_after(finished_on)

  def sibblings? = sibblings.exists?
  def predecessors? = predecessors.exists?
  def successors? = successors.exists?
  
```

so that each `*Period` class needs to implement:
   ```scope :siblings_of, ->(instance) { ... }```

and therefore each `*Period` instance can give quick access to their sibblings, predecessors and successors instances?

peteryates: It sounds like a good idea but I think it's worth doing in a separate PR because the sibling behaviour isn't always obvious. In a mentorship, for example, is it a mentorship from the ECT's point of view, or the mentor's? We have `mentee_siblings` there at the moment. Similarly training periods are also more complex.

ltello (approved):

PR #247, opened 2025-02-21
Implement the change answers in the register ect journey

leoapost:
This PR introduces change links on the Check Answers page for the Register ECT journey.

Changes:
- Added routes for the change pages
- Implemented wizard steps for the change pages
- Created reusable partials and view files for the change pages
- Added change links on the Check Answers page
- Prepopulated step attributes from the store to retain and display user-entered values
- Moved wizard step specs to a folder matching the implementation paths
- Added specs to test step attribute prepopulation
- Introduced happy path specs to verify change answer functionality
- Reused existing view specs in the change steps
- Fixed the previous step of `Schools::RegisterECTWizard::IndependentSchoolAppropriateBodyStep`

This ensures a smoother user experience and better test coverage for the Register ECT journey.

Screenshot:
screenshot


app/views/schools/register_ect_wizard/change_email_address.html.erb
```diff
@@ -0,0 +1,2 @@
+<%= render 'email_address',
+            backlink_href: schools_register_ect_wizard_check_answers_path %>
```
ltello, line 2: If you define `previous_step` in the step class `Schools::RegisterECTWizard::ChangeEmailAddressStep` then this will become `backlink_href: @wizard.previous_step_path` (exactly the same as `app/views/schools/register_ect_wizard/change_email_address.html.erb` so `backlink_href` can be moved inside the partial and the 2 view files just be:
`<%= render 'email_address' %>

config/routes.rb
```diff
@@ -111,27 +113,41 @@
...
+      post "change-state-school-appropriate-body", action: :create
 
       get "independent-school-appropriate-body", action: :new
       post "independent-school-appropriate-body", action: :create
+      get "change-independent-school-appropriate-body", action: :new
+      post "change-independent-school-appropriate-body", action: :create
 
       get "programme-type", action: :new
       post "programme-type", action: :create
+      get "change-programme-type", action: :new
```
peterdavidhamilton, line 141: My PR add PT and LP change links, so there is a duplicate here and the LP ones could be reordered to follow your pattern.

leoapost: Good catch! I didn't encounter any conflicts in the routes file and didn't notice the duplicates when I rebased my branch. I pushed another commit to remove them.

app/wizards/schools/register_ect_wizard/wizard.rb
```diff
@@ -10,8 +10,14 @@ class Wizard < DfE::Wizard::Base
           {
             already_active_at_school: AlreadyActiveAtSchoolStep,
             cannot_register_ect: CannotRegisterECTStep,
+            change_email_address: ChangeEmailAddressStep,
+            change_independent_school_appropriate_body: ChangeIndependentSchoolAppropriateBodyStep,
             change_lead_provider: ChangeLeadProviderStep,
             change_programme_type: ChangeProgrammeTypeStep,
+            change_review_ect_details: ChangeReviewECTDetailsStep,
```
ltello, line 17: Maybe `change_ect_details` makes more sense?

app/views/schools/register_ect_wizard/change_email_address.html.erb
leoapost: Initially, I was using `previous_step` methods, but I noticed that the register mentor files use path helpers instead. So, I updated it to follow the same approach for consistency.

spec/views/schools/register_ect_wizard/shared_examples/independent_school_appropriate_body_view.rb
```diff
@@ -0,0 +1,46 @@
...
+
+    it 'renders an error summary' do
+      expect(view.content_for(:error_summary)).to have_css('.govuk-error-summary')
+    end
+  end
+
+  it "includes a back button that targets #{back_step_name} page" do
+    render
+
+    expect(view.content_for(:backlink_or_breadcrumb)).to have_link('Back', href: send(back_path))
```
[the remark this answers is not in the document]
leoapost: The `back_path` includes a symbol representing a route path helper method, and `send` is used to retrieve the corresponding URL path.

It’s a pattern already used in the mentor specs. Here's an example where a [spec](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/spec/views/schools/register_mentor_wizard/change_email_address.html.erb_spec.rb#L6) includes the [shared example](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/spec/views/schools/register_mentor_wizard/shared_examples/email_view.rb#L45) that tests the title

app/wizards/schools/register_ect_wizard/wizard.rb
leoapost: Indeed it reads better, but all `change` steps follow the existing step names, prefixed with "change_", ensuring a consistent and predictable pattern.

For example:
- Current step: `review_ect_details: ReviewECTDetailsStep`
- Equivalent change step: `change_review_ect_details: ChangeReviewECTDetailsStep`

I don't have a strong opinion on this, so I'm happy to change it if we still want it.

spec/views/schools/register_ect_wizard/shared_examples/state_school_appropriate_body_view.rb
```diff
@@ -0,0 +1,46 @@
...
+  let(:appropriate_body_name) { nil }
+  let(:store) { FactoryBot.build(:session_repository, appropriate_body_name:, full_name: 'John Smith') }
+  let(:wizard) { FactoryBot.build(:register_ect_wizard, current_step:, store:) }
+
+  before do
+    assign(:wizard, wizard)
+    assign(:ect, ect)
+  end
+
+  it "sets the page title to 'Which appropriate body will be supporting John Smith's induction?'" do
```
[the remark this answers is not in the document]
leoapost: It seems we predominantly use `it "sets the page title to something" do` for title specs in our spec/view files, with a few instances of `describe 'heading and title' do`.

We could review our approach and discuss whether to standardize on describe or it. However, for this PR, I think it would be better to maintain the original format of the specs.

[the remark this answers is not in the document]
leoapost: I forgot to mention that I pushed a commit to update `eql(sanitize(title))` to `eql('string')` 👍

peterdavidhamilton (approved):

app/wizards/schools/register_ect_wizard/wizard.rb
ltello: All of them but this one in the mentor journey.

app/views/schools/register_ect_wizard/change_email_address.html.erb
ltello: I'm going to create a PR to do it in the mentor journey. 
If you merge this one as it is I can address it as well as part of that PR.

ltello (commented):
I've left a couple of new comments.
I think that for consistency with the mentor journey it would be good to rename the `ChangeReviewECTDetails` step. 
Or we can as well do it in a separate PR.

PR #250, opened 2025-02-21
Record events for admin induction period updates

peteryates:
When admins make changes to induction periods we want to record an event that tells us what changed, and later, why.

We will do this by triggering an `Events::Record` method from the service class that does the update.

We'll have lots of this style of event in the future so we spent some time ensuring it's easy, and we don't end up with too much _change describing_ code littered around.

## Changes:

* there is a new column, `modifications`, on the `events` table, it's a `varchar[]` column and is intended to contain a human readable list of changes we can display in a `<ul>`
* there's a new class called `Events::DescribeModifications` which takes a set of changes from `ActiveRecord#changes` and convers them to the human readable list. We're using the word modifications instead of changes because changes is reserved by Rails and yield 'dangers name' errors
* the `UpdateInductionPeriodService` now takes an `author` argument and passes it, along with the modifications, to the event recording

## Preview

```
ecf2_development=# select * from events order by id desc limit 1;
┌─[ RECORD 1 ]───────────────┬──────────────────────────────────────────────────────┐
│ id                         │ 1                                                    │
│ heading                    │ Induction period updated by admin                    │
│ body                       │ (null)                                               │
│ event_type                 │ admin_updates_induction_period                       │
│ happened_at                │ 2025-02-21 15:38:45.819581                           │
│ teacher_id                 │ 10                                                   │
│ appropriate_body_id        │ 4                                                    │
│ induction_period_id        │ 9                                                    │
│ induction_extension_id     │ (null)                                               │
│ school_id                  │ (null)                                               │
│ ect_at_school_period_id    │ (null)                                               │
│ mentor_at_school_period_id │ (null)                                               │
│ training_period_id         │ (null)                                               │
│ mentorship_period_id       │ (null)                                               │
│ provider_partnership_id    │ (null)                                               │
│ lead_provider_id           │ (null)                                               │
│ delivery_partner_id        │ (null)                                               │
│ user_id                    │ (null)                                               │
│ author_type                │ dfe_staff_user                                       │
│ author_id                  │ 1                                                    │
│ author_name                │ Daphne Blake                                         │
│ author_email               │ daphne@example.com                                   │
│ created_at                 │ 2025-02-21 15:38:46.118555                           │
│ updated_at                 │ 2025-02-21 15:38:46.118555                           │
│ metadata                   │ {"started_on": ["2023-06-21", "2023-05-21"]}         │
│ modifications              │ {"Started on changed from 2023-06-21 to 2023-05-21"} │
└────────────────────────────┴──────────────────────────────────────────────────────┘
```

## Notes

I left `body` clear as I think it makes sense to use that instead of introducing a new 'reason' field. The raw changes are left in `metadata` too, which could be handy later.

## Final tasks

* [ ] find and manually add events for the handful of edits we already have in prod

### DIFF
--- a/app/controllers/admin/induction_periods_controller.rb
+++ b/app/controllers/admin/induction_periods_controller.rb
@@ -8,7 +8,8 @@ module Admin
       @induction_period = InductionPeriod.find(params[:id])
       service = UpdateInductionPeriodService.new(
         induction_period: @induction_period,
-        params: induction_period_params
+        params: induction_period_params,
+        author: current_user
       )
 
       if service.update_induction!
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -1,5 +1,6 @@
 class Event < ApplicationRecord
   EVENT_TYPES = %w[
+    admin_updates_induction_period
     appropriate_body_claims_teacher
     appropriate_body_releases_teacher
     appropriate_body_fails_teacher
--- a/app/services/admin/update_induction_period_service.rb
+++ b/app/services/admin/update_induction_period_service.rb
@@ -2,20 +2,24 @@ module Admin
   class UpdateInductionPeriodService
     class RecordedOutcomeError < StandardError; end
 
-    attr_reader :induction_period, :params
+    attr_reader :induction_period, :params, :author
 
-    def initialize(induction_period:, params:)
+    def initialize(induction_period:, params:, author:)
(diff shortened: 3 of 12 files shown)


tonyheadford (approved):
Nice!

PriyaNahar2 (approved):
Looks good to me! @peteryates

PR #251, opened 2025-02-24
Allow admins to add extra context on why they're making changes

peteryates:
One of the weak areas of ECF1 was the inability to see a record of things that have happened previously to a teacher within the system. Being able to do it is vital for support teams so they can work out why records are in a particular state.

Since #250 we are recording events (DFE-Digital/register-ects-project-board#1013) when admins update induction periods.

For more information on events and how they work, see [this issue on NPQ](https://github.com/DFE-Digital/register-ects-project-board/issues/895).

## Adding extra information to the event

Events tell us what happened and who did it.

They don't tell us **why** it happened.

We have an opportunity to record a bit more info by adding an extra field (or two) to the edit forms:

* Zendesk ticket URL
* Description

## How might this information be displayed?

The original intention was to use [a timeline](https://design-patterns.service.justice.gov.uk/components/timeline/) showing the events that have happened to a record.

image

## How the form might look

This is an initial stab at adding new fields to the edit induction period form.

| New fields | Whole form |
| ----------- | ------------- |
| Screenshot From 2025-02-24 11-15-29 |Screenshot From 2025-02-24 10-43-31 |

## Extra tasks

* [ ] Think about the final design. Do the new form feels look mandatory? Should they be? Should they be sectioned off from the _main bit_ of the form?
* [ ] Which fields do we want? Can we get away with just a Zendesk URL and no place to add a reason for making the change? Should we capture the whole Zendesk URL or is the ticket ID enough?
* [ ] For devs, how do we make this easy to add to all the admin forms?

### DIFF
--- a/app/controllers/admin/induction_periods_controller.rb
+++ b/app/controllers/admin/induction_periods_controller.rb
@@ -1,5 +1,7 @@
 module Admin
   class InductionPeriodsController < AdminController
+    include EditableByAdmin
+
     def edit
       @induction_period = InductionPeriod.find(params[:id])
     end
@@ -9,7 +11,8 @@ module Admin
       service = UpdateInductionPeriodService.new(
         induction_period: @induction_period,
         params: induction_period_params,
-        author: current_user
+        author: current_user,
+        editable_by_admin_params: editable_by_admin_params(params.require(:induction_period))
       )
 
       if service.update_induction!
--- a/app/controllers/concerns/editable_by_admin.rb
+++ b/app/controllers/concerns/editable_by_admin.rb
@@ -1,0 +1,9 @@
+module EditableByAdmin
+  extend ActiveSupport::Concern
+
+  included do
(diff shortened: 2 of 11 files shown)


claire-hughez:
I think having a 'why' is sensible, particularly for situations when there isn't a Zendesk ticket URL. Would we want any validation? e.g. If Zendesk ticket URL is given, no need to give reason, otherwise reason is needed? Probably unnecessarily complicated, but I guess there's a danger of getting loads of changes at some point with no reasoning.

I think having it on the same page is helpful to have context of what they're changing tbh.

The designs for both look sensible to me, but I still want to meet for us to collaborate on the admin console as a team properly.

peterdavidhamilton:
@peteryates we discussed the potential for a `reason` on an `induction_extension` or a foreign key to link it to the induction itself. 
🎗️

peteryates:
Closing for now, can revisit later if we want this.

PR #252, opened 2025-02-24
created helper and added content for non prohibited alerts

PriyaNahar2:
Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/1225
Added some context and next steps for when non prohibited alerts are present. Kept content and link similar to prohibited alert page (to use CTR service). 

Interim solution design / content wise as it's a pretty rare occurrence so we should revisit later down the line

Can't show before screenshot as no TRNs with alerts in staging environment but it just says 'Yes' with no other context currently.
Screenshots: 
| After |
| ------ |
| Screenshot 2025-02-24 at 16 10 30 |

### DIFF
--- a/app/helpers/appropriate_body_helper.rb
+++ b/app/helpers/appropriate_body_helper.rb
@@ -56,6 +56,16 @@ module AppropriateBodyHelper
     "#{number_with_delimiter(count)} claimed #{'induction'.pluralize(count)}"
   end
 
+  def trs_alerts_text(alerts_present)
+    if alerts_present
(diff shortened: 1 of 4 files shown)


peteryates (approved):

kas-madrigal:
A TRN and date of birth of an ECT that has an alert which is not a prohibition (so we can test this isn't blocked) 

1000787 | 1984-04-01

emily-prudence-dfe (approved):
@PriyaNahar2 looks good, merge away!

PR #253, opened 2025-02-24
Skip migrating record if FIP without partnership

tonyheadford:
When extracting training periods from induction records, there are a lot of errors generated when there is no lead provider found.  This occurs when the `InductionProgramme` associated with the `InductionRecord` being processed does not have a `Partnership` assigned.

As we cannot add a `TrainingPeriod` without a valid provider, and there is no data available in ECFv1, we cannot add these so we should skip them.

### DIFF
--- a/app/migration/training_period_extractor.rb
+++ b/app/migration/training_period_extractor.rb
@@ -29,8 +29,7 @@ private
         # FIXME: recording a FIP programme without a partnership means that we cannot
         # add a training_period for it, so it doesn't make sense to include these but
         # is that the correct approach?
-        next if record_programme.training_programme == "full_induction_programme"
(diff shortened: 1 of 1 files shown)


tonyheadford:
> Is something logging the records we are skipping and why? I know this change was made when errors were reported but are we gathering metrics anywhere else?

No we're not logging things like that only failures. I'm not sure we'd need anything like that - the end goal will be that we'll want to check that what we've represented in RECT matches the data in ECFv1 when exposed by the API - we can use the parity checking approach to compare the results from both APIs (once we've built ours)

We will then need to make sure they align!

tonyheadford:
Closing because review app issue

tonyheadford:
Closing because of review app creation issues

PR #262, opened 2025-02-26
remove production from the list of restorable databases and also from the restoring job

ltello:
To prevent accidentally restore the main database in the production environment of the service:
- Remove `production` from the list of restorable databases in the `environment` input field.
- Also only allow non production databases to be restored by the the restoring job

### DIFF
--- a/.github/workflows/restore-app-main-db.yml
+++ b/.github/workflows/restore-app-main-db.yml
@@ -13,7 +13,7 @@ on:
         - sandbox
         - staging
         - test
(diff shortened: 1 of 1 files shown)


peteryates (approved):

peteryates:
Thanks @ltello. I'll sleep easier now 🙂

ltello:
> Thanks @ltello. I'll sleep easier now 🙂

Sure!. No more pills! haha

PR #264, opened 2025-02-27
Extend Interval to cope with predecessors and successors siblings of an interval

ltello:
- replace scopes of type `siblings_of(instance)` with instance methods `instance#siblings`
- throw `NotImplementedError` exception on `Interval#siblings` method to indicate that models including the `Interval` module should overide this method.
- define a set of instance methods in `Interval` to deal with predecessors and successors of a given `*Period` instance
- override `siblings` in `ECTAtSchoolPeriod` to define it based on the `ect_at_school_periods` of the `teacher` of an instance.
- same for `InductionPeriod` model.
- for `MentorAtSchoolPeriod` mode, overide it based on the `teacher` and `school` of the given instance.
- `MentorshipPeriod`: leave `#mentee_siblings` for now. Maybe in the future we should renamed it to just `siblings` and benefit from the new methods defined on `Interval` should the necessity arise.
- `TrainingPeriod`: merge some validations for ect or mentor into unique one based on trainee instead: currently for mentors the same validation was running twice. Like `MentorshipPeriod` the current `training_siblings` method on `TrainingPeriod` could be renamed to just `siblings` based on future necessities.

### DIFF
--- a/app/models/induction_period.rb
+++ b/app/models/induction_period.rb
@@ -22,14 +22,20 @@ class InductionPeriod < ApplicationRecord
 
   scope :for_teacher, ->(teacher) { where(teacher:) }
   scope :for_appropriate_body, ->(appropriate_body) { where(appropriate_body:) }
-  scope :siblings_of, ->(instance) { for_teacher(instance.teacher).excluding(instance) }
+
+  # Instance methods
+  def siblings
+    return InductionPeriod.none unless teacher
+
+    teacher.induction_periods.excluding(self)
+  end
 
 private
 
-  def valid_date_order?
-    return true if started_on.blank? || finished_on.blank?
+  def start_date_after_qts_date
+    return if teacher.blank?
 
-    started_on <= finished_on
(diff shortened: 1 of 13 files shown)


app/models/induction_period.rb
```diff
@@ -22,14 +22,20 @@ class InductionPeriod < ApplicationRecord
 
   scope :for_teacher, ->(teacher) { where(teacher:) }
   scope :for_appropriate_body, ->(appropriate_body) { where(appropriate_body:) }
-  scope :siblings_of, ->(instance) { for_teacher(instance.teacher).excluding(instance) }
+
+  # Instance methods
+  def siblings
+    return self.class.none unless teacher
```
peteryates, line 28: If it's not in a shared place is it clearer to return `InductionPeriod.none` here when there's no `teacher`?

peteryates (approved):
This is really neat. Left a super minor observation on using the class name directly to make it clearer, but a big improvement.

PR #271, opened 2025-02-27
Ensure only available email addresses can be used when registering an ECT

leoapost:
This PR introduces a new step in the Register an ECT journey to prevent the use of email addresses already associated with an ongoing school period.

Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/964

### Changes:
- Adds the `Schools::TeacherEmail` service class to check if a given email is used by any other teacher in their ongoing school period
- Introduces a new page that informs schools when an email is unavailable and prompts them to provide a different one
- Refactors the equivalent Mentor decorator to re-use the `Schools::TeacherEmail`

### Screenshot:
screenshot

### DIFF
--- a/spec/services/schools/teacher_email_spec.rb
+++ b/spec/services/schools/teacher_email_spec.rb
@@ -37,7 +37,7 @@ describe Schools::TeacherEmail do
     context "when the email has been used by the same teacher in the past" do
       let(:teacher) { ongoing_ect_at_school_period.teacher }
       let(:trn) { teacher.trn }
-      let(:email) { FactoryBot.create(:ect_at_school_period, :finished, teacher:).email }
+      let(:email) { FactoryBot.create(:ect_at_school_period, teacher:).email }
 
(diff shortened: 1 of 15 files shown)


spec/services/schools/teacher_email_spec.rb
```diff
@@ -0,0 +1,39 @@
+describe Schools::TeacherEmail do
+  let(:email) { "test@example.com" }
+  let(:trn) { "123456" }
+  subject { described_class.new(email:, trn:) }
+
+  describe "#is_currently_used?" do
```
peterdavidhamilton, line 6: Does this service spec give us sufficient confidence? It is just testing the mocked values without looking at active periods and teachers.

leoapost: @peterdavidhamilton  I've pushed a commit changing them to integration tests.

peterdavidhamilton (approved):

PR #279, opened 2025-02-27
1293: Make record view consistent and reusable across View an ECT record and Claim an ECT

ghost:
Have a reusable parent Teachers::Details view component so that we can share it across several views
https://github.com/DFE-Digital/register-ects-project-board/issues/1293

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -62,3 +62,6 @@ Lint/MissingSuper:
 
 Style/HashSyntax:
   EnforcedShorthandSyntax: always
+
+Style/Lambda:
+  EnforcedStyle: literal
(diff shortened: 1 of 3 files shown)


app/views/appropriate_bodies/claim_an_ect/register_ect/show.html.erb
```diff
@@ -4,10 +4,10 @@
 %>
 
 
-<%= govuk_panel(title_text: "You’ve successfully claimed #{PendingInductionSubmissions::Name.new(@pending_induction_submission).full_name}’s induction") %>
+<%= govuk_panel(title_text: "You've successfully claimed #{PendingInductionSubmissions::Name.new(@pending_induction_submission).full_name}'s induction") %>
```
peteryates, line 7: We should probably standardise on either curly or straight quotes, we have a mix of both. Maybe we can use a helper method or something for it.

emily-prudence-dfe:
Looks good @edujackedu !

PR #281, opened 2025-02-28
ECT show template redesign surfacing data

peterdavidhamilton:
- refactor constants to be reusable hashes
- update ECT show template design
- add extra teacher and AB helper methods
- update seed data and factories
- associate training periods to lead providers and delivery partners

### DIFF
--- a/app/helpers/teacher_helper.rb
+++ b/app/helpers/teacher_helper.rb
@@ -9,6 +9,21 @@ module TeacherHelper
     "TRN: #{teacher.trn}"
   end
 
+  # @param teacher [Teacher]
+  def teacher_induction_start_date(teacher)
+    Teachers::InductionPeriod.new(teacher).formatted_induction_start_date
+  end
+
+  # @param teacher [Teacher]
+  def teacher_induction_programme(teacher)
+    Teachers::InductionPeriod.new(teacher).induction_programme
+  end
+
+  # @param teacher [Teacher]
+  def teacher_induction_ab_name(teacher)
+    Teachers::InductionPeriod.new(teacher).appropriate_body_name
+  end
+
   def teacher_date_of_birth_hint_text
     "For example, 20 4 2001"
   end
--- a/app/services/teachers/induction_period.rb
+++ b/app/services/teachers/induction_period.rb
@@ -9,6 +9,22 @@ class Teachers::InductionPeriod
     first_induction_period&.started_on
   end
 
+  def formatted_induction_start_date
+    induction_start_date&.to_fs(:govuk)
+  end
+
+  def induction_programme
+    return unless first_induction_period
+
+    ::INDUCTION_PROGRAMMES[first_induction_period.induction_programme.to_sym]
+  end
+
+  def appropriate_body_name
+    return unless first_induction_period
+
+    first_induction_period.appropriate_body.name
+  end
+
   def active_induction_period
     # FIXME: this works if finished_on cannot be set to a future date
     # If that becomes possible, this query will need to be updated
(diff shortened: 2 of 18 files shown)


app/services/teachers/induction_period.rb
```diff
@@ -9,6 +9,22 @@ def induction_start_date
...
+  def induction_programme
+    return unless first_induction_period
+
+    ::INDUCTION_PROGRAMMES[first_induction_period.induction_programme.to_sym]
+  end
+
+  def appropriate_body_name
+    return unless first_induction_period
+
+    first_induction_period.appropriate_body.name
```
ltello, line 25: Are we getting the first_induction_period based on `started_on` order?
Shouldn't it be the last instead?

app/helpers/teacher_helper.rb
```diff
@@ -9,6 +9,21 @@ def teacher_trn(teacher)
...
+
+  # @param teacher [Teacher]
+  def teacher_induction_programme(teacher)
+    Teachers::InductionPeriod.new(teacher).induction_programme
+  end
+
+  # @param teacher [Teacher]
+  def teacher_induction_ab_name(teacher)
+    Teachers::InductionPeriod.new(teacher).appropriate_body_name
+  end
```
ltello, line 25: should we add `reported_by_ab` to these names to prevent confusion with those reported by the school?

app/views/schools/ects/show.html.erb
```diff
@@ -39,4 +29,43 @@
       ]
     )
   end
+%>
+<%=
+  if @ect.provider_led?
+    govuk_summary_card(title: 'Reported to us by your lead provider') do |card|
+      if (period = @ect.training_periods.last)
```
ltello, line 36: is the right order guaranteed here?

spec/factories/ect_at_school_period_factory.rb
```diff
@@ -4,10 +4,14 @@
   factory(:ect_at_school_period) do
     association :school
     association :teacher
+    association :lead_provider
+    association :appropriate_body
```
ltello, line 8: `lead_provider` and `appropriate_body` are not required for an ECTAtSchoolPeriod, so I wouldn't create them by default but only explicitly in the spec that needs it.

app/helpers/teacher_helper.rb
peterdavidhamilton: I think we will need something, not sure what's best at the moment; thinking it'll be easier when we have all of it populated and in use. Some naming convention, or perhaps a single point of entry where you can ask for a specific "reporter" could be useful.

peterdavidhamilton:
@mason-emily @claire-hughez this is deployed and ready for you once more

app/views/schools/ects/show.html.erb
```diff
@@ -6,41 +6,66 @@
...
-    ]
-  )
-end %>
+%>
+<h2 class='govuk-heading-m'>ECTE training details</h2>
+<%=
+  govuk_summary_card(title: 'Reported to us by your school') do |card|
+    card.with_summary_list(
+      rows: [
+        { key: { text: 'Appropriate body' }, value: { text: @ect.appropriate_body&.name } },
```
peteryates, line 26: I missed the `appropriate_body_id` column being added to the `ECTAtSchoolPeriod` model, but for clarity I think this should be more descriptive. @claire-hughez said @ltello suggested `school_recommended_appropriate_body_id`, which I think would fit the bill.

peterdavidhamilton: To ensure clarity at the db-level, is that because there will also be a corresponding `official_appropriate_body_id` (or some other prefix) on `ECTAtSchoolPeriod`?

And is this the pattern we must follow for `lead_provider_id` too?

There is also a school requested/recommended `programme_type`.

Do we need that in this PR shall we do that separately?

peterdavidhamilton:
image

PR #282, opened 2025-02-28
Set the sentry environment from the env var

peteryates:
It's currently unset and appears to default to `production` in Sentry, so all errors from review apps, staging, sandbox etc look like production ones.


peteryates:
Closing. Let's use the `SENTRY_ENVIRONMENT` var instead.

PR #285, opened 2025-03-03
Display the ECT’s TRS full name by default on the details review page

leoapost:
Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/1303

This PR fixes a bug on the ECT's review details page where the TRS full name gets replaced when the name is corrected.

### DIFF
--- a/app/wizards/schools/register_ect_wizard/ect.rb
+++ b/app/wizards/schools/register_ect_wizard/ect.rb
@@ -7,7 +7,7 @@ module Schools
       end
 
       def full_name
-        corrected_name.presence || [trs_first_name, trs_last_name].join(" ").strip
+        (corrected_name.presence || trs_full_name).strip
       end
 
       def govuk_date_of_birth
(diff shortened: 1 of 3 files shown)


app/wizards/schools/register_ect_wizard/ect.rb
```diff
@@ -89,6 +89,10 @@ def appropriate_body
       def lead_provider
         LeadProvider.find(lead_provider_id) if provider_led?
       end
+
+      def trs_full_name
+        [trs_first_name, trs_last_name].join(" ").strip
+      end
```
tonyheadford, line 95: Could we use the `Teachers::Name` service here? I know we don't have a `Teacher` at this point but the attributes have the same name so we could pass the `ECT` instance as the `teacher:` param perhaps?
Just thinking it would save duplicating building the full name?

leoapost: @tonyheadford thanks for the pointer. Wasn't aware of that service class.

tonyheadford (approved):
LGTM!

app/wizards/schools/register_ect_wizard/ect.rb
tonyheadford: There is a `full_name_in_trs` method on that service that doesn't use corrected_name @ltello

PR #287, opened 2025-03-03
Update onboarding template

avinhurry:
### Context 

- Updating the slack channels to be more specific

- Removing JIRA from the list as this is no longer required


peteryates (approved):
Makes sense.

PR #290, opened 2025-03-04
Add validation errors to guard against non-existent dates

peterdavidhamilton:
Validate a pending induction submission date prior to submitting multipart params

### DIFF
--- a/app/controllers/appropriate_bodies/claim_an_ect/register_ect_controller.rb
+++ b/app/controllers/appropriate_bodies/claim_an_ect/register_ect_controller.rb
@@ -5,18 +5,17 @@ module AppropriateBodies
         @pending_induction_submission = find_pending_induction_submission
       end
 
+      # OPTIMIZE: Pre-model validations are required to guard against
+      # ActiveRecord::MultiparameterAssignmentErrors
+      # when dates are non-existent or incorrect dates when one day from the month's end
+      # for example 31st Sept becomes 1st Oct
+      # or when negative numbers are passed in by mistake
       def update
-        register_ect = AppropriateBodies::ClaimAnECT::RegisterECT
-          .new(
-            appropriate_body: @appropriate_body,
-            pending_induction_submission: find_pending_induction_submission,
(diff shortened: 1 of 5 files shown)


app/controllers/appropriate_bodies/claim_an_ect/register_ect_controller.rb
```diff
@@ -5,18 +5,17 @@ def edit
...
       def update
-        register_ect = AppropriateBodies::ClaimAnECT::RegisterECT
-          .new(
-            appropriate_body: @appropriate_body,
-            pending_induction_submission: find_pending_induction_submission,
-            author: current_user
-          )
-
-        if register_ect.register(update_params)
+        if hash_date.valid? && register_ect.register(update_params)
```
ltello, line 14: I think this validation logic should be in the `started_on` field in `PendingInductionSubmission` model instead of here in the controller.

app/controllers/appropriate_bodies/claim_an_ect/register_ect_controller.rb
```diff
@@ -5,18 +5,17 @@ def edit
...
           redirect_to(ab_claim_an_ect_register_path(register_ect.pending_induction_submission))
         else
           @pending_induction_submission = register_ect.pending_induction_submission
+          @pending_induction_submission.errors.add(:started_on, hash_date.error_message) unless hash_date.valid?
```
ltello, line 18: We should not need this if we do what said above

app/controllers/appropriate_bodies/claim_an_ect/register_ect_controller.rb
peterdavidhamilton: It should but it doesn't catch the error that way. It is intentionally pre-validating before the params are submitted.

peterdavidhamilton:
@ltello @leoapost @avinhurry - if any of you want to work on this piece please feel free. @edujackedu has had a quick look.

PR #291, opened 2025-03-04
[1329] Add helper method to render smart quotes using RubyPants

avinhurry:
### Context

**Add helper method to convert 'dumb' quotes to 'smart' quotes using RubyPants.**

This PR introduces a helper method: `smart_quotes`, that automatically converts straight quotes (`"`, `'`) into smart quotes (`“`, `”`, `‘`, `’`). This will help to maintain a consistent gov.uk typography style across the app.

---

### Changes proposed

- Added the RubyPants gem
- Added `smart_quotes` helper method.
- Configured `RubyPants` with custom smart quote options.

---

#### In Views
You can wrap text output with the helper to apply smart quotes automatically:

```erb
<p><%= smart_quotes("He said 'hello'") %></p>
```

outputs: 

image

### Ticket

https://github.com/DFE-Digital/register-ects-project-board/issues/1329

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -42,6 +42,9 @@ gem "rotp"
 gem "rubyzip"
 gem "savon"
 
+# Render smart quotes
+gem 'rubypants'
+
 group :development do
   gem 'better_errors'
   gem 'binding_of_caller'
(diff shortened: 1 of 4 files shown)


tonyheadford (commented):
Code looks great but I'm just not sure of the need for this?  I would've thought that 99% of the time it would be static content that needed a fix rather than dynamically added data but I don't have the context of how this ticket came about?

tonyheadford (approved):
Nice one @avinhurry - I got some context from @peteryates so I understand the purpose now

PR #294, opened 2025-03-05
Interval refactor and tests

peterdavidhamilton:
- rename interval concern methods to be more idiomatic
- leverage shorthand in tests
- apply overlapping validation with distinct date error logic to all periods
- update specs to reuse period helper
- consolidate overlapping validation into concern


app/models/concerns/interval.rb
```diff
@@ -18,31 +18,55 @@ module Interval
     scope :containing_period, ->(period) { where("range @> daterange(?, ?)", period.started_on, period.finished_on) }
   end
 
-  def ongoing?
-    finished_on.nil?
+  # Validations
+  def period_dates_validation
+    return if incomplete?
+
+    errors.add(:finished_on, "The finish date must be later than the start date (#{started_on.to_fs(:govuk)})") if invalid_date_order?
```
peteryates, line 25: I think we might want to check with @PriyaNahar2 and @robthompson16 on the language here, `finish` vs `end`.

I chose to call the fields `finish` in the database because `end` is problematic in Ruby, but I think 'end' probably makes more sense to our users and we should be consistent throughout.

app/models/mentorship_period.rb
```diff
@@ -34,7 +34,7 @@ class MentorshipPeriod < ApplicationRecord
   scope :for_mentor, ->(id) { where(mentor_at_school_period_id: id) }
 
   # Instance methods
-  def mentee_siblings
+  def siblings
```
peteryates, line 37: Slightly mixed thoughts on this, because there are two kinds of siblings here - other `mentorship_periods` that belong to the same Mentor and ones that belong to the same ECT.

peteryates (approved):
Looks good, I think we could do with deciding on the language we use though and maybe adding it to the styleguide (creating one if we don't have one!)

app/models/concerns/interval.rb
PriyaNahar2: I agree and think 'end' is better and more accurate with start/end being the terminology we use so if we can, lets use 'end'?

What do you think @robthompson16?

robthompson16: Sorry only just spotted this @PriyaNahar2 sounds sensible to me!

PR #296, opened 2025-03-05
1161 separate service for updating teacher (batch update version)

ghost:


app/services/appropriate_bodies/claim_an_ect/register_ect.rb
```diff
@@ -19,9 +19,7 @@ def register(pending_induction_submission_params)
         # end
         ActiveRecord::Base.transaction do
           steps = [
-            update_name,
-            update_qts_awarded_on,
-            update_itt_provider_name,
+            update_teacher,
```
peterdavidhamilton, line 22: @peteryates was advocating for a more granular approach, if I remember correctly, with the intent that it could be reused elsewhere. Initial ideas proposed a chaining approach (if we had update methods return `self`).

peterdavidhamilton: I agree, I think it makes sense to have a single input for the params that just knows what to do with them. That's just as easy to reuse IMHO. As you say, that's the bulk upload scenario.

app/services/teachers/manage.rb
```diff
@@ -13,71 +13,116 @@ def initialize(author:, teacher:, appropriate_body:)
...
+
+  # Checks if QTS award date is changing
+  def qts_award_date_changing?(attributes)
+    attributes.key?(:trs_qts_awarded_on)
+  end
+
+  # Checks if the name has changed
   def name_changed?
-    return false if old_name.nil?
+    return false if old_name.nil? || new_name.nil?
```
peterdavidhamilton, line 91: I think you could drop the guard clause now if you wanted because the stringified comparison works with `nil`

app/services/teachers/manage.rb
```diff
@@ -13,71 +13,116 @@ def initialize(author:, teacher:, appropriate_body:)
...
+  end
 
-  attr_reader :new_name, :old_name, :new_award_date, :old_award_date
+  # Records an event for a new teacher record
+  def record_teacher_creation_event
+    Events::Record.teacher_record_created!(
+      author:,
+      teacher:,
+      appropriate_body:,
+      trn: teacher.trn
```
peterdavidhamilton, line 70: Could we avoid passing in `trn` by letting the record `heading` access `teacher.trn`

PR #303, opened 2025-03-06
Admin - Add induction period

peterdavidhamilton:
cpd-ec2-review-303-web test teacherservices cloud_admin_teachers_74_induction-periods

image

### DIFF
--- a/app/components/teachers/details/current_induction_period_component.html.erb
+++ b/app/components/teachers/details/current_induction_period_component.html.erb
@@ -13,7 +13,3 @@
     end
   end
 %>
-
-<p class='govuk-!-text-align-right govuk-!-margin-bottom-0'>
-  <%= govuk_link_to('Add an induction period', new_admin_teacher_induction_period_path(teacher_id: teacher.id)) %>
-</p>
--- a/app/components/teachers/details/current_induction_period_component.html.erb
+++ b/app/components/teachers/details/current_induction_period_component.html.erb
@@ -13,3 +13,7 @@
     end
   end
 %>
+
+<p class='govuk-!-text-align-right govuk-!-margin-bottom-0'>
+  <%= govuk_link_to('Add an induction period', new_admin_teacher_induction_period_path(teacher_id: teacher.id)) %>
+</p>
--- a/app/controllers/admin/induction_periods_controller.rb
+++ b/app/controllers/admin/induction_periods_controller.rb
@@ -9,7 +9,7 @@ module Admin
 
       if @induction_period.save
         redirect_to admin_teacher_path(@induction_period.teacher),
-                    notice: "Induction period created successfully"
+                    alert: "Induction period created successfully"
       else
         render :new
       end
@@ -29,7 +29,7 @@ module Admin
 
       if service.update_induction_period!
         redirect_to admin_teacher_path(@induction_period.teacher),
-                    notice: "Induction period updated successfully"
+                    alert: "Induction period updated successfully"
       end
     rescue UpdateInductionPeriod::RecordedOutcomeError => e
       @induction_period.errors.add(:base, e.message)
--- a/app/controllers/admin/induction_periods_controller.rb
+++ b/app/controllers/admin/induction_periods_controller.rb
@@ -5,31 +5,25 @@ module Admin
     end
 
     def create
-      @induction_period = InductionPeriod.new(teacher:, **induction_period_params)
+      service = create_induction_period_service
 
-      if @induction_period.save
-        redirect_to admin_teacher_path(@induction_period.teacher),
-                    alert: "Induction period created successfully"
-      else
-        render :new
+      if service.create_induction_period!
+        redirect_to admin_teacher_path(teacher), alert: 'Induction period created successfully'
       end
+    rescue ActiveRecord::RecordInvalid
+      @induction_period = service.induction_period
+      render :new, status: :unprocessable_entity
     end
 
     def edit
-      @induction_period = InductionPeriod.find(params[:id])
+      induction_period
     end
 
     def update
-      @induction_period = InductionPeriod.find(params[:id])
-      service = UpdateInductionPeriod.new(
-        induction_period: @induction_period,
-        params: induction_period_params,
-        author: current_user
-      )
+      service = update_induction_period_service
 
       if service.update_induction_period!
-        redirect_to admin_teacher_path(@induction_period.teacher),
-                    alert: "Induction period updated successfully"
+        redirect_to admin_teacher_path(@induction_period.teacher), alert: 'Induction period updated successfully'
       end
     rescue UpdateInductionPeriod::RecordedOutcomeError => e
       @induction_period.errors.add(:base, e.message)
@@ -51,7 +45,19 @@ module Admin
     end
 
     def teacher
-      Teacher.find(params[:teacher_id])
+      @teacher ||= Teacher.find(params[:teacher_id])
+    end
+
+    def induction_period
+      @induction_period ||= InductionPeriod.find(params[:id])
+    end
+
+    def create_induction_period_service
+      CreateInductionPeriod.new(author: current_user, teacher:, params: induction_period_params)
+    end
+
+    def update_induction_period_service
+      UpdateInductionPeriod.new(author: current_user, induction_period:, params: induction_period_params)
     end
   end
 end
--- a/app/controllers/admin/induction_periods_controller.rb
+++ b/app/controllers/admin/induction_periods_controller.rb
@@ -1,5 +1,20 @@
 module Admin
   class InductionPeriodsController < AdminController
+    def new
+      @induction_period = InductionPeriod.new(teacher:)
+    end
+
+    def create
+      @induction_period = InductionPeriod.new(teacher:, **induction_period_params)
+
+      if @induction_period.save
+        redirect_to admin_teacher_path(@induction_period.teacher),
+                    notice: "Induction period created successfully"
+      else
+        render :new
+      end
+    end
+
     def edit
       @induction_period = InductionPeriod.find(params[:id])
     end
@@ -27,8 +42,16 @@ module Admin
 
     def induction_period_params
       params.require(:induction_period).permit(
-        :started_on, :finished_on, :number_of_terms, :induction_programme
+        :started_on,
+        :finished_on,
+        :number_of_terms,
+        :induction_programme,
+        :appropriate_body_id
       )
+    end
+
+    def teacher
+      Teacher.find(params[:teacher_id])
     end
   end
 end
--- a/app/models/concerns/interval.rb
+++ b/app/models/concerns/interval.rb
@@ -22,7 +22,7 @@ module Interval
   def period_dates_validation
     return if incomplete?
 
-    errors.add(:finished_on, "The finish date must be later than the start date (#{started_on.to_fs(:govuk)})") if invalid_date_order?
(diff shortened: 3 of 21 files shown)


app/models/induction_period.rb
```diff
@@ -32,6 +33,19 @@ def siblings
...
 private
 
+  # Ensure admin users inserting new induction periods include end dates.
+  def end_date_admin_only
+    return unless inserting_induction_period? && finished_on.blank?
+
+    errors.add(:finished_on, "End date is required for inserted periods")
+  end
+
+  def inserting_induction_period?
```
avinhurry, line 43: `inserting_induction_period` is a bit confusing to me, probably because I am not as familiar with the domain.

How does `valid_for_insertion` sound to you?

app/views/admin/induction_periods/new.html.erb
```diff
@@ -0,0 +1,41 @@
...
+      <%= f.govuk_date_field :started_on,
+          legend: { text: "Start date", size: "m" } %>
+
+      <%= f.govuk_date_field :finished_on,
+          legend: { text: "End date", size: "m" } %>
+
+      <%= f.govuk_text_field :number_of_terms,
+          label: { text: "Number of terms", size: "m" },
+          hint: { text: "Enter 0 if less than one full term, or a number with up to one decimal place (for example, 1.5)" },
+          width: 4,
```
avinhurry, line 32: Since this is going to be used for integers do we know what the most common numbers will be? I would imagine it would be 1 or 2 digits. 

How would you feel about reducing the width: `width: 2`

spec/features/appropriate_bodies/claim_an_ect_spec.rb
```diff
@@ -52,6 +30,28 @@
     end
   end
 
+  describe "when the ECT is already claimed by another appropriate body" do
+    include_context 'fake trs api client that finds teacher with specific induction status', 'InProgress'
+
+    before do
+      other_body = FactoryBot.create(:appropriate_body)
+      teacher = FactoryBot.create(:teacher, trn: '1234567')
+      FactoryBot.create(:induction_period, :active, teacher:, appropriate_body: other_body)
```
avinhurry, line 39: What are your thoughts on extracting this into a method to abstract the information away from the reader, something like:

```
before do 
  create_teacher_with_induction_period
end 

private

def create_teacher_with_induction_period
  other_body = FactoryBot.create(:appropriate_body)
  teacher = FactoryBot.create(:teacher, trn: '1234567')
  FactoryBot.create(:induction_period, :active, teacher:, appropriate_body: other_body)
  teacher
end
```

spec/services/admin/current_teachers_spec.rb
```diff
@@ -1,13 +1,13 @@
...
-      it 'returns teachers with ongoing induction periods for the given appropriate bodies' do
-        ab1 = FactoryBot.create(:appropriate_body)
-        ab2 = FactoryBot.create(:appropriate_body)
-        teacher1 = FactoryBot.create(:teacher)
-        teacher2 = FactoryBot.create(:teacher)
-        teacher3 = FactoryBot.create(:teacher)
+    let(:teacher1) { FactoryBot.create(:teacher) }
+    let(:teacher2) { FactoryBot.create(:teacher) }
+    let(:teacher3) { FactoryBot.create(:teacher) }
+    let(:ab1) { FactoryBot.create(:appropriate_body) }
```
avinhurry, line 6: I know you just repositioned this into let blocks but my mind immediately thinks of Accredited body when I see AB, it would be nice to be more explicit.

```suggestion
    let(:appropriate_body) { FactoryBot.create(:appropriate_body) }
```

That would also mean you could omit the first one:

E.G

```
FactoryBot.create(
          :induction_period,
          teacher: teacher1,
          appropriate_body:)
```

spec/services/teachers/induction_spec.rb
```diff
@@ -2,91 +2,87 @@
...
-    before do
-      FactoryBot.create(:induction_period, teacher:, started_on: 2.years.ago, finished_on: 1.year.ago)
-    end
+  let(:current_period) do
+    FactoryBot.create(:induction_period, :active, teacher:, started_on: 6.months.ago)
+  end
 
-    it "returns the induction period without a finished_on date" do
-      current_period = FactoryBot.create(:induction_period, :active, teacher:, started_on: 6.months.ago)
+  let(:recent_finish) do
```
avinhurry, line 9: Recent is subjective, I had to scroll back up to see the definition, how do you feel about being more explicit here, something like:

`let(:induction_period_finished_one_year_ago)`

This will also help to reduce cognitive load, especially for someone new to the codebase.

spec/services/teachers/induction_spec.rb
```diff
@@ -2,91 +2,87 @@
...
+    FactoryBot.create(:induction_period, teacher:, started_on: 2.years.ago, finished_on: 1.year.ago)
+  end
+
+  let(:older_finish) do
```
avinhurry, line 13: Similarly for this one, it could be something like: 

`let(:induction_period_finished_two_years_ago)`

app/components/teachers/details/current_induction_period_component.html.erb
```diff
@@ -13,7 +13,3 @@
     end
   end
 %>
-
-<p class='govuk-!-text-align-right govuk-!-margin-bottom-0'>
-  <%= govuk_link_to('Add an induction period', new_admin_teacher_induction_period_path(teacher_id: teacher.id)) %>
```
avinhurry: Just noticed that you originally put the link in here. 

My mind was wondering back and forth thinking about where to put this, at first I thought the show page would be best but then I realised that the `InductionSummaryComponent` component is rendered from within another component 😅 

what was your reason for deciding to put it in the `InductionSummaryComponent` in the end?

app/controllers/admin/induction_periods_controller.rb
```diff
@@ -27,8 +42,16 @@ def update
...
+        :started_on,
+        :finished_on,
+        :number_of_terms,
+        :induction_programme,
+        :appropriate_body_id
       )
     end
+
+    def teacher
+      Teacher.find(params[:teacher_id])
```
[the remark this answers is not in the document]
peterdavidhamilton: It's currently only queried once per request. I would've otherwise unless I'm mistaken. Shall I memorize anyway do you think?

app/models/concerns/interval.rb
```diff
@@ -69,4 +69,9 @@ def has_predecessors? = predecessors.exists?
   def has_siblings? = siblings.exists?
 
   def has_successors? = successors.exists?
+
+  # Methods
+  def last_finished_sibling
+    siblings.finished.earliest_first.last
```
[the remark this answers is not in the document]
peterdavidhamilton: It probably would, I didn't think of that. I'm told performance isn't a primary concern for the service at the moment. I try this out.

app/models/induction_period.rb
peterdavidhamilton: It is attempting to articulate "are you positioning this before, or between existing periods" i.e. are you inserting. No validation is happening at this point, it is the predicate that helps us know to validate.

If you are inserting, and have also forgotten to complete the field, then we error?

The option I went with first was just `#backfilling?` which is typically succinct from me but I thought it needed expanding. 

How about `#between_other_induction_periods?`

spec/services/admin/current_teachers_spec.rb
peterdavidhamilton: I fought the urge to do just that. I will. Good call.

app/components/teachers/details/current_induction_period_component.html.erb
peterdavidhamilton: Well I have a sneaking suspicion it'll move again. @emily-prudence-dfe spotted it wasn't getting rendered for some ECTs who no longer had an ongoing induction so it needed to be in a different template. It might actually warrant its own tiny and very simple component so we can compose and position it wherever we want; but a counter argument about making it a big button at the top was thrown in the mix. I decided this is the lightest touch and gives us time to consider it.

app/views/admin/induction_periods/new.html.erb
peterdavidhamilton: The scope of the ticket started with reusing the `edit` template, but then grew once we identified the addition of the AB search input. I'd say we want consistency, and presume that width was copied from the prototype. @JSJohal is this a tweaK you'd agree with?

app/models/induction_period.rb
```diff
@@ -32,6 +33,19 @@ def siblings
 
 private
 
+  # Ensure admin users inserting new induction periods include end dates.
+  def end_date_admin_only
+    return unless inserting_induction_period? && finished_on.blank?
+
+    errors.add(:finished_on, "End date is required for inserted periods")
```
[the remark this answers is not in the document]
peterdavidhamilton: A `Start date is required` is the best I can think of, because it is not specific to admin users inserting. Is making the message conditional worth it?

spec/features/appropriate_bodies/claim_an_ect_spec.rb
peterdavidhamilton: I can improve it but I prefer using `let` and trying to float the common dependencies to the top so they're shared, and can get overridden later. There is similar code we can move out of the way but I like the code the makes each context special to be easily read in the before block. I'll push a change that does this for your consideration.

app/models/induction_period.rb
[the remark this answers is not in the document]
avinhurry: `Start date is required` sounds good to me and was my initial thought but I wasn't sure about the business logic.

I don't think we need to make the message conditional as the presence validation is already there, we are just updating the content of the message if that makes sense. It's more of just a tidy up.

spec/features/appropriate_bodies/claim_an_ect_spec.rb
peterdavidhamilton: It is lazy but effectively there is no difference here. The `before` block still creates the induction period, but when it tries to resolve the `teacher` variable for example it lazily looks up the chain, rather than having it be ready and waiting. The benefit I believe with this approach is that I can decompose dependencies and common ones rise higher in the spec, then, if need be, can get nulled or tweaked. Ultimately I've seen this pattern improve terseness and readability. I also like having RSpec use the self-documenting output wherever possible. I am trying to win people over on that one.

app/models/concerns/interval.rb
```diff
@@ -69,4 +69,9 @@ def has_predecessors? = predecessors.exists?
   def has_siblings? = siblings.exists?
 
   def has_successors? = successors.exists?
+
+  # Methods
```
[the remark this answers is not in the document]
peterdavidhamilton: True but the predicates are grouped

app/models/induction_period.rb
```diff
@@ -32,6 +33,19 @@ def siblings
 
 private
 
+  # Ensure admin users inserting new induction periods include end dates.
+  def end_date_admin_only
```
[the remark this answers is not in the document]
peterdavidhamilton: What's been bugging me about this bit is the `admin`. User roles should not be reflected here but having asked @emily-prudence-dfe this is an admin-specific validation. Can you think of an alternative, I've run out of ideas?

emily-prudence-dfe:
This looks to be working perfectly @peterdavidhamilton 👌🏻

The only part I can't test is that where the added induction period also changes the start date of the induction as a whole, that an update is sent to TRS. I can see, though, that the start date does change in service. If you can confirm the TRS API call is happening in the back end then I'm happy to merge away!

PR #307, opened 2025-03-07
Add service rules for induction extensions

peteryates:
This change adds service rules for the extensions.

It also changes the button text on the forms for 'Add extension' and' Update extension' because 'Continue' is misleading.

@PriyaNahar2 I think we need to do a bit of surgery on the edit page next week, I'm not sure it makes sense at the moment!

### DIFF
--- a/documentation/service-rules/appropriate-body-users.md
+++ b/documentation/service-rules/appropriate-body-users.md
@@ -213,7 +213,14 @@ The fail process takes place with the following steps:
 
 ## Extensions
 
-If an ECT's induction is [interrupted by more than 30 working days](/policy/induction-arrangements-for-school-teachers-in-england#para-8-1) it can be extended.
+An ECT's induction must be extended if it is [interrupted by more than 30 working days for ad hoc purposes](/policy/induction-arrangements-for-school-teachers-in-england#para-8-1).
+
+An ECT's induction may be extended if:
+- It is interrupted by statutory maternity, paternity, adoption, shared parental, parental bereavement or carer’s leave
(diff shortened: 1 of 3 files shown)


documentation/service-rules/appropriate-body-users.md
```diff
@@ -211,3 +211,40 @@ The fail process takes place with the following steps:
   - partial terms must be entered as decimals
   - value must be between 0 and 16 weeks
 
+## Extensions
+
+If an ECT's induction is [interrupted by more than 30 working days](/policy/induction-arrangements-for-school-teachers-in-england#para-8-1) it can be extended.
```
emily-prudence-dfe, line 216:
```suggestion
An ECT's induction must be extended if it is interrupted by more than 30 working days for ad hoc purposes (/policy/induction-arrangements-for-school-teachers-in-england#para-8-1)

An ECT's induction may be extended if:
- It is interrupted by statutory maternity, paternity, adoption, shared parental, parental bereavement or carer’s leave
- The ECT is determined not to have met the Teaching Standards when they were due to complete their induction, and an Appropriate Body makes a decision that an extension to their induction is appropriate.

Appropriate Bodies only need to record the latter type of extension on the service, where an induction has been extended due to the ECT failing to meet the Teaching Standards. ECTs are able to appeal this type of extension.

```

emily-prudence-dfe: This is quite wordy but more accurately represents extension scenarios and where we expect ABs to actually record an extension.

emily-prudence-dfe (changes requested):

emily-prudence-dfe (approved):

PriyaNahar2 (commented):
Service rules look good to me and agree with the 'edit extensions' page - need to revisit content.

PR #297, opened 2025-03-05
Timeline

peteryates:
Add a timeline to visualise events.

We record events when key things happen in the app, like beginning or ending an induction period.

There's no way to see them yet. This PR uses the [MoJ Timeline](https://design-patterns.service.justice.gov.uk/components/timeline/) component to do that.

For now I just added the component to the bottom of the teacher 'profile' page in the admin section.

| Zoomed in | Full page |
| ----- | ----- |
|Screenshot From 2025-03-05 13-05-22 | Screenshot From 2025-03-05 13-05-56 |

PR #310, opened 2025-03-10
[Migration] Add created_at to view components

tonyheadford:
Add `created_at` timestamp to view of `ParticipantProfile` and `InductionRecord` in the migration UI.

### DIFF
--- a/app/components/migration/induction_record_component.html.erb
+++ b/app/components/migration/induction_record_component.html.erb
@@ -61,5 +61,9 @@
       row.with_key(text: "Training status")
       row.with_value(text: induction_record.training_status)
     end
+    list.with_row do |row|
+      row.with_key(text: "Created at")
+      row.with_value(text: induction_record.created_at)
+    end
   end
 end %>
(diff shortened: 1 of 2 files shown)


app/components/migration/induction_record_component.html.erb
```diff
@@ -61,5 +61,9 @@
       row.with_key(text: "Training status")
       row.with_value(text: induction_record.training_status)
     end
+    list.with_row do |row|
+      row.with_key(text: "Created at")
+      row.with_value(text: induction_record.created_at)
```
peteryates, line 66: Do we want to format these GOV.UK style?

I added some formats here which can be cherry picked if useful (unless the timeline stuff is merged soon)

https://github.com/DFE-Digital/register-early-career-teachers-public/pull/297/files#diff-2cd7d9700ea89028447e936ab6d59ebd6187d3ee212db3c1c9791fdd309ff737R7

tonyheadford: I get the same output from no explicit format as I do from `.to_fs(:govuk)` - is that the default?
This is only so that we can see the time the objects were created for comparisons so having `2025-01-09 13:17:26 UTC` is fine

graciebrennan:
Tony to merge

PR #311, opened 2025-03-10
Make teachers refresh class use teachers manage

peteryates:
The teacher refresh class doesn't know about events but the manage one does. We want to record events when key bits of information on the teacher change.

### DIFF
--- a/app/services/appropriate_bodies/claim_an_ect/register_ect.rb
+++ b/app/services/appropriate_bodies/claim_an_ect/register_ect.rb
@@ -62,7 +62,7 @@ module AppropriateBodies
       end
 
       def teacher
-        @teacher ||= manage_teacher.teacher
+        @manage_teacher.teacher
       end
 
       def create_induction_period
@@ -77,7 +77,7 @@ module AppropriateBodies
       end
 
       def send_begin_induction_notification_to_trs
-        return true if teacher.induction_periods.any?
+        return true if ::Teachers::Induction.new(teacher).has_induction_periods?
 
         BeginECTInductionJob.perform_later(
           trn: pending_induction_submission.trn,


app/services/teachers/manage.rb
```diff
@@ -13,32 +13,72 @@ def initialize(author:, teacher:, appropriate_body:)
     @appropriate_body = appropriate_body
   end
 
+  def self.find_or_initialize_by(trn:, trs_first_name:, trs_last_name:, event_metadata:)
+    teacher = Teacher.find_or_initialize_by(trn:) do |t|
+      t.trs_first_name = trs_first_name
+      t.trs_last_name = trs_last_name
+    end
+
+    if teacher.new_record? && teacher.save
```
peterdavidhamilton, line 22: Commented on other PR asking about terseness vs verbosity on this occasion. Is it acceptable to splat event stuff because the lines above describe what it contains? Straw poll.

peteryates: Sounds good - I was going to tackle that in a subsequent PR. I'll raise a issue though and anyone's welcome to do it 🙂

peteryates: I was also going to create an `Events::Metadata` class we can use as it provides a touch more rigidity than a hash.

app/services/appropriate_bodies/claim_an_ect/register_ect.rb
```diff
@@ -53,11 +53,16 @@ def update_itt_provider_name
...
+          trn: pending_induction_submission.trn,
+          trs_first_name: pending_induction_submission.trs_first_name,
+          trs_last_name: pending_induction_submission.trs_last_name,
+          event_metadata: { author:, appropriate_body: }
+        )
       end
 
       def teacher
-        @teacher ||= Teacher.find_or_initialize_by(trn: pending_induction_submission.trn)
+        @teacher ||= manage_teacher.teacher
```
peterdavidhamilton, line 65: Was chatting with @avinhurry about memoization. I believe we can remove it here because `manage_teacher` keeps hold of the object, but generally, are we going for:
1. Do it only when required 
2. Everything, everywhere, all at once

peterdavidhamilton:
@peteryates I recently had a deployment fail to reflect recent commits. A previous web pod was struggling and rerunning the workflow wasn't sorting it. I fixed by rolling out the deployment using kubectl. 🔨

peterdavidhamilton (approved):
LGTM

app/services/appropriate_bodies/claim_an_ect/register_ect.rb
peteryates: Thanks @peterdavidhamilton, added a commit that fixes this.

PR #313, opened 2025-03-10
Refactor `ECT details` summary list into `ViewComponent`

avinhurry:
## Overview

This PR introduces the `TeacherProfileSummaryListComponent` component, which renders a summary list with key details about an ECT (Early Career Teacher). The component displays the following information:

- Teacher's name
- Email address
- Mentor information
- School start date
- Working pattern

## Changes

- Add `TeacherProfileSummaryListComponent` to render a summary list of ECT details.
- Include helper methods to format teacher name, mentor details and school start date.
- Add tests for the component to ensure it renders the correct details.

### DIFF
--- a/app/components/schools/teacher_profile_summary_list_component.html.erb
+++ b/app/components/schools/teacher_profile_summary_list_component.html.erb
@@ -1,0 +1,5 @@
+<%=
+  govuk_summary_list(
+    rows:
+  )
+%>
--- a/app/components/schools/teacher_profile_summary_list_component.html.erb
+++ b/app/components/schools/teacher_profile_summary_list_component.html.erb
@@ -1,5 +1,0 @@
-<%=
-  govuk_summary_list(
-    rows:
-  )
-%>
--- a/app/views/schools/ects/show.html.erb
+++ b/app/views/schools/ects/show.html.erb
@@ -7,17 +7,9 @@
   )
 %>
 <h2 class='govuk-heading-m'>ECT details</h2>
-<%=
-  govuk_summary_list(
-    rows: [
-      { key: { text: 'Name' },              value: { text: teacher_full_name(@ect.teacher) } },
-      { key: { text: 'Email address' },     value: { text: @ect.email } },
-      { key: { text: 'Mentor' },            value: { text: ect_mentor_details(@ect) } },
-      { key: { text: 'School start date' }, value: { text: ect_start_date(@ect) } },
-      { key: { text: 'Working pattern' },   value: { text: @ect.working_pattern&.humanize } },
-    ]
-  )
-%>
+
+<%= render Schools::TeacherProfileSummaryListComponent.new(@ect) %>
+
 <h2 class='govuk-heading-m'>ECTE training details</h2>
 <%=
   govuk_summary_card(title: 'Reported to us by your school') do |card|
--- a/app/views/schools/ects/show.html.erb
+++ b/app/views/schools/ects/show.html.erb
@@ -6,7 +6,6 @@
     caption_size: 'l',
   )
 %>
-<h2 class='govuk-heading-m'>ECT details</h2>
 
 <%= render Schools::TeacherProfileSummaryListComponent.new(@ect) %>
 
--- a/spec/components/schools/teacher_profile_summary_list_component_spec.rb
+++ b/spec/components/schools/teacher_profile_summary_list_component_spec.rb
@@ -1,12 +1,22 @@
 require "rails_helper"
 
 RSpec.describe Schools::TeacherProfileSummaryListComponent, type: :component do
-  include TeacherHelper
-  include ECTHelper
-
-  let(:ect) { FactoryBot.create(:ect_at_school_period, :active, started_on: 3.years.ago) }
+  let(:teacher) { FactoryBot.create(:teacher, trn: '9876543', trs_first_name: 'Kakarot', trs_last_name: 'SSJ') }
+  let(:teacher_2) { FactoryBot.create(:teacher, trn: '987654', trs_first_name: 'Naruto', trs_last_name: 'Ninetails') }
   let(:mentor) { FactoryBot.create(:mentor_at_school_period, :active, started_on: 3.years.ago) }
-  let!(:ongoing_mentorship) { FactoryBot.create(:mentorship_period, :active, mentee: ect, mentor:) }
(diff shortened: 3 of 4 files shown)


app/views/schools/ects/show.html.erb
```diff
@@ -7,17 +7,9 @@
...
-      { key: { text: 'Name' },              value: { text: teacher_full_name(@ect.teacher) } },
-      { key: { text: 'Email address' },     value: { text: @ect.email } },
-      { key: { text: 'Mentor' },            value: { text: ect_mentor_details(@ect) } },
-      { key: { text: 'School start date' }, value: { text: ect_start_date(@ect) } },
-      { key: { text: 'Working pattern' },   value: { text: @ect.working_pattern&.humanize } },
-    ]
-  )
-%>
+
+<%= render Schools::TeacherProfileSummaryListComponent.new(@ect) %>
```
peteryates, line 11: Make sense to include the subheading in the component too?

app/components/schools/teacher_profile_summary_list_component.html.erb
```diff
@@ -0,0 +1,5 @@
+<%=
+  govuk_summary_list(
+    rows:
+  )
+%>
```
peteryates, line 5: If this is all the component does you don't really need a template, just do

```ruby
def call
  govuk_summary_list(rows:)
end
```

Or including the header

```ruby
def call
  safe_join([tag.h2('hi'), govuk_summary_list(rows:)])
end
```

spec/components/schools/teacher_profile_summary_list_component_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  it "renders the email row with correct value" do
+    expect(page).to have_selector(".govuk-summary-list__row", text: "Email address")
+    expect(page).to have_text(ect.email)
+  end
+
+  it "renders the mentor row with correct value" do
+    expect(page).to have_selector(".govuk-summary-list__row", text: "Mentor")
+    expect(page).to have_text(ect_mentor_details(ect))
```
peterdavidhamilton, line 31: I know we aren't always doing this in other specs but I am an advocate of asserting the raw string (or partial string). As in, not including the helpers, factorying a `Teacher` with known values, and using `have_text("this is the text")`.

Yes it does mean that changes to the helper methods output can break here, but I like that

app/components/schools/teacher_profile_summary_list_component.html.erb
avinhurry: Thank you, this is awesome! 😍 

https://github.com/DFE-Digital/register-early-career-teachers-public/pull/313/commits/b046bd3da49319c6f229e0dfb9802e10d10a7f78

spec/components/schools/teacher_profile_summary_list_component_spec.rb
avinhurry: I usually like asserting against raw strings in tests too but on this occasion we are using Faker to generate things dynamically, are you suggesting that we hard code the name into the factory instead and then assert against it?

For example: https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/spec/factories/teacher_factory.rb#L4-L5

peterdavidhamilton:
```ruby
let(:teacher) { FactoryBot.create(:teacher, trn: '9876543', trs_first_name: 'Avin', trs_last_name: 'Hurry') }
let(:ect) { FactoryBot.create(:ect_at_school_period, :active, teacher:, started_on: 3.years.ago) }
```

I thought instead we could define the `Teacher` the ECT is linked to and gets the name from. If a spec also cares about the mentor you need to make two `Teacher` records with this approach, and there are examples in the suite where we do this with `mentor` and `mentee` as the variable names.

avinhurry: Ah yes that makes sense, thanks! Happy to do it that way if thats the approach we prefer but it would be nice to be consistent going forward.

I've updated it in this commit: https://github.com/DFE-Digital/register-early-career-teachers-public/pull/313/commits/a14d776d08f66cdef1146b67cd149eae377d2da9

spec/components/schools/teacher_profile_summary_list_component_spec.rb
```diff
@@ -1,12 +1,22 @@
...
+  let(:teacher_2) { FactoryBot.create(:teacher, trn: '987654', trs_first_name: 'Naruto', trs_last_name: 'Ninetails') }
   let(:mentor) { FactoryBot.create(:mentor_at_school_period, :active, started_on: 3.years.ago) }
-  let!(:ongoing_mentorship) { FactoryBot.create(:mentorship_period, :active, mentee: ect, mentor:) }
+  let(:mentor_2) { FactoryBot.create(:mentor_at_school_period, :active, teacher: teacher_2, started_on: 3.years.ago) }
+  let(:ect) do
+    FactoryBot.create(:ect_at_school_period, :active, teacher:, started_on: Date.new(2021, 9, 1),
+                                                      email: 'foobarect@madeup.com', working_pattern: 'full_time')
+  end
+
+  let!(:ongoing_mentorship) do
```
peterdavidhamilton, line 13: Is doing this any clearer?

`teacher` > `mentee_teacher`
`teacher_2` > `mentor_teacher`
`mentor` > `previous_mentor`
`mentor_2` > `current_mentor` 
 `ect` > `mentee`

And just building the `mentorship_periods` in a `before` block because we don't access them by variable name.

avinhurry: Yes, definitely! Thank you 🙏🏽 

You can clearly tell I do not know the domain 😅 

Updated here: https://github.com/DFE-Digital/register-early-career-teachers-public/pull/313/commits/9663112674b1614e637342ac73d3a29a16168044

peterdavidhamilton (approved):
Nice. Product reviewers might appreciate links or screenshots in the description when appropriate. 

The `rows` method is public and not tested. The pedant in me thinks either it should be private, or that we expect 5 rows. The latter is probably better documentation 🤷🏼‍♂️

avinhurry:
> Nice. Product reviewers might appreciate links or screenshots in the description when appropriate.
> 
> The `rows` method is public and not tested. The pedant in me thinks either it should be private, or that we expect 5 rows. The latter is probably better documentation 🤷🏼‍♂️

Excellent shout @peterdavidhamilton! Will add a test for that 🙏🏽

PR #317, opened 2025-03-11
restore from any dump in any azure storage account

ltello:
- Fix the workflow that restores the main database of an environment to be able to do it from any file stored in any other environment's Azure storage account.
- Add failure notification to the workflow that dumps an environment's main database and stores it in their Azure Storage account.


tonyheadford (approved):
Fantastic!

PR #326, opened 2025-03-12
Fix the induction event happened_at values

peteryates:
I implemented this wrong first time around because I was thinking of the event from the point of view of when it was recorded, not from the actual final timeline.

On the timeline we want the date the person started or finished an induction period, not the date the AB informed us.

We'll need to fix some records in prod but that is thankfully straightfoward.

### DIFF
--- a/app/services/appropriate_bodies/release_ect.rb
+++ b/app/services/appropriate_bodies/release_ect.rb
@@ -34,6 +34,10 @@ module AppropriateBodies
     end
 
     def ongoing_induction_period
+      @ongoing_induction_period ||= find_ongoing_induction_period
+    end
+
+    def find_ongoing_induction_period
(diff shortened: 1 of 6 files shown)


peteryates:
I think the script we'd need to run it would be essentially this:

```ruby
since_go_live_date = Date.new(2025, 2, 19)..
closure_events = %w(
  appropriate_body_releases_teacher
  appropriate_body_releases_teacher
  appropriate_body_fails_teacher
)

Event.transaction do
  Event
    .where(created_at: since_go_live_date, event_type: 'appropriate_body_claims_teacher')
    .each { |e| e.update!(happened_at: e.induction_period.started_on) }
  
  Event
    .where(created_at: since_go_live_date, event_type: closure_events)
    .each { |e| e.update!(happened_at: e.induction_period.finished_on) }
end
```

PR #332, opened 2025-03-13
Relax pre-sept-2021 validation for admins

peterdavidhamilton:
Make pre-sept-2021 validation conditional to permit historic dates for admins

### DIFF
--- a/app/models/concerns/shared_induction_period_validation.rb
+++ b/app/models/concerns/shared_induction_period_validation.rb
@@ -2,12 +2,12 @@ module SharedInductionPeriodValidation
   extend ActiveSupport::Concern
 
   included do
-    validates :appropriate_body_id,
-              presence: { message: "Select an appropriate body" }
+    validate :started_on_from_september_2021_onwards, if: -> { started_on.present? }, on: :register_ect
 
-    validate :started_on_from_september_2021_onwards, if: -> { started_on.present? }
+    validates :appropriate_body_id, presence: { message: "Select an appropriate body" }
 
     validate :started_on_not_in_future, if: -> { started_on.present? }
+
     validate :finished_on_not_in_future, if: -> { finished_on.present? }
   end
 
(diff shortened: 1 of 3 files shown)


config/initializers/constants.rb
```diff
@@ -1,3 +1,9 @@
+# ECF version 1 released: 1st September 2021
+ECF_ROLLOUT_DATE = Date.new(2021, 9, 1).freeze
+
+# Introduction of ECT statutory induction: 1st September 1999
+STATUTORY_INDUCTION_ROLLOUT_DATE = Date.new(1999, 9, 1).freeze
```
peteryates, line 5: Think we'll need these values outside of the realm of induction period validation?

app/models/concerns/shared_induction_period_validation.rb
```diff
@@ -2,17 +2,25 @@ module SharedInductionPeriodValidation
   extend ActiveSupport::Concern
 
   included do
-    validates :appropriate_body_id,
-              presence: { message: "Select an appropriate body" }
+    validate :started_on_from_september_2021_onwards, if: -> { started_on.present? }, on: :register_ect
 
-    validate :started_on_from_september_2021_onwards, if: -> { started_on.present? }
+    validate :started_on_from_september_1999_onwards, if: -> { started_on.present? }
```
peteryates, line 7: ✅

peteryates (commented):
Looks good to me. I do wonder about making those dates globals because I can't think of where else they'd be used.

peteryates (approved):

emily-prudence-dfe:
This looks good and works as expected. Just to check as I'm reviewing both at the same time - will this date validation also apply to Add an Induction Period once merged? Is there anything (aside from merging both) that needs to be done to make sure it does?

@peterdavidhamilton

peterdavidhamilton:
@emily-prudence-dfe yes this will ensure Admin inserted periods can't pre-date 1999. Just need to resolve any conflicts.

Issue #268, opened 2025-02-27
Add feasible steps functionality to the Register Mentor journey


ltello:
## What?
Even though each step of the Register Mentor journey has a `previous_step` and `next_step` logic included to redirect the user to the appropriate next step, nothing prevents them from redirecting their browser to any random step at any time or navigate though them via the back and forward buttons of the browser.

Therefore, we need a way to be able to redirect the user's browser to the appropriate next step based on the overall state in the journey and the currently visited step.

The suggestion in this ticket is to extend the journey with functionality so that it can compute the list of allowed steps (steps we consider 'visitable') based on the current overall state in the journey, the actually visited step and automatically redirect the user to the last of those allowed steps if the current one is not included in that list.

## Why?
This matches ECF1. It avoids getting weird data entering our database, or risks where we make changes when people are partway through the journey.

graciebrennan:
PR is almost ready for this

graciebrennan:
Tony to review

graciebrennan:
Tony has reviewed 

Lorenzo has added specs for further review

claire-hughez:
Make sure:
1. Registering a mentor isn't broken
2. You can't skip around from step 2 to step 5, or other hijinks, in the Register a mentor journey

claire-hughez:
I can't get past the page asking for email to test this properly, it errors as seen below.

https://github.com/user-attachments/assets/30207d55-f09a-42eb-94a2-55f5298155f3

ltello:
The issue is that if the mentor is being assigned to a school-led ECT and the next screen to be display (about confirming mentor training eligibility) assumes this mentor's ECT has a lead provider (but hasn't) and it tries to display the name (breaking the service)

claire-hughez:
That makes sense, I've made a note to talk to Emily M about school-led tickets.

However, I'm still having some issues.

I registered Theo Phillips as a mentor for someone. Then, when registering a new mentor I was trying to skip through to the check-answers page after only selecting to register a new mentor, and now every time I try to register a new mentor it just automatically assigns Theo Phillips.

https://github.com/user-attachments/assets/fde2cf60-1196-45ce-bc7e-fc1ca5aa951f

graciebrennan:
@ltello to review Claire's comment and can go to Emily for sign-off in Claire's absence

ltello:
There was a bug in the code that has been fixed by now. 
Please, re-test. 
The idea is: 
[...]

PR #337, opened 2025-03-17
Mentor wizard: redirect user to the first allowed step

ltello:
[Issue](https://github.com/DFE-Digital/register-early-career-teachers-public/issues/268)

Even though each step of the Register Mentor journey has a previous_step and next_step logic included to redirect the user to the appropriate next step, nothing prevents them from redirecting their browser to any random step at any time or navigate though them via the back and forward buttons of the browser.

Therefore, we need a way to be able to redirect the user's browser to the appropriate next step based on the overall state in the journey and the currently visited step.

The proposal in this PR is to extend the journey with functionality so that it can compute the list of allowed steps (steps we consider 'visitable') based on the current overall state in the journey, the actually visited step and automatically redirect the user to the last of those allowed steps if the current one is not included in that list.

### DIFF
--- a/app/wizards/schools/register_mentor_wizard/check_answers_step.rb
+++ b/app/wizards/schools/register_mentor_wizard/check_answers_step.rb
@@ -12,7 +12,12 @@ module Schools
     private
 
       def persist
-        AssignMentor.new(ect:, mentor: mentor.register!).assign!
+        ActiveRecord::Base.transaction do
+          AssignMentor.new(ect:, mentor: mentor.register!).assign!
+        end
+      rescue StandardError => e
+        mentor.registered = false
+        raise e
       end
     end
   end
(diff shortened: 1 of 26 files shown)


app/wizards/schools/register_mentor_wizard/check_answers_step.rb
```diff
@@ -12,7 +12,12 @@ def previous_step
     private
 
       def persist
-        AssignMentor.new(ect:, mentor: mentor.register!).assign!
+        ActiveRecord::Base.transaction do
+          AssignMentor.new(ect:, mentor: mentor.register!).assign!
+        end
```
tonyheadford, line 17: There's a transaction in the `.assign!` method, I don't think we need one here too?

tonyheadford (approved):
Nice.

PR #338, opened 2025-03-17
Improve event retrieval

peteryates:
Follows on from #297

This PR adds `Events::List`, a class responsible for pulling the events from the database for a given record (and eventually) filtering and ordering them.

### DIFF
--- a/app/services/events/list.rb
+++ b/app/services/events/list.rb
@@ -1,0 +1,13 @@
+module Events
+  class List
+    attr_accessor :scope
+
+    def initialize
+      @scope = Event.latest_first
+    end
+
+    def for_teacher(teacher)
+      scope.where(teacher:)
+    end
+  end
+end
--- a/spec/services/events/list_spec.rb
+++ b/spec/services/events/list_spec.rb
@@ -1,0 +1,15 @@
+describe 'Events::List' do
+  describe 'default scope' do
+    it 'orders by latest first by default' do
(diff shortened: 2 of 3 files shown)


app/services/events/list.rb
```diff
@@ -0,0 +1,13 @@
+module Events
+  class List
+    attr_accessor :scope
```
avinhurry, line 3: Is this used? Could it be worth waiting until we need it to add it? 

YAGNI

spec/services/events/list_spec.rb
```diff
@@ -0,0 +1,15 @@
+describe 'Events::List' do
+  describe 'default scope' do
+    it 'orders by latest first by default' do
+      expect(Events::List.new.scope.to_sql).to include(%(ORDER BY "events"."happened_at" DESC))
+    end
+  end
+
+  describe '.for_teacher' do
+    it 'selects only events with a teacher_id matching the provided teacher' do
```
avinhurry, line 9: My personal preference wording preference would be:

```suggestion
    it 'only selects events with a teacher_id matching the provided teacher' do
```

But not too fussed

avinhurry (approved):
A couple of minor suggestions/comments, otherwise looks good 👍🏽

app/services/events/list.rb
peteryates: I'm certain we'll be using it shortly, once we start filtering by event type.

PR #340, opened 2025-03-17
Refactor `govuk_summary_cards` used in `ects/show` into a ViewComponent

avinhurry:
### Context

Following on from #313, we are refactoring the `govuk_summary_cards` used in `ects/show` into a ViewComponent.

This change improves reusability and enables us to write targeted tests for the component, ensuring consistent rendering across different contexts.

### Thoughts/Considerations 🤔 

I considered making this more generic so it could be used for other summary cards, with helper methods in the view. However, that would introduce a lot of logic into the view, making it harder to maintain. Since this component is scoped to schools, I felt it was reasonable to tie it more closely to this domain.

Would a more generic summary card component be worth considering, or does keeping it scoped to schools feel like the right level of abstraction for now?

#### School Rows
- **Not `provider_led`**  
  School rows - Not provider_led

- **`provider_led`**  
  School rows - provider_led

---

#### Lead Provider Rows
- **With `training_periods`**  
  Lead provider rows

- **No `training_periods`**  

Lead provider rows - No training periods

---

#### Appropriate Body Rows
- **With `induction_periods`**  
  Appropriate body rows

- **No `induction_periods`**  
  Appropriate body rows - No induction periods

### DIFF
--- a/app/views/schools/ects/show.html.erb
+++ b/app/views/schools/ects/show.html.erb
@@ -9,69 +9,12 @@
 
 <%= render Schools::TeacherProfileSummaryListComponent.new(@ect) %>
 
-<h2 class='govuk-heading-m'>ECTE training details</h2>
-<%=
-  govuk_summary_card(title: 'Reported to us by your school') do |card|
-    if @ect.provider_led?
-      card.with_summary_list(
-        rows: [
-          { key: { text: 'Appropriate body' },
-            value: { text: appropriate_body_name(appropriate_body_type: @ect.appropriate_body_type,
-                                                 appropriate_body: @ect.appropriate_body) } },
-          { key: { text: 'Programme type' },   value: { text: programme_type_name(@ect.programme_type) } },
-          { key: { text: 'Lead provider' },    value: { text: @ect.lead_provider_name } },
-        ]
-      )
-    else
-      card.with_summary_list(
-        rows: [
-          { key: { text: 'Appropriate body' },
-            value: { text:appropriate_body_name(appropriate_body_type: @ect.appropriate_body_type,
-                                                appropriate_body: @ect.appropriate_body) } },
-          { key: { text: 'Programme type' },   value: { text: programme_type_name(@ect.programme_type)  } },
-        ]
-      )
-    end
-  end
-%>
+<h2 class="govuk-heading-m">ECTE training details</h2>
 
-<%=
-  if @ect.provider_led?
-    govuk_summary_card(title: 'Reported to us by your lead provider') do |card|
-      if (period = @ect.training_periods.earliest_first.last)
-        card.with_summary_list(
-          rows: [
-            { key: { text: 'Lead provider' },    value: { text: period.lead_provider.name } },
-            { key: { text: 'Delivery partner' }, value: { text: period.delivery_partner.name } },
-          ]
-        )
-      else
-        card.with_summary_list do |summary_list|
-          summary_list.with_row do |row|
-            row.with_value(text: 'Your lead provider has not reported any information to us yet.')
-          end
-        end
-      end
-    end
-  end
-%>
+<%= render Schools::SummaryCardComponent.new(title: 'Reported to us by your school', ect: @ect, data_source: :school) %>
 
-<%=
-  govuk_summary_card(title: 'Reported to us by your appropriate body') do |card|
-    if @ect.teacher.induction_periods.any?
-      card.with_summary_list(
-        rows: [
-          { key: { text: 'Appropriate body' },      value: { text: teacher_induction_ab_name(@ect.teacher) } },
-          { key: { text: 'Programme type' },        value: { text: teacher_induction_programme(@ect.teacher) } },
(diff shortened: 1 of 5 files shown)


app/views/schools/ects/show.html.erb
```diff
@@ -9,69 +9,14 @@
...
-          { key: { text: 'Appropriate body' },
-            value: { text:appropriate_body_name(appropriate_body_type: @ect.appropriate_body_type,
-                                                appropriate_body: @ect.appropriate_body) } },
-          { key: { text: 'Programme type' },   value: { text: programme_type_name(@ect.programme_type)  } },
-        ]
-      )
-    end
-  end
-%>
+<h2 class="govuk-heading-m">ECTE training details</h2>
```
avinhurry, line 12: I propose that we keep this out of the component and move [this one](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/components/schools/teacher_profile_summary_list_component.rb#L21) out of the component too.

The goal is to make the page easier to read and understand without needing to look inside the components. By keeping structural elements like headings in the view, it’s clearer how the page is laid out at a glance.

What are your thoughts on that @peteryates?

spec/factories/ect_at_school_period_factory.rb
```diff
@@ -25,6 +25,7 @@
 
     trait :school_led do
       programme_type { 'school_led' }
+      lead_provider { nil }
```
avinhurry, line 28: @ltello do you think it is a good idea to set the `lead_provider` to `nil` in this factory, since we are now validating that it's `nil`:

https://github.com/DFE-Digital/register-early-career-teachers-public/blob/b9ebeb4b0394c4778e45a663b5f561b1721ebb14/app/models/ect_at_school_period.rb#L50-L54

If this is not the approach we want to take globally, I can set it to `nil` in my specific test. 

Keen to hear your thoughts!

app/views/schools/ects/show.html.erb
```diff
@@ -9,69 +9,12 @@
...
+
+<%= render Schools::SummaryCardComponent.new(title: 'Reported to us by your school', ect: @ect, data_source: :school) %>
+
+<%= render Schools::SummaryCardComponent.new(title: 'Reported to us by your lead provider', ect: @ect, data_source: :lead_provider) %>
+
+<%= render Schools::SummaryCardComponent.new(title: 'Reported to us by your appropriate body', ect: @ect, data_source: :appropriate_body) %>
 
-<%=
-  if @ect.provider_led?
```
avinhurry, line 39: Am I right in thinking that if `@ect.provider_led?` is `false`, `training_periods` wouldn't exist, making the `@ect.provider_led?` check redundant?

- `ect_at_school_period` [can have many](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/b9ebeb4b0394c4778e45a663b5f561b1721ebb14/app/models/ect_at_school_period.rb#L25) `training_periods`.



- Each `training_period` [belongs to](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/models/training_period.rb#L12-L13) a `lead_provider` and `delivery_partner`

claire-hughez: This is such a good question! @peteryates @tonyheadford 

If Avin is right, then we should really rename training_periods to be reflective of the fact it's provider_led_training_periods, surely?

graciebrennan:
Lorenzo to have review

spec/factories/ect_at_school_period_factory.rb
tonyheadford: I think generally you want factories to construct valid objects so in this instance I think it's fine, a school-led training  period won't have a lead provider.

app/views/schools/ects/show.html.erb
tonyheadford: It would mean that if they switch to provider-led then we would lose the info about the time spent doing school-led unless we add an additional `ECTAtSchoolPeriod` which goes against what it represents

tonyheadford: so maybe the training_period is the right place for it?

claire-hughez: I guess we'd know because it's the difference between the start_date of the ect_at_school period, and the start date of the first provider_led_training_period.

However, if we still think they're something different, maybe it's still best to separate, but that means a training_period really belongs to an ECT, and sometimes it has a pairing (of LP and DP) associated with it?

tonyheadford (approved):
Looks great!

app/views/schools/ects/show.html.erb
peteryates: I kind of assume that if we're rendering just a group of components on a page it makes the structure clear, but I see your point.

What about

```erb
<%= render SomeComponent(header: "Widgets", other_things: @stuff)
```

avinhurry: Hmm, I would have liked: 

`<%= render SomeComponent(h2: "H2", heading: 'component heading', other_things: @stuff)`

But i'll keep it out of the component because we're now using that same component to render three of the summary cards but the same `h2` is only shown once at the top. 

I think this approach keeps it simple.

Thank you!

PR #343, opened 2025-03-17
Add starting off point for ECF1 API service rules

peteryates:
As discussed, slightly restructure the service rules by:

* adding a new empty doc for ECF1 API service rules to the site and link it from the main nav
* moving API versioning doc from the service rules to the product section


peteryates:
Replaced by #345 - review app went wonky.

PR #353, opened 2025-03-18
Add `ENVIRONMENT` var to sandbox's config

### LINKED ISSUE #270: Fix `make ci sandbox aks-ssh`
## Description

Fix `make ci sandbox aks-ssh` without breaking the build.
For some reason `sandbox` is missing in the K8 command triggered:
`kubectl -n cpd-production exec -ti --tty deployment/cpd-ec2--web -- /bin/sh`

## Note
It seem sensible to add the missing `ENVIRONMENT=sandbox` entry to `config/global_config/sandbox.sh`.
However this breaks the deploy step to `sandbox` in the build.

## Why?
This is a command that is broken in sandbox but not other environments.

peteryates:
This was missing and made the aks-ssh make command fail as it was incorrectly building this name `"cpd-ec2--web"` instead of `"cpd-ec2-sandbox-web"`.

Fixes #270

### DIFF
--- a/config/global_config/sandbox.sh
+++ b/config/global_config/sandbox.sh
@@ -1,4 +1,5 @@
 CONFIG=sandbox
+ENVIRONMENT=sandbox
 CONFIG_SHORT=sb
 AZURE_SUBSCRIPTION=s189-teacher-services-cloud-production
 AZURE_RESOURCE_PREFIX=s189p01


config/global_config/sandbox.sh
```diff
@@ -1,4 +1,5 @@
 CONFIG=sandbox
+ENVIRONMENT=sandbox
```
avinhurry, line 2: Do we need to add one for [review.sh](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/config/global_config/review.sh)?

peteryates: I don't think so, `ENVIRONMENT` is set per pull request (i.e., `REVIEW-123`) [in the Makefile](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/Makefile#L26).

avinhurry (approved):

PR #354, opened 2025-03-18
AB outcome journeys generating errors due to no active induction

peterdavidhamilton:
- consistent naming of "ongoing_induction_period"
- defend against running services that depend on an ongoing induction
- hide CTAs when not available
- redirect path when not available
- raise error consistently across AB journeys


peteryates (approved):
Looks good, definitely tidier.

PR #362, opened 2025-03-19
Simplify app setup

peteryates:
We want the process of getting the app up and running as simple as possible.

It should be no more than a case of copying the `.env.example` file to `.env` and running `bin/setup`.

## Changes

- **Disable MIGRATION settings in the .env**
- **Remove localhost gem and associated config**


.env.example
```diff
@@ -3,17 +3,17 @@ ENVIRONMENT_COLOUR=red
 ENVIRONMENT_PHASE_BANNER_CONTENT="Local development environment"
 TRS_API_BASE_URL=http://localhost:3000
 TRS_API_KEY=abc12
```
avinhurry, line 5: Is it worth a note about these two somewhere in the setup docs?

peteryates: I was thinking about how we can handle this gracefully when they aren't set (in development mode only) and was going to do the following:

1. comment those two lines out of `.env.example`
2. disable the blue helper box by default 
3. add a `TRS_API_FAKE_CLIENT=true` line instead
4. add a line to each of the env-specific configs that sets the API client class in the config (hardcoded to the real one for production and test, dependant on env vars in development)

Then someone cloning the app and copying the `.env` would get a nice fake API experience out of the box without having to set up extra things

PR #364, opened 2025-03-19
Add support for rendering `DetailsComponent` for admins

avinhurry:
### What's Changed

- Added support for passing an `is_admin` flag to `DetailsComponent`.
- Updated `InductionSummaryComponent` to handle `is_admin`, ensuring correct rendering.

### Why
The **extensions section** in the teacher view incorrectly displays **Add/View** links in the **admin console**. These links should only appear in the **Appropriate Body (AB) interface**, but the shared component wasn’t handling this correctly.

This update ensures that the `DetailsComponent` behaves correctly across different views.

[Sentry errors](https://dfe-teacher-services.sentry.io/issues/6223619216/?environment=production&project=4508369974788096&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&sort=date&stream_index=0)

### Screenshot

image

### DIFF
--- a/app/components/teachers/details_component.rb
+++ b/app/components/teachers/details_component.rb
@@ -8,8 +8,8 @@ module Teachers
       Teachers::Details::ITTDetailsComponent.new(teacher:)
     }
 
-    renders_one :induction_summary, -> {
-      Teachers::Details::InductionSummaryComponent.new(teacher:)
+    renders_one :induction_summary, ->(is_admin: false) {
+      Teachers::Details::InductionSummaryComponent.new(teacher:, is_admin:)
     }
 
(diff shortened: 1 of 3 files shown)


peteryates:
Would it make more sense to just have a separate components for the admin and school (and AB) sections of the app?

avinhurry:
> Would it make more sense to just have a separate components for the admin and school (and AB) sections of the app?

@peteryates thats a good question and one I did consider. 

Most of the logic was already structured to dynamically switch between admin and AB, and it felt solid. If we expect this component to grow significantly in different directions for AB/Support, I agree that splitting it might make sense.

For now, I didn’t see enough return on investment to justify the change - but do you think the complexity of this component will increase a lot?

PR #365, opened 2025-03-20
Fix performance cops

peterdavidhamilton:


avinhurry (approved):
Nice! 

Thank you for doing these housekeeping tasks

PR #369, opened 2025-03-21
Update api-versioning to fix typos and tweak some sentences.md

bbelward:
Made cosmetic changes so the content reads more smoothly and uses the correct terminology - e.g delivery ECF policy because deliver ECF based training. Appreciate it should probably be deliver Early Career Teacher Entitlement but given this is internal facing have tweaked slightly.


shahadRah (approved):
LGTM! Thanks for the review

avinhurry (approved):

PR #372, opened 2025-03-24
Fix Capybara Rubocop Violations

avinhurry:
### Context

We're gradually working through the rubocop violations listed in `.rubocop_todo.yml`.

### Changes proposed

Here we clean up the following violations: 

- **Abide by** [Capybara/CurrentPathExpectation](https://www.rubydoc.info/gems/rubocop-rspec/1.19.0/RuboCop/Cop/RSpec/Capybara/CurrentPathExpectation) (this change excludes the feature tests from this cop because we use Playwright (not Capybara) for those. Leaving it enabled elsewhere in case Capybara matchers are used in other test types.)

- **Exclude feature tests from** [Capybara/VisibilityMatcher](https://www.rubydoc.info/gems/rubocop-capybara/RuboCop/Cop/Capybara/VisibilityMatcher) (this change helps to avoid ambiguity by being more explicit and using the preferred `:visible` symbol instead of `true`.)

### Guidance to review

I've addressed each cop in single commit, so it would make the most sense to review commit by commit.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -54,21 +54,6 @@ RSpec/ContextWording:
 RSpec/DescribedClass:
   Enabled: false
 
-# Offense count: 11
-# This cop supports safe autocorrection (--autocorrect).
-RSpec/EmptyLineAfterFinalLet:
-  Exclude:
-    - 'spec/components/appropriate_bodies/claim_ect_actions_component_spec.rb'
-    - 'spec/requests/appropriate_bodies/claim_an_ect/check_ect_spec.rb'
-    - 'spec/requests/appropriate_bodies/claim_an_ect/register_ect_spec.rb'
-    - 'spec/services/appropriate_bodies/claim_an_ect/register_ect_spec.rb'
-    - 'spec/services/teachers/extensions/create_extension_spec.rb'
-    - 'spec/services/teachers/extensions/update_extension_spec.rb'
-    - 'spec/views/schools/register_mentor_wizard/check_answers.html.erb_spec.rb'
-    - 'spec/wizards/schools/register_ect_wizard/find_ect_step_spec.rb'
-    - 'spec/wizards/schools/register_mentor_wizard/find_mentor_step_spec.rb'
-    - 'spec/wizards/schools/register_mentor_wizard/national_insurance_number_step_spec.rb'
-
 # Offense count: 6
 # This cop supports safe autocorrection (--autocorrect).
 # Configuration parameters: CustomTransform, IgnoredWords, DisallowedExamples.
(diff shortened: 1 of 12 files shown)


.rubocop_todo.yml
```diff
@@ -58,17 +47,27 @@ RSpec/ContextMethod:
...
 # SupportedStyles: described_class, explicit
 RSpec/DescribedClass:
   Enabled: false
 
-# Offense count: 52
+# Offense count: 11
 # This cop supports safe autocorrection (--autocorrect).
 RSpec/EmptyLineAfterFinalLet:
-  Enabled: false
+  Exclude:
```
peterdavidhamilton, line 60: These can be fixed before being added to the todo

.rubocop_todo.yml
```diff
@@ -1,22 +1,11 @@
 # This configuration was generated by
-# `rubocop --auto-gen-config --no-auto-gen-timestamp`
-# using RuboCop version 1.74.0.
+# `rubocop --auto-gen-config`
```
peterdavidhamilton, line 2: There is a `bin/rubocop-todo` script

.rubocop_todo.yml
```diff
@@ -58,17 +47,27 @@ RSpec/ContextMethod:
 RSpec/ContextWording:
   Enabled: false
 
-# Offense count: 80
+# Offense count: 85
```
peterdavidhamilton, line 50: This number has increased but I don't think it should have. Maybe a rebase?

.rubocop_todo.yml
```diff
@@ -137,7 +136,7 @@ RSpec/MatchArray:
     - 'spec/models/mentor_at_school_period_spec.rb'
     - 'spec/models/provider_partnership_spec.rb'
 
-# Offense count: 552
+# Offense count: 566
```
peterdavidhamilton, line 139: Same here

peterdavidhamilton (changes requested):

.rubocop_todo.yml
avinhurry: I think it might have, it's 84 even with a rebase

.rubocop_todo.yml
avinhurry: Thanks - a rebase has fixed this one 👍🏽

.rubocop_todo.yml
avinhurry: Ah i didn't realise, thanks. I have used that now.

.rubocop_todo.yml
avinhurry: My plan was to keep this PR confined to the capybara cops but, I can make these changes here if you like: 

https://github.com/DFE-Digital/register-early-career-teachers-public/pull/372/commits/0310985494e6f65574c6fbc0f5fa6af27d9a8dcf

peterdavidhamilton (approved):

peterdavidhamilton:
🖖🏼

PR #374, opened 2025-03-24
Enable content security policy (CSP)

peteryates:
Before/after

```diff
  HTTP/2.0 200 OK
  cache-control: max-age=0, private, must-revalidate
  content-length: 0
+ content-security-policy: default-src 'self' https:; font-src 'self' https: data:; img-src 'self' https: data:; object-src 'none'; script-src 'self' https: 'nonce-'; style-src 'self' https: 'nonce-'
  content-type: text/html; charset=utf-8
  link: </assets/application-a2ba66bf.css>; rel=preload; as=style; nopush
  referrer-policy: strict-origin-when-cross-origin
  strict-transport-security: max-age=31536000; includeSubDomains
  vary: Accept
  x-content-type-options: nosniff
  x-frame-options: SAMEORIGIN
  x-permitted-cross-domain-policies: none
  x-xss-protection: 0
```

Note, [viewing in dev tools won't show nonces](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/nonce#accessing_nonces_and_nonce_hiding) (seriously it took me like an hour of debugging to work out why they were all blank 😭) but if you view source they're present.

image

### DIFF
--- a/config/initializers/content_security_policy.rb
+++ b/config/initializers/content_security_policy.rb
@@ -1,25 +1,13 @@
-# Be sure to restart your server when you modify this file.
+Rails.application.configure do
+  config.content_security_policy do |policy|
+    policy.default_src :self, :https
+    policy.font_src    :self, :https, :data
+    policy.img_src     :self, :https, :data
+    policy.object_src  :none
+    policy.script_src  :self, :https
+    policy.style_src   :self, :https
+  end
 
-# Define an application-wide content security policy.
-# See the Securing Rails Applications Guide for more information:
-# https://guides.rubyonrails.org/security.html#content-security-policy-header
-
-# Rails.application.configure do
-#   config.content_security_policy do |policy|
-#     policy.default_src :self, :https
-#     policy.font_src    :self, :https, :data
-#     policy.img_src     :self, :https, :data
-#     policy.object_src  :none
-#     policy.script_src  :self, :https
-#     policy.style_src   :self, :https
-#     # Specify URI for violation reports
-#     # policy.report_uri "/csp-violation-report-endpoint"
-#   end
-#
-#   # Generate session nonces for permitted importmap, inline scripts, and inline styles.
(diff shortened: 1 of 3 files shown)


config/initializers/content_security_policy.rb
```diff
@@ -1,25 +1,13 @@
-# Be sure to restart your server when you modify this file.
```
avinhurry, line 1: What are your thoughts around leaving these comments in?

config/initializers/content_security_policy.rb
```diff
@@ -1,25 +1,13 @@
...
+    policy.font_src    :self, :https, :data
+    policy.img_src     :self, :https, :data
+    policy.object_src  :none
+    policy.script_src  :self, :https
+    policy.style_src   :self, :https
+  end
 
-# Define an application-wide content security policy.
-# See the Securing Rails Applications Guide for more information:
-# https://guides.rubyonrails.org/security.html#content-security-policy-header
```
avinhurry, line 5: What are your thoughts around leaving these comments in?

avinhurry (approved):

config/initializers/content_security_policy.rb
```diff
@@ -1,25 +1,13 @@
...
-#   end
-#
-#   # Generate session nonces for permitted importmap, inline scripts, and inline styles.
-#   config.content_security_policy_nonce_generator = ->(request) { request.session.id.to_s }
-#   config.content_security_policy_nonce_directives = %w(script-src style-src)
-#
-#   # Report violations without enforcing the policy.
-#   # config.content_security_policy_report_only = true
-# end
+  config.content_security_policy_nonce_generator = ->(request) { request.session[:nonce] ||= SecureRandom.hex }
```
avinhurry, line 11: Just wondering if you considered generating the nonce per request instead of per session?

peteryates: It will interfere with caching in difficult to detect ways so opted per session.

PR #379, opened 2025-03-25
Wrap add induction link in role-based conditional

peterdavidhamilton:
Appropriate Body
image
Admin
image

Fixes #DFE-Digital/register-ects-project-board#1411


claire-hughez:
I've roughly prod reviewed this and can confirm:
- Add an induction period still works as intended on the admin console
- ABs can no longer see 'add an induction period'

But might want @joe-harrison-dfe or someone else in Emily P's absence to do another check that knows the AB stuff better! :)

peteryates (approved):

PR #382, opened 2025-03-25
Update email label to discourage use of generic addresses

avinhurry:
## Context

Update the email field label to discourage use of generic email addresses (e.g. admin@example.com).
## Changes proposed

- Extracted a new helper method `generic_email_label` in `ContentHelper` to render the label text.
- Included a `govuk_link_to` link styled with `govuk-link--no-visited-state`.
- Updated the content on the form page to use this new helper.
- Use smart quotes

## Before / After

| Before | After |
|--------|-------|
| image | image |

## Notes and Observations

This change affects the **Register ECT** email page in the following places:

- **When registering an ECT**  
  `app/views/schools/register_ect_wizard/_email_address.html.erb`

- **When editing a new ECT from the check answers page**  
  `app/views/schools/register_ect_wizard/change_email_address.html.erb`

---

**Mentor creation and editing** use a *different* partial (with the same filename):

- **When creating a mentor**  
  `app/views/schools/register_mentor_wizard/email_address.html.erb`

- **When editing a mentor**  
  `app/views/schools/register_mentor_wizard/change_email_address.html.erb`

### DIFF
--- a/app/helpers/content_helper.rb
+++ b/app/helpers/content_helper.rb
@@ -1,0 +1,9 @@
+module ContentHelper
+  def generic_email_label
+    safe_join([
+      "Do not use a generic email like ",
+      govuk_link_to("admin@example.com", "#", no_visited_state: true),
+      "."
+    ])
+  end
+end
--- a/app/views/schools/register_ect_wizard/_email_address.html.erb
+++ b/app/views/schools/register_ect_wizard/_email_address.html.erb
@@ -1,15 +1,17 @@
 <% page_data(
-    title: "What is #{@ect.full_name}'s email address?",
+    title: "What is #{@ect.full_name}’s email address?",
     error: @wizard.current_step.errors.present?,
     backlink_href: @wizard.previous_step_path,
 ) %>
 
-<p class="govuk-body">You must provide an email address that your ECT can access. You can update it later if needed.</p>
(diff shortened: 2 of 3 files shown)


peterdavidhamilton (approved):

app/helpers/content_helper.rb
```diff
@@ -0,0 +1,9 @@
+module ContentHelper
+  def generic_email_label
+    safe_join([
+      "Do not use a generic email like ",
+      govuk_link_to("admin@example.com", "#", no_visited_state: true),
```
ltello, line 5: Wouldn't it be more meaningful to use `headteacher@your_school.com` instead?

avinhurry: This is already merged now but we can do a follow up if needed.

Do you have any thoughts on that @robthompson16?

robthompson16: I'm fairly agnostic about what the email address says so happy for it to be changed if we think it will be beneficial to users.

Do the generic emails we have in the database tend to have the term 'headteacher' rather than 'admin'? I'm trying to think about what's more commonly used.

I like your suggestion but wonder whether we could simplify to [headteacher@school.com](mailto:headteacher@school.com)?

PR #383, opened 2025-03-25
Update the SchoolPeriods builder with the required ECTAtSchoolPeriod attributes

leoapost:
Ticket: [#1399](https://github.com/DFE-Digital/register-ects-project-board/issues/1385)

New attributes were introduced in ECTAtSchoolPeriod, which broke the migrations and led to some specs being disabled.

This PR updates the migration files to include the new attributes and re-enables the previously disabled specs.

Changes:
- Update the builder and the extractor to work with the new attributes
- Add a mapper class to map the programme type values from ECF1 to ECF2 values
- Update and re-enable the disabled specs

### DIFF
--- a/app/migration/builders/ect/school_periods.rb
+++ b/app/migration/builders/ect/school_periods.rb
@@ -14,9 +14,7 @@ module Builders
           school = School.find_by!(urn: period.urn)
           ::ECTAtSchoolPeriod.create!(teacher:,
                                       school:,
-                                      appropriate_body_type: period.appropriate_body_type,
                                       programme_type: period.programme_type,
-                                      appropriate_body_id: period.appropriate_body_id,
                                       lead_provider_id: period.lead_provider_id,
                                       started_on: period.start_date,
                                       finished_on: period.end_date,
--- a/app/migration/builders/ect/school_periods.rb
+++ b/app/migration/builders/ect/school_periods.rb
@@ -14,6 +14,10 @@ module Builders
           school = School.find_by!(urn: period.urn)
           ::ECTAtSchoolPeriod.create!(teacher:,
                                       school:,
+                                      appropriate_body_type: period.appropriate_body_type,
+                                      programme_type: period.programme_type,
+                                      appropriate_body_id: period.appropriate_body_id,
+                                      lead_provider_id: period.lead_provider_id,
                                       started_on: period.start_date,
                                       finished_on: period.end_date,
                                       ecf_start_induction_record_id: period.start_source_id,
--- a/app/migration/mappers/programme_type_mapper.rb
+++ b/app/migration/mappers/programme_type_mapper.rb
@@ -1,0 +1,19 @@
+module Mappers
+  class ProgrammeTypeMapper
+    # Maps the ECF1 values to the ECF2 values
+    MAPPING = {
+      "core_induction_programme" => "school_led",
+      "design_our_own" => "school_led",
+      "school_funded_fip" => "provider_led",
+      "full_induction_programme" => "provider_led"
+    }.freeze
+
+    def initialize(value)
+      @value = value
+    end
+
+    def mapped_value
+      MAPPING[@value]
+    end
+  end
+end
--- a/app/migration/mappers/training_programme_type_mapper.rb
+++ b/app/migration/mappers/training_programme_type_mapper.rb
@@ -1,5 +1,5 @@
 module Mappers
-  class ProgrammeTypeMapper
+  class TrainingProgrammeTypeMapper
     # Maps the ECF1 values to the ECF2 values
     MAPPING = {
       "core_induction_programme" => "school_led",
--- a/app/migration/school_period_extractor.rb
+++ b/app/migration/school_period_extractor.rb
@@ -33,7 +33,7 @@ private
     induction_records.each_with_object([]) do |induction_record, periods|
       record_school = induction_record.induction_programme.school_cohort.school
       extracted_training_programme = induction_record.induction_programme.training_programme
-      programme_type = Mappers::ProgrammeTypeMapper.new(extracted_training_programme).mapped_value
+      programme_type = Mappers::TrainingProgrammeTypeMapper.new(extracted_training_programme).mapped_value
 
       if current_school != record_school
         current_school = record_school
--- a/app/migration/school_period_extractor.rb
+++ b/app/migration/school_period_extractor.rb
@@ -32,6 +32,8 @@ private
 
     induction_records.each_with_object([]) do |induction_record, periods|
       record_school = induction_record.induction_programme.school_cohort.school
+      extracted_training_programme = tinduction_record.induction_programme.training_programme
+      programme_type = Mappers::ProgrammeTypeMapper.new(extracted_training_programme).mapped_value
 
       if current_school != record_school
         current_school = record_school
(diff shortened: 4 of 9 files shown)


app/migration/school_period_extractor.rb
```diff
@@ -40,7 +42,11 @@ def build_school_periods_from_induction_records
                                                      start_date: induction_record.start_date,
                                                      end_date: induction_record.end_date,
                                                      start_source_id: induction_record.id,
-                                                     end_source_id: induction_record.id)
+                                                     end_source_id: induction_record.id,
+                                                     appropriate_body_type: induction_record.appropriate_body.body_type,
+                                                     programme_type:,
+                                                     appropriate_body_id: induction_record.appropriate_body_id,
+                                                     lead_provider_id: nil)
```
tonyheadford, line 49: I think we need to dig more here otherwise we'll get odd failures.
 - if the AB is not set on the induction record we need to go look at the `induction_programme.school_cohort.appropriate_body` and use that if it is present
 - can `lead_provider_id` only be `nil` if `programme_type` is `school_led`?

app/migration/school_period_extractor.rb
```diff
@@ -32,6 +32,8 @@ def build_school_periods_from_induction_records
 
     induction_records.each_with_object([]) do |induction_record, periods|
       record_school = induction_record.induction_programme.school_cohort.school
+      extracted_training_programme = tinduction_record.induction_programme.training_programme
```
tonyheadford, line 35: typo here? `tinduction_record`

app/migration/mappers/programme_type_mapper.rb
```diff
@@ -0,0 +1,19 @@
+module Mappers
+  class ProgrammeTypeMapper
```
tonyheadford, line 2: Could this be named `TrainingProgrammeTypeMapper` or `InductionProgrammeTypeMapper` just to be super clear?

app/migration/mappers/programme_type_mapper.rb
```diff
@@ -0,0 +1,19 @@
...
+      "school_funded_fip" => "provider_led",
+      "full_induction_programme" => "provider_led"
+    }.freeze
+
+    def initialize(value)
+      @value = value
+    end
+
+    def mapped_value
+      MAPPING[@value]
```
tonyheadford, line 16: I wonder if this should `raise` if there is no mapping?  What do you think?

app/migration/builders/ect/school_periods.rb
```diff
@@ -14,6 +14,10 @@ def build
           school = School.find_by!(urn: period.urn)
           ::ECTAtSchoolPeriod.create!(teacher:,
                                       school:,
+                                      appropriate_body_type: period.appropriate_body_type,
+                                      programme_type: period.programme_type,
+                                      appropriate_body_id: period.appropriate_body_id,
+                                      lead_provider_id: period.lead_provider_id,
```
peteryates, line 20: It's my understanding that `appropriate_body_type`, `lead_provider_id` and `appropriate_body_id` are just temporarily present on `ECTAtSchoolPeriod` and  in the longer term they'll shift to `InductionPeriod` and `TrainingPeriod` respectively.

Maybe we can put them together here and leave a comment, just to inform other people reading it that it's not going to be there forever.

app/migration/school_period_extractor.rb
leoapost: Good catch. Fixed.

app/migration/builders/ect/school_periods.rb
leoapost: I'm wondering if it's better to add a note in the classes where the temp attributes are defined. 

I added one in the `Migration::SchoolPeriod` class for now.

app/migration/mappers/programme_type_mapper.rb
leoapost: I checked the values in the database and they look all they have one of the keys in the MAPPING. Don't expect any nil values.

app/migration/mappers/training_programme_type_mapper.rb
```diff
@@ -0,0 +1,19 @@
+module Mappers
+  class TrainingProgrammeTypeMapper
+    # Maps the ECF1 values to the ECF2 values
+    MAPPING = {
+      "core_induction_programme" => "school_led",
+      "design_our_own" => "school_led",
+      "school_funded_fip" => "provider_led",
+      "full_induction_programme" => "provider_led"
+    }.freeze
```
avinhurry, line 9: Just a question for my own understanding: 

Is `training_programme` on `InductionProgramme` from ECF1 equivalent to `programme_type` on `ECTAtSchoolPeriod` on ECF2?

app/migration/mappers/training_programme_type_mapper.rb
```diff
@@ -0,0 +1,19 @@
+module Mappers
+  class TrainingProgrammeTypeMapper
```
avinhurry, line 2: Minor naming thought - since it's already name spaced under `Mappers`, would it make sense to just call the class `TrainingProgramme` instead of `TrainingProgrammeTypeMapper`?

app/migration/mappers/training_programme_type_mapper.rb
```diff
@@ -0,0 +1,19 @@
...
+      "design_our_own" => "school_led",
+      "school_funded_fip" => "provider_led",
+      "full_induction_programme" => "provider_led"
+    }.freeze
+
+    def initialize(value)
+      @value = value
+    end
+
+    def mapped_value
```
avinhurry, line 15: If you changed `TrainingProgrammeTypeMapper` to `TrainingProgrammeType` then this could be changed to `value`.

That means:

`Mappers::TrainingProgrammeTypeMapper.new(extracted_training_programme).mapped_value`

Would become 

`Mappers::TrainingProgrammeType.new(extracted_training_programme).value` 

Which feels cleaner and a bit less repetitive.

app/migration/mappers/training_programme_type_mapper.rb
leoapost: @avinhurry correct

app/migration/mappers/training_programme_type_mapper.rb
leoapost: `mapped_value` makes it clear that you’re getting the converted type rather than the original. With just `value`, it’s less obvious, and I’d feel the need to double check the code to be sure.

app/migration/school_period_extractor.rb
```diff
@@ -66,4 +70,16 @@ def mentor?
...
+
+  def discover_programme_type(induction_record)
+    extracted_training_programme = induction_record.induction_programme.training_programme
+    Mappers::TrainingProgrammeTypeMapper.new(extracted_training_programme).mapped_value
+  end
+
+  def discover_lead_provider_id(induction_record, programme_type)
+    return nil if programme_type == 'school_led'
+
+    extracted_lead_provider_name = induction_record.induction_programme.partnership.lead_provider.name
```
tonyheadford, line 82:
```suggestion
    return nil if induction_record.induction_programme.partnership.nil?
    extracted_lead_provider_name = induction_record.induction_programme.partnership.lead_provider.name
```

tonyheadford (approved):
Nice

PR #394, opened 2025-03-27
Added support link to footer and iterated support page

PriyaNahar2:
Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/1415

Added link to footer and refined content on support page. Content reviewed by @robthompson16 

Before and after screenshots: 
| Before | After |
|  ------ | ------ |
| Screenshot 2025-03-27 at 12 08 46 | Screenshot 2025-03-27 at 12 08 52 |

Before and after screenshots: 
| Before | After |
|  ------ | ------ |
| Screenshot 2025-03-27 at 12 08 38 | Screenshot 2025-03-27 at 12 22 00 |

### DIFF
--- a/app/views/pages/support.html.erb
+++ b/app/views/pages/support.html.erb
@@ -1,5 +1,5 @@
 <% page_data(title: "Contact support", error: false) %>
 
 <p class="govuk-body">
-  If you have a question, or you’ve had a problem using this service, contact us at <%= support_mailto_link %>.
+  If you had a problem using this service or have a question about it, email <%= support_mailto_link %>.
 </p>
--- a/app/views/layouts/shared/_footer.html.erb
+++ b/app/views/layouts/shared/_footer.html.erb
@@ -4,6 +4,7 @@
       'Accessibility' => '/accessibility',
       'Cookies' => '/cookies',
       'Privacy' => '/privacy',
+      'Contact support' => '/support'
     }
   )
 %>


peteryates (approved):

app/views/pages/support.html.erb
```diff
@@ -1,5 +1,5 @@
 <% page_data(title: "Contact support", error: false) %>
 
 <p class="govuk-body">
-  If you have a question, or you’ve had a problem using this service, contact us at <%= support_mailto_link %>.
+  If you had a problem using this service or have a question about it, email <%= support_mailto_link %>.
```
peteryates, line 4: Much clearer.

joe-harrison-dfe:
All looks good to me !

joe-harrison-dfe (approved):

app/views/pages/support.html.erb
avinhurry: Just wondering about the use of past tense - should we consider present tense in case someone's currently having a problem?

I'm not sure what the intended use of this will be though so this might be a silly question.

PriyaNahar2: No that's a good point!! Me and @robthompson16 played around with the content a lot and although it's past tense it's still correct. The current ECF1 service also has similar content and other services too - so kept it in line with other services. 

Not a silly question at all... believe it or not, Rob and I went through circles for this one sentence lol

avinhurry: Thanks - I went in circles trying to figure out why it was past tense 🤣

PR #395, opened 2025-03-27
Remove unnecessary top margins

avinhurry:
### Context

This removes unnecessary `govuk-!-margin-top-5` classes that wasn't affecting the layout. Simplifies the markup slightly without changing the visual appearance.

Recommend hiding whitespace changes to make the diff clearer - it’s a very small change.


peteryates (approved):
🧹

PR #399, opened 2025-03-28
[1399] -  add type to appropriate bodies

ltello:
The first versions of register ECT journey recorded everything programme-wise into the ECTAtSchoolPeriod instance created. 

Now that we have decided the types ABs can have on `ECF2` (`teaching_school_hub`, `local_authority`, `national`), this PR will add that field to the model and remove it from  `ECTAtSchoolPeriod` and `School` (last registration choices).

This change needs the journey steps themselves to be accommodated accordingly.

[Issue](https://github.com/DFE-Digital/register-ects-project-board/issues/1399)

### DIFF
--- a/app/models/appropriate_body.rb
+++ b/app/models/appropriate_body.rb
@@ -1,6 +1,4 @@
 class AppropriateBody < ApplicationRecord
-  ISTIP = 'Independent Schools Teacher Induction Panel (ISTIP)'.freeze
-
   # Enums
   enum :body_type,
        { local_authority: 'local_authority',
@@ -38,8 +36,4 @@ class AppropriateBody < ApplicationRecord
               message: 'Must be a number between 1000 and 9999',
               allow_blank: true
             }
-
-  def self.istip
-    find_by_name(ISTIP)
-  end
 end
--- a/app/models/appropriate_body.rb
+++ b/app/models/appropriate_body.rb
@@ -1,5 +1,12 @@
 class AppropriateBody < ApplicationRecord
-  NAME_FOR_TEACHING_INDUCTION_PANEL_TYPE = 'Independent Schools Teacher Induction Panel (ISTIP)'.freeze
+  ISTIP = 'Independent Schools Teacher Induction Panel (ISTIP)'.freeze
+
+  # Enums
+  enum :body_type,
+       { local_authority: 'local_authority',
+         national: 'national',
+         teaching_school_hub: 'teaching_school_hub' },
+       validate: { message: "Must be local authority, national or teaching school hub" }
 
   # Associations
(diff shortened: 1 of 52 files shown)


app/models/appropriate_body.rb
```diff
@@ -1,5 +1,14 @@
...
+  ISTIP = 'Independent Schools Teacher Induction Panel (ISTIP)'.freeze
+
+  AppropriateBody.inheritance_column = nil
+
+  # Enums
+  enum :type,
+       { local_authority: 'local_authority',
+         national: 'national',
+         teaching_school_hub: 'teaching_school_hub' },
+       validate: { message: "Must be local authority, national or teaching school hub" }
```
tonyheadford, line 11: Could we use something other than `type` for the name e.g. `ab_type` or `body_type` instead just so we don't cause confusion and stick to the Rails conventions?

peteryates: Probably worth getting input on this from @emilyprudencedfe or @priyanahar2

ltello: No problem to rename it but Rails provides `inheritance_column` precisely for this

tonyheadford (approved):
Looks great!

app/models/appropriate_body.rb
```diff
@@ -31,4 +38,8 @@ class AppropriateBody < ApplicationRecord
               message: 'Must be a number between 1000 and 9999',
               allow_blank: true
             }
+
+  def self.istip
+    find_by_name(ISTIP)
+  end
```
peteryates, line 44: This feels a little business logic-y for a method on the model. We already have `AppropriateBodies::Search`, could this just be `AppropriateBodies::Search.new.istip`?

app/models/ect_at_school_period.rb
```diff
@@ -17,7 +11,7 @@ class ECTAtSchoolPeriod < ApplicationRecord
   # Associations
   belongs_to :school, inverse_of: :ect_at_school_periods
   belongs_to :teacher, inverse_of: :ect_at_school_periods
-  belongs_to :appropriate_body
+  belongs_to :school_reported_appropriate_body, class_name: 'AppropriateBody', optional: true
```
peteryates, line 14: We have `belongs_to_required_by_default = false` [set for the app](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/config/application.rb#L20) so we can skip the optional here. It's set that way so Rails doesn't emit errors on the relationship rather than on the foreign key (which means error summary links don't work)

app/models/appropriate_body.rb
ltello: Sure, I am not familiar with that SO.

PR #400, opened 2025-03-28
[1409] Add admin service to unclaim ECT

ghost:
https://github.com/DFE-Digital/register-ects-project-board/issues/1409

To unclaim we need to tell TRS API that the ECT's induction status should be changed to `RequiredToComplete` and set`start_date`, `completed_date` to `nil`

### DIFF
--- a/spec/jobs/purge_pending_induction_submissions_job_spec.rb
+++ b/spec/jobs/purge_pending_induction_submissions_job_spec.rb
@@ -1,5 +1,3 @@
-require "rails_helper"
-
 RSpec.describe PurgePendingInductionSubmissionsJob, type: :job do
   describe "#perform" do
     let(:appropriate_body) { FactoryBot.create(:appropriate_body) }
--- a/spec/jobs/teachers/sync_teacher_with_trs_job_spec.rb
+++ b/spec/jobs/teachers/sync_teacher_with_trs_job_spec.rb
@@ -1,5 +1,3 @@
-require "rails_helper"
-
 RSpec.describe Teachers::SyncTeacherWithTRSJob, type: :job do
   describe "#perform" do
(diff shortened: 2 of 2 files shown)


app/models/event.rb
```diff
@@ -8,6 +8,7 @@ class Event < ApplicationRecord
     appropriate_body_releases_teacher
     appropriate_body_fails_teacher
     appropriate_body_passes_teacher
+    support_revert_teacher_claim
```
peteryates, line 11: I think this should be `admin_` as the class ~should~ will soon be callable from the admin console.

app/services/appropriate_bodies/claim_an_ect/revert_claim.rb
```diff
@@ -0,0 +1,26 @@
...
+        @appropriate_body = appropriate_body
+        @teacher = teacher
+        @author = author
+        @induction_period = induction_period
+      end
+
+      def revert_claim
+        ActiveRecord::Base.transaction do
+          induction_period.destroy!
+          ResetInductionJob.perform_later(trn: teacher.trn)
```
peteryates, line 16: If the ECT has any induction periods at all they're `InProgress` from TRS's point of view, so we should only reset it if we're deleting their only `InductionPeriod`.

peteryates (changes requested):
We need to only reset them if they have no other induction periods, not whenever we delete an IP.

.rubocop/cop/rspec/require_rails_helper.rb
peteryates, line 1: Nice.

peteryates (approved):
Looks good. Chatted with @edujackedu and we're going to do a follow up that consolidates some of the event names and tidies things.

PR #403, opened 2025-03-31
[1421] Add a job to sync teachers data with TRS

ghost:
We need to schedule the TRS sync to happen regularly. We agreed that this should happen hourly and in batches.

We have approx 75k teachers currently and aim to sync them all every 3 days. If we do 1200 an hour (1 every 3 seconds) we'll be within our target.

- the hourly job queues up 1200 individual teacher sync jobs that call Teachers::RefreshAttributes spread evenly across the hour
- the sync jobs runs at a low priority

### DIFF
--- a/spec/jobs/teachers/schedule_trs_sync_job_spec.rb
+++ b/spec/jobs/teachers/schedule_trs_sync_job_spec.rb
@@ -1,5 +1,3 @@
-require "rails_helper"
-
 RSpec.describe Teachers::ScheduleTRSSyncJob, type: :job do
   describe "#perform" do
     let!(:teacher1) { FactoryBot.create(:teacher, trs_data_last_refreshed_at: 5.days.ago) }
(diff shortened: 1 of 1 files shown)


spec/jobs/teachers/schedule_trs_sync_job_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+RSpec.describe Teachers::ScheduleTRSSyncJob, type: :job do
+  describe "#perform" do
+    let!(:teacher1) { FactoryBot.create(:teacher, trs_data_last_refreshed_at: 5.days.ago) }
+    let!(:teacher2) { FactoryBot.create(:teacher, trs_data_last_refreshed_at: 4.days.ago) }
+    let!(:teacher3) { FactoryBot.create(:teacher, trs_data_last_refreshed_at: 3.days.ago) }
+    let!(:teacher4) { FactoryBot.create(:teacher, trs_data_last_refreshed_at: 2.days.ago) }
+    let!(:teacher5) { FactoryBot.create(:teacher, trs_data_last_refreshed_at: 1.day.ago) }
+
+    it "schedules sync jobs for teachers ordered by trs_data_last_refreshed_at" do
+      ordered_teachers = Teacher.order(trs_data_last_refreshed_at: :asc).limit(described_class::BATCH_SIZE)
```
peterdavidhamilton, line 12: Does defining an array of the teachers and seconds in the order you expect work? Rather than repeat the same code from job.

spec/jobs/teachers/schedule_trs_sync_job_spec.rb
```diff
@@ -0,0 +1,47 @@
+require "rails_helper"
```
peterdavidhamilton, line 1: Pete removed all the rails helpers

peterdavidhamilton (changes requested):
Looks good, just cautious about code in specs that mirrors the implementation

spec/jobs/teachers/schedule_trs_sync_job_spec.rb
peterdavidhamilton: I completely agree about keeping it dumb and not having logic in the tests; I was suggesting this. What do you think?

```ruby
    it "schedules sync jobs for teachers ordered by trs_data_last_refreshed_at" do
      [
        [teacher1, 0],
        [teacher2, 3],
        [teacher3, 6],
        [teacher4, 9],
        [teacher5, 12],
      ].each do |teacher, wait_time|
        allow(Teachers::SyncTeacherWithTRSJob).to receive(:set).with(wait: wait_time.seconds).and_return(Teachers::SyncTeacherWithTRSJob)

        expect(Teachers::SyncTeacherWithTRSJob).to receive(:perform_later).with(teacher:)
      end      

      described_class.perform_now
    end
```

peterdavidhamilton (approved):
👍🏼

PR #407, opened 2025-04-01
Add events tracking to induction extensions

peterdavidhamilton:
Redo of https://github.com/DFE-Digital/register-early-career-teachers-public/pull/370

- add events to create/update extension
- combine into a single service
- remove extensions from seeds for teachers who have no induction periods
- add rake task to backfill events


peteryates (approved):
Looks good!

PR #419, opened 2025-04-03
Fix `db:prepare` rake task when ran without database

ethax-ross:
I ran into this when I was following the setup documentation.

When setting up the app from scratch we run `bin/setup`, which calls `db:preapre`. If a database has not been created yet, it will attempt to run the `setup_search_configuration` enhancement (which is invoked _before_ `db:prepare` as we pass it as an argument to `enhance()`) and it will fail as it modifies the database (which doesn't exist yet).

The `db:prepare` command will succeed if `db:create` is ran first. We have a bit of a chicken/egg scenario here and its complicated by the fact that `db:prepare` does not invoke the `db:create` enhancements (so we can't just enhance `db:create` with `db:setup_search_configuration` in a block to ensure it runs after).

I can't see a particularly nice way around this; the best I could come up with is to modify the `setup_search_configuration` rake task to create the database if it doesn't exist. We can also enhance `db:create` with a block to setup the search config _after_ the database is created (if created via `db:create`).

This means we can run `db:preapre` when the database both exists and doesn't exist and `db:create` independently. We should be fine to drop `db:schema:load` as it will be covered by one of the other two scenarios (you'd either run `db:create` or `db:preapre` first anyway).


peteryates (approved):
Works like a charm, thanks!

Issue #417, opened 2025-04-02
Refine event names


peteryates:
We have encountered a couple of areas where the attribution portion of the event name makes recording them unclear. For example, when an `InductionPeriod` is created, which flavour of event should be triggered?

## Example of confusion

Currently, we have two services that create induction periods, `Admin::CreateInductionPeriod` (for admins) and `InductionPeriods::CreateInductionPeriod` (for ECTs).

They mostly the same thing so we should really only have `InductionPeriods::CreateInductionPeriod` for creating periods. Admins can create them in the past so we'll need the extra `notify_trs_of_new_induction_start` - but the admin service can call the generic one.

## Renaming events

If we make the event names a little more generic we don't really lose much context because the `author_type` still tells us which kind of user initiated the change.

We can rename the events , perhaps like this:

| Old | Suggested |
| --- | --------- |
| admin_creates_induction_period          | induction_period_opened  |
| admin_updates_induction_period          | induction_period_updated |
| admin_fails_teacher_induction           | teacher_fails_induction |
| admin_passes_teacher_induction          | teacher_passes_induction |
| admin_reverts_teacher_claim             | induction_period_deleted |
| admin_deletes_induction_period          | induction_period_deleted |
| appropriate_body_claims_teacher         | induction_period_opened |
| appropriate_body_releases_teacher       | induction_period_closed |
| appropriate_body_fails_teacher          | teacher_fails_induction |
| appropriate_body_passes_teacher         | teacher_passes_induction |
| teacher_name_updated_by_trs             | teacher_trs_details_refreshed |
| teacher_induction_status_updated_by_trs | teacher_trs_induction_status_updated |
| teacher_attributes_updated_from_trs     | teacher_trs_attributes_updated |
| teacher_imported_from_trs               | teacher_imported_from_trs |
| import_from_dqt                         | teacher_imported_from_dqt |

This renaming should be done as a run-once database script which can live in `db/scripts`. It's fine to use SQL strings to update them which will be quicker than looping through thousands of records. Let's test anything we do in the snapshot database first though.

graciebrennan:
going to merge the event names and will then be able to address the support ticket 1409

emily-prudence-dfe:
There will likely be another couple of sub-issues here

PR #423, opened 2025-04-03
420: Add induction_period_opened generic event and replace merged events

ghost:
## Description
- merge admin_creates_induction_period and appropriate_body_claims_teacher and replace with induction_period_opened. We decided not to specify the author type in the event name to we're creating a new generic event and merging the old ones into it.
- we have two services that create induction periods, Admin::CreateInductionPeriod (for admins) and InductionPeriods::CreateInductionPeriod (for ABs).They mostly the same thing so we have refactor the admin one to call the generic one (InductionPeriods::CreateInductionPeriod) for creating periods. Admins can create them in the past so we need the extra notify_trs_of_new_induction_start

More info on this work on parent task: https://github.com/DFE-Digital/register-early-career-teachers-public/issues/417

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -52,7 +52,7 @@ def describe_induction_period(ip)
   }
 
   Event.create!(
-    event_type: 'appropriate_body_claims_teacher',
+    event_type: 'induction_period_opened',
     induction_period: ip,
     teacher: ip.teacher,
     appropriate_body: ip.appropriate_body,
@@ -63,7 +63,7 @@ def describe_induction_period(ip)
 
   if ip.finished_on
     Event.create!(
-      event_type: 'appropriate_body_claims_teacher',
+      event_type: 'induction_period_opened',
       induction_period: ip,
       teacher: ip.teacher,
       appropriate_body: ip.appropriate_body,


peteryates (commented):
Looks good.

We should be able to drop `Events::Record.admin_creates_induction_period!` now too right?

app/services/induction_periods/create_induction_period.rb
```diff
@@ -6,28 +6,70 @@ class CreateInductionPeriod
...
       @appropriate_body = appropriate_body
       @started_on = started_on
       @induction_programme = induction_programme
+      @finished_on = finished_on
+      @number_of_terms = number_of_terms
     end
 
-    def create_induction_period(author:)
-      @induction_period = InductionPeriod.create(teacher:, appropriate_body:, started_on:, induction_programme:)
+    # @param author [Sessions::User]
```
peterdavidhamilton, line 28: 🤤

app/services/induction_periods/create_induction_period.rb
```diff
@@ -6,28 +6,70 @@ class CreateInductionPeriod
...
+        success or raise ActiveRecord::Rollback
       end
 
       @induction_period
     end
+
+  private
+
+    # @param author [Sessions::User]
+    # @param modifications [Hash{String => Array}]
```
peterdavidhamilton, line 62: 🤤

app/services/admin/create_induction_period.rb
```diff
@@ -1,64 +1,59 @@
...
+      # and the teacher hasn't already passed or failed induction
+      !teacher_has_earlier_induction_periods? && !teacher_has_passed_or_failed_induction?
+    end
+
+    def teacher_has_earlier_induction_periods?
+      InductionPeriod.where(teacher:).started_before(params[:started_on]).exists?
+    end
 
+    def teacher_has_passed_or_failed_induction?
+      # Check if the teacher has any induction periods with a pass or fail outcome
```
peterdavidhamilton, line 48: BTW if the comment is atop the definition it becomes part of the yardoc output.

peterdavidhamilton (approved):

peteryates (approved):
Nice, a good start. Left a thought, we can maybe do that as a followup or later refactoring.

PR #429, opened 2025-04-07
Bulk upload for claim, outcome and release

peterdavidhamilton:
Backend changes to introduce groups of pending submissions processed from uploaded CSV files.

Two new restful routes for `claims` and `actions` have been introduced for each type of `PendingInductionSubmissionBatch` record.

### WIP

Currently, the code converts a valid `PendingInductionSubmission` immediately into `Teacher` and `InductionPeriod` records.


spec/models/pending_induction_submission_batch_spec.rb
```diff
@@ -0,0 +1,142 @@
...
+        let(:csv_file) { Rack::Test::UploadedFile.new(Rails.root.join('spec/fixtures/foo.txt'), 'text/plain') }
+
+        specify do
+          expect(batch).not_to be_valid
+          expect(batch.errors[:csv_file]).to include("File type must be a CSV")
+        end
+      end
+
+      describe "#wrong_headers" do # softened rules so order doesn't matter - still TBC
+        let(:csv_file) { Rack::Test::UploadedFile.new(Rails.root.join('spec/fixtures/claims/invalid_missing_columns.csv'), 'text/csv') }
```
peteryates, line 92: As the CSVs are only short might it make sense to just set them in a HEREDOC, something like:

```suggestion
        let(:csv_data) do
          <<~CSV
          trn,dob,end_date,number_of_terms,objective,error
          7654321,30-06-1981,2025-01-30,0.5,,
          1234567,30/06/1981,2025-01-30,7.2,,
          CSV
        end
        let(:csv_file) { Rack::Test::UploadedFile.new(csv_data, 'text/csv') }
```

peterdavidhamilton: I'm already using HEREDOCs in other specs. I retained a mix of testing methods because this way we had the files to hand for manual tests and product review, something other colleagues can open and edit. I think it is definitely something to tidy up after. I have factories and shared specs so we can see what fits best.

peterdavidhamilton: @peteryates we will eventually get rid of most fixtures once the validation revisions are completed in a follow PR. The parsed CSV data is immediately saved to the DB meaning most specs can just create an array of hashes now.

app/models/pending_induction_submission_batch.rb
```diff
@@ -0,0 +1,198 @@
...
+
+  # @return [PendingInductionSubmissionBatch] type "action"
+  def self.new_action_for(appropriate_body:, **)
+    new(appropriate_body:, batch_type: 'action', **)
+  end
+
+  # @return [Class] A class that represents a row in the CSV file
+  # @param columns [Array<Symbol>] Array of CSV column headings
+  def self.build_row_class(columns)
+    Data.define(*columns) do
```
avinhurry, line 16: This feels very magical 🪄

Maybe it would be worth moving to a concern or helper. Something like`RowBuilder` could help make the intent clearer and keep the model focussed.

app/presenters/pending_induction_submission_batch_presenter.rb
```diff
@@ -0,0 +1,79 @@
...
+        )
+      elsif claim?
+        common_rows.push(
+          ::INDUCTION_PROGRAMMES[sub.induction_programme&.to_sym] || EMPTY_CELL,
+          sub.started_on&.to_fs(:govuk) || EMPTY_CELL,
+          sub.error_message
+        )
+      end
+    end
+  end
```
avinhurry, line 65: This method is quite big and does a lot. Could it be extracted into smaller private methods. To make it easier to understand at a glance?

app/services/appropriate_bodies/process_batch/action.rb
```diff
@@ -0,0 +1,161 @@
...
+            next
+          end
+
+          if ongoing_induction_period.blank?
+            pending_induction_submission.update(error_message: "Teacher #{name} does not have an ongoing induction")
+            next
+          end
+
+          if claimed_by_another_ab?
+            pending_induction_submission.update(error_message: "Teacher #{name} was claimed by another appropriate body")
```
avinhurry, line 37: This line is repeated 4 times here, how do you feel about extracting it out into a private method?

Something like: 

```ruby
def reject_with_error(message)
  pending_induction_submission.update(error_message: message)
end
```

app/services/appropriate_bodies/process_batch/action.rb
```diff
@@ -0,0 +1,161 @@
+module AppropriateBodies
+  module ProcessBatch
+    # Management of induction periods in bulk via CSV upload
+    # Handles Pass / Fail / Release in two stages
+    class Action < Base
```
avinhurry, line 5: This class is doing a lot, for example: 

- `fetch_trs_details`
- `validate_submission!`
- `do_action`

Might be worth exploring whether any of these could be pulled out into seperate service objects or classes to try and abide by the single responsibility principle a bit more.

What do you think?

app/services/appropriate_bodies/process_batch/claim.rb
```diff
@@ -0,0 +1,76 @@
...
+          pending_induction_submission.update(error_message: "Not found in TRS")
+          next
+        rescue ::TRS::Errors::ProhibitedFromTeaching
+          pending_induction_submission.update(error_message: "Prohibited from teaching")
+          next
+        rescue ::TRS::Errors::QTSNotAwarded
+          pending_induction_submission.update(error_message: "QTS not awarded")
+          next
+        rescue StandardError => e
+          pending_induction_submission.update(error_message: e.message)
```
avinhurry, line 28: Similar to my other comment, could these 4 could be extracted into a `reject_with_error(message)` to keep things DRY-er?

```ruby
def reject_with_error(message)
  pending_induction_submission.update(error_message: message)
end
```

app/services/appropriate_bodies/process_batch/claim.rb
```diff
@@ -0,0 +1,76 @@
...
+          next
+        end
+      rescue StandardError => e
+        pending_induction_submission_batch.update(error_message: e.message)
+      end
+
+    private
+
+      # @return [?]
```
avinhurry, line 37: Should this be filled in, instead of `[?]`?

spec/models/pending_induction_submission_batch_spec.rb
```diff
@@ -0,0 +1,156 @@
...
+
+  describe "associations" do
+    it { is_expected.to belong_to(:appropriate_body) }
+    it { is_expected.to have_many(:pending_induction_submissions) }
+  end
+
+  describe "scopes" do
+    describe ".for_appropriate_body" do
+      it "returns batched submissions for the specified appropriate body" do
+        expect(described_class.for_appropriate_body(456).to_sql).to end_with(%( WHERE "pending_induction_submission_batches"."appropriate_body_id" = 456))
```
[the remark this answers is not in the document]
peterdavidhamilton: Agreed, this was copied from another model spec 🫨, can replace.

app/presenters/pending_induction_submission_batch_presenter.rb
peterdavidhamilton: This is one of those "just there to surface the data in review methods". Anything below the line is out of scope

app/services/appropriate_bodies/process_batch/action.rb
peterdavidhamilton: That's certainly the direction I would expect. `Claim` reuses existing services for example, and here I originally tried to, but it was awkward and I really wanted to avoid making changes that might alter existing services. It's giving `monad`. Some of the CSV-level validations are moving to row-level, so the logic of the validation section is about to grow. Also I'd agree that we should avoid multiple places fetching from TRS. I am working on refactoring the validations in a separate branch and will make splitting the service part of its focus.

app/services/appropriate_bodies/process_batch/claim.rb
peterdavidhamilton: It will be once I've decided what if anything it should be returning.

spec/models/pending_induction_submission_batch_spec.rb
```diff
@@ -0,0 +1,156 @@
...
+      end
+    end
+  end
+
+  describe "callbacks" do
+    describe '#data_from_csv' do
+      it 'persists data from disk to database and deletes attachment' do
```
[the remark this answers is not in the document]
peterdavidhamilton: I'll workshop it. For context the temporary file in `/storage` is only available to the webapp and data needs to be stored in the database so the worker has the information. My goal was to make clear disk data storage is transitory and we aren't accumulating files.  There is an element of housekeeping. More suggestions welcome.

db/migrate/20250324093617_create_active_storage_tables.active_storage.rb
peteryates, line 1: I'd like to avoid active storage if possible. It adds complexity to the deployment and we'll need to set up blob storage in Azure etc. I don't think we need to store the originally uploaded CSV for any reason - and if we do could it just be a `varchar[]` on the batch table?

peterdavidhamilton: We don't store it anymore. I'll look into simplifying it.

app/jobs/process_batch_action_job.rb
```diff
@@ -0,0 +1,45 @@
+class ProcessBatchActionJob < ApplicationJob
+  retry_on StandardError, attempts: 3
+
+  # @param pending_induction_submission_batch [PendingInductionSubmissionBatch]
+  # @param author_email [String]
+  # @param author_name [String]
+  def perform(pending_induction_submission_batch, author_email, author_name)
+    if pending_induction_submission_batch.processing?
```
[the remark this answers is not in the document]
peterdavidhamilton: It would, I was trying to identify when that might happen; put the guard in anyway. Better to defend.

config/routes.rb
```diff
@@ -80,6 +80,13 @@
         get 'prohibited-from-teaching/:id', to: '/appropriate_bodies/claim_an_ect/errors#prohibited_from_teaching', as: 'prohibited'
       end
     end
+
+    constraints -> { Rails.application.config.enable_bulk_upload } do
+      namespace :process_batch, path: 'bulk', as: 'batch' do
+        resources :claims, format: %i[html csv]
+        resources :actions, format: %i[html csv]
+      end
+    end
```
peteryates, line 89: :+1:

lib/trs/teacher.rb
```diff
@@ -66,10 +66,6 @@ def prohibited_from_teaching?
 
   private
 
-    def api_client
```
peterdavidhamilton, line 69: ???

spec/presenters/pending_induction_submission_batch_presenter_spec.rb
```diff
@@ -0,0 +1,88 @@
+RSpec.describe PendingInductionSubmissionBatchPresenter do
+  subject(:presenter) { described_class.new(batch) }
+
+  let(:batch) { FactoryBot.build(:pending_induction_submission_batch, :action) }
+
+  describe 'decorated model methods' do
```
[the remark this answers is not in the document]
peterdavidhamilton: You are correct, semantically this class is a decorator which is why I wrote "decorated", but not wanting to compete with the existing folder structure I have put it in with the others. I would support a refactor renaming them all to decorators. We have a mix of view helpers, services, components and decorators and should have a think about where we can improve on this.

app/controllers/appropriate_bodies/process_batch/actions_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+            .order(id: :desc)
+            .select(:id, :batch_type, :batch_status, :error_message)
+            .map { |b| b.attributes.values.map(&:to_s) }
+      end
+
+      def create
+        @pending_induction_submission_batch = new_batch_action
+
+        if @pending_induction_submission_batch.save! &&
+            @pending_induction_submission_batch.save(context: :uploaded)
```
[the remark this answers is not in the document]
peterdavidhamilton: At the time, I had used `ActiveStorage` and yes saving twice was intentional because the first save triggered `ActiveStorage` to do its thing, making the following context validation possible. Now, the file data is parsed in memory and logic has moved into different objects.

app/controllers/appropriate_bodies/process_batch/actions_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+
+        process_batch_action
+
+        redirect_to ab_batch_action_path(@pending_induction_submission_batch), alert: 'Induction changes actioned'
+      end
+
+    private
+
+      def new_batch_action
+        PendingInductionSubmissionBatch.new_action_for(appropriate_body: @appropriate_body, **import_params)
```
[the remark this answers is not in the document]
peterdavidhamilton: A parent controller

app/controllers/appropriate_bodies/process_batch/actions_controller.rb
```diff
@@ -0,0 +1,59 @@
+module AppropriateBodies
+  module ProcessBatch
+    class ActionsController < PendingInductionSubmissionBatchController
+      def index
+        @pending_induction_submission_batches = PendingInductionSubmissionBatch
+            .for_appropriate_body(@appropriate_body)
+            .action
+            .order(id: :desc)
+            .select(:id, :batch_type, :batch_status, :error_message)
+            .map { |b| b.attributes.values.map(&:to_s) }
```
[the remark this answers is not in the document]
peterdavidhamilton: @edujackedu @avinhurry at the moment this query is for a scaffolded view to help product reviewers. We've identified that such an "Upload history" for the AB is a useful part of the UX, so they can return to an upload file in case they log out or forget to hit confirm. There needs to be a link or button to navigate them back to the processed batch because there is no need for them to reupload. This is being designed so a future ticket should do what you've both identified here.

app/models/pending_induction_submission_batch.rb
peterdavidhamilton: Anything I can still do to demystify it in my documentation?

app/jobs/process_batch_action_job.rb
```diff
@@ -0,0 +1,43 @@
...
+  # @param pending_induction_submission_batch [PendingInductionSubmissionBatch]
+  # @param author_email [String]
+  # @param author_name [String]
+  def perform(pending_induction_submission_batch, author_email, author_name)
+    return if pending_induction_submission_batch.processing?
+
+    if pending_induction_submission_batch.processed?
+      # change status first for quick response
+      pending_induction_submission_batch.completed!
+      batch_action(pending_induction_submission_batch, author_email, author_name).do!
```
avinhurry, line 11: Just flagging that `pass!` calls `record_outcome!` which includes `send_pass_induction_notification_to_trs`.

If this method raises, the internal transaction that `record_outcome!` uses will roll back DB changes inside `record_outcome!`, but **not** things like posting to the TRS API. That means we lose atomicity and if the batch was marked as `completed!` beforehand, we may not catch the inconsistency.

  Would it be safer to: 

  - call `do!` first and only mark the batch as `completed!` if it succeeds?
  - consider compensating or tracking partially applied batches incase of failure?

peterdavidhamilton: You're right to flag it. The reason is the lag in actioning the changes once the user confirms. This was done so we could render success. There is potentially a need for another status, `actioning` or `doing` 😉 , which comes before `completed`. However, the things it's attempting should all succeed without exception because they have been pre-validated and filtered.

I'll take your comments away and address them in my next CSV related branch. Thanks @avinhurry 🙇🏼‍♂️

PR #431, opened 2025-04-07
426: rename delete induction period events

ghost:
We've decided to remove actor type from event names so we need to rename them:

- rename `admin_reverts_teacher_claim` to `teacher_induction_status_reset `
- rename `admin_deletes_induction_period` to `induction_period_deleted`


peterdavidhamilton (approved):
LGTM

PR #440, opened 2025-04-07
[1416] Only show content for `provider-led` on "Programme choices used by your school previously" page

avinhurry:
### Context

This change ensures a small piece of content is only rendered if the previously chosen programme type is **provider-led**.
If any other programme type was selected, the content will not be shown.

### Before/After

| Before | After |
|--------|-------|
|image|image|

### Tech notes

The second commit is only here to get the review app deploying and will be dropped before merging. I've opened a separate PR with that fix [here](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/441).


robthompson16 (approved):
Looks good thanks

PR #446, opened 2025-04-08
Remove reference to `AppropriateBody` from old migrations

peteryates:
This was included originally because there was a stipulation that `dfe_sign_in_organisation_id` was present and we had deployed non-production environments.

It turns out that the loading of AppropriateBody at this stage is now causing issues when migrating and breaking the deployment of review apps.


ltello (approved):

db/migrate/20241025151817_add_dfe_sign_in_organisation_id_to_abs.rb
```diff
@@ -1,11 +1,8 @@
 class AddDfESignInOrganisationIdToAbs < ActiveRecord::Migration[7.2]
   def change
-    add_column :appropriate_bodies, :dfe_sign_in_organisation_id, :uuid
-
-    AppropriateBody.all.find_each { |ab| ab.update(dfe_sign_in_organisation_id: SecureRandom.uuid) }
-
-    change_column :appropriate_bodies, :dfe_sign_in_organisation_id, :uuid, null: false
-
+    # rubocop:disable Rails/NotNullColumn
```
avinhurry, line 3: How do you feel about disabling this in `rubocop.yml`, like [this one](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/.rubocop.yml#L89)?

I like seeing a list of the files where we've made exceptions - it feels neater than doing it inline to me. Maybe we could also add a short comment explaining why we've disabled it, just in case it’s useful down the line?

peteryates: In actual code I'd agree, but I think it's fine in an old migration. The commit message should give enough context as to why it was changed, not sure a comment is necessary too.

peteryates: This cop might actually be useful elsewhere too!

avinhurry: Makes sense, fair point 👌🏽

PR #447, opened 2025-04-08
Populate appropriate bodies via SQL in a migration

ltello:
Do not call `AppropriateBody` model inside migrations.
Replace it with pure SQL to prevent errors recreating the database (CI for instance) if the model code or its columns change in future migrations.

### DIFF
--- a/db/migrate/20250319234928_add_type_to_appropriate_bodies.rb
+++ b/db/migrate/20250319234928_add_type_to_appropriate_bodies.rb
@@ -5,9 +5,14 @@ class AddTypeToAppropriateBodies < ActiveRecord::Migration[8.0]
 
     add_column :appropriate_bodies, :body_type, :appropriate_body_type, default: 'teaching_school_hub'
 
-    AppropriateBody.find_each do |appropriate_body|
-      appropriate_body.update_column(:body_type, type_from_name(appropriate_body.name))
-    end
+    execute <<-SQL.squish
+      UPDATE appropriate_bodies
+      SET body_type =
+        CASE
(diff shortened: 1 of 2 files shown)


db/migrate/20250319234928_add_type_to_appropriate_bodies.rb
```diff
@@ -5,9 +5,14 @@ def up
...
     add_column :appropriate_bodies, :body_type, :appropriate_body_type, default: 'teaching_school_hub'
 
-    AppropriateBody.find_each do |appropriate_body|
-      appropriate_body.update_column(:body_type, type_from_name(appropriate_body.name))
-    end
+    execute <<-SQL.squish
+      UPDATE appropriate_bodies
+      SET body_type =
+        CASE
+          WHEN name = '#{AppropriateBodies::Search::ISTIP}' THEN 'national'::appropriate_body_type
```
tonyheadford, line 12: I would hard code the right values in the SQL and not reference the code personally

ltello: Well, it is just a constant and the class doesn't require `AppropriateBody` to be loaded but ok, I will hardcode it.

tonyheadford (approved):

db/migrate/20250319234928_add_type_to_appropriate_bodies.rb
tonyheadford: It is referenced in the methods and not sure whether it would be loaded and compiled when that file was loaded gut yes, to be on the safe-side this feels better, thanks.

PR #449, opened 2025-04-08
Bugfix: No need to specify author in event heading

ghost:
https://github.com/DFE-Digital/register-ects-project-board/issues/1409

Author types have different ways of sourcing names name/fullname and with the current implementation is erroring for adming users. The user name in the heading in redundant anyway, therefore I am removing it and adding the missing spec for the event


peterdavidhamilton (approved):
👍🏼

PR #452, opened 2025-04-08
Correct logic in inserting_induction_period?

ghost:
We should first check whether we're appending the IP before checking if inserting between siblings

This will fix several support issues such as:

https://trello.com/c/1WpDRyY9/1279-change-of-start-date-request


app/models/induction_period.rb
```diff
@@ -51,8 +51,15 @@ def end_date_admin_only
...
+    # If there are no siblings with a later start date, and the current period starts after the last finished sibling,
+    # then we're appending to the end, not inserting
+    return false if siblings.started_on_or_after(started_on).empty? &&
+      (last_finished_sibling.nil? || started_on >= last_finished_sibling.finished_on)
+
+    # Otherwise, check if we're inserting between periods
     siblings.any? do |sibling|
-      started_on.before?(sibling.started_on) || sibling.finished_on && (started_on.after?(sibling.finished_on) && !sibling.eql?(last_finished_sibling))
+      started_on.before?(sibling.started_on) ||
+        (sibling.finished_on && started_on.after?(sibling.finished_on) && !sibling.eql?(last_finished_sibling))
```
peterdavidhamilton, line 62: The logic here was originally a little more complex. With your fix I think the last bit in this block is now redundant (because it was kaput 🙍🏼). So it could be as follows and still pass specs.

```ruby
    siblings.any? do |sibling|
      started_on.before?(sibling.started_on) || (sibling.finished_on && started_on.after?(sibling.finished_on)   
    end
```

peterdavidhamilton (approved):

PR #453, opened 2025-04-09
Add separate Rails environments for each deployed env

peteryates:
We currently have the following environments in our pipeline:

* review
* staging
* sandbox
* migration
* production

They all use the same `RAILS_ENV=production`. The idea behind this was to ensure we used environment variables to apply config instead of doing `if Rails.env.sandbox?` etc.

However, this has some disadvantages, namely:

1. the rails console shows `production` when we're not in production. The prompt will look like this when we're in a review app or on staging, sandbox, migration or production. It's confusing.
  ```
  register-early-career-teachers(production)>
  ```
2. we need to override production checks when doing database tasks in review

  ```
  ActiveRecord::ProtectedEnvironmentError: You are attempting to run a destructive action against your 'production' database. (ActiveRecord::ProtectedEnvironmentError)
If you are sure you want to continue, run the same command with the environment variable:
  DISABLE_DATABASE_ENVIRONMENT_CHECK=1
  ```
  Doing this normalises the process of overriding production checks so there's a greater chance of doing it by habit and actually reseeding production. This would be bad.

## Changes

- **Sort the env variable files alphabetically**
- **Add environments configs each deployed environment**
- **Add database config for the deployed environments**
- **Add rails_env variable to terraform config**


config/terraform/application/config/migration.tfvars.json
```diff
@@ -5,5 +5,6 @@
     "enable_logit": true,
     "worker_replicas": 6,
     "worker_memory_max": "2Gi",
-    "postgres_flexible_server_sku": "GP_Standard_D4ds_v5"
+    "postgres_flexible_server_sku": "GP_Standard_D4ds_v5",
+    "rails_env": "migration"
```
RMcVelia, line 9: It might be better to add RAILS_ENV to the migration.yml in the same directory. That would then get loaded via https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/config/terraform/application/application.tf#L19 and you wouldn't need any tf variable.

peterdavidhamilton (approved):

PR #457, opened 2025-04-11
adapt interval scopes to future dates

ltello:
PR to experiment how life would be if we add pure periods with future dates to the service:

- we should try to pass dates to every service/scope to minimize the issues that time travel creates on specs.
- we should try to put special attention to add specs for all the different dates a service/scope/method might make to behave differently.
- we should time travel to different dates and check the different behaviours of end-to-end scenarios.
- queries get more complicated because we need to add conditions about date ranges most of the time. Is it worth to pay this price?
- by allowing any past, present, future periods to be persisted we can be certain we are reflecting faithfully the reality and not contraining the service at all. However, is it worth the complexity added?
- If we assumed (under investigation by Tony) that every not-pending period in the db started in the past, things would get less complicated when it comes to searching periods: all ongoing periods would have null `finished_on` date which makes our queries straight forward and the service less error-prone, for instance.
- However, is that approach too restrictive so that the data at some point doesn't fully matches the reality? I can't think of any case now but would it make the service buggy because of that for certain features?

### DIFF
--- a/app/models/concerns/interval.rb
+++ b/app/models/concerns/interval.rb
@@ -6,16 +6,18 @@ module Interval
     validate :period_dates_validation
 
     # Scopes
-    scope :overlapping_with, ->(period) { where("range && daterange(?, ?)", period.started_on, period.finished_on) }
-    scope :ongoing, -> { where(finished_on: nil) }
-    scope :finished, -> { where.not(finished_on: nil) }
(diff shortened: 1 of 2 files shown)


peteryates:
I'm going to close this now as it's old but we'll refer back to it.

PR #459, opened 2025-04-11
Allow 18 character TRS induction statuses

peteryates:
RequiredToComplete is 18 characters long.

This will eventually be a database type, so just nudging it up for now.


peterdavidhamilton (approved):
This is giving me dejavu?

PR #460, opened 2025-04-11
Spike for marking future date periods with pending flag

tonyheadford:
Spike future school joiners by using a pending flag or status attribute on the `ECTAtSchoolPeriod` and adjusting the scopes accordingly
While attempting this is felt difficult to achieve without making the other related periods `pending` because queries would always have to relate back to `ECTAtSchoolPeriod` to check whether the related models were also pending and this felt like it could be done using the `started_on` making the `pending` flag redundant.
Therefore have expanded migration to add `pending` flag to the `MentorAtSchoolPeriod`, `MentorshipPeriod` and `TrainingPeriod` also. Should've also added `InductionPeriod` too because that's probably now broken.

I don't think this is a great solution, but also I'm not sure we've understood the requirements we have around supporting these future events and how we might need to:

* handle periods added before these are made "live"
* changes to these before they are "live"
* how we should handle validations in a mix of pending and non-pending records and at what point in the lifecycle
* which scopes should filter pending records by default
* which service users care about pending records and should have visibility of them

### DIFF
--- a/app/models/concerns/interval.rb
+++ b/app/models/concerns/interval.rb
@@ -4,18 +4,20 @@ module Interval
   included do
     # Validations
     validate :period_dates_validation
-
+    #
     # Scopes
     scope :overlapping_with, ->(period) { where("range && daterange(?, ?)", period.started_on, period.finished_on) }
-    scope :ongoing, -> { where(finished_on: nil) }
+    scope :ongoing, -> { not_pending.where(finished_on: nil) }
     scope :finished, -> { where.not(finished_on: nil) }
     scope :earliest_first, -> { order(started_on: 'asc') }
     scope :latest_first, -> { order(started_on: 'desc') }
-    scope :started_before, ->(date) { where(started_on: ...date) }
-    scope :started_on_or_after, ->(date) { where(started_on: date..) }
+    scope :started_before, ->(date) { not_pending.where(started_on: ...date) }
+    scope :started_on_or_after, ->(date) { not_pending.where(started_on: date..) }
     scope :finished_before, ->(date) { where(finished_on: ...date) }
     scope :finished_on_or_after, ->(date) { where(finished_on: date..) }
     scope :containing_period, ->(period) { where("range @> daterange(?, ?)", period.started_on, period.finished_on) }
+    scope :pending, -> { where(pending: true) }
+    scope :not_pending, -> { where(pending: false) }
   end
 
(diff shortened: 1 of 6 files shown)


claire-hughez:
Some vague thoughts:

_handle periods added before these are made "live"_
Not sure I understand this one or the product requirement here - I had assumed periods would be created with any dates we had, but they would also be pending until the start_date had passed. I'm not sure how we'd save the data to be used for these periods otherwise, e.g. start_date for ect_at_school_periods or the potential LP for the training_periods.

_changes to these before they are "live"_
I think we should allow users to change these records the same we would if they were 'live' (e.g. if they had started as an ECT). For example, their email, name or other details should be allowed to be edited.

The exception to this would be anything we do with withdrawals. I don't think we should allow any records to be withdrawn by a school until their start_date has been reached.

I think for MVP, we should require any deletion/withdrawals of un-live records to be withdrawn to be done via support. Potentially a further on piece of work would be schools telling us when an ECT or mentor didn't start.

_how we should handle validations in a mix of pending and non-pending records and at what point in the lifecycle_

This would be a good one to talk about. There should be validation to prevent adding anything too far in the future (although a lot of this will be linked to if the registration window has opened). 

_which scopes should filter pending records by default_
Not sure on this one.

_which service users care about pending records and should have visibility of them_

I don't think it's too important for schools to view that a record IS pending (not live yet), they should know. But they'd expect to see these records.

I expect it's helpful for lead providers to see potential ECTs who might be starting ASAP, but with the knowledge of when they're starting - but I'd hope to expose start_dates over the API to help with this, rather than doing any extra work.

tonyheadford:
> which scopes should filter pending records by default
>> Not sure on this one.

Sorry a bit techy that one, we have quite a few scopes (sql query helpers) on our periods to help with selecting and checking for validity, what I mean is in what scenarios would we want to include or exclude pending records when querying for participant info and I'm not sure we fully know as everything service wide may need to query via these periods I think

> which service users care about pending records and should have visibility of them
>>I don't think it's too important for schools to view that a record IS pending (not live yet), they should know. But they'd expect to see these records.

Schools will need to see them if we want to allow them to make changes to them and also to see their forthcoming arrivals etc. and yes it would be odd for a school to add an ECT and then not be able to see them _somewhere_ on their dashboard

peteryates:
Closing for now.

PR #462, opened 2025-04-14
Fix status change events

peteryates:
There's a problem in production where events are always being written regardless of whether a name has changed or not.

It was caused by a mismatch between the variable names `induction_status_before` and `new_induction_status`. Because we were using instance variables they were interpolated as empty strings which were written to the events log.


peterdavidhamilton (approved):
🤞🏼

PR #463, opened 2025-04-14
Spike for adding a separate model for future joiners

tonyheadford:
This approach adds a separate model to hold the details of a future school starter.  It assumes that we will in all circumstances (new registration and existing participant) have a `Teacher` model to keep things simpler.

#### Mandatory attributes:

Teacher, School, "role_type" - ECT or mentor, programme_type - school-led or provider-led, starting_on

#### Optional attributes:

Lead Provider, Appropriate Body, MentorAtSchoolPeriod

I've added associations that permit:

- A school to see who is a future joiner
- A teacher to see their future schools and provider
- A provider to see their future joiners and schools

It is trivial to get a list of future starts by provider, school or AB:

`PendingSchoolJoiner.where(lead_provider: x)`

`PendingSchoolJoiner.where(school: x)` or `school.pending_school_joiners` or `school.pending_teachers`

`PendingSchoolJoiner.where(appropriate_body: x)`

It seems like it would be easy to make changes to it if we wanted to allow the school (or admin) that functionality

It doesn't change the existing period models so no need to calculate dates or workout what is active - if it is live it is in a period if it is pending it is in the pending model.

We'd need a job to run daily that actions the pending joiners on the `starting_on` date which would:
- add the `ECTAtSchoolPeriod` (or MentorAtSchoolPeriod` if we wanted to allow that at some point)
- add a `MentorshipPeriod` as appropriate
- add a `TrainingPeriod` as appropriate as either an EOI or linked to a confirmed partnership
- close any existing `ECTAtSchool/Mentorship/TrainingPeriod`s if present (i.e. from a previous school)
- removes the `PendingSchoolJoiner` record or marks it as done in someway

I've not added specs but have played with this on the console to get a feel for it and it seems like a nice approach

### DIFF
--- a/db/migrate/20250414131145_create_pending_school_joiners.rb
+++ b/db/migrate/20250414131145_create_pending_school_joiners.rb
@@ -1,0 +1,15 @@
+class CreatePendingSchoolJoiners < ActiveRecord::Migration[8.0]
+  def change
+    create_table :pending_school_joiners do |t|
+      t.references :teacher, foreign_key: true
+      t.references :school, foreign_key: true
+      t.string :role_type, null: false, default: "ect"
+      t.date :starting_on, null: false
+      t.enum :programme_type, enum_type: :programme_type
+      t.references :mentor_at_school_period, foreign_key: true, null: true
+      t.references :lead_provider, foreign_key: true, null: true
+      t.references :appropriate_body, foreign_key: true, null: true
+      t.timestamps
+    end
+  end
+end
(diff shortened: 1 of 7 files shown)


db/migrate/20250414131145_create_pending_school_joiners.rb
```diff
@@ -0,0 +1,15 @@
...
+  def change
+    create_table :pending_school_joiners do |t|
+      t.references :teacher, foreign_key: true
+      t.references :school, foreign_key: true
+      t.string :role_type, null: false, default: "ect"
+      t.date :starting_on, null: false
+      t.enum :programme_type, enum_type: :programme_type
+      t.references :mentor_at_school_period, foreign_key: true, null: true
+      t.references :lead_provider, foreign_key: true, null: true
+      t.references :appropriate_body, foreign_key: true, null: true
```
peteryates, line 11: Might need `:email` too.

tonyheadford: Good spot

peteryates:
Closing this as we came to a decision - see #518

PR #465, opened 2025-04-14
bugfix: admit can't add induction period

ghost:
# Fix Induction Period Creation and Error Handling

## Problem
The induction period creation process had several issues:
1. The `CreateInductionPeriod` service was inconsistently handling initialization parameters across different services
2. Error handling in the controller was not properly propagating validation errors

## Changes
1. Standardized the `CreateInductionPeriod` service to consistently use `author:`, `teacher:`, and `params:` as initialization parameters
2. Updated the `Admin::CreateInductionPeriod` and `RegisterECT` services to match this pattern
3. Fixed error handling in the controller to properly set `@induction_period` in error cases


peterdavidhamilton (approved):
👍🏼 👍🏼 👍🏼

emily-prudence-dfe:
@edujackedu tested this and the only thing I have to note is that the behaviour of "Add an Induction Period" has changed - I can now add an induction period without an end date and number of terms.

I don't think this is a blocker to merging (in fact, it's actually quite helpful in a way) so please go ahead, but wanted to make sure there was a note of this behaviour change.

PR #476, opened 2025-04-15
[1344] Update mentor list and mentor details page to show assigned ECTs

avinhurry:
### Context

This PR updates the mentors index page to display the mentor's TRN and their currently assigned ECTs. 

To keep the view clean and maintainable, the summary card has been extracted into a dedicated SummaryComponent.

**Note:** The last commit with the route, is only here to make the view work. It's part of #464, so there's no need to review that commit.

### Before/After

| Before | After |
|--------|-------|
|image|image|

### DIFF
--- a/app/components/schools/mentors/summary_component.rb
+++ b/app/components/schools/mentors/summary_component.rb
@@ -1,0 +1,50 @@
+module Schools
+  module Mentors
+    class SummaryComponent < ViewComponent::Base
+      include TeacherHelper
+
+      def initialize(mentor:, school:)
+        @mentor = mentor
+        @school = school
+      end
+
+      def call
+        govuk_summary_card(title: link_to_mentor(@mentor)) do |card|
+          card.with_summary_list(
+            classes: %w[govuk-summary-list--no-border],
+            rows: [trn_row, assigned_ects_row]
+          )
+        end
+      end
+
+    private
+
+      def link_to_mentor(mentor)
+        govuk_link_to(teacher_full_name(mentor.teacher), schools_mentor_path(mentor))
+      end
+
+      def trn_row
+        { key: { text: 'TRN' }, value: { text: trn } }
+      end
+
+      def assigned_ects_row
+        { key: { text: 'Assigned ECTs' }, value: { text: assigned_ects_summary } }
+      end
+
+      def trn
+        @mentor.teacher.trn
+      end
+
+      def assigned_ects
+        @assigned_ects ||= @mentor.currently_assigned_ects
+      end
+
+      def assigned_ects_summary
+        return "No ECTs assigned" if assigned_ects.empty?
+        return "#{assigned_ects.count} assigned ECTs" if assigned_ects.count > 5
+
+        safe_join(assigned_ects.map { |ect| teacher_full_name(ect.teacher) }, tag.br)
+      end
+    end
+  end
+end
--- a/app/components/schools/mentors/summary_component.rb
+++ b/app/components/schools/mentors/summary_component.rb
@@ -9,7 +9,7 @@ module Schools
       end
 
       def call
-        govuk_summary_card(title: link_to_mentor(@mentor)) do |card|
+        govuk_summary_card(title: link_to_mentor) do |card|
           card.with_summary_list(
             classes: %w[govuk-summary-list--no-border],
             rows: [trn_row, assigned_ects_row]
@@ -19,8 +19,8 @@ module Schools
 
     private
 
-      def link_to_mentor(mentor)
-        govuk_link_to(teacher_full_name(mentor.teacher), schools_mentor_path(mentor))
+      def link_to_mentor
+        govuk_link_to(teacher_full_name(@mentor.teacher), schools_mentor_path(@mentor))
       end
 
       def trn_row
--- a/app/components/schools/mentors/summary_component.rb
+++ b/app/components/schools/mentors/summary_component.rb
@@ -20,7 +20,7 @@ module Schools
     private
 
       def link_to_mentor
-        govuk_link_to(teacher_full_name(@mentor.teacher), schools_mentor_path(@mentor))
+        govuk_link_to(teacher_full_name(@mentor.teacher), '#')
       end
 
       def trn_row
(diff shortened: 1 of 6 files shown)


spec/components/schools/mentors/summary_component_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+
+      ects.each do |teacher|
+        expect(rendered_content).to include("#{teacher.trs_first_name} #{teacher.trs_last_name}")
+      end
+    end
+  end
+
+  context 'with more than 5 ECTs' do
+    before do
+      (1..6).each do |i|
```
peterdavidhamilton, line 36: If the index being zero based isn't an issue, could use `6.times`

spec/components/schools/mentors/summary_component_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+        create(:mentorship_period, mentor:, mentee: ect, started_on:, finished_on: nil)
+        teacher
+      end
+    end
+
+    it 'lists ECT names' do
+      render_inline(described_class.new(mentor:, school:))
+
+      ects.each do |teacher|
+        expect(rendered_content).to include("#{teacher.trs_first_name} #{teacher.trs_last_name}")
```
peterdavidhamilton, line 29: I've always preferred the string without interpolation. I think I'm in the minority

app/views/schools/mentors/show.html.erb
```diff
@@ -0,0 +1,24 @@
...
+  title: Teachers::Name.new(@teacher).full_name,
+  caption: "TRN: #{@teacher.trn}",
+  header: Teachers::Name.new(@teacher).full_name,
+  backlink_href: schools_mentors_home_path)
+%>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+      <h2 class="govuk-heading-m">Mentor details</h2>
+      <%= govuk_summary_list do |summary_list|
```
peterdavidhamilton, line 11: I was wondering if throwing this in the `mentor_helper` and reusing the other helpers for name would be better? I'd definitely leverage `teacher_full_name` now though like you've done in the component.

app/components/schools/mentors/summary_component.rb
```diff
@@ -0,0 +1,50 @@
...
+          card.with_summary_list(
+            classes: %w[govuk-summary-list--no-border],
+            rows: [trn_row, assigned_ects_row]
+          )
+        end
+      end
+
+    private
+
+      def link_to_mentor(mentor)
```
peterdavidhamilton, line 22: It is possible to remove the param and access it from the instance variable

peterdavidhamilton (changes requested):

spec/components/schools/mentors/summary_component_spec.rb
avinhurry: I think you might be but that's ok 😄  

I used interpolation here because: 

  - We are defining the factories above
  - We keep things DRY
  - It prevents unecessary coupling and allows the test to be slightly more flexible

spec/components/schools/mentors/summary_component_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+
+    it 'shows ECT count instead of listing names' do
+      render_inline(described_class.new(mentor:, school:))
+      expect(rendered_content).to have_css('.govuk-summary-list__value', text: '6 assigned ECTs')
+    end
+  end
+
+  context 'when there are multiple mentors' do
+    let(:mentor2_teacher) { create(:teacher, trs_first_name: 'Sasuke', trs_last_name: 'Uchiha') }
+    let(:mentor2) { create(:mentor_at_school_period, teacher: mentor2_teacher, school:, started_on:, finished_on: nil) }
```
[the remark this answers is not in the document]
avinhurry: Good shout - thanks. 

Updated here: https://github.com/DFE-Digital/register-early-career-teachers-public/pull/476/commits/677e0a790f8cf4890233e55120a4d63c7cbeec98

spec/components/schools/mentors/summary_component_spec.rb
avinhurry: Also a nice improvement - thank you. 

Updated here: https://github.com/DFE-Digital/register-early-career-teachers-public/pull/476/commits/677e0a790f8cf4890233e55120a4d63c7cbeec98

app/views/schools/mentors/show.html.erb
avinhurry: Not sure I follow 🤔

But just to clarify - this won't be merged in this PR. I only cherry-picked it from #464, and it’s already been refactored into a component further down the line.

I mainly needed the route here so the view wouldn't break while this PR was being reviewed - I admit I was a bit lazy 😅 I did leave a note about it in the PR description, not sure if you saw it.

app/components/schools/mentors/summary_component.rb
avinhurry: It is indeed and good shout again! I've updated here: https://github.com/DFE-Digital/register-early-career-teachers-public/pull/476/commits/16c9664e18066411c5f0591d2078913a84813033

app/views/schools/mentors/show.html.erb
peterdavidhamilton: Gotcha, I didn't realise I was ignoring this too.

spec/components/schools/mentors/summary_component_spec.rb
peterdavidhamilton: Let me try and win you over.

- factories are still needed, no change there
- it's repeating code from a service
- it's no less coupled but is less readable

avinhurry: I think testing exact strings is valuable too (and I've done that on lines 64/65) but I reckon there is a limit. 

Like, if I had to generate 500 different items, would it still make sense to hardcode 500 expectations?

At that point, interpolation feels like a better trade off, it keeps the test flexible and avoids too much repetition but still checks that the output matches the data.

peterdavidhamilton: I miss read, thought it was just one teacher you were asserting not all. I concede.

PR #480, opened 2025-04-16
Change sorting on schools' ECT index page

peteryates:
This change was more difficult that originally thought.

As outlined in [this comment](https://github.com/DFE-Digital/register-ects-project-board/issues/1266#issuecomment-2809341690), sorting ECTs by whether they have a mentor is complex.

| Before | After (2 items per page for demo purposes) |
| ----- | ------ |
| image |Screenshot From 2025-04-17 14-56-37 |
 
## Changes

1. Make the ECT index page use `Teachers::Search`
2. Allow `Teachers::Search` to find teachers who are currently ECTs at a school
3. Add a custom sort order so we can put mentorless ECTs first, and then sort by most-recently registered
4. change the global pagination 'per page' setting to 20 (ok'd with @emily-prudence-dfe)

## Notes

1. The view isn't idea. It currently is based around `@ects` rather than `@teachers` which doesn't really make sense as it's a list of teachers. I'll raise an issue to refactor that 
2. Due to the item above, it's a bit of a pain to test pagination in a view spec.
3. I added a FIXME to an unrelated pagination spec that will be done as a follow up
4. I suggest reviewing commit by commit, they're quite concise

Fixes DFE-Digital/register-ects-project-board#1266

### DIFF
--- a/app/services/teachers/search.rb
+++ b/app/services/teachers/search.rb
@@ -11,10 +11,14 @@ module Teachers
     end
 
     def search
-      scope.order(:trs_last_name, :trs_first_name, :id)
+      scope.order(order)
     end
 
   private
+
+    def order
+      %i[trs_last_name trs_first_name id]
+    end
 
     def matching(query_string)
       return if query_string == :ignore
--- a/app/services/teachers/search.rb
+++ b/app/services/teachers/search.rb
@@ -17,7 +17,20 @@ module Teachers
   private
 
     def order
-      %i[trs_last_name trs_first_name id]
+      case @sort_order
+      when :mentorless_first_then_by_registration_date
+        # mentorless teachers first, sorted by the registration date
+        # at the school, latest first
+        [
+          Arel::Nodes::Case.new
+                          .when(MentorshipPeriod.arel_table[:id].eq(nil)).then(0)
+                          .else(1)
+                          .asc,
+          { ect_at_school_periods: { created_at: 'desc' } }
+        ]
+      else
+        %i[trs_last_name trs_first_name id]
+      end
     end
 
     def matching(query_string)
@@ -53,6 +66,8 @@ module Teachers
               .where(ect_at_school_periods: { school: })
               .merge(ECTAtSchoolPeriod.ongoing)
       )
+
+      @sort_order = :mentorless_first_then_by_registration_date
     end
   end
 end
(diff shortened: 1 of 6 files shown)


spec/services/teachers/search_spec.rb
```diff
@@ -147,6 +114,64 @@
...
+
+          expect(query.to_sql).to include(%{"ect_at_school_periods"."school_id" IN (123, 456)})
+        end
+      end
+
+      context 'when absent' do
+        it 'does not join ect_at_school_periods' do
+          query = Teachers::Search.new(ect_at_school: 123).search
+
+          expect(query).not_to include('ect_at_school_periods')
```
avinhurry, line 144: I think you might be getting false positives here because your missing the `to_sql`, should it be:


```suggestion
        it 'does not join ect_at_school_periods' do
          query = Teachers::Search.new.search

          expect(query.to_sql).not_to include('ect_at_school_periods')
```

app/views/schools/ects/index.html.erb
```diff
@@ -18,18 +18,20 @@
...
 
   <div class='govuk-grid-column-two-thirds'>
-    <% @ects.map do |ect| %>
-      <%= 
+    <% @teachers.map { |t| t.ect_at_school_periods.first }.each do |ect| %>
+      <%=
         govuk_summary_card(title: link_to_ect(ect)) do |card|
           card.with_summary_list(
-            classes: %w[govuk-summary-list--no-border govuk-!-margin-bottom-0],
+            borders: false,
```
avinhurry, line 25: Nice! I didn't know you could do this.

app/services/teachers/search.rb
```diff
@@ -39,5 +57,17 @@ def where_query_matches(query_string)
 
       @scope.merge!(@scope.search(query_string))
     end
+
+    def where_ect_at(school)
+      return if school == :ignore
+
+      @scope.merge!(
+        @scope.eager_load(ect_at_school_periods: [:school, { mentorship_periods: { mentor: :teacher } }])
```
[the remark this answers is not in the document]
peteryates: Yep, that's the plan.

spec/services/teachers/search_spec.rb
peteryates: Ah well caught, thanks!

avinhurry (approved):

PR #482, opened 2025-04-16
474: event name tidy up

ghost:
## Description
We're removing author type from event names, renamed events from->to:

```
appropriate_body_passes_teacher -> teacher_passes_induction
appropriate_body_fails_teacher -> teacher_fails_induction
admin_fails_teacher_induction -> teacher_fails_induction
admin_passes_teacher_induction -> teacher_passes_induction
```

### DIFF
--- a/.github/workflows/backup-db.yml
+++ b/.github/workflows/backup-db.yml
@@ -10,6 +10,7 @@ on:
         type: choice
         options:
           - staging
+          - sandbox
           - production
       backup-file:
         description: |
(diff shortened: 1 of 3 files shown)


peterdavidhamilton (approved):
LGTM - my only suggestion (for the next ticket) is that we reorder `Events::Record` to group by type of action first, and then by author. I think it'll make it easier to see `what` is recorded alongside the versions of `who` can do it. cc @peteryates WDYT?

PR #486, opened 2025-04-17
475: event name tidy up

ghost:
renamed `teacher_induction_status_updated_by_trs` to `teacher_trs_induction_status_updated`

Group related event recording methods together to improve code readability

### DIFF
--- a/app/services/events/record.rb
+++ b/app/services/events/record.rb
@@ -71,8 +71,6 @@ module Events
 
     def record_event!
       check_relationship_attributes_are_persisted
-      # FIXME: This job is causing serialization errors when launched within a failing transaction block.
-      # We have not yet identified the root cause of this issue therefore we're going to run the job inline for now.
       RecordEventJob.perform_later(**attributes)
     end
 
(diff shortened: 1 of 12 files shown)


app/models/event.rb
```diff
@@ -2,18 +2,17 @@ class Event < ApplicationRecord
   EVENT_TYPES = %w[
     appropriate_body_adds_induction_extension
     appropriate_body_updates_induction_extension
-    import_from_dqt
```
peteryates, line 5: I think we should keep this, we used it when doing the initial import.

app/services/events/record.rb
```diff
@@ -102,6 +100,23 @@ def self.record_induction_period_closed_event!(author:, appropriate_body:, induc
...
+    def self.record_induction_period_updated!(author:, modifications:, induction_period:, teacher:, appropriate_body:, happened_at: Time.zone.now)
+      event_type = :induction_period_updated
+      heading = 'Induction period updated by admin'
+
+      new(event_type:, modifications:, author:, appropriate_body:, induction_period:, teacher:, heading:, happened_at:).record_event!
+    end
+
+    def self.record_induction_period_deleted!(author:, modifications:, teacher:, appropriate_body:, body: nil)
+      event_type = :induction_period_deleted
+      happened_at = Time.zone.now
```
peteryates, line 112: Worth making the `happened_at` consistent with the method above?

app/services/events/record.rb
```diff
@@ -170,35 +189,25 @@ def self.teacher_trs_attributes_updated!(author:, teacher:, modifications:, happ
...
-
-      event_type = :teacher_fails_induction
-      heading = "#{Teachers::Name.new(teacher).full_name} failed induction (admin)"
-      happened_at = induction_period.finished_on
-
-      new(event_type:, author:, appropriate_body:, teacher:, induction_period:, heading:, happened_at:).record_event!
-    end
+    # Admin Events
 
     def self.record_teacher_induction_status_reset_event!(author:, appropriate_body:, teacher:)
```
peterdavidhamilton, line 212: Is this event one for the `Teacher Status Events` section above, or is it different, or deserve its own section?

peterdavidhamilton:
How bothered are we about method name consistency in `Events::Record`? Some end with bangs, some with `_event`.

PR #491, opened 2025-04-19
Hide the `Event` model from the ERD

peteryates:
While it is part of the schema it's not really important from a data model point of view, and makes the diagram significantly more complicated due to its numerous relationships.

The diagram is clearer when it's not there.


peteryates:
Also `EarlyRollOutMentor` should be an attribute of teacher not its own table.

avinhurry (approved):

PR #496, opened 2025-04-22
Bulk CSV row-level validations

peterdavidhamilton:
- Refactor file-level checks to row-level
- Customise messages
- Control order and presentation


app/services/appropriate_bodies/process_batch/action.rb
```diff
@@ -0,0 +1,174 @@
...
+
+        nil
+      rescue TRS::Errors::TeacherNotFound
+        "Not found in TRS"
+      rescue TRS::Errors::ProhibitedFromTeaching
+        "Prohibited from teaching"
+      rescue TRS::Errors::QTSNotAwarded
+        "QTS not awarded"
+      rescue StandardError
+        "TRS API could not be contacted"
```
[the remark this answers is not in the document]
peterdavidhamilton: `25-28` takes the returned string and saves on the pending submission. Is that enough or can we do more?

app/models/pending_induction_submission_batch.rb
```diff
@@ -0,0 +1,184 @@
...
+    error: 'Error message',
+  }.freeze
+
+  ActionRow = build_row_class(ACTION_CSV_HEADINGS.keys)
+  ClaimRow = build_row_class(CLAIM_CSV_HEADINGS.keys)
+
+  # Associations
+  belongs_to :appropriate_body
+  has_many :pending_induction_submissions
+  has_one_attached :csv_file
```
peteryates, line 85: I don't really see why we want to hold onto the file here, and active storage adds complexity on the deployment/backup side.

The way it worked in my mind was that the upload request creates pending induction submissions which _represent_ the row.

Or is there a reason to keep the original that I missed?

peterdavidhamilton: @peteryates the relevant changes in this PR are just the newest commit but I needed it deployed for Joe and Priyah. You'd already commented about blob storage on the original PR and I am replacing active storage.

PR #498, opened 2025-04-22
Partnership model investigation

ethax-ross:
> ⚠️ Model naming TBC/open to input

[Jira-4215](https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-4215)
[Jira-4216](https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-4216)

- Add core partnership models

Add the core partnership models:

- `LeadProviderActivePeriod` (`ActiveLeadProvider` in original diagram)
- `LeadProviderDeliveryPartnership` (`AvailableProviderPairing` in original diagram)
- `SchoolPartnership`

Add minimal validation for the models and their relationships.

- Add `training_period` references to `LeadProviderActivePeriod` and `SchoolPartnership`

As per the UML diagram, we want `TrainingPeriod` to have a zero/one relationship to `SchoolPartnership` and `LeadProviderActivePeriod` (originally `ActiveLeadProvider` in the UML). Both `LeadProviderActivePeriod` and `SchoolPartnership` can have many training periods.

I've opted for the `TrainingPeriod` -> `LeadProviderActivePeriod` relationship being an `expression_of_interest` and the `TrainingPeriod` -> `SchoolPartnership` relationship as a `confirmed_school_partnership`.

There can only be one or the other, not both (and there must be at least one).

- Relax TrainingPeriod partnership validation

To avoid failures elsewhere in the codebase where we build `TrainingPeriod` from scratch (not using the factory) I'm omitting the validation that ensures a partnership is set. We can add this back in at a later date.

## Notes

- `LeadProviderActivePeriod` is used to define which of the lead providers are active for a given registration year.
- `LeadProviderDeliveryPartnership` describes which delivery partners are associated with which lead provider for each registration year.

These two models will be created up-front by RECT to align with how we setup a new academic year in ECF at the moment (which I believe involves importing the active lead providers and delivery partners from a CSV).

- `TrainingPeriod` describes a period in which a participant is in training. 
- Think of this as the induction record replacement to show how a participant's training changes over time.
- It will either have an `expression_of_interest` OR a `confirmed_school_partnership` (not both and at least one).
- When a participant is on-boarded their training period will either be:
  - Assigned to a `SchoolPartnership` through `confirmed_school_partnership` if one already exists for the desired school/delivery partner.
  - Assigned to a `LeadProviderActivePeriod` as an `expression_of_interest`
 - Once the lead provider confirms an `expression_of_interest` a `SchoolPartnership` will be created and the `confirmed_school_partnership` assigned.

We are currently expecting one delivery partner per school per lead provider, but in the future/after looking at the existing data we may need to support multiple. The schema would allow for that.

We are treating `SchoolPartnership` as a new model here, but in reality it will be a rename of the existing `ProviderPartnership` model.

```mermaid
erDiagram
    LeadProvider ||--|{ LeadProviderActivePeriod : ""
    RegistrationYear ||--|{ LeadProviderActivePeriod : ""
    LeadProviderActivePeriod ||--|{ LeadProviderDeliveryPartnership : ""
    DeliveryPartner ||--|{ LeadProviderDeliveryPartnership : ""
    LeadProviderDeliveryPartnership |o--|{ SchoolPartnership : ""
    LeadProviderActivePeriod |o..|{ TrainingPeriod : "expression_of_interest"
    School ||--|{ SchoolPartnership : ""
    SchoolPartnership |o..|{ TrainingPeriod : "confirmed_school_partnership"
```


app/models/delivery_partnership.rb
```diff
@@ -0,0 +1,8 @@
+class DeliveryPartnership < ApplicationRecord
+  belongs_to :active_period
+  belongs_to :delivery_partner
+  has_many :school_partnerships
+
+  validates :active_period, presence: true
+  validates :delivery_partner, presence: true
```
claire-hughez, line 7: Hey, so just testing my understanding..

This would be a list of active periods (where a LP is a valid LP, essentially) and during these periods which delivery partners can work with them?

They can then be selected to be used for school partnerships.

This makes sense to me, but I wonder if it's easy to conflate school_partnerhips with delivery_partnerships, and it's safer to refer to a partnership as something between a school, a delivery partner and lead provider. That's why 'pairings' was originally used I think.

app/models/active_period.rb
```diff
@@ -0,0 +1,9 @@
+class ActivePeriod < ApplicationRecord
+  belongs_to :lead_provider
+  belongs_to :academic_year
+  has_many :delivery_partnerships
+  has_many :training_periods
```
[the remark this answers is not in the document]
ethax-ross: Yes that's my understanding (I'm still trying to wrap my head around it 😅) - I agree the naming is too vague here, I'll maybe change it to `LeadProviderActivePeriod`

spec/factories/training_period_factory.rb
```diff
@@ -8,6 +8,8 @@
     started_on { generate(:base_training_date) }
     finished_on { started_on + 1.day }
 
+    association :expression_of_interest, factory: :lead_provider_active_period
```
cwrw, line 11: maybe we can add a trait for linking to a confirmed school partnership?

app/models/training_period.rb
```diff
@@ -55,6 +58,11 @@ def one_id_of_trainee_present
     errors.add(:base, "Only one id of trainee required. Two given") if ids.all?
   end
 
+  def one_partnership_present
```
cwrw, line 61: In discussions and the writeup, although we did say it's either phase one (EOI) or phase 2 (fully fledged school partnership) do we really need to clear the expression of interest, if we keep it just for the record, would it be an issue 🤔 ? I guess it would cause confusion with a circular links, but is there a way we can keep both to preserve the history of an EOI? happy to discuss

cwrw (commented):
looks great thanks Ross! just some minor comments/questions. On another note the PR for renaming has been merged, I'll review early next week so we can have an action plan as there are a lot of conflicts now

spec/factories/training_period_factory.rb
ethax-ross: I didn't add this as there isn't a need for it yet but happy to put it in up front 👍

app/models/delivery_partnership.rb
```diff
@@ -0,0 +1,8 @@
+class DeliveryPartnership < ApplicationRecord
```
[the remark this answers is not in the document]
peteryates: Maybe `ProviderPartnership` could make a quick return here (after #505) 😄

leandroalemao (commented):
Hi @ethax-ross I agree with most of it but took a while to understand TrainingPeriod -> LeadProviderActivePeriod relationship being an `expression_of_interest` and the TrainingPeriod -> SchoolPartnership relationship as a `confirmed_school_partnership`.. not very straightforward one. Maybe only a boolean flag on training period to differentiate and a `lead_provider_active_period_id`?

ethax-ross:
> Hi @ethax-ross I agree with most of it but took a while to understand TrainingPeriod -> LeadProviderActivePeriod relationship being an `expression_of_interest` and the TrainingPeriod -> SchoolPartnership relationship as a `confirmed_school_partnership`.. not very straightforward one. Maybe only a boolean flag on training period to differentiate and a `lead_provider_active_period_id`?

So if I understand you correctly then I think the problem with that is that they're pointing to different models; one is `LeadProviderActivePeriod` and the other is `SchoolPartnership`. I'm not sure how the boolean flag would help unless they were pointing to the same entity?

app/models/training_period.rb
cwrw: discussed with @ethax-ross @leandroalemao and then later @pete, this might be something we can do (and keep both records) we will discuss in the team meeting tomorrow before confirming way forward

ethax-ross:
Closing in favour of #532 which is the same changes re-applied on `main` with some minor differences.

app/models/training_period.rb
pete: > discussed with @ethax-ross @leandroalemao and then later @pete, this might be something we can do (and keep both records) we will discuss in the team meeting tomorrow before confirming way forward

I don't think I've discussed a Rails project for at least a year.

PR #500, opened 2025-04-22
Add disaster recovery hint to app config

peteryates:
Set `send_traffic_to_maintenance_page = true` can be used to send traffic to the maintenance page regardless of the app's state. Leaving it in the config as when we need it it might not be obvious where to put it, etc

### DIFF
--- a/config/terraform/application/application.tf
+++ b/config/terraform/application/application.tf
@@ -55,6 +55,11 @@ module "web_application" {
   max_memory = var.webapp_memory_max
 
   enable_logit = var.enable_logit
+
+  # Uncomment this when we want traffic to be redirected to the maintenance
+  # page during disaster recovery (i.e., while waiting for a database to be
+  # recreated)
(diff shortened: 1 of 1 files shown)


config/terraform/application/application.tf
```diff
@@ -55,6 +55,11 @@ module "web_application" {
   max_memory = var.webapp_memory_max
 
   enable_logit = var.enable_logit
+
+  # Uncomment this when we want traffic to be redirected to the maintenance
+  # page during disaster recovery (i.e., while waiting for a database to be
+  # recreated)
+  # send_traffic_to_maintenance_page = true
```
avinhurry, line 62: Curious if you considered putting this in a variable and controlling it via environment specific `.tfvars` files? E.g `staging.tfvars`?

That way we could test it out whenever we wanted to and would make accidental activation more difficult.

avinhurry (approved):

PR #502, opened 2025-04-22
[1417] Include day in ECT start date question

avinhurry:
### Context

When asking for the date that ECTs start training. We currently only ask for a month and a year. In addition to this we will also be asking for the day now.

### Changes proposed

- Collect the day from the user, in addition to the month and year
- Update error text to reflect new format
- Update hint text to reflect new format
- Use the format `"%-d %B %Y"` for the new date, e.g `2 January 1998`

### Before/After

| Before | After |
|--------|-------|
|image|image|

| Before | After |
|--------|-------|
|image|image|


ltello (approved):
Nice

PR #472, opened 2025-04-15
Rename `AcademicYear` to `RegistrationPeriod`

peteryates:
This model now holds boundaries of each registration period so the name academic year isn't a good fit.

Changes:

* rename `AcademicYear` to `RegistrationPeriod` in the model, specs, factory, etc
* add columns for `started_on`, `finished_on` and `enabled` as specified in DFE-Digital/register-ects-project-board#1476
* update the seeds to reflect new fields
   ```
  🌱 Adding registration periods
  🌱   2021 (running from 2021-06-01 until 2022-05-31)
  🌱   2022 (running from 2022-06-01 until 2023-05-31)
  🌱   2023 (running from 2023-06-01 until 2024-05-31)
  🌱   2024 (running from 2024-06-01 until 2025-05-31)
  🌱   2025 (running from 2025-06-01 until 2026-05-31)
  ```

PR #505, opened 2025-04-23
Rename `ProviderPartnership` to `SchoolPartnership`

peteryates:
Following on from #472, this isn't the final form of SchoolPartnership, just the first step in its metamorphosis towards the new EOI model.


peteryates:
Oh Rubocop probably complained because of [.rubocop_todo.yml](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/.rubocop_todo.yml), which I didn't know was a thing!

tonyheadford (approved):

PR #506, opened 2025-04-23
Fix transaction rollback and improve RecordOutcome service.

ghost:
- Fix bug where jobs were enqueued despite transaction rollback
- Refactor pass!/fail! methods to use shared record_outcome! method
- Replace case statements with dynamic method calls
- Use update! instead of update for better error handling
- Add explicit validation failure tests
- Configure ActiveJob test adapter


peterdavidhamilton (approved):

peteryates (approved):
Looks good, left one suggestion to make code easier to find.

Issue #507, opened 2025-04-23
Implementing the new EOI/partnership model


peteryates:
## ECF1

In ECF1 the `Partnership` model holds the mapping between unique combinations of:

* `School`
* `Cohort`
* `LeadProvider` 
* `DeliveryPartner`

`InductionProgramme` records belong to a single `Partnership`.

```mermaid
erDiagram
    DeliveryPartner ||--|{ Partnership : ""
    LeadProvider ||--|{ Partnership : ""
    School ||--|{ Partnership : ""
    Cohort ||--|{ Partnership : ""
    Partnership ||--|{ InductionProgramme : ""
```

## ECF2 current

In ECF2 we currently have a model like this. It's largely the same as ECF1's but doesn't involve the school yet, there's no concept of a _partnership_ between schools and provider pairings per registration period.

```mermaid
erDiagram
    DeliveryPartner ||--|{ ProviderPartnership : ""
    LeadProvider ||--|{ ProviderPartnership : ""
    AcademicYear ||--|{ ProviderPartnership : ""
    ProviderPartnership ||--|{ TrainingPeriod : ""
```

## ECF2 proposed

The current provider pairings implementation is similar to ECF1's. A row in the `provider_partnerships` table tells us which <mark>lead provider</mark> is working with which <mark>delivery partner</mark> for the <mark>academic year</mark>.

To solve the problems with this model and allow us record expressions of interest (DFE-Digital/register-ects-project-board#1212) we came up with the following model. The reasoning behind it is covered [in this discussion](https://github.com/DFE-Digital/register-ects-project-board/discussions/1364#discussioncomment-12702737).

```mermaid
erDiagram
    LeadProvider ||--|{ ActiveLeadProvider : ""
    RegistrationPeriod ||--|{ ActiveLeadProvider : ""
    ActiveLeadProvider ||--|{ AvailableProviderPairing : ""
    DeliveryPartner ||--|{ AvailableProviderPairing : ""
    AvailableProviderPairing |o--|{ SchoolPartnership : ""
    ActiveLeadProvider |o..|{ TrainingPeriod : "phase 1 (expression of interest)"
    School ||--|{ SchoolPartnership : ""
    SchoolPartnership |o..|{ TrainingPeriod : "phase 2 (confirmed)"
```

We also, separately, decided that it would make more sense for `AcademicYear` to be called `RegistrationPeriod`.

## Getting there

[...]

PR #513, opened 2025-04-24
Add `ActiveLeadProvider` model with relationships, seeds, specs etc

peteryates:
~~This follows on from #505 and is a WIP - rebase it's merged 🚧~~

It adds `ActiveLeadProvider`, a table intended to hold a mapping between lead providers and registration periods - so we know which lead providers were active when.

Refs #507


db/seeds.rb
```diff
@@ -40,6 +36,15 @@ def describe_registration_period(ay)
...
 end
 
+def describe_lead_provider(lp, years)
+  years_description = if years.any?
+                        Colourize.text(years.map(&:year).join(', '), :green)
+                      else
+                        Colourize.text('inactive', :red)
+                      end
+  print_seed_info("#{lp.name} (#{years_description})", indent: 2)
+end
```
peteryates, line 46: This prints:

image

db/seeds.rb
```diff
@@ -192,16 +197,39 @@ def describe_teacher(teacher)
...
 
-grove_institute = LeadProvider.create!(name: 'Grove Institute').tap { |dp| describe_lead_provider(dp) }
-LeadProvider.create!(name: 'Evergreen Network').tap { |dp| describe_lead_provider(dp) }
-national_meadows_institute = LeadProvider.create!(name: 'National Meadows Institute').tap { |dp| describe_lead_provider(dp) }
-LeadProvider.create!(name: 'Woodland Education Trust').tap { |dp| describe_lead_provider(dp) }
-LeadProvider.create!(name: 'Teach Orchard').tap { |dp| describe_lead_provider(dp) }
-LeadProvider.create!(name: 'Highland College University').tap { |dp| describe_lead_provider(dp) }
-wildflower_trust = LeadProvider.create!(name: 'Wildflower Trust').tap { |dp| describe_lead_provider(dp) }
-LeadProvider.create!(name: 'Pine Institute').tap { |dp| describe_lead_provider(dp) }
+grove_institute = LeadProvider.create!(name: 'Grove Institute')
```
cwrw, line 210: maybe we can use proper lead providers rather than example ones? happy to discuss

cwrw (commented):
looks good, just some minor comments

db/seeds.rb
cwrw: I was thinking this mostly because eventually we'll create API tokens to test on the review app, for ECF1 we had set and known tokens such as `ambition-token`, `ucl-token` etc for the known lead providers which made testing easy and they were remember by everyone. If we go with random lead providers this would be harder to remember. Lead providers mostly don't change so it'll just make things easier and more standard for testing I think. I think the only real data I'd suggest is provider names, everything else can be fake as it is now

peteryates: I'd agree if we were generating them with Faker, but they're hardcoded. We can definitely use the set and known tokens idea though.

I like having a clear distinction between real and fake and would prefer not to mix up.

cwrw: discussed on slack: we'll use fake names closer to the real ones. I'll open a different pr not part of this one. Thanks

cwrw (approved):

PR #517, opened 2025-04-25
add events for mentor registration and mentorship assignment

ltello:


peteryates:
I believe this functionality is now present:

1. https://github.com/DFE-Digital/register-early-career-teachers-public/commit/05e50d17a
2. https://github.com/DFE-Digital/register-early-career-teachers-public/commit/089ceeb32

PR #519, opened 2025-04-25
514: event name tidy up

ghost:
Remove author type from event names, we can infer the author from the event metadata instead


peteryates (approved):
Nice!

PR #523, opened 2025-04-25
Record an event when mentors are registered

### LINKED ISSUE #521: Record an event when a new mentor is registered into the service
When a school registers a mentor we should record an event that'll be visible in the relevant timelines.

peteryates:
Fixes #521 

## Changes

- **Reformat attr_reader list**
- **Add school_user factory**
- **Add teacher_registered_as_mentor event**
- **Make mentor reg wizard record an event on completion**

### DIFF
--- a/app/services/events/record.rb
+++ b/app/services/events/record.rb
@@ -220,6 +220,16 @@ module Events
       new(event_type:, induction_period:, modifications:, author:, appropriate_body:, teacher:, heading:, happened_at:, body:).record_event!
     end
 
+    # ECT and mentor events
+
+    def self.record_teacher_registered_as_mentor_event!(author:, mentor_at_school_period:, teacher:, school:, happened_at: Time.zone.now)
+      event_type = :teacher_registered_as_mentor
+      teacher_name = Teachers::Name.new(teacher).full_name
+      heading = "#{teacher_name} was registered as a mentor at #{school.name}"
+
+      new(event_type:, author:, heading:, mentor_at_school_period:, teacher:, school:, happened_at:).record_event!
+    end
+
   private
 
     def attributes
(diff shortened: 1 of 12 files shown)


spec/factories/sessions/school_user_factory.rb
peteryates, line 1: We should probably create factories for the other various kinds of users too, it'll definitely remove some duplication from the tests.

tonyheadford (approved):
LGTM

app/wizards/schools/register_mentor_wizard/check_answers_step.rb
```diff
@@ -15,7 +15,7 @@ def previous_step
 
       def persist
         ActiveRecord::Base.transaction do
-          AssignMentor.new(ect:, mentor: mentor.register!).assign!
+          AssignMentor.new(ect:, mentor: mentor.register!(author: wizard.author)).assign!
```
ltello, line 18:
```suggestion
          AssignMentor.new(ect:, mentor: mentor.register!(author:)).assign!
```

app/services/events/record.rb
```diff
@@ -220,6 +220,16 @@ def self.record_induction_period_reopened_event!(author:, induction_period:, mod
       new(event_type:, induction_period:, modifications:, author:, appropriate_body:, teacher:, heading:, happened_at:, body:).record_event!
     end
 
+    # ECT and mentor events
+
+    def self.record_teacher_registered_as_mentor_event!(author:, mentor_at_school_period:, teacher:, school:, happened_at: Time.zone.now)
```
ltello, line 225: We can get the :school from the : mentor_at_school_period.
Worth considering whether to remove the :school param even though that might trigger an extra query to the db?

peteryates: I'm on the fence with this, it makes things less verbose but means we need to set relationships up in the specs which has become a problem in the past.

PR #529, opened 2025-04-28
[1470] Rename `mentor_completion_*` columns on `Teacher`

avinhurry:
### Context

The current column names related to mentors are unclear. We are renaming them to be more explicit, making it easier to understand the data model at a glance.

### Changes Proposed

- Rename `mentor_completion_reason`  to `mentor_became_ineligible_for_funding_reason` on the `Teacher` table
- Rename `mentor_completion_date` to `mentor_became_ineligible_for_funding_on` on the `Teacher` table.
- Rename the associated enum from `mentor_completion_reason` to `mentor_became_ineligible_for_funding_reason`
- Update all code references.
- Re-introduce the ERD diagram (previously lost)

### Notes

We are doing the migration and code changes in one PR for the following reasons:

- the app has a small user base and this reduces complexity.

- The updated columns on the `Teacher` records are not used in production by real users.

- No need for backfilling and so a better ROI.

- Keeping it together preserves a cleaner GIT history.


app/components/schools/mentors/ect_mentor_training_details_component.rb
```diff
@@ -15,7 +15,7 @@ def render?
       end
 
       def eligible_for_training?
-        teacher.mentor_completion_date.nil?
+        teacher.mentor_became_ineligible_for_funding_on.nil?
       end
```
peteryates, line 19: Should this be checking that the date is earlier than today?

peteryates: Sorry just noticed this is still in draft - if you'd already spotted it ignore me 🫥

avinhurry: Good question, it's one I also asked: 

See my post on slack here: https://ukgovernmentdfe.slack.com/archives/C06V31R0Z17/p1744210652433659?thread_ts=1744208328.640409&cid=C06V31R0Z17

Do you think it should be changed?

app/components/migration/participant_profile_component.html.erb
```diff
@@ -40,11 +40,11 @@
     else
       list.with_row do |row|
         row.with_key(text: "Mentor completion date")
-        row.with_value(text: participant_profile.mentor_completion_date)
+        row.with_value(text: participant_profile.mentor_became_ineligible_for_funding_on)
```
avinhurry, line 43: @tonyheadford just wanted to double check that these do need to be updated?

Let me know if there is anything else I should consider.

tonyheadford: No these are models connected to ECF1 so we shouldn't change these, thanks for checking 👍

avinhurry: Nice one - i've removed. Thank you!

app/components/schools/mentors/ect_mentor_training_details_component.rb
avinhurry: Resolving this. As it was discussed on slack.

We will create a new card/cards to make things clearer here as part of future work.

PR #536, opened 2025-04-29
feat: add script to fix mismatched induction statuses for teachers

ghost:
Process:
0- Retrieve all users whose latest induction period status isn't in sync with TRS 
1- does a sync trs attributes inline (not in a job) 
2- checks if the TRS status reflects the latest outcome 
3- if it does, skip
4- if it doesn't, send the relevant update to TRS's API to sync it with the db

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -9,8 +9,6 @@ on:
     branches:
       - main
   pull_request:
-    branches:
-      - main
   merge_group:
 
 permissions:
--- a/documentation/service-rules/ecf1/lead-provider-api.md
+++ b/documentation/service-rules/ecf1/lead-provider-api.md
@@ -2,237 +2,26 @@
 title: Lead provider API
 ---
 
-Learn about how the payment engine works once it receives declarations from lead providers using the Early Career Framework API.
+Lead providers must use this API to view, submit and update participant data so they receive accurate payments from the Department for Education (DfE) for their ECF-based training.
 
-These are the steps
+Once a participant has been registered to the service by their school, data associated with them becomes available to lead providers via the API.
 
-1. Provider submits a declaration.
-2. The system validates data.
-3. If invalid, errors are flagged up with corrective action suggestions.
-4. If valid, the declaration is created and linked to an output fee calculation.
-5. Service fees and banding are determined and applied.
-6. Final payments are calculated and displayed for contract managers.
+## Contents
 
-## Declarations in a nutshell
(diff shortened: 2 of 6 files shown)


db/scripts/fix_mismatched_induction_statuses.rb
```diff
@@ -0,0 +1,81 @@
...
+    t.trs_induction_status
+  FROM teachers t
+  INNER JOIN latest_induction_periods lip ON t.id = lip.teacher_id
+  WHERE lip.outcome IN ('pass', 'fail')
+  AND t.trs_induction_status = 'InProgress'
+SQL
+
+mismatched_teachers = Teacher.find_by_sql(sql)
+
+Rails.logger.debug "Found #{mismatched_teachers.count} teachers with mismatched statuses"
```
peteryates, line 38: It feels a _little_ odd here that it's doing `Teacher.find_by_sql` but the SQL doesn't return a teacher row.

Maybe if the query above was amended so it just returned the `t.id as teacher_id`

```sql
  SELECT t.id as teacher_id
  FROM teachers t
  INNER JOIN latest_induction_periods ...
```

then we could pull the IDs out into an array:

```
mismatched_teacher_ids = ActiveRecord::Base.connection.execute(sql).map { |r| r["id"] }
```

Then we could pull them out cleanly:

```ruby
mismatched_teachers = Teacher.where(id: mismatched_teacher_ids)
```

db/scripts/fix_mismatched_induction_statuses.rb
```diff
@@ -0,0 +1,81 @@
...
+
+  Rails.logger.debug "Processing teacher #{teacher.trn}..."
+
+  # Sync with TRS
+  Rails.logger.debug "  Syncing with TRS..."
+  Teachers::RefreshTRSAttributes.new(teacher).refresh!
+
+  # Check if TRS status now matches outcome
+  teacher.reload
+  if teacher.trs_induction_status == teacher_data.outcome.capitalize
```
peteryates, line 51: Not sure this'll work, our outcomes our `pass` and `fail` but the statuses [returned from the API](https://preprod.teacher-qualifications-api.education.gov.uk/swagger/index.html) are

```
[ None, RequiredToComplete, Exempt, InProgress, Passed, Failed, FailedInWales ]
```

Might need to just hardcode them.

peteryates (commented):
Looks good, sorry for reviewing before it's made it out of draft but left a couple of suggestions 🙂

PR #544, opened 2025-05-01
Partnership endpoints investigation

leandroalemao:
### Context

- Ticket: [CPDLP-4218](https://dfedigital.atlassian.net/browse/CPDLP-4218)

We're investigating how the partnerships endpoints would look like with the new models.

- Create `GET /api/v3/partnerships` and `GET /api/v3/partnerships/{id}`
- Add query for multiple partnerships
- Add query for single partnership
- Check performance of query
  - ~50ms in comparison to ~80ms in ECF1
- Basic serialiser to surface data - we’ll build the proper one later
  - look into existing models only, for instance if induction tutor isn’t there we won’t investigate
- are there any issues with showing relationship partnerships or challenged partnerships or any risks
  - we don't surface relationship partnerships in ECF1 `where(relationship: false)`, but we display when a partnership is challenged. I don't see any risks in displaying relationship partnerships in ECF2 but it needs to be assessed by the business.


Gemfile
```diff
@@ -45,6 +45,11 @@ gem "savon"
 # Render smart quotes
 gem 'rubypants'
 
+# For the API
+gem "blueprinter"
+gem "oj"
+gem "pagy"
```
peteryates, line 51: I know this is just a spike but we don't need to specify `pagy` as it's already pulled in by `govuk-components`.

app/serializers/api/partnership_serializer.rb
```diff
@@ -0,0 +1,23 @@
+module API
+  class PartnershipSerializer < Blueprinter::Base
+    class AttributesSerializer < Blueprinter::Base
+      exclude :id
+
+      field(:cohort) { |p, _| p.registration_period.year.to_s }
+      field(:urn) { |p, _| p.school.urn.to_s }
+      field(:school_id) { |p, _| p.school.id }
```
cwrw, line 8: same here for school id and delivery partner id and uuids

app/serializers/api/partnership_serializer.rb
```diff
@@ -0,0 +1,23 @@
...
+      field(:delivery_partner_id) { |p, _| p.delivery_partner.id }
+      field(:delivery_partner_name) { |p, _| p.delivery_partner.name }
+
```
cwrw, line 11: what about challenged partnerships? we need to show the old information from ECF1 but also we don't support it in the new world? can we add those for now and have that as part of the discussion as well?

db/migrate/20250429103738_create_api_tokens.rb
```diff
@@ -0,0 +1,13 @@
+class CreateAPITokens < ActiveRecord::Migration[8.0]
+  def change
+    create_table :api_tokens do |t|
```
cwrw, line 3: discussed in person: this wasn't part of the spike but we can keep for now among a few extra things to get the API going :)

spec/features/appropriate_bodies/claim_an_ect_spec.rb
```diff
@@ -119,7 +119,7 @@ def now_i_should_be_on_the_claim_an_ect_register_page
   end
 
   def when_i_enter_the_start_date
-    page.get_by_label('Day').fill('3')
+    page.get_by_label('Day').fill('7')
```
cwrw, line 122: not sure why this needed to change 🤔

app/serializers/api/partnership_serializer.rb
```diff
@@ -0,0 +1,23 @@
+module API
+  class PartnershipSerializer < Blueprinter::Base
+    class AttributesSerializer < Blueprinter::Base
+      exclude :id
```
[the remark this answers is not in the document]
cwrw: discussed in person: we will need to bring those values in and do something similar to NPQ on autogeneration of uuid. We can discuss in the session the preferred naming and hear thoughts from RECT

app/serializers/api/partnership_serializer.rb
cwrw: discussed in person: there are pros/cons to showing challenged vs none. We will be asking providers on this data, and bring up in the next session with RECT

app/serializers/api/partnership_serializer.rb
```diff
@@ -0,0 +1,23 @@
...
+
+      field(:cohort) { |p, _| p.registration_period.year.to_s }
+      field(:urn) { |p, _| p.school.urn.to_s }
+      field(:school_id) { |p, _| p.school.id }
+      field(:delivery_partner_id) { |p, _| p.delivery_partner.id }
+      field(:delivery_partner_name) { |p, _| p.delivery_partner.name }
+
+      field :created_at
+      field :updated_at
```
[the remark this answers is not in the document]
cwrw: discussed in person: we will add a new field that we can update when necessary to avoid a touch model

cwrw (approved):

ethax-ross (approved):

PR #545, opened 2025-05-01
Record events when mentorships are formed

### LINKED ISSUE #527: Record an event when a mentorship is created
When a new mentorship is created either via the register ECT or register mentor or assign mentor features in the service, an event should be created.

Note, there's some extra complexity here because we only link a single teacher to an event. We should probably create two similar events here, one for the mentor and one for the mentee, with headings something like this:

* John Smith began mentoring Julie Jones
* Julie Jones is being mentored by John Smith

peteryates:
When a mentor is assigned to an ECT (the mentee) we want to record events so the timeline is populated properly.

As two teachers are involved we will record two events, one for the mentor and one for the mentee. This solves problems of _ownership_, because a mentor might be at a different school to the mentee and the event should be visible for both.

Fixes #527.

### DIFF
--- a/spec/services/events/record_spec.rb
+++ b/spec/services/events/record_spec.rb
@@ -623,6 +623,7 @@ RSpec.describe Events::Record do
           heading: "Rhys Ifans started mentoring Steffan Rhodri",
           event_type: :teacher_starts_mentoring,
           happened_at: Time.zone.now,
+          metadata: { mentor_id: teacher.id, mentee_id: mentee.id },
           **author_params
         )
       end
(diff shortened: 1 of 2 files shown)


spec/services/events/record_spec.rb
```diff
@@ -602,4 +602,58 @@
       end
     end
   end
+
+  describe '.record_teacher_starts_mentoring_event!' do
+    let(:started_on_param) { { started_on: 2.years.ago.to_date } }
+    let(:school) { FactoryBot.create(:school) }
```
avinhurry, line 608: Just FYI we've included `FactoryBot::Syntax::Methods` now, so you should be able to just do:

```suggestion
    let(:school) { create(:school) }
```

peteryates: We decided to omit `FactoryBot::Syntax::Methods` on purpose because otherwise it's not obvious where `#create` is defined or what it's doing. I think we should stick to being explicit.

ltello (approved):

PR #549, opened 2025-05-02
Support creating/updating partnerships

ethax-ross:
[Jira-4222](https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-4222)

### Support creating/updating partnerships

Add support for creating and updating partnerships by replicating the service behaviour in ECF.

Add `ecf_id` to `School`, `DeliveryPartner` and `LeadProvider` models.

Add convenience methods to `School` for eligibility checks.

Add `Create` and `Update` services.

Accept expressions of interest for the same school when creating a `SchoolPartnership`.

### Notes on implementation

There's going to be some difficulty with keeping parity with ECF and also adopting the RECT modelling/terminology. I've opted to keep the service request/attribute naming consistent with RECT, however we'll need to translate the attributes back into what we currently return in ECF (along with any error messages - replacing `registration year` with `cohort`, for example). We may be able to do something around custom I18n backends/middleware to achieve this globally (not quite the same, but similar to [this in NPQ reg](https://github.com/DFE-Digital/npq-registration/blob/main/lib/i18n/backend/npq.rb)).

In general the mix of terminology is going to be confusing; we may want to look into seeing if we can reference the RECT terminology instead of the ECF terms for 'cohort/registration year', 'partnership/school partnership' for example.

In RECT we can't tell which school is intending to offer DfE funding for which registration period as there is no notion of a `SchoolCohort`. Instead, we have dropped the validation for this as well as the restriction on a single LP/School/Registration Year partnership (off the back of Nathan's suggestion). To determine if we should allow a partnership at all we are now checking if there are any `TrainingPeriod` records with the school relationship, which implies a FIP participant as CIP don't get training periods in RECT.

I have not gone as far as implementing partnership reminders/notifications in this spike.

Relationships and challenged partnerships are not going to be supported in RECT.

The behaviour of expressions of interests is still TBC, but for the sake of this spike we will accept expressions of interests for a given school if they have the same LP/registration year combination and we ignore any others.

A couple of pain points I ran into with this spike:

- Training period factories are a bit iffy; you can't just `create(:training_period)` as often the date ranges don't align with the associated ECT/mentor at school periods. I found it would sometimes work locally but then fail in CI, so I have had to manually create the at school periods with started on/finished on dates and set the training period dates manually as well to ensure they will align.
- Creating schools in a certain state is hard because of how the school/GIAS school factory is setup. To create a school with a custom GIAS school you need to `nil` it in the factory and assciate it with the `urn` which is confusing, for example: `build(:school, gias_school: nil, urn: gias_school.urn)
- The randomness in the school factory made lots of my seed data flakey until I made them explicit - I think we should get rid of the `[true false].sample` stuff in the seeds, personally.

### DIFF
--- a/app/services/api/school_partnerships/create.rb
+++ b/app/services/api/school_partnerships/create.rb
@@ -1,0 +1,116 @@
+# frozen_string_literal: true
+
+module API::SchoolPartnerships
+  class Create
+    include ActiveModel::Model
+    include ActiveModel::Attributes
+
+    attribute :registration_year
+    attribute :school_ecf_id
+    attribute :lead_provider_ecf_id
+    attribute :delivery_partner_ecf_id
+
+    validates :registration_year, :school_ecf_id, :lead_provider_ecf_id, :delivery_partner_ecf_id, presence: true
+    validate :registration_period_exists
+    validate :lead_provider_exists
+    validate :school_exists
+    validate :school_is_not_cip_only
+    validate :school_is_eligible
+    validate :delivery_partner_exists
+    validate :lead_provider_delivery_partnership_exists
+    validate :school_partnership_does_not_already_exists
+    validate :school_training_period_exists
+
+    def create
+      return false unless valid?
+
+      SchoolPartnership.create!(
+        school:,
+        lead_provider_delivery_partnership:
+      ).tap { |school_partnership| accept_expressions_of_interest(school_partnership) }
+    end
+
+  private
+
+    def accept_expressions_of_interest(school_partnership)
+      expressions_of_interest = school_partnership.lead_provider_active_period.expressions_of_interest
+      expressions_of_interest.for_school(school).each do |expression_of_interest|
+        expression_of_interest.update!(school_partnership:)
+      end
+    end
+
+    def registration_period
+      @registration_period ||= RegistrationPeriod.find_by(year: registration_year) if registration_year
+    end
+
+    def lead_provider
+      @lead_provider ||= LeadProvider.find_by(ecf_id: lead_provider_ecf_id) if lead_provider_ecf_id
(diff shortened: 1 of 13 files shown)


app/services/api/school_partnerships/create.rb
```diff
@@ -0,0 +1,116 @@
...
+
+    def school_partnership_does_not_already_exists
+      return unless school && lead_provider_delivery_partnership
+
+      existing_school_partnership = school.school_partnerships.exists?(lead_provider_delivery_partnership:)
+
+      errors.add(:school_ecf_id, "School partnership already exists for the lead provider, delivery partner and registration year") if existing_school_partnership
+    end
+
+    def school_training_period_exists
```
cwrw, line 86: not sure about this - let's discuss as the school might not have set up participants yet and when they do they can link to the partnership 🤔

app/services/api/school_partnerships/create.rb
```diff
@@ -0,0 +1,116 @@
...
+        school:,
+        lead_provider_delivery_partnership:
+      ).tap { |school_partnership| accept_expressions_of_interest(school_partnership) }
+    end
+
+  private
+
+    def accept_expressions_of_interest(school_partnership)
+      expressions_of_interest = school_partnership.lead_provider_active_period.expressions_of_interest
+      expressions_of_interest.for_school(school).each do |expression_of_interest|
```
cwrw, line 37: is it worth checking those expressions don't have existing school partnerships for whatever reason 🤔

ethax-ross: Yeah good shout, I've updated - not sure about the `pending` terminology but it'll do for now

PR #558, opened 2025-05-06
Add more error types to TRS API client

### LINKED ISSUE #540: Deal with deactivated teachers better
We run a job every minute that [refreshes the details](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/services/teachers/refresh_trs_attributes.rb) for a few teachers by pulling updates from TRS. All teachers will be refreshed every couple of days.

Some requests to TRS are met with a 404 though. This happens when a teacher is deactivated - either removed entirely or two records merged into one.

I had a chat with @gunndabad he said there will eventually be a webhook broadcast mechanism so they can inform consuming services when something changes. This is reliant on them storing a link in the data that would provide a way to link the old deactivated TRN to the new one when merged. Once this is in place the API could respond with a [301 (moved permenantly)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/301).

Until then though, James suggested responding with a [410 (gone)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/410) instead.

We should

* [x] update our `TRS::APIClient#find_teacher` method so that it reacts to `410` by marking the teacher record as deactivated in our database. I suggest setting a `trs_deactivated_at` timestamp for this rather than a boolean
* [x] add a scope to teachers called `.active_in_trs` that omits records that have a value in `trs_deactivated_at`
* [x] update the `Teachers::ScheduleTRSSyncJob` to omit teachers who are deactivated in TRS so we don't get the same exception over and over

Additionally, we'll need to take an action when teachers are deactivated so these records should show up _somewhere_ (maybe Blazer for now, low-fi)

peteryates:
We have some teachers who have been deactivated in TRS. Currently the TRS API returns a `404 Not found` but soon (hopefully) it'll return a `410 Gone` instead.

We can use this to mark teachers who have been deactivated in TRS as deactivated in our database. Each one will need to be manually checked to see what action needs to be taken, thankfully it's rare.

Once we've marked teachers as `trs_deactivated` we won't try to refresh their attributes any more, which should cut plenty of noise out of Sentry.

## Changes

* Make the TRS API client understand error codes other than `404`
* Add a `trs_deactivated` flag (I chose a flag rather than a timestamp as we already have `trs_details_last_refreshed_at` which will be the same)
* Update the sync job to mark teachers who come back with a `410` as deactivated
* Update the sync job scheduler to omit deactivated teachers

Fixes #540


peterdavidhamilton (approved):

peteryates:
Chatted with @gunndabad - he suggested [this alternative](https://github.com/DFE-Digital/teaching-record-system/pull/2013) approach to informing us about errors because introducing `410` would be a version-breaking API change. I said we're happy to hold on until their next version ships and stick with HTTP codes as I think it's easier for us to stay on the same page.

No reason why this can't be merged now though, so let's go 🚢

PR #563, opened 2025-05-07
Spike partnerships data compatibility

ethax-ross:
[Jira-4221](https://dfedigital.atlassian.net.mcas.ms/browse/CPDLP-4221)

Leaving this as draft as we probably don't want to merge this one down.

- Minor updates to validation

Disable the training period validation as we haven't brought this across from ECF yet.

- Add partnerships data migration script

A spike for migrating partnerships data.

Migrates delivery partners, lead providers and cohorts.

Expects GIAS import to have been ran locally, then imports schools from ECF.

Creates partnerships.

#### Notes

Most of the partnerships pull across fine, there are a handful with school-related errors that we will need to dig into further when we investigate the schools data:

```
schools_not_found.count
=> 5 (out of 54796 total)

schools_not_found.values.map(&:full_messages).flatten.tally
=> {"Ukprn has already been taken"=>5}

invalid_school_partnerships.values.map(&:full_messages).flatten.tally
=> 
{"School ECF School is not eligible"=>399,
 "School ECF School is CIP only"=>69,
 "School ECF School does not exist"=>14}
```

If we omit the school validation we only get errors for schools that don't exist:

```
=> {"School ECF School does not exist"=>15}
```

### DIFF
--- a/app/models/migration/school.rb
+++ b/app/models/migration/school.rb
@@ -2,5 +2,15 @@ module Migration
   class School < Migration::Base
     has_many :school_cohorts
     has_many :partnerships
+
+    def in_england?
+      english_district_code?(administrative_district_code)
+    end
+
+    def english_district_code?(district_code)
+      # expanded to include the 9999 code which seems to have crept in and is preventing a couple of schools onboarding
+      # the establishment codes should filter out any that should not come in that are 9999 district
+      district_code.to_s.match?(/^([Ee]|9999)/)
+    end
   end
 end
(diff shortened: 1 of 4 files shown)


cwrw:
summary from Ross:
- 63,397 partnerships migrate successfully and 482 fail. Of those;
  - 399 fail as the school is not 'eligible' in its current state
  - 69 fail as the school is CIP-only
  - 14 fail as the school does not exist, most likely because:
- There are 5 schools that seem to exist in RECT matching on UKPRN but have a different URN
- Two school categories don't exist in RECT but do in ECF (these may or may not have partnerships):
  - Children's centre linked site
  - Children's centre

The validation around training periods was disabled for the run as the data is not yet migrated (this determines if a school can have a partnership/intends to delivery DfE funded training - if none exists it would result in a validation error on creating the partnership)

ethax-ross:
Closing this PR off as we aren't going to merge it into the base branch.

PR #565, opened 2025-05-08
Participants endpoint investigation

leandroalemao:
### Context

- Ticket: [CPDLP-4223](https://dfedigital.atlassian.net/browse/CPDLP-4223)

We're investigating how the participants endpoints would look like with the new models.

- This is just a spike and for discussion
- Create `GET /api/v3/participants` and `GET /api/v3/participants/{id}`
- Add query for multiple partnerships
- Add query for single partnership
- Check performance of query
  - ~90ms in comparison to ~300ms in ECF1
- Basic serialiser to surface data - we’ll build the proper one later

### DIFF
--- a/app/serializers/api/participant_serializer.rb
+++ b/app/serializers/api/participant_serializer.rb
@@ -1,0 +1,50 @@
+module API
+  class ParticipantSerializer < Blueprinter::Base
+    class AttributesSerializer < Blueprinter::Base
+      exclude :id
+
+      field(:full_name) { |p, _| [p.trainee.teacher.trs_first_name, p.trainee.teacher.trs_last_name].join(' ') }
+      field(:trn) { |p, _| p.trainee.teacher.trn }
+
+      field(:ecf_enrolments) do |object, options|
+        object.trainee.training_periods.map do |training_period|
+          next unless training_period.lead_provider == options[:lead_provider]
+
+          {
+            training_record_id: training_period.id,
+            email: training_period.trainee.email,
+            mentor_id: training_period.for_ect? ? training_period.trainee&.mentors&.map(&:id) : nil,
+            school_urn: training_period.school_partnership&.school&.urn,
+            participant_type: training_period.for_ect? ? :ect : :mentor,
+            cohort: training_period.school_partnership&.registration_period&.year.to_s,
+            training_status: training_period.trainee.teacher.trs_induction_status,
+            participant_status: nil,
+            teacher_reference_number_validated: nil,
+            eligible_for_funding: nil,
+            pupil_premium_uplift: nil,
+            sparsity_uplift: nil,
+            schedule_identifier: nil,
+            delivery_partner_id: training_period.delivery_partner&.id,
+            withdrawal: [],
+            deferral: [],
+            created_at: training_period.created_at.rfc3339,
+            induction_end_date: training_period.finished_on&.strftime("%Y-%m-%d"),
+            mentor_funding_end_date: nil,
+            cohort_changed_after_payments_frozen: nil,
+            mentor_ineligible_for_funding_reason: training_period.for_mentor? ? training_period.trainee.teacher.mentor_became_ineligible_for_funding_reason : nil,
+          }
+        end
+      end
+
+      field :created_at
+      field :updated_at
+    end
+
+    identifier :id
+    field(:type) { "participant" }
+
+    association :attributes, blueprint: AttributesSerializer do |participant|
+      participant
+    end
+  end
+end
--- a/app/serializers/api/participant_serializer.rb
+++ b/app/serializers/api/participant_serializer.rb
@@ -1,50 +1,63 @@
 module API
   class ParticipantSerializer < Blueprinter::Base
+    identifier :ecf_user_id, name: :id
+    field(:type) { "participant" }
+
     class AttributesSerializer < Blueprinter::Base
       exclude :id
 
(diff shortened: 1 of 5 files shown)


app/serializers/api/participant_serializer.rb
```diff
@@ -0,0 +1,50 @@
...
+
+      field(:full_name) { |p, _| [p.trainee.teacher.trs_first_name, p.trainee.teacher.trs_last_name].join(' ') }
+      field(:trn) { |p, _| p.trainee.teacher.trn }
+
+      field(:ecf_enrolments) do |object, options|
+        object.trainee.training_periods.map do |training_period|
+          next unless training_period.lead_provider == options[:lead_provider]
+
+          {
+            training_record_id: training_period.id,
```
cwrw, line 14: we need to think about if training record will match participant profile id in ECF1, or will it be different? where would we put participant profile id going forward as training record is closer to an induction record than a profile?

app/serializers/api/participant_serializer.rb
```diff
@@ -0,0 +1,50 @@
...
+            email: training_period.trainee.email,
+            mentor_id: training_period.for_ect? ? training_period.trainee&.mentors&.map(&:id) : nil,
+            school_urn: training_period.school_partnership&.school&.urn,
+            participant_type: training_period.for_ect? ? :ect : :mentor,
+            cohort: training_period.school_partnership&.registration_period&.year.to_s,
+            training_status: training_period.trainee.teacher.trs_induction_status,
```
cwrw, line 20: I think training status will be `nil` and this is for participant status

app/serializers/api/participant_serializer.rb
```diff
@@ -0,0 +1,50 @@
+module API
+  class ParticipantSerializer < Blueprinter::Base
+    class AttributesSerializer < Blueprinter::Base
+      exclude :id
```
cwrw, line 4: I was wondering if we should be querying by teacher, as that's the main id similar to user on ecf1?

app/services/participants/query.rb
```diff
@@ -0,0 +1,74 @@
...
+    end
+
+    def participant(id: nil)
+      scope.where(id:)
+    end
+
+  private
+
+    def all_participants
+      TrainingPeriod.from("(#{ect_participants.to_sql} UNION #{mentor_participants.to_sql}) as training_periods")
```
cwrw, line 21: similar to the questions above, I think we should be querying by teacher as thats the same as User in ecf1? happy to chat

cwrw (commented):
thanks Leandro a few observations and questions

app/services/participants/query.rb
```diff
@@ -0,0 +1,74 @@
...
+    def latest_mentor_training_periods_join
+      TrainingPeriod
+      .select(Arel.sql("DISTINCT FIRST_VALUE(training_periods.id) OVER (#{latest_mentor_training_period_order}) AS latest_id"))
+      .joins(:school_partnership)
+      .merge!(SchoolPartnership.for_lead_provider(lead_provider.id))
+    end
+
+    def latest_ect_training_period_order
+      <<~SQL
+        PARTITION BY(training_periods.ect_at_school_period_id) ORDER BY
```
ethax-ross, line 50: Maybe one for its own spike, but I think we should aim to avoid these kinds of complex queries with logic in them and maybe explore some alternative options

app/serializers/api/participant_serializer.rb
```diff
@@ -0,0 +1,50 @@
...
+            teacher_reference_number_validated: nil,
+            eligible_for_funding: nil,
+            pupil_premium_uplift: nil,
+            sparsity_uplift: nil,
+            schedule_identifier: nil,
+            delivery_partner_id: training_period.delivery_partner&.id,
+            withdrawal: [],
+            deferral: [],
+            created_at: training_period.created_at.rfc3339,
+            induction_end_date: training_period.finished_on&.strftime("%Y-%m-%d"),
```
ethax-ross, line 31: Just a minor thing and not a problem for now, but I think we should centralise/standardise the date formats when we get to implementing the serializers down the line

app/serializers/api/participant_serializer.rb
```diff
@@ -0,0 +1,50 @@
+module API
+  class ParticipantSerializer < Blueprinter::Base
+    class AttributesSerializer < Blueprinter::Base
```
ethax-ross, line 3: I liked the [view structure we used in NPQ reg](https://github.com/DFE-Digital/npq-registration/blob/main/app/serializers/api/participant_serializer.rb#L9), I think we should use it again for the API versions here

app/serializers/api/participant_serializer.rb
```diff
@@ -0,0 +1,63 @@
...
+      view :v3 do
+        field(:full_name) { |t, _| Teachers::Name.new(t).full_name }
+        field(:trn) { |t, _| t.trn }
+
+        field(:ecf_enrolments) do |teacher, _options|
+          (teacher.ect_at_school_periods + teacher.mentor_at_school_periods).map do |school_period|
+            # Better using uuids instead as we don't know whether latest_training_period is an id from ect_at_school_periods or mentor_at_school_periods
+            training_period = school_period.training_periods.find { |training_period| teacher.latest_training_period.include?(training_period.id) }
+
+            next if training_period.blank?
```
mooktakim, line 18: fyi for anyone looking at this in the future
this will create nil value in the `map` result.

leandroalemao:
Closing this for now as was just an investigation.

PR #573, opened 2025-05-09
Make `early_roll_out_mentor` an attribute of `Teacher`

### LINKED ISSUE #492: Make `early_roll_out_mentor` an attribute of `Teacher`
It's a table on its own at the moment containing just a TRN.

It should be a boolean column on the `teachers` table called `early_roll_out_mentor`.

_Originally posted by @peteryates in https://github.com/DFE-Digital/register-early-career-teachers-public/issues/491#issuecomment-2816670862_

peteryates:
Having a separate table that only contains a list of TRNs seems weird when we have a teachers table where TRN is an attribute.

If a teacher was an early roll out mentor, it should be an attribute of the teacher record.

This change drops the early roll out mentors table and model and replaces it with a flag on the teacher record. It also introduces a trait called `:early_roll_out_mentor` on the teachers factory we can use in tests to set them accordingly.

The seeds have been modified so the same records are marked as being ERO:

image

## Review notes

Pay particular attention to the changes I made in `app/wizards/schools/register_mentor_wizard/mentor.rb`. I didn't want to change logic outside of the service classes but other specs failed. I'm going  to do a follow up PR which moves the eligibility logic to its own service class. 

## Changes

- **Drop early_roll_out_mentors table, add teacher flag**
- **Remove early roll out mentors model and factory**
- **Update the seeds to populate the ERO mentor flag**
- **Update ERO tests to work with the teacher record**

Fixes #492


peteryates:
Closed as this took the wrong approach. There's already a way of checking if someone's ERO on teacher :man_facepalming:

PR #578, opened 2025-05-12
1538 - content updates to align with prototype

StephenC-V1:
1538 - content updates to align with prototype


tonyheadford (approved):
LGTM - just a question about should we be using `’` and `‘` instead of `'` in our content?

PR #582, opened 2025-05-12
1392 - Block ECTs from being registered for training if they have failed induction

ltello:
[Issue](https://github.com/DFE-Digital/register-ects-project-board/issues/1392)

image

Video:

https://github.com/user-attachments/assets/defff024-f466-46c2-a0c1-d8645c60d28c

### DIFF
--- a/app/components/test_guidance_component.rb
+++ b/app/components/test_guidance_component.rb
@@ -15,6 +15,9 @@ class TestGuidanceComponent < ViewComponent::Base
       [
         ["1000890", "15/01/1997", "JL049125A", "Induction completed"],
         ["3002564", "05/07/1991", "JK028223B", "Exempt from induction"],
+        ["3001899", "16/07/1991", "JL056123C", "Failed induction"],
+        ["3003943", "02/02/1964", "AB722128C", "No QTS"],
+        ["1000787", "01/04/1984", "JL056123B", "Alerts but not prohibited"],
         ["3002586", "03/02/1977", "OA647867D", ""],
         ["3002585", "02/01/1966", "MA251209B", ""],
         ["3002584", "24/11/1955", "RE937588C", ""],
--- a/spec/views/schools/register_ect_wizard/induction_failed.html.erb_spec.rb
+++ b/spec/views/schools/register_ect_wizard/induction_failed.html.erb_spec.rb
@@ -1,0 +1,20 @@
+RSpec.describe "schools/register_ect_wizard/induction_failed.html.erb" do
+  let(:ect) { double('ECT', full_name: 'John Doe') }
+
+  before do
+    assign(:ect, ect)
+    render
+  end
+
+  it "sets the page title" do
+    expect(sanitize(view.content_for(:page_title))).to eql('You cannot register John Doe')
+  end
+
+  it "displays the ECT's full name in the body" do
+    expect(rendered).to have_text("Our records show John Doe has failed their induction and is not eligible for further ECT training.")
+  end
+
+  it "includes a link to register another ECT" do
+    expect(rendered).to have_link('Register another ECT', href: schools_register_ect_wizard_find_ect_path)
+  end
+end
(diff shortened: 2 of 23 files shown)


spec/views/schools/register_ect_wizard/induction_failed.html.erb_spec.rb
```diff
@@ -0,0 +1,20 @@
+RSpec.describe "schools/register_ect_wizard/induction_failed.html.erb" do
+  let(:ect) { double('ECT', full_name: 'John Doe') }
+
+  before do
+    assign(:ect, ect)
+    render
+  end
+
+  it "sets the page title" do
+    expect(sanitize(view.content_for(:page_title))).to eql('You cannot register John Doe')
```
avinhurry, line 10: I don't think `sanitize` is needed here, `page_title` is just a plain string in the view helper, could it be simplified to: 

```suggestion
    expect(view.content_for(:page_title)).to eql('You cannot register John Doe')
```

spec/wizards/schools/register_ect_wizard/national_insurance_number_step_spec.rb
```diff
@@ -0,0 +1,199 @@
...
+        allow(fake_client).to receive(:find_teacher).and_return(
+          TRS::Teacher.new(
+            'trn' => '1234568',
+            'firstName' => 'Kirk',
+            'lastName' => 'Van Houten',
+            'dateOfBirth' => '1977-02-03',
+            'induction' => {
+              'status' => 'Failed'
+            }
+          )
```
avinhurry, line 127: This is repeated 3 times with the only difference being the induction status. 

It could be an idea to DRY it up by extracting it into a let block and then over-riding the induction status as required: 

```
let(:induction_status) { 'Failed' }

let(:teacher_data) do
  {
    'trn' => '1234568',
    'firstName' => 'Kirk',
    'lastName' => 'Van Houten',
    'dateOfBirth' => '1977-02-03',
    'induction' => {
      'status' => induction_status
    }
  }
end
```

avinhurry (approved):
Really good PR! Clear structure, good tests and covers edge cases.

Just a left a couple of totally optional suggestions on making the tests a bit cleaner.

app/components/test_guidance_component.rb
```diff
@@ -15,6 +15,9 @@ def rows
       [
         ["1000890", "15/01/1997", "JL049125A", "Induction completed"],
         ["3002564", "05/07/1991", "JK028223B", "Exempt from induction"],
+        ["3001899", "16/07/1991", "JL056123C", "Failed induction"],
+        ["3003943", "02/02/1964", "AB722128C", "No QTS"],
+        ["1000787", "01/04/1984", "JL056123B", "Alerts but not prohibited"],
```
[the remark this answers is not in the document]
ltello: These are ones that Claire wants to include for future scenarios

spec/views/schools/register_ect_wizard/induction_failed.html.erb_spec.rb
ltello: It is still needed. Special characters in people's names will be escaped by Rails and would not match the expected string.

spec/wizards/schools/register_ect_wizard/national_insurance_number_step_spec.rb
ltello: I just added a new commit with all your spec suggestions and a few more.

avinhurry: Nice - specs look much neater now. Thanks @ltello! 🚢

avinhurry (approved):
🚢

claire-hughez:
This looks great Lorenzo, I think I've found some unrelated issues though I'll share on Slack.

Image

Is this the name that TRS has stored in the pre-prod API lol?

Normally they have normal names so I was just surprised!

The link to check a teacher's record is incorrect. I feel like this has come up before. We might need a ticket to fix this separately if it's broken in other places. It should be https://www.gov.uk/guidance/check-a-teachers-record. I'm happy to have a search in the repo to fix it elsewhere if needed if you're ok to fix in this ticket?

Our 'block' to stop registration for the exempt record TRN 3002564 and the completed induction record 
TRN 1000890 is not working. This is pre-production data. This is the case in both RIAB for ABs and RECT for schools. This could be because:
1) In Pre-prod, the record is no longer exempt
2) We have some sort of bug somewhere on this

The pre-prod data for prohibited teacher records still works.

I'll check with TRS colleagues. This also highlights them changing the status of pre-prod records is might happen quite frequently as it happened with the failed record too.

PR #586, opened 2025-05-13
Run migrations before trying to boot the app

peteryates:
This is wholesale lifted from https://github.com/DFE-Digital/itt-mentor-services/.

There it was intended to ensure there are no problems caused by multiple containers trying to run the same migrations simultaneously. Here we'll benefit from that too, but also it will simplify our Dockerfile and mean we only need to worry about starting the app, not doing database migration stuff too.

### DIFF
--- a/config/terraform/application/application.tf
+++ b/config/terraform/application/application.tf
@@ -33,87 +33,26 @@ module "application_configuration" {
   )
 }
 
-# Run database migrations
-# https://guides.rubyonrails.org/active_record_migrations.html#preparing-the-database
-# Terraform waits for this to complete before starting web_application and worker_application
-resource "kubernetes_job" "migrations" {
-  metadata {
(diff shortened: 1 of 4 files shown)


config/terraform/application/application.tf
```diff
@@ -33,8 +33,86 @@ module "application_configuration" {
...
+    name      = "${var.service_name}-${var.environment}-migrations"
+    namespace = var.namespace
+  }
+
+  spec {
+    template {
+      metadata {
+        labels = { app = "${var.service_name}-${var.environment}-migrations" }
+        annotations = {
+          "logit.io/send" = "true"
```
RMcVelia, line 50: can you add "fluentbit.io/exclude" = "true"
as it stops us writing azure container logs alongside logit logs.

RMcVelia (approved):

PR #590, opened 2025-05-15
API docs skeleton

peteryates:
This change introduces some placeholder pages which can be used for documenting the API.

There's a landing page with a masthead which borrows heavily from [NPQ's API guidance](https://register-national-professional-qualifications.education.gov.uk/api/guidance).

| Main page | Content page |
| ------ | ----- |
| Screenshot From 2025-05-15 11-40-45 | Screenshot From 2025-05-15 11-40-54 |

## Release notes

| Rlease notes | Release note data |
| ------ | ----- |
|image | image |

Fixes DFE-Digital/register-ects-project-board#1615, DFE-Digital/register-ects-project-board#1617, DFE-Digital/register-ects-project-board#1618


app/controllers/api/guidance_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+    end
+
+    def page
+      path = params[:page]
+
+      template = {
+        "page-1" => "page_1",
+        "page-2" => "page_2",
+        "page-3" => "page_3",
+      }.fetch(path)
```
peteryates, line 17: I hardcoded these for now. When we have a better idea of which pages we'll have this can be moved to a service object that gives us a list of paths/templates/page titles etc we can use here and above in `#navigation_items`.

.env.example
```diff
@@ -11,6 +11,7 @@ ENABLE_SCHOOLS_INTERFACE=false
 ENABLE_LOCAL_TLS=true
 TEST_GUIDANCE=true
 ENABLE_BLAZER=true
+ENABLE_API=true
```
cwrw, line 14: as discussed we'll probably need to move this to a feature flag to avoid deploying to enable/disable at the time of migration

app/views/api/guidance/release_notes.yml
```diff
@@ -0,0 +1,47 @@
+# - title: This is an example note
```
cwrw, line 1: So I guess you went with yaml for the structure, but I wonder if we end up tying our hands if we need more custom stylings, maybe its ok for now

app/controllers/api/guidance_controller.rb
cwrw: I prefer service object to model (which we use in NPQ)

app/controllers/api/guidance_controller.rb
```diff
@@ -0,0 +1,29 @@
+module API
+  class GuidanceController < ApplicationController
```
cwrw, line 2: can we add request specs for the guidance pages?

cwrw (commented):
looks great, just some minor comments

leandroalemao:
Can mermaid.js be integrated as well if needed in the future?

.env.example
peteryates: I think adding feature flags is a bigger change, happy for it to be done following this though.

app/views/api/guidance/release_notes.yml
peteryates: Yeah, markdown is supported here though - not sure what extra styling we might need in a release note.

app/controllers/api/guidance_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+    skip_before_action :authenticate
+
+    layout 'api_guidance', only: 'show'
+
+    def show
```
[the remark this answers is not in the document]
peteryates: Yeah - it's easy enough to change once we get some actual content though.

.env.example
cwrw: makes sense to be separate 👍

peteryates:
I'm sure we can add some mermaidjs in there if we need it @leandroalemao :smile:

cwrw (approved):
🙌

PR #599, opened 2025-05-16
[RECT-1607] Add investigation on API Token

leandroalemao:
https://github.com/DFE-Digital/register-ects-project-board/issues/1607

We use bearer tokens to connect to the Lead provider API. Those are stored in the database, where we generate API keys for providers. We will need to create this model before we build the API authentication.

- Investigate the API token model and any fields we need to bring over from ECF1
- Do we need to make this table agnostic to the lead provider API, incase we create other APIs in RECT? think of scope or a different solution (similar to NPQ) https://github.com/DFE-Digital/npq-registration/blob/b14ef7b8d54eba446ced1e4f99f29697cd40260f/app/models/api_token.rb#L1
- Investigate any validations required on the associated fields
- There is a spike for API tokens for reference https://github.com/DFE-Digital/register-early-career-teachers-public/pull/544/files
- Present the findings to the team so they can pick up the next ticket

----
#### Option 1 (`APIToken`) using scopes Aas per NPQ reg:

APIToken is scoped by one of the hardcoded options
```
enum :scope, {
    lead_provider: "lead_provider",
    teacher_record_service: "teacher_record_service",
  }, validate: true
```
- it has a direct `belongs_to` association with the LeadProvider model, as it's the main user of the api for now.
- the token generation and validation logic is tied to specific scopes
----

#### Option 2 (`NewAPIToken`) using polymorphic associations

- any model can have associated api tokens
- removed the need for enums and conditional validations
- new api types can be added without changing the token schema
- each api owner model is responsible for its own tokens
- token generation and validation logic is done the same across all api types

```
class LeadProvider < ApplicationRecord
  has_many :api_tokens, as: :tokenable, dependent: :destroy
end

class AdminAPI < ApplicationRecord
  has_many :api_tokens, as: :tokenable, dependent: :destroy
end
```


app/models/api_token.rb
```diff
@@ -0,0 +1,27 @@
+class APIToken < ApplicationRecord
+  has_secure_token :hashed_token, length: 32
+  encrypts :hashed_token, deterministic: true
+
+  belongs_to :lead_provider, optional: true
+
+  enum :scope, {
```
cwrw, line 7: do we have other options of this implementation, as this is similar to NPQ?

db/seeds.rb
```diff
@@ -736,3 +736,9 @@ def describe_teacher(teacher)
 end
 
 print_seed_info('Adding funding exemptions:')
+
+print_seed_info('Adding APITokens:')
```
cwrw, line 740: After Mook merges his pr for providers, we can add a token for each provider, but will wait for that one

lib/tasks/api_token.rake
```diff
@@ -0,0 +1,27 @@
...
+    task :generate_token, %i[lead_provider_id] => :environment do |_t, args|
+      logger = Rails.env.test? ? Rails.logger : Logger.new($stdout)
+
+      lead_provider = LeadProvider.find_by(id: args.lead_provider_id)
+      raise("Unknown lead_provider_id") unless lead_provider
+
+      scope = APIToken.scopes[:lead_provider]
+      unhashed_token = APIToken.create_with_random_token!(scope:, lead_provider:)
+
+      logger.info "API Token created: #{unhashed_token}"
```
cwrw, line 23: we should include the name of the provider here, but one for discussion on if we need a rake task vs doing manually

ethax-ross (approved):
Looks good past Abeer's comments

mooktakim (approved):

app/models/api_token.rb
leandroalemao: Hi @cwrw I've added a new option using polymorphic association

cwrw:
thanks @leandroalemao for the second polymorphic option, let's chat tomorrow briefly before summarising with the team

peteryates:
Polymorphic relationships aren't enforceable at the database level so let's avoid them if pos.

db/migrate/20250519155554_create_new_api_tokens.rb
```diff
@@ -0,0 +1,15 @@
+class CreateNewAPITokens < ActiveRecord::Migration[8.0]
+  def change
+    create_table :new_api_tokens do |t|
+      t.references :tokenable, polymorphic: true, index: true
+
+      t.string :hashed_token, null: false
+      t.string :description
```
tonyheadford, line 7: Could `:description` be made mandatory?  Would make like a bit easier when looking at data perhaps if we by default include the `LeadProvider.name` when generating for example?  Is it overkill - what do you think?

ethax-ross:
> Polymorphic relationships aren't enforceable at the database level so let's avoid them if pos.
> 
> I'd rather have a `lead_provider_id` and, if we need one later (say for ABs) an `appropriate_body_id` - each with their own foreign key to the appropriate table. We can always validate that only one is present if needed (we could do this [at the database level too](https://api.rubyonrails.org/v6.1/classes/ActiveRecord/ConnectionAdapters/SchemaStatements.html#method-i-add_check_constraint) if we wanted).

I’d usually agree with avoiding polymorphic relationships, but in this case I think it’s the better approach. Using multiple foreign keys and check constraints introduces more complexity and tightly couples the model to current use cases. With polymorphism its keeping things cleaner and more extensible; especially if we might associate with more types in the future. We can still ensure data integrity with Rails validations and tests.

db/migrate/20250519155554_create_new_api_tokens.rb
leandroalemao: Yeah agree!

leandroalemao:
Let's say we have 10 types of API Keys.. how the table look like? and the validations? and the constraints? It's hard to extend the API token functionality using multiple foreign keys

cwrw:
agree with choosing polymorphism as well for extensibility and easily adding different types in the future. The NPQ model was done this way as an after thought, as it was already implemented just for lead providers so I think it'll be quite confusing and complex in the future

mooktakim:
Polymorphism makes sense here as we're trying to recreate it with scopes anyway.
And the description is 👍

I'd go with Option 2

cwrw:
discussed in person: we're going to go with simple foreign key for now, however when a new API comes in we'll transition to polymorphic when needed

PR #607, opened 2025-05-19
[1639] Updated lead provider names, 5 providers only, included 2021 cohort

mooktakim:
Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/1639

* Updated lead provider names with 5 pretty unique names that are hard to recognise.
* Added 2021 for Ambitious and Better practice network.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -244,24 +244,18 @@ rp_2025 = RegistrationPeriod.create!(year: 2025, started_on: Date.new(2025, 6, 1
 
 print_seed_info("Adding lead providers")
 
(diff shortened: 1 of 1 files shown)


db/seeds.rb
```diff
@@ -245,43 +239,43 @@ def describe_teacher(teacher)
 
 grove_artisan_partnership_2021 = SchoolPartnership.create!(
```
cwrw, line 240: I think we might need to rename the partnerships here and everywhere else to the new names?

leandroalemao (approved):

PR #608, opened 2025-05-19
Make statement seed output tabular

peteryates:
This follows on from #604 and changes the seed output so it's in a table rather than a list. It displays the same information.

## Changes

- **Tabularize the statements seed output**
- **Print blank line between groups of statements**


ethax-ross (approved):
Looks good 👍 the lead provider name colourisation maybe isn't needed now, I found it helped when you had multiple in groups to distinguish them but I don't think its adding much here as they aren't repeated now

db/seeds.rb
```diff
@@ -126,6 +132,41 @@ def describe_teacher(teacher)
   print_seed_info("#{teacher_name} (early roll out mentor: #{ero_status})", indent: 2)
 end
 
+def describe_group_of_statements(lead_provider, statements, month_col_width: 15, year_col_width: 10)
```
ethax-ross, line 135: Just a heads up that this will probably break when you rebase; off the back of a suggestion by Abeer I'm now generating two years worth of statements for each lead provider active period:

https://github.com/DFE-Digital/register-early-career-teachers-public/pull/604/files#diff-c61a903a1201b0c7f076b0270377c20390c26f655ce63689d7d843296fcf63f3R767

This is because we accept payments across multiple years, so if a lead provider is active for the 2024 registration period they will have statements in 2024 and 2025.

peteryates: It did indeed :joy:

I just repushed.

peteryates: Ah I see now, one would overwrite the other. I'll concat them.

peteryates: image

ethax-ross (approved):

PR #612, opened 2025-05-20
[1468] - Rename chosen_* fields on schools table to last_chosen_*

ltello:
[Ticket](https://github.com/DFE-Digital/register-ects-project-board/issues/1468)

Rename `School#chosen_*` fields to `School#last_chosen_*` to clearly mean they are populated after the choices made by the school for its last teacher registered.

Also its association columns and other methods derived from their name.


app/services/schools/register_ect.rb
```diff
@@ -45,7 +45,7 @@ def register!
       not_registered_as_an_ect!
 
       ActiveRecord::Base.transaction do
-        update_school_choices!
+        update_school_last_choices!
```
peteryates, line 48: `#update_last_school_choices` probably makes more sense here

peteryates (approved):
Looks good, left a minor suggestion.

app/services/schools/register_ect.rb
ltello: Funny. I named it initially as you suggest but them I thought it would mean: update the choices of the last school.

PR #613, opened 2025-05-20
Spike: API request testing

ethax-ross:
Spike into different ways we could go about testing the API requests; we wanted to look into end to end testing in feature specs, mocking out the underlying services and end to end testing in request specs.

This is in an effort to avoid getting into a similar state as we find ourselves in for ECF request specs where the tests are long, hard to read and maintain and have a lot of boilerplate/duplication. See [this as an example](https://github.com/DFE-Digital/early-careers-framework/blob/main/spec/requests/api/v3/ecf/participants_spec.rb).

When we tackled this in NPQ registration we went with a mixture of mocking out some of the underlying services (such as the service calls to create/update information) and also calling the whole stack at times (for the GET endpoints). We extracted as much as we could to shared contexts. This is an improvement on ECF (see [the corresponding spec file](https://github.com/DFE-Digital/npq-registration/blob/main/spec/requests/api/v3/participants_spec.rb)) however the underlying shared contexts got quite complicated (see [this as an example](https://github.com/DFE-Digital/npq-registration/blob/main/spec/support/shared_examples/api_create_endpoint_support.rb#L5-L12)).

---

[End to end testing in feature specs](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/613/commits/2c6b6d52fcefd7b660d0fab6e938f42f237ba07c) 

This turned out to be a bit of a non-starter. Rails feature tests are heavily geared towards driving the tests through a browser as a user would. They don't really fit the scenario of making API calls and you have to work against Rails to do it (accessing the page driver directly or by spinning up a request context in playwright).

**Pros**
Inline with existing end to end tests

**Cons**
Fighting with how Rails wants feature tests to work

---

[Request tests with mocking](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/613/commits/e3e5f9f557d1904797e8824e661a5751ba02b7aa) 

These tests aim to isolate and test what the controller is responsible for, mocking out the underlying services. Again, this feels like we're working against Rails where typically a request spec would exercise the whole stack. An example of where this falls down is trying to mock the pagination where you have to fallback on `expect_any_instance_of`. The old Rails controller tests would be better suited to this approach as you could stub out controller methods. We end up which quite complex shared contexts going this route, as we did for some of the NPQ reg tests.

**Pros**
Faster as we're not exercising the whole stack
We're not over-testing underlying services

**Cons**
We're not exercising the whole stack
Setup is more complex
Some things are hard/hacky to mock (pagination)
Tests need to know a lot about the implementation

---

[Request specs without mocking](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/613/commits/c863af2fdb45ebc7af973c3dc81cbad59541f261) 

This setup feels like a nice balance and more inline with how Rails is guiding you to test. Its slightly harder to ensure coverage of an underlying service call than with mocking, but by asserting on the created object we should be able to verify a service is called correctly. The aim would be to only test the controller responsibilities, so not making a request for every conceivable input and instead relying on the underlying unit tests to provide robust coverage for service behaviour.

**Pros**
Works the whole stack
Setup is less complex (especially for simpler GET calls)
Tests aren't tightly coupled to the implementation details

**Cons**
Ensuring services are called with correct arguments is not explicit/may be prone to gaps
Tests will be slower


peteryates:
Thanks @ethax-ross - this is a great overview.

I thought we could just do something along the lines of `type: :api_feature` and make them a bit more request like (so split by what they do rather than how they do it) - but if it means we're swimming upstream against Rails agree it's a non-starter.

On 3., this sounds like the way to go.

> The aim would be to only test the controller responsibilities, so not making a request for every conceivable input and instead relying on the underlying unit tests to provide robust coverage for service behaviour.

If we can stick to that, and just have a handful of specs per endpoint to make sure the parts are individually plugged in correctly - but with most of the testing of those parts going on at the unit level - we'll avoid most repetition I was worried about.

I'm happy with that approach. It should be fast enough not to have to break out any parallelised rspec hacks too :crossed_fingers:

avinhurry:
Thanks for laying all of this out so clearly @ethax-ross, it’s a really helpful comparison.

My vote is also for request specs without mocking. It feels like the most realistic and rails esk approach and uses the full stack without getting convoluted in mocks or brittle test setups.

The shared examples here keep things clean and DRY, which should help us avoid the duplication and complexity that has been been seen in the ECF 1 specs. If we focus on controller responsibilities and use unit tests for the service logic, I think this strikes a good balance between test coverage and maintainability.

I also think the pros of clearer, more realistic and less fragile tests outweigh the cons of slightly slower test speeds.

tonyheadford:
Nice one @ethax-ross 🎉  another vote here for option 3

mooktakim:
Option 3  👍

leandroalemao:
Option 3 🏆

ltello:
Thank you @ethax-ross for the great analysis.
Yes, option 3 looks much more sensible to me too. Loads of things to mock on option 2, very implementation dependent and kind of more complex code even thought it would run faster.

PR #614, opened 2025-05-20
1526: Allow edits of induction period with outcomes

ghost:
## Description

We've built Edit Induction Period for open inductions, and enabled editing of terms for closed inductions, in our admin console.

We don't allow users right now to edit dates of induction periods for closed inductions. We have had support tickets about this. To reduce dev effort on support, we need to make this function available to admin users.

## User story

**As** a support user
**I need** to be able to update dates on closed inductions
**So that** I can help to resolve support issues from users more quickly and without the need for a developer

## Acceptance criteria

- [x] 1. "Edit Induction Period" option is available for any induction period, regardless of whether the induction is closed
- [x] 2. All validation rules for existing Edit Induction Period are in place (e.g. overlapping induction period not allowed)
- [x] 3. When overall induction Start Date or End Date are changed, there is an API call to update the relevant field in TRS too.
- [x] 4. An event is recorded in the database


app/services/admin/update_induction_period.rb
```diff
@@ -16,20 +16,20 @@ def initialize(author:, induction_period:, params:)
...
 
-      ActiveRecord::Base.transaction do
-        success = [
-          induction_period.save!,
-          record_event(modifications),
-          notify_trs_of_start_date_change(previous_start_date)
-        ].all?
+      # Prevent setting dates to nil for induction periods with outcomes
+      if induction_period.outcome.present? && params.key?(:finished_on) && params[:finished_on].nil?
+        raise ActiveRecord::RecordInvalid.new(induction_period), "End date cannot be set to nil for induction periods with outcomes"
```
peterdavidhamilton, line 26: I don't think this validation scenario is in the feature spec but it'd be good to add. If the user can read this, we should reword it (unless @PriyaNahar2 disagrees)
 
`End date cannot be removed once the induction period has an outcome`

spec/services/admin/update_induction_period_spec.rb
```diff
@@ -34,34 +36,111 @@
...
           )
         end
+
+        context "when trying to set end date to nil" do
+          let(:params) { { finished_on: nil } }
+
+          it "raises an error" do
+            expect { service.update_induction_period! }.to raise_error(
+              ActiveRecord::RecordInvalid,
+              "End date cannot be set to nil for induction periods with outcomes"
```
peterdavidhamilton, line 80: 👀

peterdavidhamilton (commented):
Small content change might be needed

peterdavidhamilton (approved):

PriyaNahar2:
@emily-prudence-dfe just tagging you here -

Agree with @peterdavidhamilton that content needs rework to make the error message clear however, don't think it's priority for this /next sprint. Maybe this PR can be merged and then ticket to be raised for the backlog to iterate content and monitor if users need more clarity.

emily-prudence-dfe:
I was able to remove an end date from an induction period with an outcome in testing @edujackedu 

I couldn't find any other issues otherwise

PR #621, opened 2025-05-21
1510: Set teacher trs status to in progress after claiming them

ghost:
After we claim an ECT, their status should always be 'InProgress'.

Fixing the bug introduced here https://github.com/DFE-Digital/register-early-career-teachers-public/pull/581/files#diff-86bd7e6d615fbb6566173a35519c938b7dee3668ca39cc18c5913b5bb3aca8e8R51

### DIFF
--- a/.github/workflows/brakeman.yml
+++ b/.github/workflows/brakeman.yml
@@ -25,7 +25,7 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Setup Ruby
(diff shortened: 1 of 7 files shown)


peterdavidhamilton (approved):
👍🏼

emily-prudence-dfe:
Tested this - found that when claiming an ECT who starts in "Required to Complete", their status updates straight away to "in progress"

Merge away @edujackedu

PR #622, opened 2025-05-21
[1648] Statement serializer with jbuilder

mooktakim:
### Context

* Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/1648

* Statements serializer using `jbuilder`
* rspec view test for the serializer

### DIFF
--- a/app/views/api/v3/statements/show.json.jbuilder
+++ b/app/views/api/v3/statements/show.json.jbuilder
@@ -1,0 +1,13 @@
+json.id @statement.api_id
+json.type "statement"
+
+json.attributes do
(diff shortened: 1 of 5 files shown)


app/views/api/v3/statements/show.json.jbuilder
```diff
@@ -0,0 +1,13 @@
+json.id @statement.api_id
+json.type "statement"
+
+json.attributes do
+  json.month Date::MONTHNAMES[@statement.month]
```
ethax-ross, line 5: How would we handle different serialisations for the versions of the API with jbuilder? it looks like its scoped to the API version here, so would sharing/extending for other versions be a pain?

PR #625, opened 2025-05-22
[1647] Create Statement query service

leandroalemao:
### Context

[Issue #1647](https://github.com/DFE-Digital/register-ects-project-board/issues/1647)

We need a query service in order to return statement responses.

- [X] Add a new query service for retrieving statements
- [X] Allow the retrieval of multiple and single statements
- [X] Follow the pattern set in NPQ: https://github.com/DFE-Digital/npq-registration/blob/main/app/services/statements/query.rb this includes the location of the query service, as well as filters
- [X] We can also include the same ordering as well from NPQ
- [X] We can leave the type out for now as that won't be included

### Guidance to review

Review app


spec/models/statement_spec.rb
```diff
@@ -14,6 +16,24 @@
     it { is_expected.to validate_uniqueness_of(:api_id).case_insensitive.with_message("API id already exists for another statement") }
   end
 
+  describe "scopes" do
+    describe ".with_state" do
+      it "selects only statements with states matching the provided name" do
+        expect(Statement.with_state("foo").to_sql).to include(%(WHERE "statements"."state" = 'foo'))
```
ethax-ross, line 22: I get that we do these tests elsewhere, but it might be worth us getting consistent on how we test scopes cc @cwrw

spec/services/statements/query_spec.rb
```diff
@@ -0,0 +1,265 @@
...
+
+          expect(query.statements).to eq([])
+        end
+
+        it "does not filter by lead provider if none is supplied" do
+          condition_string = %(" WHERE \"lead_providers\".\"id")
+
+          query = described_class.new
+
+          expect(query.scope.to_sql).not_to include(condition_string)
```
ethax-ross, line 45: I would vote that we avoid these kind of assertions. It doesn't feel right to be asserting on raw SQL when we're two levels removed from it (by ActiveRecord and then the query service).

cwrw: agree although we did this before I'd rather we don't do this here as well

spec/models/statement_spec.rb
cwrw: agree, having a look at other models with scope tests they are testing which records are returned, can we do something similar here please?

mooktakim: agree, we shouldn't be testing SQL

app/services/api/concerns/filter_ignorable.rb
```diff
@@ -0,0 +1,13 @@
...
+  module Concerns
+    module FilterIgnorable
+      extend ActiveSupport::Concern
+
+      def ignore?(filter:)
+        filter == :ignore || (!filter.nil? && filter.blank?)
+      end
+    end
+  end
+end
```
mooktakim, line 13: I'm just not sure about the location of this concern, no reason other than "looks weird" lol

Maybe `app/services/concerns` is better location, that way we can use `include FilterIgnorable`

app/services/statements/query.rb
```diff
@@ -0,0 +1,61 @@
...
+    def statements
+      scope.order(payment_date: :asc)
+    end
+
+    def statement(id: nil, api_id: nil)
+      return scope.find_by!(api_id:) if api_id.present?
+      return scope.find(id) if id.present?
+
+      fail(ArgumentError, "id or api_id needed")
+    end
```
mooktakim, line 27: this might require more discussion, not a huge fan of the multiple id options.
Should we break it into two methods instead?
`def statement_by_api_id`

spec/services/statements/query_spec.rb
mooktakim: 💯

app/services/statements/query.rb
leandroalemao: Yeah agree.. tks @mooktakim I've made the change.. I can't see the need of looking for a statement other than by `api_id` in ECF2. Maybe we had this need in ECF1 as lead provider would search by either id or ecf_id when we flipped ids..

app/services/statements/query.rb
```diff
@@ -0,0 +1,60 @@
...
+      where_updated_since(updated_since)
+      where_state_is(state)
+      where_output_fee_is(output_fee)
+    end
+
+    def statements
+      scope.order(payment_date: :asc)
+    end
+
+    def statement_by_api_id(api_id)
```
mooktakim, line 22: We still should have find by id:
```ruby
def statement(id)
   scope.find(id)
end
```

cwrw (approved):

PR #627, opened 2025-05-22
Add `APIToken` model

ethax-ross:
[Issue #1608](https://github.com/DFE-Digital/register-ects-project-board/issues/1608)

### Context

We want to add a token-based authentication method for API requests. The tokens should be secure/hashed when stored in the database and it should be easy to retrieve a lead provider for a given token.

- Add APIToken model

Add the `APIToken` model that we will use to authenticate lead providers with the API.

Includes a `description` field to explain the purpose of a token if its not clear.


app/models/api_token.rb
```diff
@@ -0,0 +1,10 @@
+class APIToken < ApplicationRecord
+  has_secure_token :token, length: 32
+  encrypts :token, deterministic: true
+
+  belongs_to :lead_provider
+
+  validates :lead_provider, presence: { message: "Lead provider must be specified" }
```
leandroalemao, line 7: can we add the error messages to en.yml maybe?

leandroalemao (commented):
Hi @ethax-ross any seed data to be added as well?

app/models/api_token.rb
ethax-ross: There is no `en.yml` in RECT as it was decided to use inline error messages

leandroalemao: Think it should be ok to create the en.yml file

ethax-ross: We could re-add it, but I mean it was intentionally removed as a team decision before we got here so we'd need to rehash that discussion 🙂  https://github.com/DFE-Digital/register-early-career-teachers-public/commit/0ffc5448c0266e393d461d21c006848e934fe748

cwrw (approved):

app/models/api_token.rb
peteryates: Yeah it's easier for everyone when the messages are alongside the errors.

peteryates: A bonus benefit is that it makes spotting `"is missing"` errors much easier to spot.

leandroalemao (approved):

mooktakim (approved):

PR #629, opened 2025-05-22
Use `allow_value` instead of shoulda matcher `validate_inclusion_of`

### LINKED ISSUE #628: Fix test that's emitting warnings by checking boolean values
There's a warning when running the specs on main at the moment.

Not sure where it's coming from exactly.

```
************************************************************************
Warning from shoulda-matchers:

You are using `validate_inclusion_of` to assert that a boolean column
allows boolean values and disallows non-boolean ones. Be aware that it
is not possible to fully test this, as boolean columns will
automatically convert non-boolean values to boolean ones. Hence, you
should consider removing this test.
************************************************************************
```

We should remove this assertion.

cwrw:
### Context

We are getting a warning from using `validate_inclusion_of` on a boolean value in the model

Change spec to `allow_value` instead and add a test for when the value is `nil` as well 

### Guidance to review
Rerun test `spec/models/statement_spec.rb` and there should be no warning

### DIFF
--- a/spec/models/statement_spec.rb
+++ b/spec/models/statement_spec.rb
@@ -7,7 +7,8 @@ describe Statement do
   describe "validations" do
     subject { FactoryBot.create(:statement) }
 
-    it { is_expected.to validate_inclusion_of(:output_fee).in_array([true, false]).with_message("Output fee must be true or false") }
+    it { is_expected.to allow_values(true, false).for(:output_fee).with_message("Output fee must be true or false") }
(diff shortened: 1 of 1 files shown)


ethax-ross (approved):

mooktakim (approved):

ethax-ross (approved):

cwrw:
close in favour of https://github.com/DFE-Digital/register-early-career-teachers-public/pull/634

PR #631, opened 2025-05-22
[1641] Add search functionality to the ECT and Mentor index pages

avinhurry:
### Context

This PR introduces search functionality to the ECT and Mentors index pages.

### Functionality introduced

search_functionality

### Changes proposed

- Add `TopSearchComponent` to encapsulate reusable search UI

- Utilise the `TopSearchComponent` on the ECT and Mentor index pages

- Refactor the Mentors index action to return a list of teachers who are mentors at the school (via `Teachers::Search`), instead of `MentorAtSchoolPeriod` records

- Render text if search returns no results


app/controllers/schools/ects_controller.rb
```diff
@@ -3,7 +3,8 @@ class ECTsController < SchoolsController
     layout "full"
 
     def index
-      @pagy, @teachers = pagy_array(Teachers::Search.new(ect_at_school: school).search)
+      @teachers = Teachers::Search.new(ect_at_school: school).search
+      @pagy, @filtered_teachers = pagy_array(Teachers::Search.new(ect_at_school: school, query_string: params[:q]).search)
```
peteryates, line 7: I think we should be able to get away with firing one search here, omitting the query string if it's blank.

PR #632, opened 2025-05-23
Add `APIToken` functionality/seeds

ethax-ross:
### Context

We want to use the new `APIToken` model in order to authenticate API requests. We also want to include known tokens for seeded lead providers so that we can easily test in development.

- Add methods to create/find API tokens

Add a method to create a new `APIToken` for a `LeadProvider`. It will default to a randomly generated `token` if not specified and a predefined `description`.

Add method to retrieve an `APIToken` for a given `token`. It will touch the `last_used_at` timestamp when doing so.

- Add APIToken seeds for each LeadProvider

Generate a known token for each lead provider in the seeds.

- Authenticate API controllers

Add `API::TokenAuthenticatable` concern and include it in the `API::BaseController`.

Add `ApiHelper` to easily make authenticated requests in the specs.

Add shared context to easily test authentication for each type of endpoint (get, post, put).

- Add events for creating/revoking access tokens

Add events to `Event::Record` for creating and revoking lead provider access tokens.

- Extract APIToken methods into API::TokenManager service

Pulls the `APIToken` methods for creating, revoking and retrieving a token into a new `API::TokenManager` service.

Integrates the actions into the newly added events.

- Namespace APIToken as API::Token

Unlike other models this is purely for the API and to mimic the `API::TokenManager` namespacing it should live under `API::`

### DIFF
--- a/app/models/api_token.rb
+++ b/app/models/api_token.rb
@@ -9,20 +9,4 @@ class APIToken < ApplicationRecord
   validates :token, uniqueness: { message: "Hashed token must be unique" }
 
   scope :lead_provider_tokens, -> { where.not(lead_provider: nil) }
-
-  class << self
-    def create_lead_provider_api_token!(lead_provider:, token: nil, description: nil)
-      description = "A lead provider token for #{lead_provider.name}" if description.nil?
-
-      create!(
-        lead_provider:,
-        token:,
-        description:
-      )
-    end
-
-    def find_lead_provider_api_token(token:)
-      find_by(token:).tap { |api_token| api_token&.touch(:last_used_at) }
-    end
-  end
 end
--- a/app/models/api_token.rb
+++ b/app/models/api_token.rb
@@ -7,4 +7,22 @@ class APIToken < ApplicationRecord
   validates :lead_provider, presence: { message: "Lead provider must be specified" }
   validates :token, presence: { message: "Hashed token must be specified" }
   validates :token, uniqueness: { message: "Hashed token must be unique" }
+
+  scope :lead_provider_tokens, -> { where.not(lead_provider: nil) }
+
+  class << self
+    def create_lead_provider_api_token!(lead_provider:, token: nil, description: nil)
+      description = "A lead provider token for #{lead_provider.name}" if description.nil?
+
+      create!(
+        lead_provider:,
+        token:,
+        description:
+      )
+    end
+
+    def find_lead_provider_api_token(token:)
+      find_by(token:).tap { |api_token| api_token&.touch(:last_used_at) }
+    end
+  end
 end
--- a/app/services/events/record.rb
+++ b/app/services/events/record.rb
@@ -307,6 +307,26 @@ module Events
       new(event_type:, author:, appropriate_body: batch.appropriate_body, heading:, happened_at: Time.zone.now, metadata:).record_event!
     end
 
+    # API Token Events
+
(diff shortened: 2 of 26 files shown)


spec/support/requests/api_helper.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  def authenticated_api_put(path, token: nil)
+    put path, headers: api_headers(token:)
+  end
+
+  def api_headers(token: nil)
+    {
+      Authorization: "Bearer #{token || generate_api_token.token}"
+    }
+  end
```
mooktakim, line 18: I was planning on adding the `rspec-default_http_header` gem for headers like npq/ecf1.
What do you think?

app/models/api_token.rb
```diff
@@ -7,4 +7,20 @@ class APIToken < ApplicationRecord
...
+      description = "A lead provider token for #{lead_provider.name}" if description.nil?
+
+      create!(
+        lead_provider:,
+        token:,
+        description:
+      )
+    end
+
+    def find_lead_provider_api_token(token:)
```
cwrw, line 22: do we want to limit this by lead provider as well?

spec/support/requests/api_helper.rb
cwrw: maybe we hold off for now, seems to be cleanly done?

ethax-ross: Yeah I'm not sure its worth it for the use case we have at the moment. Happy to revisit if this gets out of hand though

app/models/api_token.rb
ethax-ross: I was in two minds about this, it felt a little pointless given we only support lead providers at the moment but maybe its worth it to avoid an inadvertent issue down the line if/when we add another foreign key 🤔 I'll add it in.

spec/support/requests/api_helper.rb
mooktakim: Sounds good, will do without gem 👍

app/services/events/record.rb
peteryates, line 1: :ok_hand:

app/services/events/record.rb
```diff
@@ -307,6 +307,26 @@ def self.record_bulk_upload_completed_event!(author:, batch:)
       new(event_type:, author:, appropriate_body: batch.appropriate_body, heading:, happened_at: Time.zone.now, metadata:).record_event!
     end
 
+    # API Token Events
+
+    def self.record_lead_provider_api_token_created_event!(author:, api_token:)
+      event_type = :lead_provider_api_token_created
```
cwrw, line 313: testing this locally we need to add:
```
lead_provider_api_token_created
lead_provider_api_token_revoked
```
to event types here: https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/models/event.rb#L2

cwrw: also worth making sure we test this

leandroalemao (approved):
Hi @ethax-ross Looks good 👍 tkss

app/services/events/record.rb
ethax-ross: Fixed in https://github.com/DFE-Digital/register-early-career-teachers-public/pull/632/commits/70d7f8d33333657227b7bc5fbd044eacd82f83fa

cwrw (approved):

PR #635, opened 2025-05-23
Allow fake TRS API client use by the app

peteryates:
### Context

We run into problems when testing because we have lots of environments talking to TRS' pre-prod API, including

* developer machines
* review apps (often 10-20)
* staging
* sandbox

Each of these apps has its own set of teachers and statuses, and they expect the state to match whatever's in TRS. This obviously doesn't work.

The aim here is to reduce the number of applications talking unnecessarily to TRS.

We already had a fake TRS API client that we use for testing. Moving it from `/spec` to `/lib` isn't ideal but it does remove our reliance on data that's not under our control.

It's not feature complete but it does the basics and works for registration. The name of the returned ECT will be randomly selected from a hardcoded list so we don't need to worry about Faker.

### Guidance to review

Feedback welcome, this needs more work before it's ready.


lib/trs/api_client.rb
```diff
@@ -11,6 +13,18 @@ def initialize
...
 
+    def self.build
+      if Rails.application.config.enable_fake_trs_api
+        Rails.logger.warn("Using TRS::FakeAPIClient")
+        require Rails.root.join('spec/support/api/trs/fake_api_client')
+
+        return TRS::FakeAPIClient.new(random_names: true)
+      end
+
+      Rails.logger.warn("Using real TRS::APIClient")
```
peteryates: I'll remove these logs, useful when testing.

lib/trs/api_client.rb
peteryates, line 1: If there's a nicer way to do this do let me know. There must be a better word than `.build` too.

PR #639, opened 2025-05-23
[1655] Rate limiting for /api/ with rack attack

mooktakim:
### Context

Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/1655

* Updated rack attack config to throttle `/api/` requests, 1000 requests per 5 minutes
* Added fall back throttle for 1000 requests per 5 minutes
* Added `remote_ip` changes from https://github.com/DFE-Digital/early-careers-framework/pull/4085
* Added rspec tests

### DIFF
--- a/config/initializers/rack_attack.rb
+++ b/config/initializers/rack_attack.rb
@@ -1,4 +1,19 @@
 class Rack::Attack
+  class Request < ::Rack::Request
+    def remote_ip
+      # This will always be present because the ActionDispatch::RemoteIp
+      # middleware runs long before this middleware. Use #fetch here
+      # so that we bail quickly if that middleware goes away or changes
+      # the field name. Have preferred this to instantiating a whole
+      # ActionDispatch::Request as that's a whole lot of work and this happens
+      # on every request.
+      #
+      # Favour the X-Real-IP header if set by Azure, if not fallback to
+      # ActionDispatch#remote_ip which is more reliable than Rack's ip method.
+      @remote_ip ||= env.fetch("x-real-ip", nil).presence || env.fetch("action_dispatch.remote_ip").to_s
+    end
+  end
+
   PROTECTED_ROUTES = %w[
     /otp-sign-in
     /otp-sign-in/code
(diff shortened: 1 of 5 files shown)


config/initializers/rack_attack.rb
```diff
@@ -12,9 +12,22 @@ def self.protected_path?(request)
...
+    request.get_header("HTTP_AUTHORIZATION")
+  end
+
   throttle("protected routes (OTP)", limit: 5, period: 20.seconds) do |request|
     request.ip if protected_path?(request)
   end
+
+  # Throttle /api requests by auth token (1000 requests per 5 minutes)
+  throttle("API requests by auth token", limit: 1000, period: 5.minutes) do |request|
+    auth_token(request) if api_request?(request) && !protected_path?(request)
```
peteryates, line 29: Do we need the `&& !protected_path?(request)`?

Not really sure what it's guarding against.

config/initializers/rack_attack.rb
```diff
@@ -12,9 +12,22 @@ def self.protected_path?(request)
     PROTECTED_ROUTES.any? { |route| request.path.start_with?(route) }
   end
 
+  def self.api_request?(request)
+    request.path.starts_with?("/api/")
+  end
+
+  def self.auth_token(request)
+    request.get_header("HTTP_AUTHORIZATION")
```
leandroalemao, line 20: Do we need to fallback to the ip in case someone tries to hit the api without sending a token?

config/initializers/rack_attack.rb
mooktakim: We did something similar with NPQ, we probably don't need this for rect, i'll remove it.

config/initializers/rack_attack.rb
mooktakim: Added fallback throttle to limit to 1000 reqs within 5min

config/initializers/rack_attack.rb
peteryates: Ah yeah I see now on NPQ - having different rate limit rules for `/api/guidance`.

spec/requests/rate_limiting_spec.rb
```diff
@@ -34,6 +34,51 @@
     end
   end
 
-  xit 'POST /otp-sign-in'
-  xit 'POST /otp-sign-in/verify'
+  context "rate limit /api/ endpoints by auth token" do
+    it "throttles over 1000 requests within 5 minutes" do
+      freeze_time do
+        1000.times do
```
ethax-ross, line 40: I think when the statements endpoint actually does something that hits the DB these test will be pretty slow. In NPQ reg we mock out the limits to get around that, is it worth doing something similar here?

mooktakim: Updated the tests to what we have for NPQ

cwrw (approved):

PR #642, opened 2025-05-27
Add edit and delete extensions for admins

ghost:
## Description

Currently, AB users can add and edit extensions, but support users can't.

Added functionality to allow admin users to edit and delete extensions. Reused existing update extension logic from AB and added delete logic just for admin users.

## User story

**As** an admin user
**I need** to be able to administer extension data
**So that** I can solve more support issues myself without needing a dev to be involved

## Prototype, screenshots & content

Screenshots of relevant screens:

Image

### DIFF
--- a/app/components/teachers/details/admin_induction_summary_component.html.erb
+++ b/app/components/teachers/details/admin_induction_summary_component.html.erb
@@ -25,7 +25,7 @@
       <%= induction_extensions.formatted_number_of_terms %>
     <% end %>
 
-    <%= row.with_action(text: extensions_action_text, visually_hidden_text: "extensions", href: extensions_path, classes: 'govuk-link--no-visited-state') %>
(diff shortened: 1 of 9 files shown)


app/components/teachers/details/induction_summary_component.rb
```diff
@@ -17,14 +17,18 @@ def extensions_action_text
...
       is_admin == true
     end
 
+    def extensions_path
+      if is_admin
+        helpers.admin_teacher_extensions_path(teacher)
+      else
+        helpers.ab_teacher_extensions_path(teacher)
+      end
+    end
```
peteryates, line 30: I think we should split these out into separate components now to avoid this kind of complexity within.

peteryates (approved):
Looks good :+1:

PR #643, opened 2025-05-27
[1640] - Update ECT list to show AB, LP and DP details

ltello:
### Context

[Issue](https://github.com/DFE-Digital/register-ects-project-board/issues/1640)

This story is to update the ECT list page to show each ECT's AB, LP and DP details, based on feedback that schools users provided earlier on in the journey.

# BEFORE
image

# AFTER
image

### DIFF
--- a/app/views/schools/ects/index.html.erb
+++ b/app/views/schools/ects/index.html.erb
@@ -6,54 +6,27 @@
     caption_size: 'l'
   )
 %>
-<<<<<<< HEAD
-<%= govuk_button_link_to(@number_of_teachers.zero? ? "Register an ECT starting at your school" : "Add an ECT", schools_register_ect_wizard_start_path) %>
-<hr class="govuk-section-break--m" />
-<div class="govuk-grid-row">
-  <% if @number_of_teachers.zero? %>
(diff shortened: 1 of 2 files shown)


ltello:
You are right @claire-hughez. 
I know about this issue.
The thing is that (because we were waiting until a final decision on school partnerships and all the training data model were made) we store the provider type and lead provider chosen by the school in the ECTAtSchoolPeriod created instead of a  TrainingPeriod.
The code I have written for this ticket is to pick that information from their future right place (the training record and their ActiveLeadProvider connected).
Until we create TrainingPeriods (as a result of a teacher registration) in a new ticket (that surely we need to write), the data is displayed as blank.
I will speak with Emily M. about it.

claire-hughez:
@ltello If it points at the future right location, that's fine with me, but I imagine we should write some tickets to make sure to go back to this and make those changes to where LP is saved in the data model (if Emily hasn't already given how on it she is).

mason-emily:
I'll add a ticket to make sure come back to this!

leoapost (approved):

ltello:
...merging now. 
Thanks all!

PR #645, opened 2025-05-27
Flesh out the statements show/index endpoints

ethax-ross:
### Context

We want to pull together the other changes for serializers/filtering/querying statements and flesh out the `show` and `index` actions of the `StatementsController`.

- Update rescuable errors

Remove unused filter validation error; as this is the same as the `UnpermittedParameters` exception we can just raise it as such to avoid a duplicate error handler.

Add `RecordNotFound` error handler; we will need this for show endpoints as the underlying query will raise when not found/matched on ID.

- Enable overriding the lead provider in request specs

We need to be able to specify a particular lead provider when making a request to the API in request specs. Use `active_lead_provider` by default, if defined.

Expose `params` in helper methods so we can send filters/a request body.

- Rename registration_period_start_years to registration_period_years

Rename the attribute to be inline with the model naming.

- Remove error rescuable specs

These tests will be covered by the other request specs that test filtering/creating/updating.

- Add concerns to support filtering by updated_since/cohort

Add concerns for filtering by `updated_since` and `cohort`.

Add request spec shared contexts for covering endpoints.

- Implement StatementsController actions

Hook up the `show` and `index` controller actions for statements to the existing query, serializer and filtering concerns.

Update the route to accept the `api_id` as the `id` attribute.

Add generic tests to cover these types of endpoints.

- Fix includes on `has_one through:`

Rails can deal with `includes` on a `has_one though:` relationship on simple queries but it seems to struggle when used in more complex queries like the `Statements::Query`.

Update `includes` to use the direct relationships.

### Guidance to review

Pulls in changes from #655


spec/support/shared_contexts/api/filterable_endpoint.rb
```diff
@@ -0,0 +1,93 @@
...
+    travel_to(1.year.ago) { create_resource(active_lead_provider:) }
+
+    authenticated_api_get(path, params: { filter: { updated_since: 2.months.ago.iso8601 } })
+
+    expect(response).to have_http_status(:ok)
+    expect(response.content_type).to eql("application/json; charset=utf-8")
+    expect(response.body).to eq(serializer.render(apply_expected_order([resource_updated_one_week_ago, resource_updated_one_month_ago]), root: "data"))
+  end
+
+  it "calls the correct query" do
```
cwrw, line 64: same here

app/controllers/api/v3/statements_controller.rb
```diff
@@ -1,12 +1,40 @@
 module API
   module V3
     class StatementsController < BaseController
+      include DateFilterable
+      include RegistrationPeriodFilterable
+
       def index
-        # Temp to demonstrate pagination.
-        render json: { data: paginate(Statement.all) }.to_json
+        render json: to_json(paginate(statements_query.statements))
```
cwrw, line 8: we might need to have a look at the serialiser:
```
09:50:33 web.1  |   ↳ app/serializers/statement_serializer.rb:7:in `block in <class:AttributesSerializer>'
09:50:33 web.1  |   CACHE ActiveLeadProvider Load (0.0ms)  SELECT "active_lead_providers".* FROM "active_lead_providers" WHERE "active_lead_providers"."id" = $1 LIMIT $2  [["id", 232], ["LIMIT", 1]]
```
Although its cached and could be ok not an n+1, worth revisiting incase

app/services/statements/query.rb
```diff
@@ -5,11 +5,11 @@ class Query
 
     attr_reader :scope
 
-    def initialize(lead_provider: :ignore, registration_period_start_years: :ignore, updated_since: :ignore, state: :ignore, output_fee: true)
+    def initialize(lead_provider: :ignore, registration_period_years: :ignore, updated_since: :ignore, state: :ignore, output_fee: true)
       @scope = Statement.distinct.includes(:lead_provider, :registration_period)
```
leandroalemao, line 9: `@scope = Statement.distinct.includes(active_lead_provider: %i[lead_provider registration_period])`

cwrw (commented):
looks great, just some minor comments

spec/support/shared_contexts/api/filterable_endpoint.rb
```diff
@@ -0,0 +1,93 @@
...
+
+    expect(response).to have_http_status(:ok)
+    expect(response.content_type).to eql("application/json; charset=utf-8")
+    expect(response.body).to eq(serializer.render(apply_expected_order([previous_registration_period_resource, next_registration_period_resource]), root: "data"))
+  end
+
+  it "calls the correct query" do
+    registration_period_years = "2023,2024"
+
+    expect(query).to receive(:new).with(a_hash_including(
```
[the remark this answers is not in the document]
ethax-ross: Yeah true, I thought these were cleaner to test with mocking as we have the end to end test above, but I'll make them all consistent 👍

cwrw (approved):
🙌

PR #654, opened 2025-05-28
Update navigation links for 'ECTs' and 'mentors' for schools to match prototype

claire-hughez:
I noticed these were different from the prototype (and also the capitalisation was driving me bonkers).

Fix the name of links in the main school nav.

### Guidance to review
Make sure I didn't break any tests or devvy things please!

### DIFF
--- a/app/components/navigation/primary_navigation_component.rb
+++ b/app/components/navigation/primary_navigation_component.rb
@@ -43,8 +43,8 @@ module Navigation
           { text: "Organisations", href: admin_organisations_path, active_when: '/admin/organisations' },
         ],
         school_user: [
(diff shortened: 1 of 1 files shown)


leoapost (changes requested):
It looks like a test is still expecting to find “Your ECTs”. Updating the text on line 52 of the [primary_navigation_component_spec.rb](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/fix-navigation-links-for-schools/spec/components/navigation/primary_navigation_component_spec.rb#L52) file should resolve it.

leoapost (approved):
I pushed a fix for the failing spec.

The PR looks good now.

PR #664, opened 2025-05-30
Add admin finance page and statements list

peteryates:
### Context

This is a rough draft containing statements pages.

The design will be iterated but for now they just list the statements as per the prototype and add a finance page that has a link to the statements index.

It's intended to be a starting off point and to allow us to review data in the app.

### Differences to the prototype

* no filters - we need to decide how they're going to work, hopefully avoiding the current broken state where you can select incompatible years/statements and get no results
* no secondary horizontal nav - this is in the prototype but for now I've reused the extra nav page style from organisations. It keeps the hierarchy in tact and doesn't require new code/styles.

- **Fix the organisations layout**
- **Add routes for finance and statements**
- **Add boolean to yes or no helper**
- **Add statement presenter and associated tests**
- **Add finance/statements controllers, views and specs**

### Preview

| Finance page | Statements list | Statement view |
| ---- |---- | ---- |
| image | Screenshot From 2025-05-30 14-42-57 | image |

Fixes DFE-Digital/register-ects-project-board#1697

### DIFF
--- a/app/presenters/admin/statement_presenter.rb
+++ b/app/presenters/admin/statement_presenter.rb
@@ -1,0 +1,47 @@
+module Admin
+  class StatementPresenter < SimpleDelegator
+    def self.wrap(collection)
+      collection.map { |teacher| new(teacher) }
+    end
+
+    def month_and_year
+      month_name = Date::MONTHNAMES.fetch(statement.month)
+
+      "#{month_name} #{statement.year}"
+    end
+
+    def status_tag_kwargs
+      colour = { 'open' => 'blue', 'payable' => 'yellow', 'paid' => 'green' }.fetch(statement.state)
+
+      { text: statement.state.capitalize, colour: }
+    end
+
+    def page_title
+      lead_provider_name = statement.active_lead_provider.lead_provider.name
+
+      "#{lead_provider_name} - #{month_and_year}"
+    end
+
+    def registration_period_year
+      statement.active_lead_provider.registration_period.year.to_s
+    end
+
+    def lead_provider_name
+      statement.active_lead_provider.lead_provider.name
+    end
+
+    def formatted_deadline_date
+      statement.deadline_date.to_fs(:govuk)
+    end
+
+    def formatted_payment_date
+      statement.payment_date.to_fs(:govuk)
+    end
+
+  private
+
+    def statement
+      __getobj__
+    end
+  end
+end
--- a/app/presenters/admin/statement_presenter.rb
+++ b/app/presenters/admin/statement_presenter.rb
@@ -1,7 +1,7 @@
 module Admin
   class StatementPresenter < SimpleDelegator
     def self.wrap(collection)
-      collection.map { |teacher| new(teacher) }
+      collection.map { |statement| new(statement) }
     end
 
     def month_and_year
--- a/app/views/admin/finance/statements/index.html.erb
+++ b/app/views/admin/finance/statements/index.html.erb
@@ -8,7 +8,7 @@
       head: ["ID", "Lead provider", "Contract year", "Statement", "Status"],
       rows: @statements.map { |s| [
         govuk_link_to(s.id, admin_finance_statement_path(s)),
-        govuk_link_to(s.lead_provider_name, '#', text_colour: true),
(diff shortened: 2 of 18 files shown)


app/presenters/admin/statement_presenter.rb
```diff
@@ -0,0 +1,47 @@
+module Admin
+  class StatementPresenter < SimpleDelegator
+    def self.wrap(collection)
+      collection.map { |teacher| new(teacher) }
```
cwrw, line 4:
```suggestion
      collection.map { |statement| new(statement) }
```

app/presenters/admin/statement_presenter.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    def registration_period_year
+      statement.active_lead_provider.registration_period.year.to_s
+    end
+
+    def lead_provider_name
+      statement.active_lead_provider.lead_provider.name
+    end
+
+    def formatted_deadline_date
```
cwrw, line 33: is it worth adding tests for the date methods?

app/views/admin/finance/statements/show.html.erb
```diff
@@ -0,0 +1,41 @@
...
+      row.with_key(text: "Lead provider")
+      row.with_value(text: govuk_link_to(@statement.lead_provider_name, '#', text_colour: true))
+    end
+
+    sl.with_row do |row|
+      row.with_key(text: "Status")
+      row.with_value(text: govuk_tag(**@statement.status_tag_kwargs))
+    end
+
+    sl.with_row do |row|
```
cwrw, line 15: could we add the contract year here as well?

spec/views/admin/finance/show.html.erb_spec.rb
```diff
@@ -0,0 +1,7 @@
+RSpec.describe 'admin/finance/show.html.erb' do
```
cwrw, line 1: maybe we can add a feature spec as well? curious around the balance your keeping between view specs vs feature specs

cwrw (commented):
looks great just some minor comments

app/views/admin/finance/statements/show.html.erb
peteryates: Hmm yeah - just with a placeholder value you mean?

spec/views/admin/finance/show.html.erb_spec.rb
peteryates: > curious around the balance your keeping between view specs vs feature specs

Good question. Usually the features test that a process that needs multiple pages works and a view spec just ensures the right thing's on the page in a particular state. But I don't think there's a hard and fast rule, sometimes a feature works better as it also verifies the routing etc (or JS etc).

app/views/admin/finance/statements/show.html.erb
cwrw: sorry meant `registration_period_year`

app/views/admin/finance/statements/index.html.erb
```diff
@@ -0,0 +1,20 @@
...
+
+<% if @statements.none? %>
+  <p class='govuk-body'>There are no statements</p>
+<% else %>
+  <%=
+    govuk_table(
+      head: ["ID", "Lead provider", "Contract year", "Statement", "Status"],
+      rows: @statements.map { |s| [
+        govuk_link_to(s.id, admin_finance_statement_path(s)),
+        govuk_link_to(s.lead_provider_name, '#', text_colour: true),
```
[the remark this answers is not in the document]
peteryates: Was awaiting the lead providers page to be added so we can link them together, but I'll drop for now.

app/views/admin/finance/statements/index.html.erb
```diff
@@ -0,0 +1,20 @@
+<% page_data(title: "Statements") %>
+
+<% if @statements.none? %>
+  <p class='govuk-body'>There are no statements</p>
+<% else %>
+  <%=
+    govuk_table(
```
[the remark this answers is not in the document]
peteryates: What should be in the header? Not sure how a headerless column would be from an accessibility POV.

PR #667, opened 2025-06-02
Add specs for the registration period migrators

leoapost:
### Context

Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/1669

This PR primarily adds specs for the registration period migrator.

It also updates the migrator to include the `started_on` and `finished_on` values when creating RegistrationPeriod records. The started_on is set to the cohort’s registration_start_date, and the finished_on is calculated as one year later, minus one day.

- Add factory to create ecf cohorts
- Add specs for the registration period migrator
- Update the migrator to include the missing started_on and finished_on fields

### DIFF
--- a/spec/migration/migrators/registration_period_spec.rb
+++ b/spec/migration/migrators/registration_period_spec.rb
@@ -1,0 +1,57 @@
+RSpec.describe Migrators::RegistrationPeriod do
+  describe '.record_count' do
+    it 'returns the count of cohorts' do
+      FactoryBot.create_list(:migration_cohort, 2, :with_sequential_start_year)
+      expect(described_class.record_count).to eq(2)
+    end
+  end
+
+  describe '.model' do
+    it 'returns :registration_period' do
+      expect(described_class.model).to eq(:registration_period)
+    end
+  end
+
+  describe '.cohorts' do
+    it 'returns all cohorts' do
+      cohort = FactoryBot.create(:migration_cohort)
+      expect(described_class.cohorts).to include(cohort)
+    end
+  end
+
+  describe '.reset!' do
+    before { allow(::RegistrationPeriod).to receive_message_chain(:connection, :execute) }
+
+    context 'when migration testing is enabled' do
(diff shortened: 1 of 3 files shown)


leoapost:
The specs revealed a bug in the migrator, so I've pushed a commit to fix it.

spec/migration/migrators/registration_period_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+  end
+
+  describe '.reset!' do
+    before do
+      FactoryBot.create(:registration_period)
+      allow(Rails.application.config).to receive(:enable_migration_testing).and_return(enabled_migration_testing)
+    end
+
+    context 'when migration testing is enabled' do
+      let(:enabled_migration_testing) { true }
```
peteryates, line 29: Might be a tiny bit clearer here to mirror the setting in the variable name and call it `enable_migration_testing`.

spec/migration/migrators/registration_period_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    let!(:cohort1) { FactoryBot.create(:migration_cohort, start_year: 2021) }
+    let!(:cohort2) { FactoryBot.create(:migration_cohort, start_year: 2022) }
+    let!(:data_migration) { FactoryBot.create(:data_migration, model: :registration_period) }
+
+    it 'creates registration periods with cohort start years as IDs' do
+      expect {
+        described_class.new(worker: 0).migrate!
+      }.to change(RegistrationPeriod, :count).by(2)
+      expect(RegistrationPeriod.pluck(:year)).to contain_exactly(2021, 2022)
+    end
```
peteryates, line 55: Should we be checking the start and finish dates are set here too?

spec/migration/migrators/registration_period_spec.rb
leoapost: This was meant to be `enable_migration_testing`. Cheers!

peteryates (approved):

PR #676, opened 2025-06-03
1680 bugfix: User cannot upload a file after uploading a failing one

ghost:
Resolved the issue of the service not recognising the upload of a new file after uploading one with an error.

Also updated links to overview page to point to the bulk upload index pages


app/controllers/appropriate_bodies/process_batch/actions_controller.rb
```diff
@@ -21,7 +21,9 @@ def create
 
           redirect_to ab_batch_action_path(@pending_induction_submission_batch)
         else
-          @pending_induction_submission_batch = csv_data
+          csv_data.errors.each do |error|
```
leandroalemao, line 24: How about creating a service class with these actions so we avoid fat controllers?

leandroalemao (approved):

peteryates (approved):
Looks good to me - agree with @leandroalemao's comment that some of the logic could be moved out to its own class though.

PR #677, opened 2025-06-03
Enable monitoring in staging and sandbox

peteryates:
We currently don't receive alerts when staging or sandbox go down. This change enables monitoring and sets the healthcheck endpoint and statuscake contact group up so those in the 'CPD' group will be notified if one of these environments goes down.


cwrw (approved):

peteryates:
@cwrw - fyi force pushed without the `statuscake_contact_groups`

RMcVelia (approved):

PR #682, opened 2025-06-04
[1688] Add task to set up API tokens

leandroalemao:
### Context

[Issue #1688](https://github.com/DFE-Digital/register-ects-project-board/issues/1688)

We need to setup API tokens for providers to test the API in sandbox.

- Create a task to generate API tokens for providers;

### Guidance to review

- Generates tokens for all lead providers: `rake 'lead_provider:generate_api_token'`
- Generates token for a specific lead provider by name: `rake 'lead_provider:generate_api_token[Ambitious Institute]'`
- Generates token for a specific lead provider by id: `rake 'lead_provider:generate_api_token[1]'`

### DIFF
--- a/lib/tasks/api_token.rake
+++ b/lib/tasks/api_token.rake
@@ -4,14 +4,11 @@ namespace :api_token do
     task :generate_token, %i[lead_provider_name_or_id] => :environment do |_t, args|
       logger = Rails.env.test? ? Rails.logger : Logger.new($stdout)
 
-      lead_provider_name_or_id = args.lead_provider_name_or_id
+      API::AddAPITokenToLeadProvider.new(
+        lead_provider_name_or_id: args.lead_provider_name_or_id,
+        logger:
+      ).add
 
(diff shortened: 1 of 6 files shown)


lib/tasks/api_token.rake
```diff
@@ -0,0 +1,18 @@
+namespace :api_token do
+  namespace :lead_provider do
+    desc "Generate a new API token for the Lead Providers API"
+    task :generate_token, %i[lead_provider_name_or_id] => :environment do |_t, args|
+      logger = Rails.env.test? ? Rails.logger : Logger.new($stdout)
```
peteryates, line 5: Any reason not to just use the `Rails.logger` here regardless of env?

peteryates (approved):
Looks good to me.

lib/tasks/api_token.rake
```diff
@@ -0,0 +1,18 @@
...
+
+      lead_provider_name_or_id = args.lead_provider_name_or_id
```
cwrw, line 7: althought pretty small maybe we can move this into a service, test the service and the rake task can accept a lead provider or loop through all lead providers otherwise?

cwrw (approved):

PR #685, opened 2025-06-04
Add seed data to the sandbox environment

ethax-ross:
### Context

We want to be able to generate seed data in sandbox.

The seed data initially will be registration periods, lead providers and statements. The statements data should be close to what we have in production.

We want the data to remain consistent if we run the seeds multiple times (we should end up with the same registration periods, lead providers and statements).

- Add seed data to the sandbox environment

Add sandbox seed data services. A `Base` service is used to centralise logging and environment checks.

Add a rake task for convenience; `bundle exec rake sandbox_seed_data:generate`.

### Guidance to review

Screenshot 2025-06-05 at 11 20 10

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -67,6 +67,16 @@ Style/HashSyntax:
 Style/Lambda:
   EnforcedStyle: literal
 
+Rails/UnknownEnv:
+  Environments:
+    - production
+    - staging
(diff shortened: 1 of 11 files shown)


leandroalemao (approved):

cwrw (approved):
🙌

PR #689, opened 2025-06-05
Change content from Programme type to Training programme

leoapost:
Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/1491

This PR renames the 'programme type' to 'training programme' throughout the service.


app/migration/mappers/training_training_programme_mapper.rb
```diff
@@ -1,5 +1,5 @@
 module Mappers
-  class TrainingProgrammeTypeMapper
+  class TrainingTrainingProgrammeMapper
```
peteryates, line 2:
```suggestion
  class TrainingProgrammeMapper
```

leoapost: Good catch. It seems this happened in a few more files, so I pushed a commit to fix them all.

leoapost:
I'm closing and re-opening the PR to reset the Review app resources.

app/controllers/appropriate_bodies/teachers/initial_teacher_training_records_controller.rb
peteryates, line 1: Even though this is placeholder I think we should leave it as is because this data will be retrieved from the TRS API.

```json
  "initialTeacherTraining": [
    {
      "provider": {
        "name": "string",
        "ukprn": "string"
      },
      "qualification": {
        "name": "string"
      },
      "startDate": "2025-06-13",
      "endDate": "2025-06-13",
      "programmeType": "Apprenticeship",
      "programmeTypeDescription": "string",
      "result": "Pass",
      "ageRange": {
        "description": "string"
      },
      "subjects": [
        {
          "code": "string",
          "name": "string"
        }
      ]
    }
  ],
```

PR #692, opened 2025-06-05
Use TRS API version 20250203 instead of Next in Review

peteryates:
We had this set to Next for API testing early on when releasing RIAB but now there's something causing it to crash, so for now push this back to the stable verison.

Fixes DFE-Digital/register-ects-project-board#1723


ltello (approved):
This one was easy!

PR #693, opened 2025-06-06
Fix flakey statements sandbox seed spec

ethax-ross:
When generating seed data for statements in sandbox we only get a `payable` statement if the random payment date includes the current date.

This has a chance of happening for a number of statements in the seeds, but isn't guaranteed to and so it causes the test that ensures we generate statements with all states to be flakey.

Instead, we adjust the window for statements with dates near the current date to ensure they will always end up as `payable`.


peteryates (approved):
:shipit:

PR #694, opened 2025-06-06
Fix bug in `#page_data` caused by early return

peteryates:
There was a bug in #page_data where if header was false it would return early, before the caption, backlinks, breadcrumbs etc had been dealt with.

Originally the method was much simpler but extra functionality has been added and now the return doesn't make sense - sometimes we might want a page with a manually-set header to have a back link for example.

Now, the default value for header is `:use_title` instead of `nil`.

This means we don't need to worry about differentiating between nil and false and can be more explicit.

I've also rewritten the tests to cover all of the functionality properly and no longer return a hash of values (which originally were there to make testing easier) in favour of having the tests pull data out of content_for to ensure it's set correctly.

### Guidance to review

The main risk here is that we end up rendering two headers, one manually and one from `#page_data`.


app/helpers/application_helper.rb
```diff
@@ -1,26 +1,20 @@
 module ApplicationHelper
   include Pagy::Frontend
 
-  def page_data(title:, header: nil, header_size: "l", error: false, backlink_href: nil, caption: nil, caption_size: 'm', header_classes: [])
+  def page_data(title:, header: :use_title, header_size: "l", error: false, backlink_href: nil, caption: nil, caption_size: 'm', header_classes: [])
     page_title = title_with_error_prefix(title, error:)
-
     content_for(:page_title) { page_title }
 
-    return { page_title: } if header == false
```
peteryates, line 9: This meant when `header: false` nothing below was set.

ltello (approved):

PR #697, opened 2025-06-06
Make the email page more accessible

peteryates:
On the ECT registration flow the label wasn't really describing the question, it was more of a hint.

I've made the page heading the label and set the old label text as the hint.

With this being fixed we can also change the features to use the label text when selecting the input.


tonyheadford (approved):
I think it makes sense, only doubt in my mind is whether the text should it be an instruction rather than a hint?

peteryates:
Yeah - good point. Perhaps it should go back to being a regular `<p>`.

It feels a bit warningy but using warning text would clutter the page.

PR #704, opened 2025-06-09
Add lead provider delivery partnerships

peteryates:
This is part 4 of #507.

It aims to complete complete the lead provider/delivery partner/registration period section of the schema.

| Before | After |
| ----- | ------ |
| image | image |

## Notes

There was more stuff to change than expected, I didn't realise we'd actually plumbed in the showing of selected delivery partner, lead provider and registration period before the model was complete. As a result, there are some additional methods added to make the tests pass without this spiralling into a massive PR.

## Remaining tasks

* [x] Fix the seeding process by removing the temp commit and adjusting the seeds accordingly
* [ ] Rework `ECTAtSchoolPeriods::Training` so it knows about the new schema and what to delegate where

## Review notes

Go commit by commit.

## Release notes

This will need some handholding to deploy on staging/sandbox, we'll want to flush any existing data first because of the `not null` constraints in `available_provider_pairings` and `school_partnerships`.


app/models/school_partnership.rb
```diff
@@ -1,24 +1,8 @@
 class SchoolPartnership < ApplicationRecord
   # Associations
-  belongs_to :registration_period, inverse_of: :school_partnerships
-  belongs_to :lead_provider, inverse_of: :school_partnerships
-  belongs_to :delivery_partner, inverse_of: :school_partnerships
+  belongs_to :available_provider_pairing, inverse_of: :school_partnerships
   has_many :events
```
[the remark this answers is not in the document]
peteryates: Yeah that'll follow in a separate PR

app/models/training_period.rb
```diff
@@ -53,6 +50,15 @@ def siblings
     trainee.training_periods.excluding(self)
   end
 
+  # FIXME: temporary
+  def lead_provider
+    school_partnership&.available_provider_pairing&.active_lead_provider&.lead_provider
```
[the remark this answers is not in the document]
peteryates: Yeah entirely agree - this was to get the tests passing. I think we can make encapsulate this logic in a service class when we get round to actually doing the EOI route.

peteryates:
Added an 'embolden' helper method in the seeds. It makes text bold without needing to specify a colour. To make it work I had to adjust the `Colourize.text` method to take keyword args so the colour could be omitted.

Screenshot From 2025-06-10 19-01-38

db/seeds/available_provider_pairings.rb
```diff
@@ -0,0 +1,39 @@
...
+teach_fast = LeadProvider.find_by!(name: 'Teach Fast')
+better_practice_network = LeadProvider.find_by!(name: 'Better Practice Network')
+
+active_lead_providers = ActiveLeadProvider
+  .eager_load(:registration_period, :lead_provider)
+  .index_by { |alp| [alp.lead_provider, alp.registration_period.year] }
+
+ambitious_institute_2021 = active_lead_providers.fetch([ambitious_institute, 2022])
+ambitious_institute_2022 = active_lead_providers.fetch([ambitious_institute, 2023])
+ambitious_institute_2023 = active_lead_providers.fetch([ambitious_institute, 2024])
```
ltello, line 16: Shouldn't these names be ending in *_2022, *_2023 and *_2024 instead?

db/migrate/20250605101802_create_available_provider_pairings.rb
```diff
@@ -0,0 +1,10 @@
+class CreateAvailableProviderPairings < ActiveRecord::Migration[8.0]
```
cwrw, line 1: can we use the name `LeadProviderDeliveryPartnership`, we landed on this in the end as it was clearer in meaning https://github.com/DFE-Digital/register-early-career-teachers-public/pull/498/files#diff-dc044e66140dddc0daf216e146ff14a4a49cb4eacc9329dc242a3fc09548cf98R1

db/seeds/available_provider_pairings.rb
peteryates: Ah yeah good point. Will fix.

db/migrate/20250605101802_create_available_provider_pairings.rb
peteryates: Oops I didn't update my diagram. Yeah I'll swap it over.

app/models/school_partnership.rb
```diff
@@ -1,24 +1,8 @@
...
-            presence: true
-
-  validates :delivery_partner_id,
-            presence: true
-
-  # Scopes
-  scope :for_registration_period, ->(year) { where(registration_period_id: year) }
-  scope :for_lead_provider, ->(lead_provider_id) { where(lead_provider_id:) }
-  scope :for_delivery_partner, ->(delivery_partner_id) { where(delivery_partner_id:) }
+  validates :lead_provider_delivery_partnership_id, presence: true
```
cwrw, line 7: are we holding off on the validation for school here till the next pr?

spec/models/lead_provider_delivery_partnership_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+    subject { FactoryBot.create(:lead_provider_delivery_partnership) }
+
+    it { is_expected.to validate_presence_of(:active_lead_provider_id).with_message('Select an active lead provider') }
+    it { is_expected.to validate_presence_of(:delivery_partner_id).with_message('Select a delivery partner') }
+    it { is_expected.to validate_uniqueness_of(:delivery_partner_id).scoped_to(:active_lead_provider_id).with_message('Delivery partner and active lead provider pairing must be unique') }
+  end
+
+  describe 'scopes' do
+    describe '.with_delivery_partner' do
+      it 'selects records with a matching delivery partner id' do
```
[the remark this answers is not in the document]
peteryates: I have enough trust in PostgreSQL that if we query it with this clause it will return the right record. Not really sure what the point of creating records and ensuring they come back is.

app/models/school_partnership.rb
peteryates: That'll be covered by #729

spec/models/lead_provider_delivery_partnership_spec.rb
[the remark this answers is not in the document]
cwrw: testing the query is testing the implementation rather than the behaviour. Even though we won't ever not use PostgreSQL we should focus on testing behaviour and the outcome of the scope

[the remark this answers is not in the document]
peteryates: The construction of the query is the behaviour though.

Made the change regardless

PR #707, opened 2025-06-10
Add `api_id` to the `LeadProvider` model

ethax-ross:
Add the `api_id` - which maps to the `id` in ECF) to the `LeadProvider` model.

Update the `Migrators::LeadProvider` to populate the `api_id` as part of the migration.


ethax-ross:
Closing as review app is bust

PR #720, opened 2025-06-11
[1593] Update registration period migrator

leandroalemao:
### Context

[Issue #1593](https://github.com/DFE-Digital/register-ects-project-board/issues/1593)

We need a migrator for `cohorts` from ECF1 to ECF2 `registration_period`.

Tweak existing migrator that creates and populates a new `RegistrationPeriod` record in RECT for every `Cohort` record in ECF1.


tonyheadford (approved):
LGTM!

app/models/migration/cohort.rb
```diff
@@ -1,4 +1,7 @@
 module Migration
   class Cohort < Migration::Base
+    def next
```
ethax-ross, line 3: I wonder if its worth adding tests for the migration models if we're adding in logic 🤔 I think we did in NPQ reg towards the end. It can be one for another ticket though.

ethax-ross (approved):

cwrw (approved):
🙌

PR #723, opened 2025-06-12
Add parity check models

ethax-ross:
### Context

A parity check will consist of three models; the `Run` which represents the overall 'run', the `Request` of which there will be one per endpoint in the YAML file and the `Response` which will contain the response information for each request.

- Add ParityCheck::Run model

Add the model that will store the overall run; this is mainly used for tracking the active/completed parity checks that have been ran and grouping the runs.

- Add ParityCheck::Request model

A parity check will create multiple `Request` models; these describe the requests that will be made for each parity check run.

- Add ParityCheck::Response model

Each request will create a `Response` model (or multiple responses when we introduce pagination). The responses contain the information about the responses for ECF and RECT.


cwrw (approved):

ethax-ross:
Re-opening as review app is busted

PR #736, opened 2025-06-13
Add expression of interest relationship

### LINKED ISSUE #735: Allow `TrainingPeriod` to be linked to both `ActiveLeadProvider` and `SchoolPartnership`

peteryates:
### Context

The EOI represents the preference shown by a school to work with a lead provider. It's explicitly named 'expression_of_interest' in both the database and the models to avoid confusion between it and the **actual** lead provider that will be linked to a traiing period (via the school partnership).

Fixes #735

### DIFF
--- a/spec/factories/training_period_factory.rb
+++ b/spec/factories/training_period_factory.rb
@@ -3,10 +3,23 @@ FactoryBot.define do
 
   factory(:training_period) do
     for_ect
-    association :school_partnership
+    with_school_partnership
 
     started_on { generate(:base_training_date) }
     finished_on { started_on + 1.day }
+
+    trait :with_school_partnership do
+      association :school_partnership
+    end
+
+    trait :with_expression_of_interest do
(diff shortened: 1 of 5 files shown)


spec/factories/training_period_factory.rb
```diff
@@ -3,11 +3,26 @@
...
+    trait :with_expression_of_interest do
+      association :expression_of_interest, factory: :active_lead_provider
+    end
+
+    trait :with_only_expression_of_interest do
+      school_partnership_id { nil }
+      association :expression_of_interest, factory: :active_lead_provider
+    end
+
+    trait :expression_of
```
mooktakim, line 24: A misstype?

mooktakim (approved):
just one minor comment

spec/factories/training_period_factory.rb
peteryates: Ah yeah, thanks for catching that!

PR #743, opened 2025-06-16
Add migrator for statement adjustments

tonyheadford:
### Context

Add a migrator for ECF `Finance::Adjustment` records. These are attached to a `Statement` and so are dependent on the statement migrator being run beforehand.

There are 2 migrators for this:
1. a migrator that migrates the existing `Finance::Adjustment` records from ECF and creates `Statement::Adjustment` records in RECT
2. a migrator that creates `Statement::Adjustment` records for ECF `Finance::Statement` records that have a non-zero `reconcile_amount`

Statements with a `reconcile_amount` value was how an adjustment was performed prior to the introduction of the adjustment record functionality in ECF. There are only 8 statements with a non-zero `reconcile_amount` value in ECF

I split this into 2 migrators because they are performing different queries against different tables so more difficult to get record counts for the migrator jobs and to combine into 1 migrator without adding a lot of conditional logic and complexity.

If we wanted to combine, it makes more sense to incorporate the "reconcile_amount" migrator into the `Statement` migrator.

Adds 2 migrators:
* one to migrate statement adjustments from ECF to RECT
* one to create new statement adjustments in RECT for ECF statements with reconcile amounts.

### DIFF
--- a/app/migration/migrators/reconcile_adjustment.rb
+++ b/app/migration/migrators/reconcile_adjustment.rb
@@ -19,7 +19,7 @@ module Migrators
 
     def self.reset!
       if Rails.application.config.enable_migration_testing
-        ::StatementAdjustment.connection.execute("TRUNCATE #{::StatementAdjustment.table_name} RESTART IDENTITY CASCADE")
+        ::Statement::Adjustment.connection.execute("TRUNCATE #{::Statement::Adjustment.table_name} RESTART IDENTITY CASCADE")
       end
     end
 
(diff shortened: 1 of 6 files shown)


app/migration/migrators/reconcile_adjustment.rb
```diff
@@ -0,0 +1,41 @@
...
+      statements_with_adjustments.count
+    end
+
+    def self.model
+      # this is really :statement_adjustment but if we use that it won't queue both
+      :reconcile_adjustment
+    end
+
+    def self.statements_with_adjustments
+      Migration::Statement.where.not(reconcile_amount: 0.0)
```
leandroalemao, line 13: Should we exclude statements with adjustments from statements migrator then?

```
def self.statements
  Migration::Statement.where(reconcile_amount: 0.0)
end
```

leandroalemao (approved):

PR #744, opened 2025-06-16
Change time zone in rails to BST/London

cwrw:
### Context

We want to keep database values to the default UTC, however want the time zone to be BST/London time to avoid confusion.

- Change Time zone to London (BST)
- Keep db times at UTC (default)

### Guidance to review

I set the active record `default_timezone` to `:utc` to be explicit however that is the default. Happy to remove it if people think otherwise


peteryates (approved):
Looks good pending failing specs

leandroalemao (approved):

cwrw:
@peteryates I've updated the client and specs, I couldn't find anywhere else this would matter, but let me know if anything comes to mind that needs `utc` conversion

peteryates (approved):
All looks good.

Nowhere else springs to mind, the vast majority is date rather than time focussed.

leandroalemao (approved):

PR #746, opened 2025-06-16
Add `ParityCheckRequestJob`

ethax-ross:
### Context

We want to add a job to process the requests created in the parity check runner. The job will run the request through the client and persist the response.

- Switch `RequestBuilder` to accept `Request`

We can make the interface cleaner by switching to accepting a `request` instead of the underlying `lead_provider` and `endpoint`.

- Return `Response` from `ParityCheck::Client`

Instead of using a hash we can return the `Response` that will otherwise need to be constructed from it.

Switch to accepting `request` instead of `lead_provider` and `endpoint`.

- Add factories for parity check models

Add the factories for parity check models to make the tests easier to setup.

- Add `ParityCheck::RequestHandler`

The request handler will be called with a `Request` which it will hand off to the `Client` and then associate the response and save. This will be called from the `ParityCheckRequestJob`.

- Add `ParityCheckRequestJob`

We will enqueue one of these jobs for each request created in the `Runner`. The job hands off to the `RequestHandler`.

- Enqueue `ParityCheckRequestJob` in `Runner`

Each time we create a new `Request` we want to enqueue a `ParityCheckRequestJob` to process the request.

### Guidance to review

Arguably the job could call the `Client` directly, but I like the idea of having something in between and I expect the `RequestHandler` may end up with more responsibilities.

### DIFF
--- a/app/migration/parity_check/runner.rb
+++ b/app/migration/parity_check/runner.rb
@@ -13,7 +13,8 @@ module ParityCheck
 
       Run.create!(started_at: Time.current).tap do |run|
         lead_providers.to_a.product(endpoints).each do |lead_provider, endpoint|
-          run.requests.create!(lead_provider:, endpoint:)
+          request = run.requests.create!(lead_provider:, endpoint:)
+          ParityCheckRequestJob.perform_later(request_id: request.id)
         end
       end
     end
--- a/spec/migration/parity_check/runner_spec.rb
+++ b/spec/migration/parity_check/runner_spec.rb
@@ -40,6 +40,16 @@ RSpec.describe ParityCheck::Runner do
       )
     end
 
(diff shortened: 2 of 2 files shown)


cwrw (approved):
looks good, although I think we touched on this before around queueing everything at one and that messing with the performance of the endpoint? and possibly the post/put getting muddled/returning different results?

ethax-ross:
> looks good, although I think we touched on this before around queueing everything at one and that messing with the performance of the endpoint? and possibly the post/put getting muddled/returning different results?

Yeah this is something I need to think about; I'll maybe add another ticket for it; I think optimally we could run the GET requests in parallel and the POST/PUT one at a time after all the GETs and that would keep everything consistent for the run. Having it configurable for when we want to do a 'performance' run would be ideal as well so we can run everything serially.

PR #753, opened 2025-06-17
Bulk index page redesign

peterdavidhamilton:
### Context

The claims and outcomes journey have a "history" page for an ABs upload activity


joe-harrison-dfe:
All looks good to me here!

PR #757, opened 2025-06-17
Rename Lead Provider `api_id` to `ecf_id`

cwrw:
### Context

We will be importing the `uuid` from ECF1 into RECT. IDs required for the API will be stored in columns called `api_id` Those values will autogenerate when not set and will never be `null`. 

IDs required for the migration and to make moving data easier, however not needed for the API and likely to be removed after will be called `ecf_id`. Those IDs are not auto generated and can be `null`

Lead provider ids are not surfaced in the API so can be renamed to `ecf_id`

- Rename the column `api_id` to `ecf_id` in Lead providers
- Rename all occurrences of this column name in migrators and parity check to `ecf_id`

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -28,6 +28,6 @@ GIAS_API_PASSWORD=
 GIAS_API_USER=
 GIAS_EXTRACT_ID=
 ENABLE_TRS_TEACHER_REFRESH=true
-PARITY_CHECK_KEYS='{"lead_provider_api_id":"token"}'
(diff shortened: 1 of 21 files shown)


leandroalemao (commented):
Think we need to change here as well? https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/.env.example#L31

ethax-ross (approved):

leandroalemao (approved):

PR #764, opened 2025-06-18
Switch parity check from HTTParty to Faraday

ethax-ross:
We are switching the parity check to use Faraday to be consistent with other areas of the app.

It makes sense to take advantage of the `in_parallel` support Faraday offers to make the parity check requests in parallel (to both services at the same time).  This should make the overall parity check run quicker.

Extracts the response benchmarking to a Faraday middleware to make the client cleaner.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -25,7 +25,6 @@ gem "govuk_markdown"
 
 gem "mail-notify"
 
-gem "httparty"
 gem "sentry-rails"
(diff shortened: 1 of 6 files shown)


peteryates (approved):
Nice. Wasn't aware of `Async::HTTP::Faraday` :ok_hand:

cwrw (approved):

PR #766, opened 2025-06-18
Add admin views to review batch uploads

ghost:
This PR add a page for admins to review bulk uploads and an index page to access the bulk uploads to review:

Screenshot 2025-06-23 at 11 56 05

Screenshot 2025-06-23 at 11 55 50

### DIFF
--- a/app/helpers/batch_helper.rb
+++ b/app/helpers/batch_helper.rb
@@ -45,11 +45,12 @@ module BatchHelper
         'Created',
         'CSV Rows',
         'Processed',
-        'Errors'
+        'Errors',
+        'Action'
       ],
       rows: batches.map do |batch|
         [
-          govuk_link_to(batch.id, admin_bulk_batch_path(batch)),
+          batch.id.to_s,
           batch.appropriate_body.name,
           govuk_tag(text: batch.batch_type, colour: batch.batch_type == 'claim' ? 'blue' : 'green'),
           batch_status_tag(batch),
@@ -57,7 +58,8 @@ module BatchHelper
           batch.created_at.to_fs(:govuk),
           (batch.data&.count || 0).to_s,
           batch.pending_induction_submissions.count.to_s,
-          batch.pending_induction_submissions.with_errors.count.to_s
+          batch.pending_induction_submissions.with_errors.count.to_s,
+          govuk_link_to('View', admin_bulk_batch_path(batch))
         ]
       end
     )


spec/helpers/appropriate_body_helper_spec.rb
```diff
@@ -97,19 +105,22 @@
...
-      it 'returns yes' do
-        expect(trs_alerts_text(true)).to include('Yes')
-      end
+      it 'displays yes with a link to teacher record service' do
+        result = trs_alerts_text(true)
 
-      it 'also returns a sentence about getting more info' do
-        expect(trs_alerts_text(true)).to include('Use the', 'to get more information')
-        expect(trs_alerts_text(true)).to include('https://www.gov.uk/guidance/check-a-teachers-record')
+        expect(result).to have_text('Yes')
```
peterdavidhamilton, line 111: I'd favour writing the specs this way, there is overlap, but I think it communicates things more clearly.

```ruby
expect(result).to have_text("Use the Check a teacher's record service to get more information")
expect(result).to have_link("Check a teacher's record service", href: "https://www.gov.uk/guidance/check-a-teachers-record")
```

app/helpers/batch_helper.rb
```diff
@@ -69,19 +99,19 @@ def batch_progress_card(batch)
       },
       {
         key: { text: 'Number of CSV rows' },
-        value: { text: batch.rows.count }
+        value: { text: batch.data&.count || 0 }
```
peterdavidhamilton, line 102: I think we should stick to accessing through `#rows` becasue it is lazy there shouldn't be a performance issue

app/helpers/batch_helper.rb
```diff
@@ -98,52 +128,68 @@ def batch_raw_data_table(batch)
   # Temporary method helpful for debugging during development
```
peterdavidhamilton, line 128: We should remove this comment from any temp method that now has a permanent home in the admin console

app/helpers/batch_helper.rb
```diff
@@ -98,52 +128,68 @@ def batch_raw_data_table(batch)
...
+
     govuk_table(
-      caption: "Downloadable bad CSV data (#{batch.failed_submissions.count} rows)",
-      head: batch.row_headings.values,
-      rows: batch.failed_submissions
+      caption: "Submissions with errors (#{error_submissions.count} rows)",
+      head: ['TRN', 'Date of birth', 'Error messages'],
+      rows: error_submissions.map do |submission|
+        [
+          submission.trn,
```
peterdavidhamilton, line 153: Important distinction about your change, is when you get a row with an invalid TRN with 8 characters for example.

The TRN here will have been "sanitised" in order to be persisted, so the error won't match what the support agent sees.

The method name was chosen because it details what was available to the user as a downloadable CSV file, which the table doesn't do anymore. Now it basically repeats the table above, filtering out the successes. I think this is less useful for support, if not redundant.

peterdavidhamilton (changes requested):

peterdavidhamilton (approved):

PR #777, opened 2025-06-19
[1776] Filter financial statements

mooktakim:
### Context

Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/1776

* New `Admin::Statements::Filter` component
* New `Admin::Statements::Selector` component
* Updated `Statements::Query` service class to
  * allow `statement_date` filtering
  * allow `order_by` ordering for statement_date order or payment_date order
  * `output_fee` filter accepts `false` as value

### DIFF
--- a/app/views/admin/finance/statements/show.html.erb
+++ b/app/views/admin/finance/statements/show.html.erb
@@ -1,4 +1,4 @@
-<% page_data(title: @statement.page_title) %>
+<% page_data(title: @statement.page_title, backlink_href: (request.referer || admin_finance_statements_path)) %>
 
 <%= render Admin::Statements::SelectorComponent.new(statement: @statement) %>
 


app/components/admin/statements/filter.html.erb
```diff
@@ -0,0 +1,45 @@
+<%= form_with url: admin_finance_statements_path, method: :get do |f| %>
+  <div class="admin-statements-filter">
+      <div class="govuk-form-group filter-form-group">
```
peteryates, line 3: Indentation here jumps from 2 to 4 spaces.

mooktakim: removed spacing

peteryates (approved):
Nice - looks good and all works as expected.

PR #784, opened 2025-06-20
Rename `statements.state` to `status`

### LINKED ISSUE #715: Rename `state` to `status` on `statements`
We have a field called `state` in the `statements` table but the other state/status columns in the app are called `status`.

It'll be easier to remember if we keep them aligned and rename it to `status` and the type `statement_states` to `statement_statuses` (which can be done with [rename_enum](https://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/PostgreSQLAdapter.html#method-i-rename_enum)).

peteryates:
We have lots of status columns in the app already and having one called 'state' (as it is in ECF1) will likely be confusing.

This change renames the column, enumerated type and all refences to 'state' to 'status'

Fixes #715

## Review notes

Did I miss anywhere? It's hard to search for with `state` being the first half of `statement` and a common general term :grimacing:

### DIFF
--- a/app/migration/migrators/statement.rb
+++ b/app/migration/migrators/statement.rb
@@ -37,7 +37,7 @@ module Migrators
           payment_date: ecf_statement.payment_date,
           marked_as_paid_at: ecf_statement.marked_as_paid_at,
           output_fee: ecf_statement.output_fee,
-          state: state(ecf_statement),
+          status: status(ecf_statement),
           created_at: ecf_statement.created_at,
           updated_at: ecf_statement.updated_at
         )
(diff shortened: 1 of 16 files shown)


avinhurry (commented):
Do we also need to update the parameter and column references in:

- https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/services/statements/query.rb

It still uses `state` as a param and in method names like `where_state_is`, `Statement.with_state`, etc.

I've forgotten how to read code so please do tell me i've misunderstood or missed something 😄

PR #791, opened 2025-06-23
Add parity check start page

ethax-ross:
### Context

We want to add a page available to DfE users that allows us to kick off a parity check. It should only be accessible if you are a DfE user and the parity check is enabled on the environment. We want to list and allow the user to select endpoints for the run.

- Support multiple endpoints with the same method/path

At the moment if we define two endpoints with the same path under the same request method in the YAML file we only get one after parsing the file due to it being structured in a hash.

To avoid this, we can change the endpoints to be array items; this will allow duplicates. The use case for this is when testing, for example, an index action once with pagination and a separate time with filters.

- Add group_name and description to Endpoint

When we come to render an endpoint in a view we want to be able to describe the endpoint and group them.

Add `description` which will be used for the checkbox labels.

Add `group_name` which will allow us to group endpoints sensibly in the UI.

Add validation to ensure endpoint paths do not contain query parameters (these should be specified under the `options[:query]` in the YAML file.

- Add dependent: destroy to parity check to_many relationships

This is just for convenience, so we can clear a run with `run.destroy!` and it will take care of the related run data.

- Add admin page to start a parity check

Add the initial page to start a parity check; this is only accessible to administrators and only if the environment supports parity checks.

The `Runner` has been updated so we can use it as a form object.

If no endpoints are selected the user will see an error message.

On submitting a run we show a flash notice to inform the user it has been started.

Endpoints are grouped to make selection easier.

### Guidance to review

| Run - no endpoints    | Run - endpoints | Run - after submit |
| -------- | ------- | ------- |
| Screenshot 2025-06-23 at 14 09 27  | Screenshot 2025-06-25 at 14 47 03  | Screenshot 2025-06-25 at 14 47 36 |

### DIFF
--- a/app/controllers/migration/parity_checks_controller.rb
+++ b/app/controllers/migration/parity_checks_controller.rb
@@ -1,0 +1,28 @@
+class Migration::ParityChecksController < ::AdminController
+  before_action :load_endpoints, only: %i[new create]
+
+  def new
+    @runner = ParityCheck::Runner.new
+  end
+
+  def create
+    @runner = ParityCheck::Runner.new(runner_params)
+    if @runner.valid?
+      @runner.run!
+      flash[:notice] = "Parity check has been started."
+      redirect_to new_migration_parity_check_path
+    else
+      render :new
+    end
+  end
+
+private
+
+  def load_endpoints
+    @endpoints = ParityCheck::Endpoint.all
+  end
+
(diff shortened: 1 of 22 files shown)


leandroalemao (approved):
Tested locally here all good tks 👏 👍

PR #796, opened 2025-06-24
1702 API front page navigation

mark-avery-dfe:

I've updated the primary navigation tabs, the banner text/description and the H2s.


app/views/api/guidance/show.html.erb
mark-avery-dfe, line 1: Need to delete references to v1 and v2 of the Swagger documentation because only v3 is being carried over to the new world.

PR #798, opened 2025-06-24
Updated .tool-versions to nodejs 22.5.1

mooktakim:
Updated `.tool-versions` to nodejs 22.5.1 as 22.5.0 isn't working locally


peteryates:
Thinking about it - is there any need for us to double up?

We have `.ruby-version`, `.node-version` _and_ `.tool-versions`. I'd rather not force people into choosing a particular version manager setup though :thinking:

mooktakim:
I deleted `.tool-versions` and updated `.node-version`

Personally prefer to keep to one file with only `.tool-versions`

peteryates:
Had a play with covering this in specs in #812

PR #805, opened 2025-06-25
Replace `output_fee` boolean with `fee_type`

peteryates:
This change is an attempt to make statement fee types a bit clearer. In ECF1 the logic was that if `output_fee == true` the statement was an output fee, and if `output_fee == false` it's a service fee.

It's not intuitive or searchable, and given the schema it'd be impossible for someone to work out what a service fee is.

This change replaces the output_fee column with fee_type, which is an enumerated type where the options are `service` or `output`.

### DIFF
--- a/db/seeds/statements.rb
+++ b/db/seeds/statements.rb
@@ -47,7 +47,8 @@ grouped_active_lead_providers.each do |lead_provider, active_lead_providers|
     years.product(months).map do |year, month|
       deadline_date = Time.zone.local(year, month).end_of_month
       payment_date = Time.zone.at(rand(deadline_date.to_i..(deadline_date + 2.months).to_i))
-      status = if payment_date < Date.current
+      fee_type = %w[service output].sample
+      status = if payment_date < Date.current && fee_type == 'output'
                  :paid
                elsif Date.current.between?(deadline_date, payment_date)
                  :payable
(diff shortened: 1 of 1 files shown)


app/models/statement.rb
```diff
@@ -1,20 +1,26 @@
...
   validates :month, numericality: { in: 1..12, only_integer: true, message: "Month must be a number between 1 and 12" }
   validates :year, numericality: { greater_than_or_equal_to: 2020, is_less_than_or_equal_to: :maximum_year, only_integer: true, message: "Year must be on or after 2020 and on or before #{maximum_year}" }
   validates :active_lead_provider_id, uniqueness: { scope: %i[year month], message: "Statement with the same month and year already exists for the lead provider" }
   validates :api_id, uniqueness: { case_sensitive: false, message: "API id already exists for another statement" }
 
-  scope :with_output_fee, ->(output_fee: true) { where(output_fee:) }
+  scope :with_fee_type, ->(fee_type) { where(fee_type:) }
   scope :with_status, ->(*status) { where(status:) }
 
+  enum :fee_type, { service: "service", output: "output" }
```
peteryates, line 22: If we add a suffix here the enum-provided methods might be a bit clearer :thinking: 

```suggestion
  enum :fee_type, { service: "service", output: "output" }, suffix: 'fee'
```

eg.

```ruby
s1 = Statement.find(1)
s1.output_fee?
```

mooktakim (approved):

Rmcnaney:
Will be merged today.

PR #807, opened 2025-06-25
Show completed parity checks

ethax-ross:
> ⚠️ WIP

### Context

We want to be able to view all the completed parity check runs, so that we can later drill down into the run details.

- Support breadcrumbs in `page_data`

It looks like support for this was intended/started but not finished.

Add `breadcrumbs` option to `page_data` which will render out the GOV UK style breadcrumbs.

This is going to be useful for the parity check navigation.

- Add completed parity checks to seed data

This makes developing the parity check locally easier if we can have completed parity checks planted.

I've not planted any pending/in-progress runs as the would just be instantly picked up by the job workers and become completed when the app boots anyway - we can easily kick these off manually.

- Add convenience methods/extra validation to `Run`/`Request`/`Response`

Add convenience methods for metrics and data we will want to surface as part of the completed parity checks page, namely:

- Completed parity checks
- Request groups
- Performance gain ratio
- Match rate

Update `Request` to validate responses are present when moving to a `completed`
state.

Update `Run` to validate all requests have been completed when moving to a `completed` state.

- Add completed parity checks page

Add a new route/page to show completed parity check runs.

Link to the new page from the start page when there are completed runs present.

Display key/high-level information about each run.

Paginate the results in case we end up with a lot of retained runs.

Add breadcrumbs to make navigation easier (eventually the parity check views will become quite nested when we drill down a run -> requests -> responses -> diffs).

### Guidance to review

| Seed output    | Start page (with completed runs link) | Completed page | Completed page (no runs yet) |
| -------- | ------- | ------- | ------- | 
| Screenshot 2025-06-25 at 14 43 28 | Screenshot 2025-06-25 at 14 43 56 | Screenshot 2025-06-25 at 14 44 44 | Screenshot 2025-06-25 at 14 45 15 |

### DIFF
--- a/app/controllers/migration/parity_checks_controller.rb
+++ b/app/controllers/migration/parity_checks_controller.rb
@@ -1,7 +1,8 @@
 class Migration::ParityChecksController < ::AdminController
   layout "full"
 
-  before_action :load_endpoints, :load_runs, only: %i[new create]
+  before_action :load_endpoints, :load_pending_and_in_progress_runs, only: %i[new create]
+  before_action :load_completed_runs, only: %i[new create completed]
 
   def new
     @runner = ParityCheck::Runner.new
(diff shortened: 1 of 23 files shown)


leandroalemao (approved):
Looking good and all tested as well tkss 👏 👍

PR #808, opened 2025-06-25
[1781] Authorise for payment action on statement

mooktakim:
### Context

Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/1781

* Created `Admin::Statements::PaymentAuthorisationComponent` component
* Created `Statements::AuthorisePayment` service class
* Controller action to authorise payment

### Guidance to review

* In DfE admin, go to statements
* Select a `payable` statement

### DIFF
--- a/app/components/admin/statements/payment_authorisation_component.html.erb
+++ b/app/components/admin/statements/payment_authorisation_component.html.erb
@@ -1,0 +1,5 @@
+<% if can_authorise_payment? %>
+  <%= govuk_button_to "Authorise for payment", authorise_payment_admin_finance_statement_path(statement) %>
+<% elsif marked_as_paid? %>
+  <%= tag.p(marked_as_paid_text) %>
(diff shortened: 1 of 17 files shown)


peteryates (approved):
This looks great.

Some thoughts, not on the implementation but generally on the interaction:

* clicking authorise authorises instantly and there's no way to undo/unauthorise, and no confirmation screen
* clicking 'Authorise for payment' and then having the status go directly to 'Paid' makes me think "wow, they paid that quickly!" - I think that status does need renaming to `authorised_for_payment`

peteryates:
I kind of missed the 'Authorised for payment' line here:

image

Feels like maybe it should be in the summary list, **Authorised for payment at** 10:02pm on 25 Jun 2025

Also, we're using the short date there but the long one just above

Screenshot From 2025-06-25 22-15-08 (Edit)

PR #809, opened 2025-06-25
[1705] Create `TrainingPeriod` when registering an ECT

avinhurry:
### Context

Currently, we store the `LeadProvider` directly on the `ECTAtSchoolPeriod` when registering an ECT. This is inflexible and doesn't support scenarios where a teacher changes provider partway through their training. 

To improve this, we're introducing `TrainingPeriod` records to represent periods of training with a given provider.

- Creates a `TrainingPeriod` when registering an ECT via `Schools::RegisterECT`
- Associates it with the relevant `ECTAtSchoolPeriod`
- Links to a `SchoolPartnership` if one exists for the selected `LeadProvider`, otherwise creates an expression of interest
- Links the registration event to the `training_period`


app/services/schools/register_ect.rb
```diff
@@ -68,6 +70,44 @@ def create_teacher!
...
 
+    def registration_period
+      @registration_period ||= RegistrationPeriod.for_date(started_on)
+    end
+
+    def active_lead_provider
+      @active_lead_provider ||= ActiveLeadProvider.find_by(
+        lead_provider:,
+        registration_period:
+      ) || raise("Missing ActiveLeadProvider for #{lead_provider&.name} in #{registration_period&.year}")
```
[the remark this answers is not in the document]
peteryates: I don't think so; the contracts etc will be in place by the time the `RegistrationPeriod` opens so the `ActiveLeadProvider` and `LeadProviderDeliveryPartnership` records will all have been set up.

app/models/registration_period.rb
```diff
@@ -21,6 +21,10 @@ class RegistrationPeriod < ApplicationRecord
   validates :started_on, presence: { message: "Enter a start date" }
   validates :finished_on, presence: { message: "Enter an end date" }
 
+  def self.for_date(date)
+    where('range @> ?::date', date).first
```
peteryates, line 25: If you rebase to bring in the changes made in #776 this can be

```
find_by(*date_in_range(date))
```

peteryates: Also this feels like it's going to be useful - but it's not suitable for `Interval` because some `Period` types (mentors) can overlap.

I wonder if there should be another module for date functions common to exclusive intervals, `ExclusiveInterval` perhaps?

Not yet, just thinking.

app/services/schools/register_ect.rb
```diff
@@ -68,6 +70,44 @@ def create_teacher!
...
+    end
+
+    def create_training_period!
+      @training_period = ::TrainingPeriod.create!(
+        ect_at_school_period:,
+        started_on: ect_at_school_period.started_on,
+        school_partnership:,
+        expression_of_interest:
+      )
+    end
```
peteryates, line 109: I wouldn't worry yet but maybe when we add the mentor version of this it might make sense to deal with training period creation in its own service class. There will be a few other things that will trigger their creation too.

app/models/registration_period.rb
avinhurry: Ah missed that. Nice! I've updated to use it and renamed the method for clarity.

app/services/schools/register_ect.rb
```diff
@@ -48,6 +49,7 @@ def register!
         update_school_last_choices!
         create_teacher!
         @ect_at_school_period = start_at_school!
+        create_training_period! if lead_provider.present?
```
peteryates, line 52: It's not clear what the intent behind the `if` statement is here, it could be made a little more obvious that we only want to create them for `provider-led` training.

peteryates (approved):

app/services/schools/register_ect.rb
avinhurry: Good shout - have updated 👍🏽

PR #814, opened 2025-06-26
Fix some specs caused by incorrect merge

peteryates:
These specs were broken but [the branch](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/805) was automatically merged by GitHub even though the requirements weren't met!

### DIFF
--- a/app/controllers/admin/finance/statements_controller.rb
+++ b/app/controllers/admin/finance/statements_controller.rb
@@ -34,7 +34,7 @@ module Admin::Finance
         lead_provider:,
         registration_period_years:,
         statement_date:,
-        output_fee:,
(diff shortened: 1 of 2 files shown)


app/controllers/admin/finance/statements_controller.rb
```diff
@@ -53,16 +53,16 @@ def statement_date
       filter_params[:statement_date].presence
     end
 
-    def output_fee
+    def fee_type
```
peteryates, line 56: This struck me as a little odd, if we want it to default to 'Output' we should just pre-set the filter I think.

leandroalemao (approved):

PR #817, opened 2025-06-26
Relax TRS name validation

peteryates:
### Context

In TRS names aren't validated using the same rules we use in ECF2. We mandate the presence of both first and last names but in TRS they can be `null` or blank. This means when we run our nightly sync job, some records fail because validation isn't successful.

There are a few kinds of 'odd' records:

* records that are missing a first name
* records that are missing a last name
* records that have a full stop `.` instead of a name

The records aren't necessarily untrue; some teachers are from countries/cultures that only have one name. We need to follow TRS's lead so this change brings us in line with them. However, it also makes an effort to display things nicely.

- **Remove the 'not null' from trs first and last names**
- **Remove teacher first/last name presence validation**
- **Update Teachers::Name so it deals with partial names**

### Displaying partial names

Blank names have either a `null` value, an empty string `''` or a full stop `'.'`.

When the record has the values:

* `trs_first_name: 'John', trs_last_name: ''`, we'll show 'John'
* `trs_first_name: '.', trs_last_name: 'Jones'`, we'll show 'Jones'
* `trs_first_name: 'Edwina', trs_last_name: null`, we'll show 'Edwina'

When the record has no values:

* `trs_first_name: '.', trs_last_name: '.'`, we'll show 'Unknown' (any combinations of `nil`, `''` and `'.'` have the same behaviour here)

### Final steps

* ~~what do we do when both names are blank? Can this happen? cc @gunndabad~~


gunndabad:
@peteryates we do have a handful of records that are active, have a TRN but have neither a first name nor a last name (12 to be exact).

peteryates:
Thanks @gunndabad. I've made it show 'Unknown' on the chance we get a teacher with neither name.

db/migrate/20250626150847_remove_not_null_from_trs_teacher_names.rb
peteryates, line 1: Just making the name columns not null here, need to replace the search col as it's built using the names.

db/migrate/20250626150847_remove_not_null_from_trs_teacher_names.rb
```diff
@@ -0,0 +1,22 @@
...
+    remove_column :teachers, :search, :tsvector, type: :tsvector, as: "to_tsvector('unaccented', #{tsvector_columns})", stored: true
+    change_column :teachers, :trs_first_name, :string, null: true
+    change_column :teachers, :trs_last_name, :string, null: true
+    add_column :teachers, :search, :tsvector, type: :tsvector, as: "to_tsvector('unaccented', #{tsvector_columns})", stored: true
+    add_index :teachers, :search, using: :gin
+  end
+
+  def down
+    remove_column :teachers, :search, :tsvector, type: :tsvector, as: "to_tsvector('unaccented', #{tsvector_columns})", stored: true
+    change_column :teachers, :trs_first_name, :string, null: false, limit: 80
```
avinhurry, line 13: There is no limit on the `up` migration or the schema but the down migration includes `limit: 80`, did you mean to add this here?

Also is `add_index :teachers, :search, using: :gin` missing from `down` too?

spec/services/teachers/name_spec.rb
```diff
@@ -50,5 +74,47 @@
...
+
+    describe 'partial names in TRS' do
+      subject { service.full_name_in_trs }
+
+      let(:teacher) { FactoryBot.build(:teacher, **names) }
+
+      context 'when the first name is nil' do
+        let(:names) { { trs_first_name: nil, trs_last_name: 'Rigg' } }
+
+        it('returns only the first name') { is_expected.to eql(teacher.trs_last_name) }
```
avinhurry, line 86:
```suggestion
        it('returns only the last name') { is_expected.to eql(teacher.trs_last_name) }
```

spec/services/teachers/name_spec.rb
```diff
@@ -50,5 +74,47 @@
...
+      end
+
+      context 'when the first name is blank' do
+        let(:names) { { trs_first_name: '', trs_last_name: 'Rigg' } }
+
+        it('returns only the first name') { is_expected.to eql(teacher.trs_last_name) }
```
avinhurry, line 92:
```suggestion
        it('returns only the last name') { is_expected.to eql(teacher.trs_last_name) }
```

spec/services/teachers/name_spec.rb
```diff
@@ -50,5 +74,47 @@
...
+      end
+
+      context 'when the first name is .' do
+        let(:names) { { trs_first_name: '.', trs_last_name: 'Rigg' } }
+
+        it('returns only the first name') { is_expected.to eql(teacher.trs_last_name) }
```
avinhurry, line 98:
```suggestion
        it('returns only the last name') { is_expected.to eql(teacher.trs_last_name) }
```

avinhurry (commented):
Really well tested and very helpful description - thanks!

Just a couple of small questions/typos in line.

PR #826, opened 2025-06-27
[1811] -  add comparison migrator for schools

ltello:
### Context

[Issue](https://github.com/DFE-Digital/register-ects-project-board/issues/1811)

We need to establish whether we have the correct schools in RECT and whether their attributes match.

We can use the migrator framework to perform a comparison of schools between RECT and ECF.

For this we can build a migrator that compares the data from ECF against that in RECT and it should log errors if it finds a missing school or an attribute mismatch between schools. We can then use these errors to analyse what we have and whether we need to adjust the GIAS import to include different schools.

Note that some attribute naming is different between ECF and RECT models and not all ECF School attributes exist in RECT. Attributes that are not present in the schema shouldn't be logged as errors.

- model to connect to the ECF schools named Migration::School
- migrator to do the comparison against the GIAS::School records in RECT
- log data discrepancies
- don't log any schools that are the same
- populate the api_id for matching schools

The migrator is not be dependent on any other migrators.

The GIAS school import will need to be manually performed in the RECT environment before the migration is run and a fresh ECF snapshot used as school changes are processed daily.

## Schools imported
In ECF there are a large number of schools that are now closed that probably should not be included in the query.

We import only `Migration::School.eligible_or_cip_only` schools.


app/migration/migrators/school.rb
```diff
@@ -0,0 +1,79 @@
...
+      postcode: :postcode,
+      primary_contact_email: :primary_contact_email,
+      secondary_contact_email: :secondary_contact_email,
+      section_41_approved: :section_41_approved,
+      status: :status,
+      type_name: :school_type_name,
+      ukprn: :ukprn_to_i
+    }.freeze
+
+    MISMATCH_FIELD_MESSAGE = ->(school, field, gias_value, ecf_value) { "School #{school.urn} (#{school.name}) mismatch value on field named '#{field}': #{ecf_value} on ECF whilst #{gias_value} expected on RECT!" }
```
peteryates, line 23: Probably a bit easier to read with quotes around the values:

```suggestion
    MISMATCH_FIELD_MESSAGE = ->(school, field, gias_value, ecf_value) { "School #{school.urn} (#{school.name}) mismatch value on field named '#{field}': '#{ecf_value}' on ECF whilst '#{gias_value}' expected on RECT!" }
```

app/models/migration/school.rb
```diff
@@ -1,6 +1,57 @@
...
+    def eligible_type? = !NON_ELIGIBLE_TYPE_CODES.include?(school_type_code)
+
+    def funding_eligibility
+      return 'eligible_for_fip' if open? && in_england? && (eligible_type? || (independent_school_type? && section_41_approved?))
+      return 'eligible_for_cip' if open? && cip_only_type? && !section_41_approved?
+
+      'ineligible'
+    end
+
+    def induction_eligibility = %w[eligible_for_fip eligible_for_cip].include?(funding_eligibility)
```
peteryates, line 37: Easier to follow if this is `funding_eligibility != 'ineligible'`?

peteryates (approved):
The logic seems fine but I don't really see why we need to do this, other than any fields which we write in ECF1.

Aren't we just doing this?

```mermaid
flowchart TD
    gias -- import --> ecf1
    gias -- import --> ecf2
    ecf1 -- migrate --> ecf2
```

In my mind we should just be checking that all the schools that exist in ECF1 are in ECF2, and fix our ECF2 GIAS import to bring in any that aren't.

tonyheadford (approved):
Looks great, agree with @peteryates suggestions too for clarity

PR #827, opened 2025-06-27
Allow Playwright slowmo to be set with SLOWMO

peteryates:


spec/support/rspec_playwright.rb
```diff
@@ -11,7 +11,7 @@ def self.start_browser
     Playwright.create(playwright_cli_executable_path: PLAYWRIGHT_CLI_EXECUTABLE_PATH)
               .playwright
               .chromium
-              .launch(headless:)
+              .launch(headless:, slowMo:)
```
avinhurry, line 14: Shall we make this Ruby-esque?
```suggestion
              .launch(headless:, slowMo: slow_mo)
```

spec/support/rspec_playwright.rb
```diff
@@ -29,6 +29,12 @@ def self.headless
     end
   end
 
+  # rubocop:disable Naming/MethodName
+  def self.slowMo
+    ENV.fetch('SLOWMO', 0).to_i
+  end
+  # rubocop:enable Naming/MethodName
```
avinhurry, line 36: Then there's no need to disable rubocop: 
```suggestion
  def self.slow_mo
    ENV.fetch('SLOWMO', 0).to_i
  end
```

avinhurry (approved):
This is really cool! Left a small suggestion - otherwise looks great! Thanks

PR #830, opened 2025-06-30
Show a parity check request in detail

ethax-ross:
### Context

We want to add a page that shows the details of a parity check request and links in to each of the responses.

- Add view helpers for parity check responses

Add view helpers to render out details of a parity check response; we want to show the status code as a colour-coded tag and the comparison emoji for matching/different responses.

- Expose different? method on parity check response

To avoid `unless matching?` in various places its going to be helpful to have `different?` exposed publicly.

- Don't set page on single responses

When a request only has a single response its most likely not a paginated result; seed single responses with a `nil` page for more realistic data.

- Improve parity check request factory

When generating a completed request we should make sure the run is also in a relevant state (defaulting to completed as well).

If the caller passes `responses` we should ignore `response_types`.

- Add view to show a parity check request

We want to allow users to drill into a particular request; here we can describe the request at a high-level and list the responses associated with the request and some key metrics of those.

### Guidance to review

| Single successful response    | Multiple successful responses | Multiple mix of responses | Paginated responses |
| -------- | ------- | ------- | ------- | 
| Screenshot 2025-06-30 at 08 52 14  | Screenshot 2025-06-30 at 08 52 43   | Screenshot 2025-06-30 at 08 53 00 | screencapture-0-0-0-0-3000-migration-parity-checks-2-requests-26-2025-06-30-08_53_24 |

### DIFF
--- a/app/assets/stylesheets/migration/parity_check.scss
+++ b/app/assets/stylesheets/migration/parity_check.scss
@@ -8,7 +8,7 @@
   }
 }
 
-.parity-check-table {
+.parity-check-requests-table {
   tbody:nth-child(odd) {
     tr:last-child .govuk-table__cell {
       border: none;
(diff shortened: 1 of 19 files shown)


cwrw (approved):
🚀

PR #831, opened 2025-06-30
[1705] Create `TrainingPeriod` when registering a mentor

avinhurry:
### Context

Following on from #809 we're adding`TrainingPeriod` records to when registering a mentor.

- Add `ect_id` to store In order to infer the `lead_provider` via the `ect` in the `register_mentor` service. 

**Note that the above will change shortly, once we remove `provider_id` from `ECTAtSchoolPeriod`**

- Creates a `TrainingPeriod` when registering a mentor via `Schools::RegisterMentor`
- Associates it with the relevant `MentorAtSchoolPeriod`
- Links to a `SchoolPartnership` if one exists for the selected `LeadProvider`, otherwise creates an expression of interest
- Links the registration event to the `training_period`
- Add `TrainingPeriod#create` service
- Refactor shared methods into `training_period_sources.rb`

### DIFF
--- a/app/models/concerns/training_period_sources.rb
+++ b/app/models/concerns/training_period_sources.rb
@@ -6,10 +6,7 @@ module TrainingPeriodSources
   end
 
   def active_lead_provider
-    @active_lead_provider ||= ActiveLeadProvider.find_by(
-      lead_provider:,
-      contract_period:
-    ) || raise("Missing ActiveLeadProvider for #{lead_provider&.name} in #{contract_period&.year}")
+    @active_lead_provider ||= ActiveLeadProvider.find_by!(lead_provider:, contract_period:)
   end
 
   def school_partnership
--- a/spec/services/schools/register_ect_spec.rb
+++ b/spec/services/schools/register_ect_spec.rb
@@ -33,8 +33,8 @@ RSpec.describe Schools::RegisterECT do
     let(:lead_provider) { FactoryBot.create(:lead_provider) }
 
     context 'when no ActiveLeadProvider exists for the contract_period' do
-      it 'raises an error explaining the missing ActiveLeadProvider' do
-        expect { service.register! }.to raise_error(RuntimeError, /Missing ActiveLeadProvider/)
+      it 'raises an error' do
+        expect { service.register! }.to raise_error(ActiveRecord::RecordNotFound)
       end
     end
 
--- a/spec/services/schools/register_mentor_spec.rb
+++ b/spec/services/schools/register_mentor_spec.rb
@@ -26,8 +26,8 @@ RSpec.describe Schools::RegisterMentor do
 
   describe '#register!' do
     context 'when no ActiveLeadProvider exists for the registration period' do
-      it 'raises an error explaining the missing ActiveLeadProvider' do
-        expect { service.register! }.to raise_error(RuntimeError, /Missing ActiveLeadProvider/)
+      it 'raises an error' do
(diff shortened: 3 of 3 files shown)


app/services/training_periods/create.rb
```diff
@@ -0,0 +1,29 @@
+module TrainingPeriods
+  class Create
+    def initialize(period:, started_on:, school_partnership:, expression_of_interest:)
+      @period = period
+      @started_on = started_on
+      @school_partnership = school_partnership
+      @expression_of_interest = expression_of_interest
+    end
+
+    def call
```
peteryates, line 10: `#create_training_period` might be a bit more descriptive?

avinhurry: I did wonder about that but I went with `.call` mainly to keep it aligned with common service object patterns where the purpose of the class already describes the action (i.e 'Create' is already in the name).
  

`TrainingPeriods::Create.new(...).call` felt a bit more idiomatic and avoids the duplication that `TrainingPeriods::Create.new(...).create_training_period` has.

Happy to go with the longer form if you think clarity would be improved but wanted to flag the reason behind my decision.

Let me know your thoughts.

app/services/training_periods/create.rb
```diff
@@ -0,0 +1,29 @@
...
+      ::TrainingPeriod.create!(
+        period_type_key => @period,
+        started_on: @started_on,
+        school_partnership: @school_partnership,
+        expression_of_interest: @expression_of_interest
```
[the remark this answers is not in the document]
avinhurry: Were you thinking something like: 

```
if period.is_a?(ECTAtSchoolPeriod)
  TrainingPeriods::CreateFromEctPeriod.new(period:, started_on:, school_partnership:, expression_of_interest:).call
elsif period.is_a?(MentorAtSchoolPeriod)
  TrainingPeriods::CreateFromMentorPeriod.new(period:, started_on:, school_partnership:, expression_of_interest:).call
```

I guess its more simple but comes with more repetition. I suppose my version doesn't follow the SRP thinking about it, but it does keep things nice and encapsulated 🤔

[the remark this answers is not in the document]
peteryates: Yeah, I'd have done something likestarted_on: @started_on,
        school_partnership: @school_partnership,
        expression_of_interest: @expression_of_interest

app/services/training_periods/create.rb
peteryates: I think the reason is sound but generally prefer the clearer version. I know we're unlikely to do it here in this case but if you had `a = TrainingPeriods::Create.new(...)` then several lines later `a.call`, it would be less clear than `a.create_training_period`

app/models/concerns/training_period_sources.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  def contract_period
+    @contract_period ||= ContractPeriod.containing_date(started_on)
+  end
+
+  def active_lead_provider
+    @active_lead_provider ||= ActiveLeadProvider.find_by(
+      lead_provider:,
+      contract_period:
+    ) || raise("Missing ActiveLeadProvider for #{lead_provider&.name} in #{contract_period&.year}")
```
peteryates, line 12: Is the custom error here worth it, vs just a `find_by!`?

```suggestion
    @active_lead_provider ||= ActiveLeadProvider.find_by!(lead_provider:, contract_period:)
```

peteryates (approved):
Looks good to me!

PR #833, opened 2025-06-30
[1788] added/updated/deleted events for statement adjustments

mooktakim:
### Context

Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/1788

* Added/updated/deleted events for statement adjustments

### DIFF
--- a/app/services/events/record.rb
+++ b/app/services/events/record.rb
@@ -21,6 +21,9 @@ module Events
                 :school_partnership,
                 :lead_provider,
                 :delivery_partner,
+                :active_lead_provider,
+                :statement,
+                :statement_adjustment,
                 :user,
                 :modifications,
                 :metadata
@@ -43,6 +46,9 @@ module Events
       school_partnership: nil,
       lead_provider: nil,
       delivery_partner: nil,
+      active_lead_provider: nil,
+      statement: nil,
+      statement_adjustment: nil,
       user: nil,
       modifications: nil,
       metadata: nil
@@ -64,6 +70,9 @@ module Events
       @school_partnership = school_partnership
       @lead_provider = lead_provider
       @delivery_partner = delivery_partner
+      @active_lead_provider = active_lead_provider
+      @statement = statement
+      @statement_adjustment = statement_adjustment
       @user = user
       @modifications = DescribeModifications.new(modifications).describe
       @metadata = metadata || modifications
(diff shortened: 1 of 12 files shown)


spec/features/admin/finance/adjustments/update_adjustment_spec.rb
```diff
@@ -22,6 +24,7 @@
     and_i_see_new_adjustment_values
     and_i_see_new_adjustment_total
     and_adjustment_should_have_been_updated
+    and_an_adjustment_updated_event_recorded
```
avinhurry, line 27:
```suggestion
    and_an_adjustment_updated_event_is_recorded
```

spec/features/admin/finance/adjustments/update_adjustment_spec.rb
```diff
@@ -107,6 +110,13 @@ def when_i_fill_in(label:, value:)
   alias_method :and_i_fill_in, :when_i_fill_in
 
   def and_i_click_button(name)
-    page.get_by_role('button', name:).click
+    perform_enqueued_jobs do
+      page.get_by_role('button', name:).click
+    end
+  end
+
+  def and_an_adjustment_updated_event_recorded
```
avinhurry, line 118:
```suggestion
  def and_an_adjustment_updated_event_is_recorded
```

avinhurry (commented):
Had a question in line and a tiny suggestion.

Looks good otherwise 👍🏽

app/services/events/record.rb
```diff
@@ -343,6 +352,82 @@ def self.record_lead_provider_api_token_revoked_event!(author:, api_token:)
...
+      new(
+        event_type:,
+        author:,
+        heading:,
+        statement:,
+        active_lead_provider:,
+        lead_provider:,
+        happened_at: Time.zone.now,
+        metadata:
+      ).record_event!
```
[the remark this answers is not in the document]
peteryates: I think if we're deleting it there won't be an `id`, so it makes sense to be omitted.

app/services/events/record.rb
```diff
@@ -343,6 +352,82 @@ def self.record_lead_provider_api_token_revoked_event!(author:, api_token:)
       new(event_type:, author:, heading:, lead_provider:, happened_at: Time.zone.now, metadata:).record_event!
     end
 
+    # Statement Adjustment Events
+
+    def self.record_statement_adjustment_added_event!(author:, statement_adjustment:)
```
peteryates, line 357: Does it make sense to pass all the values in here rather than traverse from `statement_adjustment`? The chances are we already have access to them in the controller and it generally makes testing easier. No strong feelings either way here though because there are only a couple being set.

mooktakim: i think this keeps the controller clean, let the service class do all the work. adjustment knows all the information

app/services/events/record.rb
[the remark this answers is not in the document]
mooktakim: We won't need to as the obj still exists with all the information we need.
Also, its inside a transaction now.

avinhurry (approved):

peteryates (approved):

PR #840, opened 2025-07-01
[1783] Add api schools serialiser

leandroalemao:
### Context

Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/1783

We would like to add a serializer for the schools endpoint that is similar to [ECF1 ](https://github.com/DFE-Digital/early-careers-framework/blob/main/app/serializers/api/v3/ecf/school_serializer.rb). This will tie into building the main endpoint.

- Add api id to schools
- Add service to return the training programme for a given school and registration period
- Add api schools serialiser

### Guidance to review

Dev review

### DIFF
--- a/app/models/school.rb
+++ b/app/models/school.rb
@@ -121,8 +121,6 @@ class School < ApplicationRecord
   end
 
   def training_programme_for(contract_period_id:)
-    return transient_training_programme if respond_to?(:transient_training_programme)
-
     Schools::TrainingProgramme.new(school: self, contract_period_id:).training_programme
   end
 
--- a/app/services/schools/query.rb
+++ b/app/services/schools/query.rb
@@ -10,7 +10,8 @@ module Schools
       @scope = eligible_schools.not_cip_only.select(
         "schools.*",
         transient_in_partnership,
-        transient_training_programme(lead_provider.id),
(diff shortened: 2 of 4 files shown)


cwrw (commented):
let's pause this for now until we get answers and finalise all the other tickets, thanks leandro :)

app/serializers/school_serializer.rb
```diff
@@ -0,0 +1,43 @@
...
+        return false if options[:registration_period_id].blank?
+
+        school.school_partnerships.for_registration_period(options[:registration_period_id]).exists?
+      end
+
+      def updated_at(school, options)
+        (
+          (school.school_partnerships.for_registration_period(options[:registration_period_id]) || []).map(&:updated_at) +
+          [school.updated_at]
+        ).compact.max
```
peteryates, line 32: This working out of the most recent `updated_at` is quite _heavy_ for a serialiser - perhaps it should be in a service class of its own?

cwrw: we discussed this on the spike and we'll be implementing another field for updated at we explicitly update when needed rather than following this pattern which is based off ECF1, it'll be a different ticket and we'll pick this back up once implemented :) NO TOUCH MODEL

PR #841, opened 2025-07-01
Rename `RegistrationPeriod` to `ContractPeriod`

peteryates:
### Context

This renames `RegistrationPeriod` to `ContractPeriod`.

We've had several discussions (and [a poll](https://github.com/DFE-Digital/register-ects-project-board/discussions/1886)) on the topic of changing this, but the main reasons for the switch is that the date isn't really related to when the person was registered, but when their training starts - i.e., which contract does it fall under.

- **Rename registration_periods to contract_periods**
- **Replace registration period with contract period**


Rmcnaney:
Someone to review today plsss

cwrw (approved):

PR #849, opened 2025-07-02
AB teachers query

peterdavidhamilton:
### Context

[Sentry alert](https://dfe-teacher-services.sentry.io/issues/6621363705/events/f38e59cba8234f3aaa3f44a5bb5463d9/)

- add development gem
- diagnose query on AB homepage

### Guidance to review

Locally, select an AB persona. Without the addition in the controller you will see a browser/console warning


cpjmcquillan:
With `prosopite`, these are the logs **before** the fix

image

And these are the logs **after** the fix

image

---
It looks like there is another N+1 to unpick there, but on first glance it doesn't seem trivial.

- Essentially we are returning a bunch of `Teacher records`, then iterating them. 
- For each `Teacher`, we render a row in the table. 
- When we render the "Induction start date", the service object makes a query to retrieve the first `InductionPeriod` record **ordered** by `started_on` date.

I can think of a couple straightforward ways we might fix this, but I'd like to hear people's thoughts.

1. **Order the association, so we don't need to make a separate query in the service object.** 
Ordering induction periods by the date they started on makes sense to me, but does this make sense in all contexts? Are there situations where this is unhelpful?
2. **Define "first" (and "last") associations on the `Teacher` model**
This has a smaller blast radius, since we can "include" `first_induction_period` with the search query for this specific controller without impacting **everything**. But maybe `first_induction_period` isn't important enough to be defined as a distinct relationship in the `Teacher` model...

What do people think? Is there another approach we might prefer? cc @peterdavidhamilton

peteryates (approved):
:zap:

cpjmcquillan:
Looking at #843 (which solves a similar N+1 issue), I'm tempted to go for option 2 🔼

peterdavidhamilton:
> Looking at #843 (which solves a similar N+1 issue), I'm tempted to go for option 2 🔼

Go for it @cpjmcquillan 🦾

PR #852, opened 2025-07-02
[1705] Remove `lead_provider_id` from `ECTAtSchoolPeriod`

avinhurry:
### Context

We are removing `lead_provider_id` from `ECTAtSchoolPeriod`. `TrainingPeriod` will now be the source of truth for the `LeadProvider`.

- Remove `lead_provider_id` from `ECTAtSchoolPeriod`
- Update teacher history seeds with training periods


app/migration/builders/ect/school_periods.rb
```diff
@@ -15,7 +15,6 @@ def build
           ::ECTAtSchoolPeriod.create!(teacher:,
                                       school:,
                                       training_programme: period.training_programme,
-                                      lead_provider_id: period.lead_provider_id,
```
avinhurry, line 18: Will this change cause any issues for the migration work @tonyheadford?

app/models/ect_at_school_period.rb
```diff
@@ -61,8 +48,17 @@ class ECTAtSchoolPeriod < ApplicationRecord
...
+    training_periods.latest_first.first
+  end
+
+  def lead_provider
+    latest_training_period&.lead_provider
+  end
+
+  def delivery_partner
+    latest_training_period&.delivery_partner
+  end
```
ltello, line 61: We have a specific SO for the training stuff of an ECTAtSchoolPeriod.
https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/services/ect_at_school_periods/training.rb

ltello (commented):
I just left a few comments @avinhurry.
Hopefully that helps

app/services/schools/register_ect.rb
```diff
@@ -86,15 +86,14 @@ def active_lead_provider
...
       SchoolPartnership
         .joins(:lead_provider_delivery_partnership)
         .find_by(
           school:,
           lead_provider_delivery_partnerships: {
-            active_lead_provider_id: provider.id
+            active_lead_provider_id: active_lead_provider.id
           }
         )
     end
```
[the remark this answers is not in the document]
avinhurry: Ah great spot and thank you for reviewing and flagging @ltello!

Do you have any suggestions? We could order by `created_at` temporarily but thats very risky. 

Would it make sense to add a `started_at` column to `SchoolPartnership` and use this to decide which one to choose.

Did you have any thoughts on this @peteryates?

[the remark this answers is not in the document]
peteryates: We had a chat about this recently when discussing #856 and @easeynathan's steer was to use the earliest created school partnership in the registration period. In that PR there's a `.earliest_first` scope to help do that.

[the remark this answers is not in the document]
avinhurry: Would this be a temporary solution? If we ever need to import or correct data later on, `created_at` might not reflect the actual start order of the partnerships.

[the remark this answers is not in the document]
ltello: If the first ECT registered for that contract period wants to keep their LP-DP from previous school that will basically screw up this school for all the rest of ECTs/Mentors for that contract period.
I guess that will create multiple request on support.

Would it be possible to leave these participants linked only to an expresion of interest instead? So the LP will eventually resolve it and decide the school partnership to link the participant to?

app/models/ect_at_school_period.rb
```diff
@@ -61,8 +48,13 @@ class ECTAtSchoolPeriod < ApplicationRecord
...
 
-  # lead_provider_name
-  delegate :name, to: :lead_provider, prefix: true, allow_nil: true
+  def lead_provider
+    training_service.latest_lead_provider
+  end
+
+  def delivery_partner
+    training_service.latest_delivery_partner
+  end
```
ltello, line 57: Not sure we want to include this here and call it just `lead_provider` and `delivery_partner` instead of `latest_lead_provider` or `current_lead_provider`...

I wouldn't include these methods in the model and call the SC `ECTAtSchoolPeriods::Training.new...` instead wherever the training information of an ECTAtSchoolPeriod is needed throught the codebase.

avinhurry: Ah yeah, you're right - I was being a bit lazy. It looked like more needed changing based on the failing tests, but it's definitely better to remove it now before it gets used more widely. Good shout.

All updated 👍🏽

app/views/schools/register_mentor_wizard/check_answers.html.erb
```diff
@@ -30,7 +30,7 @@
     if @wizard.ect.provider_led_training_programme? && @mentor.funding_available?
       summary_list.with_row do |row|
         row.with_key(text: 'Lead provider')
-        row.with_value(text: @wizard.ect.lead_provider.name)
+        row.with_value(text: ECTAtSchoolPeriods::Training.new(@wizard.ect).latest_lead_provider&.name )
```
ltello, line 33:
```suggestion
        `row.with_value(text: ECTAtSchoolPeriods::Training.new(@wizard.ect).latest_lead_provider_name)`
```

app/components/schools/summary_card_component.rb
```diff
@@ -54,16 +58,16 @@ def school_rows
         { key: { text: 'Appropriate body' }, value: { text: @ect.school_reported_appropriate_body_name } },
         { key: { text: 'Training programme' }, value: { text: training_programme_name(@ect.training_programme) } }
       ].tap do |rows|
-        rows << { key: { text: 'Lead provider' }, value: { text: @ect.lead_provider_name } } if @ect.provider_led?
+        rows << { key: { text: 'Lead provider' }, value: { text: training.latest_lead_provider&.name } } if @ect.provider_led?
```
ltello, line 61:
```suggestion
        rows << { key: { text: 'Lead provider' }, value: { text: training.latest_lead_provider_name } } if @ect.provider_led?
```

app/components/schools/summary_card_component.rb
```diff
@@ -54,16 +58,16 @@ def school_rows
...
 
     def lead_provider_rows
-      return NO_INFORMATION_REPORTED[:lead_provider] unless (period = @ect.training_periods.earliest_first.last)
+      return NO_INFORMATION_REPORTED[:lead_provider] unless @ect.provider_led?
 
       [
-        { key: { text: 'Lead provider' }, value: { text: period.lead_provider.name } },
-        { key: { text: 'Delivery partner' }, value: { text: period.delivery_partner.name } }
+        { key: { text: 'Lead provider' }, value: { text: training.latest_lead_provider&.name || 'Not available' } },
+        { key: { text: 'Delivery partner' }, value: { text: training.latest_delivery_partner&.name || 'Not available' } }
```
ltello, line 70:
```suggestion
        { key: { text: 'Lead provider' }, value: { text: training.latest_lead_provider_name || 'Not available' } },
        { key: { text: 'Delivery partner' }, value: { text: training.latest_delivery_partner_name || 'Not available' } }
```

app/views/schools/register_mentor_wizard/confirmation.md.erb
```diff
@@ -21,7 +21,8 @@ They do not need to provide us with any further details.
 
 <% if @wizard.ect.provider_led_training_programme? %>
 <% if @mentor.funding_available? %>
-We'll pass on their details to <%= @wizard.ect.lead_provider.name %> who will contact them to arrange mentor training.
+We'll pass on their details to <%= ECTAtSchoolPeriods::Training.new(@wizard.ect).latest_lead_provider&.name %>
```
ltello, line 24:
```suggestion
We'll pass on their details to <%= ECTAtSchoolPeriods::Training.new(@wizard.ect).latest_lead_provider_name %>
```

app/views/schools/register_mentor_wizard/review_mentor_eligibility.html.erb
```diff
@@ -7,7 +7,7 @@
 
 <p class="govuk-body">Our records show that <%= @mentor.full_name %> can get up to 20 hours of ECTE mentor training as your school is working with a DfE-funded training provider.</p>
 
-<p class="govuk-body">We'll pass on their details to <%= @wizard.ect.lead_provider.name %> who will contact them to arrange the training.</p>
+<p class="govuk-body">We'll pass on their details to <%= ECTAtSchoolPeriods::Training.new(@wizard.ect).latest_lead_provider&.name %> who will contact them to arrange the training.</p>
```
ltello, line 10:
```suggestion
<p class="govuk-body">We'll pass on their details to <%= ECTAtSchoolPeriods::Training.new(@wizard.ect).latest_lead_provider_name %> who will contact them to arrange the training.</p>
```

app/wizards/schools/register_mentor_wizard/mentor.rb
```diff
@@ -71,7 +71,9 @@ def ect
       end
 
       def lead_provider
-        ect&.lead_provider
+        return unless ect
+
+        ECTAtSchoolPeriods::Training.new(ect).latest_lead_provider
```
ltello, line 76:
```suggestion
        ECTAtSchoolPeriods::Training.new(ect).latest_lead_provider if ect
```

app/components/schools/mentors/ect_mentor_training_details_component.rb
```diff
@@ -19,16 +19,18 @@ def eligible_for_training?
...
 
       def assigned_ects
         @assigned_ects ||= @mentor.currently_assigned_ects
       end
+
+    private
+
+      def lead_provider
+        ECTAtSchoolPeriods::Training.new(assigned_ects.select(&:present?).min_by(&:started_on))&.latest_lead_provider
+      end
```
ltello, line 33:
```suggestion
        assigned_ects
          .order(:started_on)
          .detect do |ect|
            name = ECTAtSchoolPeriods::Training.new(ect).latest_lead_provider_name
            return name if name
          end
      end

      def assigned_ects
        @assigned_ects ||= @mentor.currently_assigned_ects
      end        
```

avinhurry:
Thank you for your feedback @ltello! I have made all your suggested changes

app/services/schools/register_ect.rb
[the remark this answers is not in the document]
claire-hughez: I'm not sure this works and it might warrant a conversation. It's worth bearing in mind this barely ever happens though!

>So the LP will eventually resolve it via the API and decide which school partnership to link the participant to?

There is no way for a lead provider to link participants to a particular delivery partner over the API. Just a school.

>They wouldn't be creating a partnership, they'd need to assign an ECT to an existing one. Is that possible?

We have a design and mapping for this I believe. 

But there's still an issue where the partnership doesn't exist at the point of registration. I think we could:
- Just show a list of the different delivery partners it could be
- Inform lead providers to pass this on in support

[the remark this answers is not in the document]
avinhurry: Resolving this - the discussion was taken offline. 

**Next steps:** 

1) Flesh out @mason-emily's  ticket on rolling over and pick up next sprint
2) Add ticket on 2+ DPs and 1LPs for admin tooling backlog

PR #856, opened 2025-07-02
Add `School::LatestRegistrationChoices` service

peteryates:
### Context

This service class will, given a `school` and `registration_period`, allow someone to retrieve:

* the last chosen appropriate body
* the last chosen lead provider and, if a school partnership already exists between the lead provider and school in the registration period, the associated delivery partner 

If there is no previous choice recorded against the school, the `#appropriate_body` and `#lead_provider_and_delivery_partner` methods will return `nil`

### Questions

* Should we return something when the `last_chosen_lead_provider` is present on the school but isn't in a partnership with the school? @mason-emily 

### Review guidance

Commit by commit is the way to go. The `LeadProviders::Active` service added in bc4be3e isn't used here but, depending on the answer to the question above could be. If we don't need it I'll drop it from the PR (although we'll likely need it or something like it at some point).


mason-emily:
> Should we return something when the last_chosen_lead_provider is present on the school but isn't in a partnership with the school?
So there's an EOI for that provider? Yes we should return the LP only in that case - see below the different versions of the previous choices page for schools

image

tonyheadford (approved):
LGTM

PR #860, opened 2025-07-02
Relax trs name validation

peteryates:
This is a replacement for #817 - I can't reopen it because I force pushed to the branch while it was closed. 

### Context

In TRS names aren't validated using the same rules we use in ECF2. We mandate the presence of both first and last names but in TRS they can be `null` or blank. This means when we run our nightly sync job, some records fail because validation isn't successful.

There are a few kinds of 'odd' records:

* records that are missing a first name
* records that are missing a last name
* records that have a full stop `.` instead of a name

The records aren't necessarily untrue; some teachers are from countries/cultures that only have one name. We need to follow TRS's lead so this change brings us in line with them. However, it also makes an effort to display things nicely.

- **Remove the 'not null' from trs first and last names**
- **Remove teacher first/last name presence validation**
- **Update Teachers::Name so it deals with partial names**

### Displaying partial names

Blank names have either a `null` value, an empty string `''` or a full stop `'.'`.

When the record has the values:

* `trs_first_name: 'John', trs_last_name: ''`, we'll show 'John'
* `trs_first_name: '.', trs_last_name: 'Jones'`, we'll show 'Jones'
* `trs_first_name: 'Edwina', trs_last_name: null`, we'll show 'Edwina'

When the record has no values:

* `trs_first_name: '.', trs_last_name: '.'`, we'll show 'Unknown' (any combinations of `nil`, `''` and `'.'` have the same behaviour here)

### Final steps

* ~~what do we do when both names are blank? Can this happen? cc @gunndabad~~


spec/services/teachers/name_spec.rb
```diff
@@ -50,5 +74,47 @@
...
+      context 'when the last name is nil' do
+        let(:names) { { trs_first_name: 'Diana', trs_last_name: nil } }
+
+        it('returns only the last name') { is_expected.to eql(teacher.trs_first_name) }
+      end
+
+      context 'when the last name is blank' do
+        let(:names) { { trs_first_name: 'Diana', trs_last_name: '' } }
+
+        it('returns only the last name') { is_expected.to eql(teacher.trs_first_name) }
```
avinhurry, line 110:
```suggestion
        it('returns only the first name') { is_expected.to eql(teacher.trs_first_name) }
```

avinhurry (approved):
Just a few more types - but otherwise 👍🏽

PR #862, opened 2025-07-03
Recheck a bulk upload before completing

peterdavidhamilton:
### Context

If a bulk upload were to sit in a processed state and not progress to completion promptly, and any of these happens before the batch is completed:

1. the user manually processes the same TRNs
2. another user at the same AB alters ECT records
3. an admin or automation altered records
4. another AB altered records
5. an update to TRS for the TRNs

Then the promised outcomes may no longer be achievable, and unexpected validation error messages may be seen, when the batch CTA completes the previously valid submissions.

Before completing a batch, revalidate previously valid submissions, and render changes to the user

### Guidance to review

Invalidate a previously valid submission somehow, before the CTA, then when proceeding to complete the change should be caught by a second round of validation

### DIFF
--- a/app/jobs/process_batch_job.rb
+++ b/app/jobs/process_batch_job.rb
@@ -18,7 +18,19 @@ class ProcessBatchJob < ApplicationJob
     author = event_author(pending_induction_submission_batch, author_email, author_name)
     batch_service = self.class.batch_service.new(pending_induction_submission_batch:, author:)
 
-    batch_service.complete! if pending_induction_submission_batch.processed? || pending_induction_submission_batch.completing?
+    # confirm and complete
+    if pending_induction_submission_batch.processed?
+      if batch_service.revalidated?
+        batch_service.complete!
+      else
+        pending_induction_submission_batch.pending!
+        self.class.perform_later(pending_induction_submission_batch, author_email, author_name)
+      end
+    end
+
+    # resume completing
+    batch_service.complete! if pending_induction_submission_batch.completing?
+    # process or resume processing
     batch_service.process! if pending_induction_submission_batch.pending? || pending_induction_submission_batch.processing?
   rescue StandardError => e
     Rails.logger.debug("Attempt #{executions}: #{e.message}")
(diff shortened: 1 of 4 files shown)


app/services/appropriate_bodies/process_batch/base.rb
```diff
@@ -32,6 +32,18 @@ def process!
...
+      def revalidated?
+        pending_induction_submission_batch.pending_induction_submissions.without_errors.all? do |pending_induction_submission|
+          @row = pending_induction_submission_batch.row_by(trn: pending_induction_submission.trn)
+          @pending_induction_submission = pending_induction_submission
+
+          next(false) if fails_pre_checks?
+
+          validate_submission!.nil?
+        end
+      end
```
cpjmcquillan, line 45: I wonder whether revalidating submissions might be more straightforward if we were processing individual rows 💭

Is this something we have considered before?

cpjmcquillan (approved):
I added a couple thoughts, but they aren't directly related to this change  💭

app/services/appropriate_bodies/process_batch/base.rb
peterdavidhamilton: Not sure I'm picturing it. Here we revalidate only those rows that were previously valid, in the same (two-stage) way they got validated before. Before I merge can we chat to make sure I'm not missing a trick?

app/jobs/process_batch_job.rb
```diff
@@ -18,7 +18,19 @@ def perform(pending_induction_submission_batch, author_email, author_name)
...
-    batch_service.complete! if pending_induction_submission_batch.processed? || pending_induction_submission_batch.completing?
+    # confirm and complete
+    if pending_induction_submission_batch.processed?
+      if batch_service.revalidated?
+        batch_service.complete!
+      else
+        pending_induction_submission_batch.pending!
+        self.class.perform_later(pending_induction_submission_batch, author_email, author_name)
+      end
+    end
```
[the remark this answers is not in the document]
cpjmcquillan: Yeah I think the nested conditional is a good visual cue that there is something complicated going on 👌🏼

PR #864, opened 2025-07-03
Make links to statements more accessible

### LINKED ISSUE #713: Improve accessibilty of statements index
The statements index page currently has a load of links that just say "View". A screen reader will just read this out as "View, View, View, View".

Image

This should be more informative. There needs to be some [visually hidden text](https://govuk-components.netlify.app/helpers/visually-hidden-text/) that explains what the person will be viewing if they click the link.

At a minimum it should be "View statement 123" but it's likely going to be more useful if it says "View statement for [Lead Provider Name] in [Month and Year]".

cpjmcquillan:
### Context

#713 

Before, the links to statements in the admin finance dashboard weren't helpful for users using assistive technologies (i.e they all said "View").

This updates the links to include some visually hidden content that more clearly communicates the intent of each link to these users.

I've followed the approach suggested in #713.

Closes #713.

### Guidance to review

Inspect the links to statements on the admin finance page.

image

### DIFF
--- a/spec/views/admin/finance/statements/index.html.erb_spec.rb
+++ b/spec/views/admin/finance/statements/index.html.erb_spec.rb
@@ -24,7 +24,11 @@ RSpec.describe 'admin/finance/statements/index.html.erb' do
   it 'has a link to each statement' do
     render(locals:)
 
-    raw_statements.each { |s| expect(rendered).to have_link('View', href: admin_finance_statement_path(s)) }
+    presented_statements = Admin::StatementPresenter.wrap(raw_statements)
+    presented_statements.each do |s|
+      link_text = "View statement for #{s.lead_provider_name} in #{s.month_and_year}"
+      expect(rendered).to have_link(link_text, href: admin_finance_statement_path(s))
+    end
   end
 
   it 'has the right number of rows' do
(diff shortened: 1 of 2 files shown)


spec/views/admin/finance/statements/index.html.erb_spec.rb
```diff
@@ -24,7 +24,11 @@
   it 'has a link to each statement' do
     render(locals:)
 
-    raw_statements.each { |s| expect(rendered).to have_link('View', href: admin_finance_statement_path(s)) }
+    presented_statements = Admin::StatementPresenter.wrap(raw_statements)
+    presented_statements.each do |s|
+      link_text = "View statement for #{s.lead_provider_name} in #{s.month_and_year}"
+      expect(rendered).to have_link(link_text, href: admin_finance_statement_path(s))
+    end
```
peteryates, line 31: A little wordier but maybe something along the lines of this?

```suggestion    
    raw_statements.each do |s|
      statement_date = "#{Date::MONTHNAMES.fetch(s.month)} #{s.year}"

      expect(rendered).to have_link("View statement for #{s.lead_provider.name} in #{statement_date}", href: admin_finance_statement_path(s))
    end
```

peteryates: Perhaps moving the statement date building out into its own class would make sense? Then it can be used everywhere while being fairly concise.

peteryates: That doesn't need to happen now though, this is still an improvement over what was there.

PR #872, opened 2025-07-03
Introduce Mission Control

cpjmcquillan:
### Context

We want to know more about our background jobs easily. When someone asked "what's happening with OTP", my instinct was to reach for Sidekiq UI (or similar), but we didn't have anything setup.

Mission Control looks to be Rails' take on a basic jobs dashboard, and it supports Solid Queue.

This introduces the [mission control] gem from Rails so that we can use the lightweight dashboard to see more about the status of our background jobs.

I tried to keep the setup straight-forward, so the dashboard is enabled in all environments.

Admins are the only users who can access the dashboard, and they should authenticate in the usual way. This adds some complexity (the default is HTTP basic auth), but I think it's worthwhile in the interests of security.

Closes #1929.

[mission control]: https://github.com/rails/mission_control-jobs

### Guidance to review

- Get the dashboard running locally, play around with it etc
- Use the dashboard in a review app

image

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -27,9 +27,12 @@ gem "mail-notify"
 
 gem "sentry-rails"
 gem "sentry-ruby"
-gem "solid_queue"
 gem "stackprof"
 gem "state_machines-activerecord"
+
+# Background jobs
+gem "mission_control-jobs"
+gem "solid_queue"
 
 # DfE Sign-In
 gem "omniauth"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -326,6 +326,10 @@ GEM
     immutable-ruby (0.2.0)
       concurrent-ruby (~> 1.1)
       sorted_set (~> 1.0)
+    importmap-rails (2.1.0)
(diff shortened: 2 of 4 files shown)


app/services/super_admin/access.rb
```diff
@@ -0,0 +1,19 @@
...
+
+    def can_access?
+      "super_admin".in?(dfe_roles)
+    end
+
+  private
+
+    def dfe_roles
+      user&.user&.dfe_roles&.map(&:role_type)
+    end
```
peteryates, line 17: Might make sense to add a `super_user?` method to [Sessions::Users::DfEUser](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/services/sessions/users/dfe_user.rb) and the correpsonding [Persona](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/services/sessions/users/dfe_persona.rb).

graciebrennan:
@edujackedu to take a look to approve

peterdavidhamilton (approved):

PR #876, opened 2025-07-04
Introduce `Statements::Period` class

cpjmcquillan:
### Context

Following on from #864 and based on this [discussion]

This introduces a small `Statements::Period` class that returns a statement period.

I'm not completely convinced this is the best approach, so keen to hear people's thoughts.

Some decisions worth discussing:

- `Statements::Period` as a name: "Period" makes sense to me, but maybe the verbosity of "YearMonth" is more explicit and clearer
- The API: `.for(statement)` feels ergonomic to me, but perhaps we'd prefer to initialise the object and call `to_s` explicitly
- `.to_s`: I feel like this captures the intention, but maybe it's confusing. I also *vaguely* remember this being bad practice, but I can't remember *why*

Also, are there any other places this might be useful?

[discussion]: https://github.com/DFE-Digital/register-early-career-teachers-public/pull/864#discussion_r2182402696

### Guidance to review

- [ ] Take a look at the statement filter
- [ ] Take a look at the statement selector
- [ ] Take a look at the statements page

### DIFF
--- a/app/components/admin/statements/filter_component.rb
+++ b/app/components/admin/statements/filter_component.rb
@@ -29,7 +29,7 @@ module Admin
 
         dates.map do |date|
           statement_date.new(
-            name: "#{month_name(date[1])} #{date[0]}", # January 2025
+            name: ::Statements::Period.from_year_and_month(date.first, date.second), # January 2025
             id: date.join("-") # 2025-01
           )
         end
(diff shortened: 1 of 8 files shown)


peteryates (commented):
Looks great to me, nice to see little bits of repeated logic removed from various places.

Just missing specs for `Statements::Period`.

peteryates (approved):

PR #878, opened 2025-07-04
Make db test prepare work properly

peteryates:
The changes in #871 made things better but not perfect.

This change does the following:

* switch enhancing from `db:test:prepare` to `db:test:purge` - `db:test:prepare` always throws the database away and recreates it, so if the enhancement happens before it's disregarded. `db:test:purge` happens within `db:test:prepare`, before the schema is loaded or migrations run
* simplifies the job by removing the checks for database presence, we're hooking on to events that happen when the database exists so it's unnecessary

### DIFF
--- a/lib/tasks/db.rake
+++ b/lib/tasks/db.rake
@@ -36,21 +36,8 @@ end
 namespace :db do
   desc 'Add search config'
   task setup_search_configuration: :environment do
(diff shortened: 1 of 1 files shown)


cwrw (approved):
thanks Pete ⭐ !

PR #881, opened 2025-07-07
Fix flaky test

cpjmcquillan:
This test [failed recently in CI][fail] since the events were returned in a different order.

Since we're not explictly ordering the Events when we call `Event.all`, the order isn't guaranteed.

To avoid this, we can use `contain_exactly` instead of `eq`.

[fail]: https://github.com/DFE-Digital/register-early-career-teachers-public/actions/runs/16076939703/job/45374245792?pr=872


peteryates (approved):

peteryates:
Thanks! Looks like it's the same flavour as #854. Wonder if there are any more lurking.

Issue #761, opened 2025-06-18
Description bug in events with empty string


peteryates:
I noticed we had an event where a blank value in the event description.

![Image](https://github.com/user-attachments/assets/d848877e-1fb8-48b4-88d2-f238dd560c28)

I'd have expected the logic in `Events::DescribeModifications` to record this in the `Attribute set to 'Y'` format rather than `Attribute changed from '' to 'Y'`.

In this example a new teacher was claimed by an AB, their record was imported from TRS and their status changed to 'InProgress'.

## Example record

https://www.register-early-career-teachers.education.gov.uk/admin/teachers/76191/timeline

cpjmcquillan:
This isn't actually to do with an event's modification descriptions, but how we set the event heading when events are recorded.

In a couple places we don't apply any logic to format the content, so the heading is always `"Induction status changed from '#{old_induction_status}' to '#{new_induction_status}'"` —even when that doesn't make sense semantically.

I'm going to extract a small class to format these messages and use it consistently throughout.

We can then run a rake task that uses the class to update any offending events 👍🏼

joe-harrison-dfe:
reviewed and all looks good to me

kas-madrigal:
whenever this is ready to merge, go for it, it's reviewed on product side

PR #891, opened 2025-07-08
Describe event transitions consistently

cpjmcquillan:
### Context

#761 

Before, we described event modifications one way, but described some event headings another way.

At first we thought this was due to a bug #761, but in fact it was a consequence of this inconsistency.

This introduces an `Events::TransitionDescription` service that returns a human-readable description of an event transition. It gracefully handles blank values like the existing code, as well as handling some extra edge cases.

Now, we can use this to describe event modifications and for event headings when reporting events. Event timelines should now be clearer in all circumstances!

### Guidance to review

- Record some teacher name change and induction status change events
- Look at the timeline!

### DIFF
--- a/app/services/events/transition_description.rb
+++ b/app/services/events/transition_description.rb
@@ -1,0 +1,39 @@
+module Events
+  class TransitionDescription
+    def self.for(...) = new(...).describe
+
+    def initialize(attribute_name, from:, to:)
+      @attribute_name = attribute_name
+      @from = from
+      @to = to
+    end
+
+    def describe
+      case
+      when @from.blank? && @to.blank?
+        "#{@attribute_name.humanize} is not set"
+      when @from.blank?
+        "#{@attribute_name.humanize} set to #{format(@to)}"
+      when @to.blank?
+        "#{@attribute_name.humanize} #{format(@from)} removed"
+      when @from == @to
+        "#{@attribute_name.humanize} #{format(@from)} unchanged"
+      else
+        "#{@attribute_name.humanize} changed from #{format(@from)} to #{format(@to)}"
+      end
+    end
+
+  private
+
+    def format(value)
+      formatted_value = case value
+                        when Date
+                          value.to_formatted_s(:govuk_short)
+                        else
+                          value
+                        end
+
+      %('#{formatted_value}')
+    end
+  end
+end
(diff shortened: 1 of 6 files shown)


db/scripts/fix_event_headings_with_empty_strings.rb
cpjmcquillan, line 1: Alternatively, we could do this in SQL with something like this

```sql
BEGIN;

-- First update: "changed from ''" to "set"
UPDATE events 
SET heading = REPLACE(heading, 'changed from ''''', 'set') 
WHERE heading LIKE '%changed from ''''%';

-- Second update: "changed from X to ''" to "X removed"
UPDATE events 
SET heading = REPLACE(REPLACE(heading, 'changed from ', ''), ' to ''''', ' removed') 
WHERE heading LIKE '%changed from % to ''''%';

COMMIT;
```

Which approach do people prefer? 💭

cpjmcquillan, line 1: Alternatively, we could do this in SQL with something like this

```sql
BEGIN;

-- First update: "changed from ''" to "set"
UPDATE events 
SET heading = REPLACE(heading, 'changed from ''''', 'set') 
WHERE heading LIKE '%changed from ''''%';

-- Second update: "changed from X to ''" to "X removed"
UPDATE events 
SET heading = REPLACE(REPLACE(heading, 'changed from ', ''), ' to ''''', ' removed') 
WHERE heading LIKE '%to ''''%';

COMMIT;
```

Which approach do people prefer? 💭

db/scripts/fix_event_headings_with_empty_strings.rb
```diff
@@ -0,0 +1,23 @@
...
+end
+
+# Retrieve all the events with headings that "changed from `Something` to ''"
+#
+# Before: "Induction status changed from 'InProgress' to ''"
+# After: "Induction status 'InProgress' removed"
+Event.where("heading LIKE ?", "%to ''").find_each do |event|
+  heading = event.heading.gsub("changed from ", "").gsub("to ''", "removed")
+  event.update_column(:heading, heading)
+end
```
cpjmcquillan, line 23: I'm not sure we need to do this part.

In Production there are 719 events `WHERE heading LIKE '%changed from ''''%';` and 719 events `WHERE heading LIKE '%''''%';`.

db/scripts/fix_event_headings_with_empty_strings.rb
peteryates: The SQL is nice and concise here.

I think it's clearer to [dollar quote](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-DOLLAR-QUOTING) though.

peteryates (approved):

app/services/events/transition_description.rb
peteryates, line 1: Super neat.

PR #894, opened 2025-07-08
1890: allow admins to import teachers

ghost:
### Context
https://github.com/DFE-Digital/register-ects-project-board/issues/1890

## Description

We have seen support issues like [this one](https://trello.com/c/mnhV4Vdm/1612-record-not-found-in-riab) where a record is in TRS but not in the RECT database (in this case, correctly, as the record was in Failed status when we migrated)

In this case, the support user is unable to edit the record in RIAB as it has not been created. They then need to be able to add a closed induction period to the record, so replicating the "claim" journey for admin users would not quite fit this scenario.

Instead, it would be most flexible to do the two things in the "Claim" journey as 2 separate steps

1. Import a record from TRS (no induction period, no status update) - covered by this PR 
2. Add an induction period (either closed or open) - covered by another ticket

- Added a new flow to import ECTs without creating an induction period as part of the process
- changed pending induction submissions to allow `nil` appropriate body id

### Guidance to review
- this feature partially mirrors the claim ECT flow for ABs

### DIFF
--- a/app/controllers/admin/import_ect/check_ect_controller.rb
+++ b/app/controllers/admin/import_ect/check_ect_controller.rb
@@ -1,0 +1,31 @@
+module Admin
+  module ImportECT
+    class CheckECTController < AdminController
+      def edit
+        @pending_induction_submission = find_pending_induction_submission
+        @teacher = Teacher.find_by(trn: @pending_induction_submission.trn)
+      end
+
+      def update
+        @pending_induction_submission = find_pending_induction_submission
+
+        check_ect = Admin::ImportECT::CheckECT.new(
+          pending_induction_submission: @pending_induction_submission
+        )
+
+        if check_ect.import
+          redirect_to(admin_import_ect_register_path(check_ect.pending_induction_submission))
+        else
+          @pending_induction_submission = check_ect.pending_induction_submission
+          render :edit
+        end
+      end
+
+    private
+
+      def find_pending_induction_submission
+        PendingInductionSubmission.find(params[:id])
+      end
+    end
+  end
+end
--- a/app/controllers/admin/import_ect/find_ect_controller.rb
+++ b/app/controllers/admin/import_ect/find_ect_controller.rb
@@ -1,0 +1,45 @@
+module Admin
+  module ImportECT
+    class FindECTController < AdminController
+      def new
+        @pending_induction_submission = PendingInductionSubmission.new
+      end
+
+      def create
+        find_ect = Admin::ImportECT::FindECT.new(
+          pending_induction_submission: PendingInductionSubmission.new(
+            **pending_induction_submission_params
+          )
+        )
+        @pending_induction_submission = find_ect.pending_induction_submission
+
+        if find_ect.import_from_trs!
+          redirect_to(edit_admin_import_ect_check_path(find_ect.pending_induction_submission))
+        else
+          render(:new)
+        end
+      rescue TRS::Errors::QTSNotAwarded
+        @pending_induction_submission.save!
+        redirect_to admin_import_ect_errors_no_qts_path(@pending_induction_submission)
+      rescue TRS::Errors::ProhibitedFromTeaching
+        @pending_induction_submission.save!
+        redirect_to admin_import_ect_errors_prohibited_path(@pending_induction_submission)
+      rescue TRS::Errors::InductionAlreadyCompleted
+        @pending_induction_submission.save!
+        redirect_to admin_import_ect_errors_already_complete_path(@pending_induction_submission)
+      rescue TRS::Errors::TeacherNotFound, TRS::Errors::TeacherDeactivated
+        @pending_induction_submission.errors.add(:base, "No teacher with this TRN and date of birth was found")
+        render(:new)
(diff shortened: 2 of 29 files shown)


app/views/admin/import_ect/register_ect/show.html.erb
```diff
@@ -0,0 +1,27 @@
...
+  You've successfully imported <%= @pending_induction_submission.trs_first_name %> <%= @pending_induction_submission.trs_last_name %> into the system. They are now available for appropriate bodies to claim.
+</p>
+
+<h2 class="govuk-heading-m">What happens next</h2>
+
+<p>
+  Use this service to manage <%= @pending_induction_submission.trs_first_name %> <%= @pending_induction_submission.trs_last_name %>'s induction. For example, if:
+</p>
+
+<ul class="govuk-list govuk-list--bullet">
```
peterdavidhamilton, line 18: Can we use the list helper again

spec/features/admin/import_ect_spec.rb
```diff
@@ -0,0 +1,214 @@
...
+  end
+
+  def and_i_should_see_the_ect_details
+    expect(page.get_by_text('Kirk Van Houten')).to be_visible
+    expect(page.get_by_text('1234567')).to be_visible
+    expect(page.get_by_text('1 February 2003')).to be_visible
+  end
+
+  def when_i_click_import_ect
+    # Try the most common submit button text
```
peterdavidhamilton, line 155: Is wording still being considered?

peterdavidhamilton (approved):
Looks familiar and like it makes sense

db/schema.rb
cpjmcquillan, line 1: Should this change be in here?

app/views/admin/import_ect/register_ect/show.html.erb
```diff
@@ -0,0 +1,27 @@
...
+<%= govuk_list([
+  "they need their details updated",
+  "they change appropriate body",
+  "their induction programme changes or ends"
+], type: :bullet) %>
+
+<ul class="govuk-list">
+  <li><%= govuk_link_to("Import another ECT", new_admin_import_ect_find_path, no_visited_state: true) %></li>
+  <li><%= govuk_link_to("Back to teachers", admin_teachers_path, no_visited_state: true) %></li>
+</ul>
```
cpjmcquillan, line 27: Should we use the [`govuk_list`](https://govuk-components.netlify.app/helpers/list/) helper here too?

app/views/admin/import_ect/errors/induction_already_completed.md.erb
```diff
@@ -0,0 +1,7 @@
+---
+title: You cannot register <%= PendingInductionSubmissions::Name.new(@pending_induction_submission).full_name %>
```
cpjmcquillan, line 2: I wonder if it's worth defining a `full_name` method on the submission? I appreciate we want don't want massive models, but the ergonomics of this feels a bit janky IMO.

We could have something like this

```ruby
# app/models/pending_induction_submission.rb

class PendingInductionSubmission
  def full_name
    PendingInductionSubmissions::Name.new(self).full_name
  end
end
```

and use it in views like this

```erb
<%= @pending_induction_submission.full_name %>
```

What do you think? 💭

app/controllers/admin/import_ect/find_ect_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+
+    private
+
+      def pending_induction_submission_params
+        params.require(:pending_induction_submission).permit(:trn, :date_of_birth)
+      end
+
+      def pending_induction_submission_attributes
+        { appropriate_body_id: nil }
+      end
```
cpjmcquillan, line 47: Do we need to set this explicitly? 💭

app/controllers/admin/import_ect/check_ect_controller.rb
```diff
@@ -0,0 +1,31 @@
...
+      end
+
+      def update
+        @pending_induction_submission = find_pending_induction_submission
+
+        check_ect = Admin::ImportECT::CheckECT.new(
+          pending_induction_submission: @pending_induction_submission
+        )
+
+        if check_ect.begin_import!
```
cpjmcquillan, line 16: Typically I'd expect methods ending in `!` to do something surprising or destructive or non-normal in some way.

Since we're not using `update!` under the hood, this feels little misleading here.

What do you think?

app/controllers/admin/import_ect/find_ect_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+        )
+        @pending_induction_submission = find_ect.pending_induction_submission
+
+        if find_ect.import_from_trs!
+          redirect_to(edit_admin_import_ect_check_path(find_ect.pending_induction_submission))
+        else
+          render(:new)
+        end
+      rescue TRS::Errors::QTSNotAwarded
+        @pending_induction_submission.save!
```
cpjmcquillan, line 23: What's the reason we want to save the submission after these sorts of errors? 🤔

cpjmcquillan (approved):
Left a few comments, but most trying to understand what's going on! 💚

PR #895, opened 2025-07-09
[1784] Add query service for schools endpoint

leandroalemao:
### Context

Ticket: [1784](https://github.com/DFE-Digital/register-ects-project-board/issues/1784)

We would like to create the query similar to [ECF1](https://github.com/DFE-Digital/early-careers-framework/blob/main/app/services/api/v3/ecf/schools_query.rb) to drive the GET schools endpoint

- Add a schools query which accepts similar filters for cohort/urn/updated since and returns the correct results. Please remember that cohort is a required filter


app/services/schools/query.rb
```diff
@@ -0,0 +1,175 @@
...
+
+    def initialize(lead_provider_id: :ignore, urn: :ignore, updated_since: :ignore, contract_period_id: :ignore, sort: nil)
+      @scope = default_scope(contract_period_id).select(
+        "schools.*",
+        transient_in_partnership(contract_period_id),
+        transient_mentors_at_school(contract_period_id),
+        transient_ects_at_school_training_programme(contract_period_id),
+        transient_expression_of_interest(lead_provider_id, contract_period_id)
+      ).or(schools_with_existing_partnerships(contract_period_id))
+            .or(schools_with_expression_of_interests(lead_provider_id, contract_period_id))
```
cwrw, line 17: do we need this? Im thinking the default scope would include those? 🤔

app/services/schools/query.rb
```diff
@@ -0,0 +1,175 @@
...
+
+      ects_eoi = School.select('1 AS one').from('schools s')
+                  .joins(ect_at_school_periods: { training_periods: { expression_of_interest: :contract_period } })
+                  .where(contract_periods: { year: contract_period_id })
+                  .where(expression_of_interest: { lead_provider_id: })
+                  .where('schools.id = s.id')
+                  .limit(1)
+                  .to_sql
+
+      "CASE
```
cwrw, line 151: maybe this final logic can be in ruby and we just include the ects/mentors as transient values?

app/models/concerns/gias_helpers.rb
```diff
@@ -9,6 +9,9 @@ module GIASHelpers
     scope :in_england, -> { in_gias_schools.where(gias_school: { in_england: true }) }
     scope :section_41, -> { in_gias_schools.where(gias_school: { section_41_approved: true }) }
     scope :eligible, -> { currently_open.eligible_establishment_type.in_england.or(currently_open.in_england.section_41) }
+    scope :cip_only, -> { currently_open.where(gias_school: { type_name: GIAS::Types::CIP_ONLY_TYPES }) }
+    scope :eligible_or_cip_only, -> { eligible.or(cip_only) }
```
cwrw, line 13: do we need this?

cwrw (commented):
thanks Leandro for this! just a few comments, happy to chat if there are any questions

app/services/gias/types.rb
```diff
@@ -1,5 +1,5 @@
 module GIAS
-  class Types
+  module Types
```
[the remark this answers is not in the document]
leandroalemao: Because it groups related constants... so it looks like a module rather than a class... I can change it back to class 👍

app/services/schools/query.rb
peteryates, line 1: This feels like we're forcing an ORM to do things ORMs aren't meant to do.

The resulting object isn't really a `School` record.

Should this just be an actual query?

app/models/concerns/gias_helpers.rb
```diff
@@ -0,0 +1,20 @@
...
+    scope :in_gias_schools, -> { includes(:gias_school).references(:gias_schools) }
+
+    scope :currently_open, -> { in_gias_schools.where(gias_school: { status: GIAS::Types::OPEN_STATUSES }) }
+    scope :eligible_establishment_type, -> { in_gias_schools.where(gias_school: { type_name: GIAS::Types::ELIGIBLE_TYPES }) }
+    scope :in_england, -> { in_gias_schools.where(gias_school: { in_england: true }) }
+    scope :section_41, -> { in_gias_schools.where(gias_school: { section_41_approved: true }) }
+    scope :eligible, -> { currently_open.eligible_establishment_type.in_england.or(currently_open.in_england.section_41) }
+    scope :cip_only, -> { currently_open.where(gias_school: { type_name: GIAS::Types::CIP_ONLY_TYPES }) }
+    scope :eligible_or_cip_only, -> { eligible.or(cip_only) }
+    scope :not_cip_only, -> { where.not(id: cip_only) }
```
peteryates, line 14: If you look in [GIAS::SchoolRow](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/services/gias/school_row.rb) we do lots of this eligibility checking up front when we import/update schools from GIAS.

We should be able to use that (it might need to be tweaked/extended) but it'll save having to use complex logic/queries.

app/services/schools/query.rb
cwrw: We are returning a school object with extra "transitional" fields which will help with calculations due to the complexity of getting to the data.
I think it is quite a complex query however still a query. I think we can definitely do improvements, but I'd leave it for further iterations maybe

app/models/concerns/gias_helpers.rb
peteryates: If for any reason we can't use it as-is, we should either update what we have or add to it so we can apply the logic once at import and then use it via a single scope.

spec/services/schools/query_spec.rb
```diff
@@ -0,0 +1,442 @@
...
+      }
+    end
+
+    context "when no params is sent" do
+      it "returns no schools" do
+        expect(described_class.new.schools).to be_empty
+      end
+    end
+
+    it "returns all eligible schools" do
```
cwrw, line 17: can we add a spec for existing partnerships schools, that might be ineligible?

cwrw (approved):

cwrw (approved):

PR #899, opened 2025-07-09
Use dates instead of days of the week for backup filenames

peteryates:
The infra team recently made a change to the storage account where backup files can't be deleted/overwritten for a certain period of time, after which they'll be automatically deleted.

We hit a problem today where the `Wednesday` backup couldn't be overwritten so the backup failed.

This change switches the format from:

`cpdec2_production_nightly_copy_Wednesday` to `cpdec2_production_nightly_copy_2025-07-09`


.github/workflows/nightly-copy-production_db_into_snapshot_db_and_store_dump.yml
```diff
@@ -9,34 +9,34 @@ permissions:
...
     outputs:
-      today_day: ${{ steps.get-day.outputs.today_day }}
+      today: ${{ steps.get-date.outputs.today }}
     steps:
-      - name: Get today's day of the week
-        id: get-day
-        run: echo "today_day=$(date +'%A')" >> $GITHUB_OUTPUT
+      - name: Get today's date
+        id: get-date
+        run: echo "today=$(date +'%F')" >> $GITHUB_OUTPUT
```
avinhurry, line 19: Maybe we could omit the special characters to make this more robust for storage?

```suggestion
        run: echo "today=$(date +'%Y%m%d')" >> $GITHUB_OUTPUT
```

peteryates: We're using `+'F'` elsewhere so i left it for consistency, but if we were starting fresh I'd agree.

PR #905, opened 2025-07-10
[1502] Add extra validation checks for start date when registering an ECT

avinhurry:
### Context

We are adding two new validation checks for the start date when registering an ECT. 

#### Prevent users entering a start date that is more than 2 registration periods ago.

image

#### If the ECT has been registered at another school before we will prevent them from adding a start date that is before the  previous registration start date. 

image

### Guidance to review

I considered extracting the previous registration logic into a `ECTRegistrationHistory` service object. However the existing methods in the ECT wrapper already provided the neccssary functionality.  Also a lack of a `Teacher` at times during the wizard flow would have added complexity so I decided against it in the end but happy to revisit if others feel strongly.

### DIFF
--- a/app/wizards/schools/register_ect_wizard/start_date_step.rb
+++ b/app/wizards/schools/register_ect_wizard/start_date_step.rb
@@ -40,7 +40,7 @@ module Schools
       def add_start_date_too_early_error(period)
         errors.add(
           :start_date,
-          "This ECT was previously registered at #{period.school&.name} (#{period.started_on.to_formatted_s(:govuk)}). Enter a later date."
+          "Our records show that #{wizard.ect.full_name} started teaching at #{period.school&.name} on #{period.started_on.to_formatted_s(:govuk)}. Enter a later start date."
         )
       end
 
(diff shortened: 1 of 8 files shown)


app/wizards/schools/register_ect_wizard/start_date_step.rb
```diff
@@ -21,6 +22,32 @@ def previous_step
 
     private
 
+      def start_date_not_before_previous_school
+        return if skip_start_date_validation?
```
mooktakim, line 26: to avoid showing two error messages, can we add `return if errors.any?`

avinhurry: It's already in the `skip_start_date_validation?` method (line 47)

```
 def skip_start_date_validation?
        start_date.blank? || errors[:start_date].any?
      end
```

lib/schools/validation/ect_start_date.rb
```diff
@@ -21,6 +21,17 @@ def formatted_date
       def date_missing?
         super || date_as_hash.values_at(1, 2, 3).all?(&:nil?)
       end
+
+      def extra_validation_error_message
+        return unless earliest_permitted_date
+        return unless value_as_date < earliest_permitted_date
+
+        "Start date cannot be earlier than the last 2 registration periods. Enter a date later than #{earliest_permitted_date.to_formatted_s(:govuk)}"
```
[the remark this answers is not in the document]
avinhurry: Thanks - i've fixed and also updated the screenshot in the description.

PR #906, opened 2025-07-10
[DNM] testing analytics

cwrw:


app/models/migration/base.rb
```diff
@@ -2,7 +2,7 @@ module Migration
   class Base < ApplicationRecord
     self.abstract_class = true
 
-    connects_to database: { reading: :legacy, writing: :legacy } unless Rails.env.test?
+    connects_to database: { reading: :legacy, writing: :legacy } if Rails.env.migration? || Rails.env.development?
```
peteryates, line 5: Can we control this using an environment variable rather than checking the Rails envs?

peteryates (approved):

PR #907, opened 2025-07-10
Fix N+1 queries on the admin teachers page

cpjmcquillan:
### Context

This addresses some N+1 queries on the admin teachers page.

1. Calling `count` on an ActiveRecord association **always** makes a database request. If we include `induction_periods` when retrieving the Teacher records, we can use `size` instead.

2. Similarly to #858, if we define a `current_appropriate_body` association, we can leverage ActiveRecord and `include` it without make an extra trip to the database.

### Guidance to review

**Before**
image
image

**After**
image

### DIFF
--- a/app/models/teacher.rb
+++ b/app/models/teacher.rb
@@ -17,7 +17,10 @@ class Teacher < ApplicationRecord
   has_many :induction_periods
   has_one :first_induction_period, -> { order(started_on: :asc) }, class_name: "InductionPeriod"
   has_one :last_induction_period, -> { order(started_on: :desc) }, class_name: "InductionPeriod"
+  has_one :ongoing_induction_period, -> { ongoing }, class_name: "InductionPeriod"
+
   has_many :appropriate_bodies, through: :induction_periods
+  has_one :current_appropriate_body, through: :ongoing_induction_period, source: :appropriate_body
 
   has_many :events
 
--- a/app/controllers/admin/teachers_controller.rb
+++ b/app/controllers/admin/teachers_controller.rb
@@ -7,7 +7,12 @@ module Admin
       @pagy, @teachers = pagy(
         ::Teachers::Search.new(
           query_string: params[:q]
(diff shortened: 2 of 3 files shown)


app/models/teacher.rb
```diff
@@ -17,7 +17,10 @@ class Teacher < ApplicationRecord
   has_many :induction_periods
   has_one :first_induction_period, -> { order(started_on: :asc) }, class_name: "InductionPeriod"
   has_one :last_induction_period, -> { order(started_on: :desc) }, class_name: "InductionPeriod"
+  has_one :ongoing_induction_period, -> { ongoing }, class_name: "InductionPeriod"
```
cpjmcquillan, line 20: There are a few places we might want to use this. Do we want to do that here or later? 💭

app/models/teacher.rb
```diff
@@ -17,7 +17,10 @@ class Teacher < ApplicationRecord
...
+
   has_many :appropriate_bodies, through: :induction_periods
+  has_one :current_appropriate_body, through: :ongoing_induction_period, source: :appropriate_body
```
avinhurry, line 23: Would you mind adding some tests please to confirm it returns the correct appropriate body (or `nil`) depending on whether the teacher has an ongoing induction period?

app/models/teacher.rb
avinhurry: I'm not in tomorrow but FWIW, depends how many there are - might be nicer to get this in then create follow ups?

avinhurry (commented):
Nice improvement! 👍🏽 

Just 1 comment around tests in line.

app/models/teacher.rb
cpjmcquillan: Added ✅

app/models/teacher.rb
cpjmcquillan: Opened #919

PR #909, opened 2025-07-11
Add rate limiting

RMcVelia:
### Context

Add front door rate limiting 

Add for production, sandbox and staging
rate_limit_max is the max per source IP per 5 minute period
Number given is double what is seen for a 10 minute period via logit and the nginx ingress

### Guidance to review

check you are ok with the max setting
make env domains-plan


config/terraform/domains/environment_domains/config/sandbox.tfvars.json
```diff
@@ -12,5 +12,6 @@
       "environment_short": "sb",
       "origin_hostname": "cpd-ec2-sandbox-web.teacherservices.cloud"
     }
-  }
+  },
+  "rate_limit_max": 150
```
cwrw, line 16: can we do staging as well, as we had an outage there as well?

cwrw (approved):

PR #910, opened 2025-07-11
Fix issues with `MentorshipPeriod` migrator

tonyheadford:
### Context

After recent split up of `Teacher` migrators I noticed that the `MentorshipPeriod` migrator had incorrect dependencies which meant that it ran before all it's dependencies were met.  Also refactored to add model and perform migration in the migrate! method and add missing specs

- Fix dependencies for migrator
- Refactor migrate method
- Add specs

### DIFF
--- a/app/migration/builders/mentorship_periods.rb
+++ b/app/migration/builders/mentorship_periods.rb
@@ -37,12 +37,12 @@ module Builders
           next
         end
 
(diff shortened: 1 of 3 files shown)


gen-sole:
@ltello to review

ltello (approved):

PR #912, opened 2025-07-11
Improving the fake TRS API client

peteryates:
### Context

We've had several problems caused by several using and overwriting the same data in TRS's preprod API.

For example, when someone's induction is marked as complete in RIAB, they're no longer claimable in RECT. This meant team members had to reset/reseed `staging` to be able to re-claim teachers.

### Changes

This PR splits the `TRS::FakeAPIClient` in two:
  1. the `TRS::TestAPIClient` which is used by tests and has values set on initialization
  2. the `TRS::FakeAPICLient` which can be used by the app when `config.enable_fake_trs_api = true` (and will fail in production regardless of the setting)

The main reason for the split was the original approach of having an `app_mode` setting made the behaviour of the class more complex and changes could potentially either break the tests or the fake behaviour, adding risk.

Now they are separate the risk is reduced and the classes are much simpler.

### Using the test API client

The intended way of using `TRS::TestAPIClient` is by stubbing the call before running the test, so to return a teacher who has alerts which prohibit them from teaching we can do this:

```ruby
  before do
    test_api_client = TRS::TestAPIClient.new(is_prohibited_from_teaching: true)
    allow(TRS::APIClient).to receive(:new).and_return(test_api_client)
  end
```

### Using the fake API client

First set `ENABLE_FAKE_TRS_API=true` in your `.envrc`/`.env` file.

Now restart the application. The blue help box will look like this:

image

These special TRNs will illicit a predetermined response. All other TRNs will return a new `TRS::Teacher` with a [fake name](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/lib/trs/fake_api_names.yml).

### Making the fake API remember changes

Until now the fake API didn't actually do anything when inductions are started, passed, failed or reset. This means when a teacher is 'Passed' their TRS induction status remains `InProgress`, which would be confusing and unrealistic.

Additionally, if teachers exist in an environment and the `Teachers::RefreshTRSAttributes` job runs, their status would be overwritten.

We already have Redis deployed in all environments (thus far only used for recording Rack Attack hits) and this functionality uses it. It's probably going to surprise devs when they pull these changes so I'll post about it in the channel first. It should just be a case of installing Valkey/Redis and setting `REDIS_CACHE_URL='redis://localhost:6379'`.

https://github.com/user-attachments/assets/f507c9db-9544-4cc2-b017-a52e95315222

### Review notes

Go commit by commit. It looks like lots has changed but nearly all of it is changing references to the fake API with the test API.

### DIFF
--- a/lib/trs/fake_api_client.rb
+++ b/lib/trs/fake_api_client.rb
@@ -1,26 +1,12 @@
 module TRS
   class FakeAPIClient
     class FakeAPIClientUsedInProduction < StandardError; end
-    attr_reader :random_names
 
-    def initialize(
-      raise_not_found: false,
-      raise_deactivated: false,
-      include_qts: true,
-      include_itt: true,
-      prohibited_from_teaching: false,
-      induction_status: nil,
-      random_names: false
-    )
+    def initialize(raise_not_found: false, raise_deactivated: false)
       fail(FakeAPIClientUsedInProduction) if Rails.env.production?
 
       @raise_not_found = raise_not_found
       @raise_deactivated = raise_deactivated
-      @include_qts = include_qts
-      @include_itt = include_itt
-      @prohibited_from_teaching = prohibited_from_teaching
-      @induction_status = induction_status
-      @random_names = random_names
     end
 
     def find_teacher(trn:, date_of_birth: "1977-02-03", national_insurance_number: nil)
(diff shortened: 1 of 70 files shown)


lib/trs/fake_api_client.rb
```diff
@@ -29,112 +15,213 @@ def find_teacher(trn:, date_of_birth: "1977-02-03", national_insurance_number: n
...
+    def override_data_for_special_trns(trn)
       case trn.to_i
-      when 7_000_001 then raise(TRS::Errors::QTSNotAwarded)
+      when 7_000_001 then @has_qts = false
       when 7_000_002 then raise(TRS::Errors::TeacherNotFound)
-      when 7_000_003 then raise(TRS::Errors::ProhibitedFromTeaching)
+      when 7_000_003 then @is_prohibited_from_teaching = true
       when 7_000_004 then raise(TRS::Errors::TeacherDeactivated)
+      when 7_000_005 then @has_alerts_but_not_prohibited = true
+      when 7_000_006 then nil # teacher with TRN 7000006 is seeded as an early roll out mentor
```
peteryates, line 121: These changes mean the fake API client returns data instead of raising early, so it works with both the RIAB and RECT flows.

spec/support/api/trs/test_api_client_spec.rb
peteryates, line 1: Unconventional location, but wasn't really sure where to put this!

lib/trs/fake_api_client.rb
```diff
@@ -29,112 +15,213 @@ def find_teacher(trn:, date_of_birth: "1977-02-03", national_insurance_number: n
...
     end
 
-    def induction_status
-      return {} unless @induction_status
-
+    def other_alert_data
       {
-        'induction' => { 'status' => @induction_status },
+        # Conditional Registration Order - unacceptable professional conduct
+        'alerts' => [{ 'alertType' => { 'alertCategory' => { 'alertCategoryId' => '5562a5b7-3e32-eb11-a814-000d3a23980a' } } }]
```
peterdavidhamilton, line 179: Did you want to get this uuid into a constant too?

tonyheadford (commented):
Could we have a means to set the induction data to a particular state for specs and maybe with a predetermined TRN/DoB?  I'm thinking more for when we want to test adding ECTs and Mentors and checking eligibility and that kind of thing (on ECF that was the magic date thing really - being able to specify the induction start date via magic DoB and then checking the participant has been added to the correct contract year and schedule for example)

peteryates:
Yeah - we can use `#override_data_for_special_trns` to do that, it could be extended to do anything given a TRN, date of birth and national insurance number.

peteryates:
The intention is to use `TRS::TestAPIClient` in tests. It can be set up in any required state by passing values into its initializer. There should be no need to use TRN/dates/national insurance numbers in the specs.

In the `TRS::FakeAPIClient` (used in the app for manual testing purposes) we can define all the special cases we need to in `#override_data_for_special_trns`.

tonyheadford (approved):
Nice!

PR #917, opened 2025-07-13
Remove the debugging screenshot from parity check spec

peteryates:
This file kept creeping into my commits when it's saved to the root of the app, moving it to the tmp dir means it'll be ignored by git.


ethax-ross:
The screenshot has been accidentally committed, would you mind just removing that from the spec, or I’ll do it when I’m back 🙏

peteryates:
Ah I'll update this PR to remove it - wondered if it was there for a reason

ethax-ross (approved):
Thanks Pete 👍

PR #920, opened 2025-07-14
Add `rspec` tag to fake API data-saving specs

peteryates:
This will allow developers to run the specs without having to set up and configure Redis/Valkey locally by setting the contents of `.rspec-local` to:

```bash
--tag ~redis
```

PR #924, opened 2025-07-14
Add bang to reset teacher induction method

peteryates:
Add a exclamation mark to the `#reset_teacher_induction!` method in the TRS API client so it's consistent with its siblings.

Also while I was here I noticed the `#reopen_teacher_induction!` method didn't exist in the test or fake clients, so I added them.

### Context

- **Add a ! to #reset_teacher_induction**
- **Add missing #reopen_teacher_induction! to fake api client**
- **Rename test API client contexts file**


.rspec-local
```diff
@@ -0,0 +1 @@
+--tag ~redis
```
avinhurry, line 1: Did you intend to commit this? 

It was added to `.gitignore` in #920, so I assume it’s meant to be local-only.

avinhurry (approved):
1 question about the `rspec-local` file in line, otherwise looks good!

.rspec-local
peteryates: Ah damn well caught. Yep ignored in one branch and committed in the other 😭

PR #943, opened 2025-07-16
Add induction records analysis

leoapost:
### Context

Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/1269

This PR adds functionality to analyze participants' induction records grouped by lead providers, to identify timeline gaps, overlaps, and data quality issues in ECF participant induction records that we will need to migrate over.

- Add service class to orchestrate the analysis
- Add class to export the results in csv
- Add class to print the results
- Add model associations to `Migration::InductionRecord` required for the analysis

### Guidance to review

1. Get a list of participants you want to analyse i.e. participants with multiple induction records with blank end dates
```
participant_profiles = ::Migration::ParticipantProfile.joins(:induction_records).where(induction_records: { end_date: nil }).group("participant_profiles.id").having("COUNT(induction_records.id) > 1")
```
2. Run the analysis, print, export and keep the results in a var
```
results = Migration::AnalyseInductionSequences.call(participant_profiles, verbose: true, export: true)
```
The printed results look like this:
Screenshot 2025-07-16 at 10 13 36

### DIFF
--- a/app/models/migration/induction_record.rb
+++ b/app/models/migration/induction_record.rb
@@ -6,5 +6,11 @@ module Migration
     belongs_to :mentor_profile, class_name: "Migration::ParticipantProfile"
     belongs_to :schedule
     belongs_to :preferred_identity, class_name: "Migration::ParticipantIdentity"
+
+    has_one :school_cohort, through: :induction_programme
+    has_one :school, through: :school_cohort
+    has_one :partnership, through: :induction_programme
+    has_one :delivery_partner, through: :partnership
+    has_one :lead_provider, through: :partnership
   end
 end
--- a/app/services/migration/induction_sequence_exporter.rb
+++ b/app/services/migration/induction_sequence_exporter.rb
@@ -16,45 +16,16 @@ module Migration
 
       Rails.logger.info("InductionSequenceExporter: Exporting analysis to CSV: #{file_path}")
 
-      CSV.open(file_path, "w", write_headers: true, headers: csv_headers) do |csv|
-        @results.each do |result|
-          write_result_to_csv(csv, result)
-        end
+      data = @results.flat_map do |result|
+        result[:provider_periods].map { |period| build_csv_row(result, period) }
+      end
+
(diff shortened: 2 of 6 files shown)


app/services/migration/induction_sequence_exporter.rb
```diff
@@ -0,0 +1,112 @@
...
+        @results.each do |result|
+          write_result_to_csv(csv, result)
+        end
+      end
+
+      Rails.logger.info("InductionSequenceExporter: Export completed: #{file_path}")
+      file_path.to_s
+    end
+
+    def csv_headers
```
peteryates, line 29: Because there are lots of columns here it might be easier (if a little less efficient) to create an array of hashes and pass them in one at a time, something like this:

```ruby
data = [{a: "b", c: "d", e: "f"}, {a: "b", c: "d", e: "f"}, {a: "b", c: "d", e: "f"}]

puts CSV.generate(headers: data.first.keys, write_headers: true) { |csv| data.each { |r| csv << r } }
```

Otherwise it's really tricky to tell if they line up correctly!

app/models/migration/induction_record.rb
```diff
@@ -6,5 +6,11 @@ class InductionRecord < Migration::Base
     belongs_to :mentor_profile, class_name: "Migration::ParticipantProfile"
     belongs_to :schedule
     belongs_to :preferred_identity, class_name: "Migration::ParticipantIdentity"
+
+    has_one :school_cohort, through: :induction_programme
+    has_one :school, through: :school_cohort
+    has_one :partnership, through: :induction_programme
+    has_one :delivery_partner, through: :partnership
+    has_one :lead_provider, through: :partnership
```
peteryates, line 14: Can we add specs for these?

leoapost: The test strategy for migration models is a bit unclear, and coverage is pretty minimal. I think that's ok for now since these models mainly just fetch ECF records and a few associations - there's not much business logic involved, so the risk is likely low.

Nonetheless, I have added some tests to ensure coverage for this model, and we can follow up if we feel we need better overall coverage.

PR #944, opened 2025-07-16
Admins can add induction periods for Teachers without data

cpjmcquillan:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/1891

Before, Admins could only add an induction period for a Teacher if they already had some induction period data.

This meant support were unable to add induction periods for "empty" Teachers. "Empty" Teachers can happen when the only induction period is deleted or they are imported from TRS.

This adds a link to the teacher page when there are no induction periods.

As part of this change, a bug was fixed where the induction period form was using traning programme choices instead of induction period choices. It looks like this was introduced in #765 (specifically
2d53a517accd4933f502a5dfcb1011069812ff03).

### Guidance to review

When logged in as an admin user:

- [ ] Add an induction period is available on records where there is no other induction period data
- [ ] I am able to add a closed induction period to an "empty" teacher record (i.e. start date and end date)
- [ ] I am able to add an open induction period to an "empty" teacher record (i.e. start date, no end date)
- [ ] The service updates the induction start date in TRS to the start date of the added induction period
- [ ] The service updates the induction status in TRS to "in progress"

### DIFF
--- a/app/views/admin/induction_periods/new.html.erb
+++ b/app/views/admin/induction_periods/new.html.erb
@@ -33,7 +33,7 @@
           width: 4,
           inputmode: "numeric" %>
 
-      <%= f.govuk_collection_radio_buttons :induction_programme, training_programme_choices, :identifier, :name, legend: { text: 'Induction programme' }  %>
+      <%= f.govuk_collection_radio_buttons :induction_programme, induction_programme_choices, :identifier, :name, legend: { text: 'Induction programme' }  %>
 
       <%= f.submit "Save", class: "govuk-button" %>
     <% end %>
--- a/app/views/admin/teachers/show.html.erb
+++ b/app/views/admin/teachers/show.html.erb
@@ -23,4 +23,10 @@ end %>
   <%= govuk_link_to('View change history', admin_teacher_timeline_path(@teacher)) %>
 </p>
 
+<% if @teacher.induction_periods.none? %>
+  <p class="govuk-body">
+    <%= govuk_link_to("Add an induction period", new_admin_teacher_induction_period_path(@teacher)) %>
+  </p>
+<% end %>
+
 <%= govuk_warning_text(text: "Some of this teacher's records could not be migrated") if @teacher.has_migration_failures? %>
--- a/spec/features/admin/teachers/add_induction_period_spec.rb
+++ b/spec/features/admin/teachers/add_induction_period_spec.rb
@@ -1,0 +1,160 @@
+describe "Admin adds an induction period" do
+  before { sign_in_as_dfe_user(role: :admin) }
+
+  context "when the teacher has no induction periods" do
+    it "adds a closed induction period" do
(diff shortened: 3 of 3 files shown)


spec/features/admin/teachers/add_induction_period_spec.rb
cpjmcquillan, line 1: I never write tests like this usually, so I've done my best 😆🥒

app/views/admin/induction_periods/new.html.erb
```diff
@@ -33,7 +33,7 @@
           width: 4,
           inputmode: "numeric" %>
 
-      <%= f.govuk_collection_radio_buttons :induction_programme, training_programme_choices, :identifier, :name, legend: { text: 'Induction programme' }  %>
+      <%= f.govuk_collection_radio_buttons :induction_programme, induction_programme_choices, :identifier, :name, legend: { text: 'Induction programme' }  %>
```
cpjmcquillan, line 36: As far as I can tell, this wasn't intentional 🤔

The request spec used `fip` so never failed, but there was never an associated feature spec.

cc @peterdavidhamilton

peterdavidhamilton: Good catch. On it.

peterdavidhamilton: https://github.com/DFE-Digital/register-early-career-teachers-public/pull/946

app/views/admin/teachers/show.html.erb
```diff
@@ -23,4 +23,10 @@ end %>
   <%= govuk_link_to('View change history', admin_teacher_timeline_path(@teacher)) %>
 </p>
 
+<% if @teacher.induction_periods.none? %>
+  <p class="govuk-body">
+    <%= govuk_link_to("Add an induction period", new_admin_teacher_induction_period_path(@teacher)) %>
```
peterdavidhamilton, line 28: We could update the view spec with two contexts, with and without `induction_period`, and get this covered there too.

spec/features/admin/teachers/add_induction_period_spec.rb
cpjmcquillan: I've tried to follow the cucumber-style (`defining_descriptive_methods`, and `Given, When, Then`) we're using in feature specs, but also tried to tell as much of a story as I could.

Passing the teacher and appropriate body names explicitly as arguments made sense to me because we're asserting on them. This way there aren't any [Mystery Guests](http://xunitpatterns.com/Obscure%20Test.html#Mystery%20Guest).

I find `let` typically makes both those things more difficult to achieve in my tests.

If we're all in on `let` though and me not using it makes people's lives more difficult, I can easily make adjustments!

peterdavidhamilton: You've achieved that; it would be lacking something if it didn't start with `Given`. I think we have a mix of styles but, this is true 🥒

PR #948, opened 2025-07-16
[1767] Add release note tags functionality

leandroalemao:
### Context

Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/1767

Opening a new PR as https://github.com/DFE-Digital/register-early-career-teachers-public/pull/942 's review app was broken.

When release notes are created we need to add labels to them to show their main purpose (like has been done for 2025 release notes in ECF1).

- Add release note tags functionality

Screenshot 2025-07-17 at 10 41 17

### Guidance to review

[Review app](https://cpd-ec2-review-948-web.test.teacherservices.cloud/api/guidance/release-notes)


app/components/api/guidance/release_notes/tags_component.rb
```diff
@@ -0,0 +1,48 @@
...
+
+          render_tags
+        end
+
+      private
+
+        def render_tags
+          sorted_tags = tags.sort_by { |tag| TAG_MAPPINGS.keys.index(tag) }
+
+          %(<div class="release-notes-tags"><div class="tag-group">#{sorted_tags.map { |tag| render_tag(tag) }.join}</div></div>).html_safe
```
peteryates, line 36: Now we're in a component we can use Rails' tag helpers to make this a bit more concise:


```suggestion
          tag.div(class: "release-notes-tags") do
            tag.div(safe_join(sorted_tags.map { |tag| render_tag(tag) }, "\n"), class: "tag-group")
          end 
```

app/views/api/guidance/release_notes.yml
```diff
@@ -20,6 +23,9 @@
 
 - title: View financial statements in test environment
   date: 2025-06-23
+  tags:
+    - "#sandbox-release"
+    - "#new-feature"
```
peteryates, line 28: I'd be tempted to drop the `#` prefix here, then the `.tr("#", "")` above can be removed and the prefix from each `TAG_MAPPING` entry.

app/components/api/guidance/release_notes/tags_component.rb
```diff
@@ -0,0 +1,48 @@
...
+
+        def initialize(tags:)
+          @tags = tags
+        end
+
+        def all_tags
+          return if tags.blank?
+
+          render_tags
+        end
```
peteryates, line 29: I think this can be replaced with [render?](https://viewcomponent.org/guide/conditional_rendering.html), something like:

```
def render?
  tags.any?
end
```

app/assets/stylesheets/api_guidance.scss
```diff
@@ -57,3 +57,14 @@ $x-govuk-masthead-text-colour: govuk-colour("white");
...
+
+.release-notes-tags {
+  .tag-group {
+    margin: -15px 0 20px;
+
+    .govuk-tag {
+      font-size: 1em;
+      margin: 0 10px 8px 0;
+    }
+  }
```
peteryates, line 69: Providing there's a new line between the govuk tags this can be simplified to just use line height

```
.release-notes-tags {
  .tag-group {
    line-height: 2em;
  }
}
```

Image

peteryates (approved):
Looks great, thanks @leandroalemao!

Left some really minor comments.

app/assets/stylesheets/api_guidance.scss
leandroalemao: Tks @peteryates I've followed all your comments but left `margin: 0 10px 8px 0;` to give a bit more space between tags.. it was too tight one next to the other in my opinion.. tks

app/components/api/guidance/release_notes/tags_component.rb
```diff
@@ -0,0 +1,54 @@
...
+        class UnknownTagError < StandardError; end
+
+        TAG_MAPPINGS = {
+          "breaking-change" => "red",
+          "new-feature" => "green",
+          "bug-fix" => "yellow",
+          "new-field" => "turquoise",
+          "data-update" => "blue",
+          "contract-closure" => "orange",
+          "new-course" => "pink",
```
claire-hughez, line 14: When would this one be used? @mark-avery-dfe @andycrawshaw @shahadRah

app/assets/stylesheets/api_guidance.scss
mark-avery-dfe: @leandroalemao, is it possible to reduce the gap between the release note heading and the tags + add some space between the tags and the copy?

mark-avery-dfe:
> Looks good to me!
> 
> I was intrigued why we have 'new course' and I do have a contenty question for Mark on 'new feature', as statements isn't really a new feature it's just been rebuilt, but we can tweak those things later if we want.
> 
> I think we probably should allow for these labels on the frontpage too, so maybe we should add them to the release notes summary dev ticket in next sprint. What do you think @mark-avery-dfe ?

'New course' is an NPQs tag, so we don't need it in RECT. 
Good point re. 'New feature'. This should only be reserved for genuinely new, err, features. 
I agree that we should have the tags on the front/landing page.

claire-hughez:
Can I suggest then....
1. We remove the new feature tag from the most recent release note
2. We change the pink release tag to 'Removed field' maybe, and add that to the most recent release note?

app/components/api/guidance/release_notes/tags_component.rb
claire-hughez, line 14:
```suggestion
          "removed_field" => "pink",
```

app/views/api/guidance/release_notes.yml
```diff
@@ -20,6 +23,9 @@
 
 - title: View financial statements in test environment
   date: 2025-06-23
+  tags:
+    - sandbox-release
+    - new-feature
```
claire-hughez, line 28:
```suggestion
    - removed_field```

Comment (claire-hughez) [inline app/views/api/guidance/release_notes.yml:28]:
```suggestion
    - removed-field
```

cwrw (approved):

PR #953, opened 2025-07-16
Do not return schools in the query when contract period param is invalid

leandroalemao:
### Context

In ECF1, we don't return schools in the query when cohort does not exist.

Return no schools when the `contract_period_id` sent in the params is from an invalid/non-existent `ContractPeriod`.


spec/factories/contract_period_factory.rb
```diff
@@ -10,5 +10,9 @@
     initialize_with do
       ContractPeriod.find_or_create_by(year:)
     end
+
+    trait :consecutive_years do
```
cwrw, line 14: why do we need an extra trait? can we not use contract period as the setup is similar without the rest?

cwrw (approved):

PR #959, opened 2025-07-17
Removed fields we don't need in BQ

t-phillips-dfe:
We need to remove these fields from hidden_pii as they are also in the blocklist


peteryates:
Tested locally and all looks good :+1:

peteryates (approved):

PR #962, opened 2025-07-17
Clear User OTP tokens when copying data from prod to migration

cwrw:
### Context
On every data copy from prod to migration, we also copy over Users and their encrypted OTP tokens. This causes errors when trying to access the users even in the console.

When refreshing migration, add a step to clear OTP tokens on all users to avoid doing manually.

### Guidance to review
@tonyheadford suggested using:
```
ActiveRecord::Encryption.without_encryption do
  my_user.update!(otp_secret: nil)
end
```

It seems using `update_all` is ok and doesn't need the without encryption block.

I'll test this on migration before merging


tonyheadford (approved):
Nice!

ethax-ross (approved):

PR #973, opened 2025-07-18
Simplify `Teachers::Details::PastInductionPeriodsComponent`

cpjmcquillan:
As highlighted by Avin in #936, we can simplify the logic for rendering action links in the summary list.

This refactors the component and adds some tests for when editing is enabled and when it isn't.

### DIFF
--- a/app/components/teachers/details/past_induction_periods_component.rb
+++ b/app/components/teachers/details/past_induction_periods_component.rb
@@ -26,16 +26,23 @@ module Teachers::Details
       end
     end
 
-    def edit_link(period)
-      return unless enable_edit
+    delegate :govuk_link_to, to: :helpers
 
-      helpers.govuk_link_to('Edit', helpers.edit_admin_teacher_induction_period_path(teacher_id: teacher.id, id: period.id), no_visited_state: true)
-    end
+    def actions_for(period)
+      return [] unless enable_edit
 
-    def delete_link(period)
-      return unless enable_edit
-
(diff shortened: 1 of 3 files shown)


app/components/teachers/details/past_induction_periods_component.rb
```diff
@@ -26,16 +26,24 @@ def induction_period_programme(period)
...
-    def edit_link(period)
-      return unless enable_edit
-
-      helpers.govuk_link_to('Edit', helpers.edit_admin_teacher_induction_period_path(teacher_id: teacher.id, id: period.id), no_visited_state: true)
-    end
-
-    def delete_link(period)
-      return unless enable_edit
-
-      helpers.govuk_link_to('Delete', helpers.confirm_delete_admin_teacher_induction_period_path(teacher_id: teacher.id, id: period.id), method: :get, class: 'govuk-link--destructive', no_visited_state: true)
```
cpjmcquillan, line 38: These are links, and the route is defined as a `GET`, so the `method: :get` felt superfluous.

Perhaps I'm missing something, though 🤔

app/components/teachers/details/past_induction_periods_component.rb
```diff
@@ -26,16 +26,24 @@ def induction_period_programme(period)
...
+      [
+        govuk_link_to(
+          'Edit',
+          helpers.edit_admin_teacher_induction_period_path(teacher, period),
+          no_visited_state: true
+        ),
+        govuk_link_to(
+          'Delete',
+          helpers.confirm_delete_admin_teacher_induction_period_path(teacher, period),
+          class: 'govuk-link--destructive',
```
peteryates, line 43: Not sure what `govuk-link--destuctive` is doing here, can't see any mention of it in `govuk-frontend` 🤔

peteryates (approved):
Looks good - think we can drop the destructive class though.

PR #974, opened 2025-07-18
Upgrade to Playwright 1.53.0

peteryates:
Note, currently not quite working, YusukeIwaki/playwright-ruby-client#343


peteryates:
It looks like the `have_url` matcher is broken in playwright-ruby-client, here's the first failing spec.

https://github.com/DFE-Digital/register-early-career-teachers-public/blob/a2021b5d817ee9d359fc3d0dbef3d2613d3323a1/spec/features/admin_manages_ect_extensions_spec.rb#L20

cpjmcquillan (approved):

PR #982, opened 2025-07-21
Add training programme to training periods

peteryates:
This implements the first part of DFE-Digital/register-ects-project-board#2013 by:

* adding `training_programme` to `TrainingPeriod`
* updating the ECT and mentor registration flows so they write the training programme to the `training_period`
* adding validation to `TrainingPeriod` which only allows 

What it doesn't do (and will be addressed in a later PR):

* remove the `training_programme` from `ECTAtSchoolPeriod`
* change the pages that show teacher info over to use the training programme from the `TrainingPeriod` 
* update the `Schools::Query` object so it retrieves teh `training_programme` from the `training_periods` rather than the `ect_at_school_periods`


spec/migration/migrators/training_period_spec.rb
peteryates, line 1: Might need a hand with this tomorrow @tonyheadford - don't want to dive in without guidance!

spec/services/training_periods/create_spec.rb
```diff
@@ -67,4 +69,34 @@
       }.to raise_error(ArgumentError, /Unsupported period type/)
     end
   end
+
+  describe '.school_led' do
+    let(:period) { FactoryBot.create(:ect_at_school_period) }
+
+    it 'calls new with the school_led arguments' do
+      allow(TrainingPeriods::Create).to receive(:new).with(any_args).and_call_original
```
avinhurry, line 77: Is it worth being specific about the args in here since we do care about them?

We could use something like: 

```
expect(TrainingPeriods::Create).to receive(:new).with(
  period:,
  started_on:,
  training_programme: 'school_led'
).and_call_original
```

That way the test a bit tighter.

app/models/training_period.rb
```diff
@@ -87,4 +95,10 @@ def at_least_expression_of_interest_or_school_partnership_present
 
     errors.add(:base, 'Either expression of interest or school partnership required')
   end
+
+  def only_provider_led_mentor_training
+    if mentor_at_school_period.present? && school_led_training_programme?
```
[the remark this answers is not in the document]
peteryates: Yeah, lots of teachers will end up having both and I think in some circumstances, a few will have both at once. There's no rule that mandates a mentor has completed their own training.

spec/services/training_periods/create_spec.rb
peteryates: It's a bit tighter but I don't think it makes any difference with regards to the assertion.

Ill get rid of the `with(any_args)` as that's not adding anything to it either and just go with this, so it's clear:

```
allow(TrainingPeriods::Create).to receive(:new).and_return(true)
```

avinhurry (approved):
I haven't looked into the migration changes in detail, but everything looks good from the schools side.

Just a reminder to clear out any `training_periods` in sandbox/staging before merging, since the new column is non nullable.

PR #985, opened 2025-07-22
Always assign an induction programme to Induction Periods

cpjmcquillan:
We validate an `InductionPeriod` has an `induction_programme` and we enforce this at the database level with a non-null constraint.

Before #765 this wasn't an issue, because we always assigned an induction programme.

Since then, we have only been assigning an induction programme if `ENABLE_BULK_CLAIM` is false. Otherwise, we have been assigning only a `training_programme`.

This results in validation errors wherever we try to create a new `InductionPeriod` (e.g Admins, ABs).

Not good!

This overrides the `training_programme` setter to ensure that even when `ENABLE_BULK_CLAIM` is true we still assign an `induction_programme`. We are using the `PROGRAMME_MAPPER` to do this.

Now, users can create induction periods with the flag enabled and disabled without error.

This is one approach, but I'm not sure whether it makes the most sense or not.

Alternatively we could only validate `induction_programme` when the flag is disabled. To do this we would also need to remove the database constraint.

### DIFF
--- a/app/models/induction_period.rb
+++ b/app/models/induction_period.rb
@@ -57,6 +57,14 @@ class InductionPeriod < ApplicationRecord
     super || ::PROGRAMME_MAPPER[induction_programme]
   end
 
+  def training_programme=(value)
+    if Rails.application.config.enable_bulk_claim
+      self.induction_programme = ::PROGRAMME_MAPPER[value]
+    end
+
+    super
+  end
+
 private
 
   # Ensure admin users inserting new induction periods include end dates.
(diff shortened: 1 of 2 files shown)


peterdavidhamilton:
Currently, these two services do the mapping rather than a setter:
- app/services/admin/update_induction_period.rb (is moving out of this namespace in a new PR)
- app/services/appropriate_bodies/claim_an_ect/register_ect.rb

I agree it makes sense to consolidate and just do it with the getter.

PR #988, opened 2025-07-22
Allow ABs to edit active induction periods

peterdavidhamilton:
### Context

Only Admins are currently able to edit induction periods. ABs have requested this functionality for active periods.

- separate editing flag, adding deleting flag, in component
- use mode to change context in component
- add controllers and routes for ABs editing induction periods
- add some view and feature specs
- copy view templates from admins to appropriate bodies
- adopt govuk view helpers where possible
- move update induction service out of admin namespace to be shared
- use consistent pattern in induction controllers
- add new event heading for AB edits
- align component structure and tests

### Guidance to review

Use an AB persona and navigate to an ECT with an active induction

cpd-ec2-review-988-web test teacherservices cloud_appropriate-body_teachers_28

### DIFF
--- a/spec/requests/appropriate_bodies/induction_periods_spec.rb
+++ b/spec/requests/appropriate_bodies/induction_periods_spec.rb
@@ -1,14 +1,13 @@
 RSpec.describe 'Appropriate body editing an active induction period', type: :request do
   include_context 'sign in as non-DfE user'
 
-  let(:teacher) { FactoryBot.create(:teacher, trs_qts_awarded_on: 1.year.ago) }
   let(:appropriate_body) { FactoryBot.create(:appropriate_body) }
+  let(:teacher) { FactoryBot.create(:teacher, trs_qts_awarded_on: 1.year.ago) }
+  let!(:induction_period) do
+    FactoryBot.create(:induction_period, :active, teacher:, started_on: 9.months.ago)
+  end
 
   describe "GET /appropriate-body/teachers/:teacher_id/induction-periods/:id/edit" do
-    let(:induction_period) do
-      FactoryBot.create(:induction_period, teacher:, started_on: 9.months.ago, finished_on: 6.months.ago)
-    end
-
     it "returns success" do
       get edit_ab_teacher_induction_period_path(induction_period.teacher, induction_period)
       expect(response).to be_successful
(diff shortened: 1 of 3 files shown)


spec/components/teachers/details_component_spec.rb
```diff
@@ -22,64 +22,36 @@
...
-        expect(result.to_html).not_to be_empty
-      end
-    end
+  it "has slots" do
+    expect(component).to respond_to(:personal_details)
+    expect(component).to respond_to(:induction_summary)
+    expect(component).to respond_to(:current_induction_period)
+    expect(component).to respond_to(:itt_details)
+    expect(component).to respond_to(:past_induction_periods)
+    expect(component).to respond_to(:induction_outcome_actions)
```
cpjmcquillan, line 31: Is the intention of these tests to verify the API of the component?

I wonder how valuable they are (i.e empty slots would go ✅), but the alternative (e.g asserting against the content of each) feels cumbersome.

I don't feel like I've ever landed on a perfect way to test these sorts of things in components, so interested to see how this feels for us! 💡

peterdavidhamilton: I guess my thinking was the previous assertions only checked some slots, and did so by asserting content and an associated teacher. This pattern does work if we add the missing child components and also duplicates specs in their dedicated tests which require multiple contexts. My goal was to both lightly test and document that these are the available nested components and any mode based logic therein.

spec/requests/appropriate_bodies/induction_periods_spec.rb
```diff
@@ -0,0 +1,281 @@
...
+            { induction_period: { started_on: 11.months.ago } }
+          end
+
+          it "enqueues BeginECTInductionJob" do
+            patch(ab_teacher_induction_period_path(teacher, earliest_period), params:)
+            expect(BeginECTInductionJob).to have_received(:perform_later).with(
+              trn: teacher.trn,
+              start_date: params[:induction_period][:started_on].to_date
+            )
+          end
```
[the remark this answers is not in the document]
peterdavidhamilton: Following the pattern from the existing admin version from which this is taken and ran out of steam refactoring. I do agree.

> There is a gem for that

As the saying goes

> There is a matcher for that

PR #998, opened 2025-07-23
Add `api_updated_at` into school API serialiser logic

leandroalemao:
### Context

New field called `api_updated_at` was added in https://github.com/DFE-Digital/register-early-career-teachers-public/pull/971

- Add `api_updated_at` into school serialiser logic

### DIFF
--- a/app/serializers/school_serializer.rb
+++ b/app/serializers/school_serializer.rb
@@ -17,7 +17,7 @@ class SchoolSerializer < Blueprinter::Base
       expressions_of_interest?(school)
     end
     field :created_at
-    field :updated_at
+    field(:api_updated_at, name: :updated_at)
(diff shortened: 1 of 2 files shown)


app/serializers/school_serializer.rb
```diff
@@ -17,7 +17,12 @@ class AttributesSerializer < Blueprinter::Base
       expressions_of_interest?(school)
     end
     field :created_at
-    field :updated_at
+    field(:updated_at) do |school|
+      [
+        school.api_updated_at,
```
cwrw, line 22: can we only use `api_updated_at` please? as anything required to be updated is pushed there regardless?

cwrw (approved):
🙌

PR #1000, opened 2025-07-23
Fix Sentry config

cpjmcquillan:
In #981, I missed the `.rails`.

We didn't catch it because Sentry is disabled in review apps, but it blew up on Staging.

This fixes things.

Verified locally with `ENABLE_SENTRY=true bin/dev` before/after the change.

### DIFF
--- a/config/initializers/sentry.rb
+++ b/config/initializers/sentry.rb
@@ -7,6 +7,6 @@ if Rails.application.config.enable_sentry
     config.release = ENV["COMMIT_SHA"]
     config.traces_sample_rate = 0.1
     config.profiles_sample_rate = 0.1
-    config.active_job_report_on_retry_error = true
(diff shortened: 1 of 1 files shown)


config/initializers/sentry.rb
```diff
@@ -7,6 +7,6 @@
     config.release = ENV["COMMIT_SHA"]
     config.traces_sample_rate = 0.1
     config.profiles_sample_rate = 0.1
-    config.active_job_report_on_retry_error = true
+    config.rails.active_job_report_on_retry_error = true
```
peteryates, line 10: Looks good to me [based on what's in the gem](https://github.com/getsentry/sentry-ruby/blob/818b6c0f6c2c717a697ee67ecaf0ac78feea842f/sentry-rails/lib/sentry/rails/active_job.rb#L92).

PR #1002, opened 2025-07-23
[1123] admin console organisations delivery partners

mooktakim:
### Context

Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/1123

* New delivery partners sections in organisations tab

### DIFF
--- a/app/services/delivery_partners/search.rb
+++ b/app/services/delivery_partners/search.rb
@@ -1,19 +1,17 @@
 module DeliveryPartners
   class Search
     def initialize(query_string = nil)
-      @scope = DeliveryPartner
-
       @query_string = query_string
     end
 
     def search
-      query = if @query_string.blank?
-                @scope.all
-              else
-                @scope.where("name ILIKE ?", "%#{@query_string}%")
-              end
+      scope = DeliveryPartner.all
 
-      query.order(name: 'asc')
+      if @query_string.present?
+        scope = scope.where("name ILIKE ?", "%#{@query_string}%")
+      end
+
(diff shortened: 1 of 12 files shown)


app/services/delivery_partners/search.rb
```diff
@@ -0,0 +1,19 @@
...
+      query = if @query_string.blank?
+                @scope.all
+              else
+                @scope.where("name ILIKE ?", "%#{@query_string}%")
+              end
+
+      query.order(name: 'asc')
+    end
+  end
+end
```
avinhurry, line 19: Does this work? It simplifies the code by removing an unnecessary instance variable.

```suggestion
module DeliveryPartners
  class Search
    def initialize(query_string = nil)
      @query_string = query_string
    end

    def search
      scope = DeliveryPartner.all

      if @query_string.present?
        scope = scope.where("name ILIKE ?", "%#{@query_string}%")
      end

      scope.order(name: :asc)
    end
  end
end
```

app/views/admin/delivery_partners/index.html.erb
```diff
@@ -0,0 +1,41 @@
...
+        results found for "<%= params[:q] %>"
+      </p>
+    <% end %>
+
+    <%= govuk_table do |table|
+      table.with_head do |head|
+        head.with_row { |row| row.with_cell(text: "Delivery partner") }
+      end
+
+      table.with_body do |body|
```
avinhurry, line 32: Do we want to hide this table and return `No results found` if there are no results?

PR #1003, opened 2025-07-23
[1441] Allow schools to re use lead provider and delivery partner choices for a new ECT where possible

avinhurry:
#### Context

This PR introduces support for the "rollover" concept from ECF1, where schools can reuse previously selected training programme choices (e.g. lead provider and delivery partner) when registering a new ECT - but only when those choices are still valid for the current contract period.

#### Screenshots

image

####  Changes in this PR

- **Check Your Answers (CYA) Logic**  
  - The Delivery Partner row now appears on the CYA page **only if**:
    - `use_previous_ect_choices` is `true`, and
    - there is a confirmed school partnership between the previous lead provider and the school for the relevant contract period.
  - This ensures we only show delivery partner info when it reflects an actual partnership.

- **Use Previous Choices View**
  - Now conditionally shows the lead provider and delivery partner from the previous ECT if a confirmed partnership exists.
  - Falls back to the expression of interest lead provider if no confirmed partnership is found.

- **Wizard Logic**
  - Adds `#lead_provider_has_confirmed_partnership_for_contract_period?` to encapsulate the above check.
  - Adds `#previous_eoi_lead_provider_name` to support fallback scenarios.

- **Specs**
  - FactoryBot updates to support `contract_period:` in test setup.

#### ❗️What this **doesn’t** do (by design)

- This PR **does not persist** the delivery partner or school partnership in the database.
  - The school never explicitly chooses a delivery partner - it’s inferred based on the confirmed partnership with the lead provider.
  - Persisting or recreating partnerships is part of separate ongoing work.

### DIFF
--- a/app/views/schools/register_ect_wizard/_use_previous_ect_choices.html.erb
+++ b/app/views/schools/register_ect_wizard/_use_previous_ect_choices.html.erb
@@ -4,29 +4,43 @@
 
 <p class="govuk-body">You used these programme choices the last time you registered an ECT.</p>
 
-<%= govuk_summary_list do |summary_list|
-  summary_list.with_row do |row|
-    row.with_key(text: 'Appropriate body')
-    row.with_value(text: @school.last_chosen_appropriate_body_name)
-  end
+<%= govuk_summary_list do |summary_list| %>
+  <% summary_list.with_row do |row| %>
+    <% row.with_key(text: 'Appropriate body') %>
+    <% row.with_value(text: @school.last_chosen_appropriate_body_name) %>
+  <% end %>
 
-  summary_list.with_row do |row|
-    row.with_key(text: 'Training programme')
-    row.with_value(text: training_programme_name(@school.last_chosen_training_programme))
-  end
+  <% summary_list.with_row do |row| %>
+    <% row.with_key(text: 'Training programme') %>
+    <% row.with_value(text: training_programme_name(@school.last_chosen_training_programme)) %>
+  <% end %>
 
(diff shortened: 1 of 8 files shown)


app/views/schools/register_ect_wizard/_use_previous_ect_choices.html.erb
```diff
@@ -4,29 +4,44 @@
...
+        <% row.with_key(text: 'Delivery partner') %>
+        <% row.with_value(text: @ect.previous_delivery_partner_name) %>
+    <% end %>
+    <% else %>
+      <% summary_list.with_row do |row| %>
+        <% row.with_key(text: 'Lead provider') %>
+        <% row.with_value(text: @ect.previous_eoi_lead_provider_name) %>
+      <% end %>
+    <% end %>
+  <% end %>
```
peteryates, line 35: The indentation's slightly off here (line 28). I think the erb tags make it hard to see and I don't think they're necessary for this as we're just calling stuff on the component (and not using blocks or complex content).

I think this could be simplified a bit by calling it like this:

```ruby
  <%
    if @school.provider_led_training_programme_chosen?
      if @ect.lead_provider_has_confirmed_partnership_for_contract_period?(@school)
        summary_list.with_row do |row|
          row.with_key(text: 'Lead provider')
          row.with_value(text: @ect.previous_lead_provider_name)
        end

        summary_list.with_row do |row|
          row.with_key(text: 'Delivery partner')
          row.with_value(text: @ect.previous_delivery_partner_name)
        end
      else
        summary_list.with_row do |row|
          row.with_key(text: 'Lead provider')
          row.with_value(text: @ect.previous_eoi_lead_provider_name)
        end
      end
    end
  %>
```

PR #1006, opened 2025-07-24
Group Ruby and JS Playwright updates

avinhurry:
## Group Ruby and JS Playwright updates in Dependabot

We can leverage [multi-ecosystem updates](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/configuring-multi-ecosystem-updates) which became available at the [beginning of this month ](https://github.blog/changelog/2025-07-01-single-pull-request-for-dependabot-multi-ecosystem-support/) to ensure that updates to `playwright-ruby-client` (Ruby) and `playwright` (npm) are grouped into a single PR by Dependabot. 

This prevents version mismatches that cause spec failures, such as `PlaywrightMajorVersionMismatch`.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,10 +1,16 @@
 version: 2
+multi-ecosystem-groups:
+  playwright-group:
+    schedule:
+      interval: daily
+    open-pull-requests-limit: 5
 updates:
   - package-ecosystem: bundler
     directory: "/"
     schedule:
       interval: daily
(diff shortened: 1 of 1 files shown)


cpjmcquillan (approved):
Love it!

peteryates:
Do these groups definitely work across package ecosystems? I seem to remember trying this in the past but the Ruby and NPM things wouldn't play nicely together.

PR #1007, opened 2025-07-24
Spike into join table for school-level data

ethax-ross:
### Context

We want to investigate using join tables to store contextual data relative to a school (or other instances). As all the data in our database is at the participant-level, it makes retrieving school-level details to return from the lead provider API complex.

If we can add join tables that maintain this hard to compute data it will make the queries/serializers/services elsewhere in the application much simpler.

- Add Metadata::SchoolLeadProviderContractPeriod

Add a model to encapsulate metadata about a school/lead provider/contract period relationship.

This is school-level data/an aggregation of scoped participant data relative to the school.

Name-spaced under `Metadata` to ensure its not used inappropriately; it is also locked down so that only the `Metadata` namespace/services can updated the metadata models.

- Add service to manage metadata creation/updates

Add a `Metadata::Manager` that can be called from any service in order to create/update metadata for a given object(s).

Add `Metadata::Resolver` which maps a given object to its handler.

Add `Metadata::Handler::School` to update school-specific metadata, currently the `Metadata::SchoolLeadProviderContractPeriod`.

#### Class diagram

```mermaid
classDiagram
    namespace Metadata {
        class SchoolLeadProviderContractPeriod {
            -boolean in_partnership
            -boolean expression_of_interest
            -enum induction_programme_choice
            -datetime updated_at
        }
    }

    class ContractPeriod {
        -integer year
    }

    class School {
        -string name
    }

    class LeadProvider {
        -string name
    }

    SchoolLeadProviderContractPeriod --> School : belongs_to
    SchoolLeadProviderContractPeriod --> LeadProvider : belongs_to
    SchoolLeadProviderContractPeriod --> ContractPeriod : belongs_to

    School --> "0..*" SchoolLeadProviderContractPeriod : has_many
```

#### Service diagram

```mermaid
flowchart TD

Service["Service"]
Manager["Metadata::Manager"]
Resolver["Metadata::Resolver"]

subgraph Metadata::Handler
  HandlerSchool["School"]
  HandlerDeclaration["Declaration"]
  HandlerStatement["Statement"]
end

subgraph Metadata
  SchoolLeadProviderContractPeriod
  LeadProviderDeclaration
end

Service -->|create/update_metadata| Manager
Manager <-->|resolve_handler| Resolver

Manager -->|create/update_metadata| HandlerSchool
Manager -->|create/update_metadata| HandlerStatement
Manager -->|create/update_metadata| HandlerDeclaration

HandlerSchool --> SchoolLeadProviderContractPeriod
HandlerDeclaration --> LeadProviderDeclaration
```

#### General thoughts

I think this pattern would work for us; it relies on the manager being called at the appropriate times the objects, however. If we think we will always be updating the models via services we can call out from the relevant services, but if a model is likely to change via other means (console edits) we would be better to hook into the object lifecycle and call out from there (IMO this is going to be more bullet-proof but will start to muddy our models).

The performance of this solution will depend on the underlying calculations that need to be made. We could always offload some of the work to jobs if it becomes a problem, but then the data may become stale/out dated for a time and lead providers could miss updates.


cwrw (commented):
Thanks @ethax-ross this is very cleanly and beautifully implemented:
- Namespace to show intention
- Locked down to avoid manual updates via the restrict error
- can be cleanly added where required - I think adding to the model for now until services are built is ok, but I can see it being a problem as we'll need to add it to quite a few models to catch everything (at least for schools), I would opt to have it in services instead in the end. There needs to be a lot of enforcement though and I feel like things will slip
- is it safe to assume there are no competing updates/records are safe if added in multiple places
- as you mentioned there might be performance implications - but that depends on how many models are updated, and if done in bulk vs not. We can adapt depending on the use case as it comes. For migration we'll need to disable this

Update after discussing in person:
- There is a suggestion of running a nightly job to perform full resyncs of models and to flag if anything changed to ensure we don't miss anything
- services is the preference
- we can call multiple handlers for the same model if necessary, and look to sequence to avoid any issues

PR #1010, opened 2025-07-24
Update school queries to use the training programme from training periods

peteryates:
### Context

We're moving the `training_programme` field from `ECTAtSchoolPeriod` to `TrainingPeriod` bit by bit.

#982 added the `training_programme` field to `TrainingPeriod` and this change updates the school queries so they use it instead of the one on `ECTAtSchoolPeriod`.

Thankfully it was quite minimal as the query objects had joined the relevant tables already.

Refs DFE-Digital/register-ects-project-board#2013

### DIFF
--- a/app/services/schools/training_programme.rb
+++ b/app/services/schools/training_programme.rb
@@ -40,7 +40,10 @@ module Schools
     def ect_training_programme
       return school.transient_ects_at_school_training_programme if school.respond_to?(:transient_ects_at_school_training_programme)
 
-      ECTAtSchoolPeriod.joins(:training_periods).where(id: ects_expressions_of_interest.pluck(:id) + ect_at_school_periods.pluck(:id)).order(training_programme: :asc).pick('training_periods.training_programme')
+      TrainingPeriod
+        .where(ect_at_school_period_id: ects_expressions_of_interest.pluck(:id) + ect_at_school_periods.pluck(:id))
+        .order(training_programme: :asc)
+        .pick(:training_programme)
     end
 
     def not_yet_known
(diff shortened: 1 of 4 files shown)


app/services/schools/training_programme.rb
```diff
@@ -40,7 +40,7 @@ def mentors_at_school?
     def ect_training_programme
       return school.transient_ects_at_school_training_programme if school.respond_to?(:transient_ects_at_school_training_programme)
 
-      ECTAtSchoolPeriod.where(id: ects_expressions_of_interest.pluck(:id) + ect_at_school_periods.pluck(:id)).order(training_programme: :asc).pick(:training_programme)
+      ECTAtSchoolPeriod.joins(:training_periods).where(id: ects_expressions_of_interest.pluck(:id) + ect_at_school_periods.pluck(:id)).order(training_programme: :asc).pick('training_periods.training_programme')
```
peteryates, line 43: The qualified column name is needed here because for now there's a `training_programme` field in both `training_periods` and `ect_at_school_periods`.

peteryates, line 43: Also this is probably cleaner coming from the `TrainingPeriod` point of view now


```suggestion
      TrainingPeriod
        .joins(:ect_at_school_period)
        .where(ect_at_school_period: { id: ects_expressions_of_interest.pluck(:id) + ect_at_school_periods.pluck(:id) })
        .order(training_programme: :asc)
        .pick(:training_programme)
```

leandroalemao: With no joins maybe better?

```
TrainingPeriod.where(ect_at_school_period_id: ects_expressions_of_interest.pluck(:id) + ect_at_school_periods.pluck(:id)).order(training_programme: :asc).pick(:training_programme)
```

avinhurry (approved):

Issue #941, opened 2025-07-16
Drop unnecessary columns from `appropriate_bodies`


peteryates:
We imported some data from DQT that's not being used. It should be deleted so it doesn't cause confusion.

Fields to delete:

* `dqt_id` - the ID the appropriate body had in the Database for Qualified Teachers, before it was decommissioned
* `local_authority_code` - [the code used to identify the LA the appropriate body belonged to](https://get-information-schools.service.gov.uk/Guidance/LaNameCodes)
* `establishment_number` - another ID used by DfE to identify organisations, used in combination with the `local_authority_code` to produce the 'DfE number' in a `XXX/YYYY` format.

The `local_authority_code` and `establishment_number` were imported by the AB portal's CSV importer but we're not using them now.

cpjmcquillan:
@t-phillips-dfe @consult1815 👋🏼 are you still using `dqt_id` for identifying appropriate bodies in the pipeline?

graciebrennan:
Emma has confirmed the DQT_ID is still in use so we will keep this in place

emily-prudence-dfe:
Just needs dev review

PR #1011, opened 2025-07-24
Remove redundant columns from appropriate bodies

cpjmcquillan:
### Context

https://github.com/DFE-Digital/register-early-career-teachers-public/issues/941

We're not using this data anywhere (we just imported it from DQT as a starting point).

This removes it in the interests of maintenance!

### Guidance to review

~~- [ ] Data pipelines not borked!~~ We're leaving `dqt_id` since the analytics team are still using it to associate historical induction periods to appropriate bodies

### DIFF
--- a/documentation/domain-model.md
+++ b/documentation/domain-model.md
@@ -253,6 +253,7 @@ erDiagram
     datetime created_at
     datetime updated_at
     uuid api_id
+    datetime api_updated_at
   }
   Declaration {
     integer id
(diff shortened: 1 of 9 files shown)


documentation/domain-model.md
```diff
@@ -253,6 +253,7 @@ erDiagram
     datetime created_at
     datetime updated_at
     uuid api_id
+    datetime api_updated_at
```
cpjmcquillan, line 256: This isn't part of this change, but it was missing and was auto-generated!

PR #1014, opened 2025-07-25
Use dashes in Register ECT wizard

### LINKED ISSUE #556: Make `already_active_at_school` route use dashes
All of the ones surrounding it use dashes and it looks weird with underscores.

https://github.com/DFE-Digital/register-early-career-teachers-public/blob/7f0a65b320681bdb06a3b47c1460ca823c5eb5df/config/routes.rb#L117

cpjmcquillan:
### Context

https://github.com/DFE-Digital/register-early-career-teachers-public/issues/556

This path was the only route_with_underscores in the wizard.

Now it-has-dashes-like-the-others.

Closes #556

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -187,7 +187,7 @@ Rails.application.routes.draw do
       get "cannot-register-ect", action: :new
       get "cannot-register-ect-yet", action: :new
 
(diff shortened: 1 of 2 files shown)


peteryates (approved):
Nice, thanks

PR #1015, opened 2025-07-25
Add script to delete failed bulk uploads

peterdavidhamilton:
### Context

13 from initial feature release

Convenience housekeeping script


peterdavidhamilton:
> Worth maybe adding a job that runs occasionally and cleans up failed jobs after a certain period of time?

I'd feel safer not having that at the moment because we never want to see `failed` jobs, it means a dev needs to fix something.

PR #1018, opened 2025-07-25
Use non-root user

vipin-dfe:
### Context
To restrict user to non root and only allow non read permission on directories which require write access

Add a new non-root user

### Guidance to review
Application should work without any errors

### Checklist
- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history
- [x] Tested by running locally


peteryates (approved):
Looks good, definitely safer.

PR #1022, opened 2025-07-28
[2015] - Migrate closed schools on ECF1 with induction records

ltello:
### Context
[Issue](https://github.com/DFE-Digital/register-ects-project-board/issues/2015)

We are importing only open schools from GIAS into RECT.
Similarly, we are comparing only open schools on ECF1 with those in RECT.

However, some closed schools still registered have school partnerships defined.
As those partnerships will need to be migrated into RECT, we need their schools to be migrated as well.

Extend the schools comparision migrator to migrate these school to RECT.

### DIFF
--- a/app/migration/migrators/school.rb
+++ b/app/migration/migrators/school.rb
@@ -63,32 +63,6 @@ module Migrators
       }.all?
     end
 
-    def create_school_from!(ecf_school)
-      GIAS::School.create!(
-        address_line1: ecf_school.address_line1,
-        address_line2: ecf_school.address_line2,
-        address_line3: ecf_school.address_line3,
-        administrative_district_name: ecf_school.administrative_district_name,
-        api_id: ecf_school.id,
-        establishment_number: ecf_school.urn,
-        funding_eligibility: ecf_school.funding_eligibility,
-        induction_eligibility: ecf_school.induction_eligibility,
-        in_england: ecf_school.in_england?,
-        local_authority_code: ecf_school.local_authority_code,
-        name: ecf_school.name,
-        phase_name: ecf_school.school_phase_name,
-        postcode: ecf_school.postcode,
-        primary_contact_email: ecf_school.primary_contact_email,
-        secondary_contact_email: ecf_school.secondary_contact_email,
-        section_41_approved: ecf_school.section_41_approved?,
-        status: ecf_school.status,
-        type_name: ecf_school.school_type_name,
-        ukprn: ecf_school.ukprn,
-        urn: ecf_school.urn,
-        website: ecf_school.school_website
-      ).tap(&:create_school!)
-    end
-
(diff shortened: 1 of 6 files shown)


app/migration/migrators/school.rb
```diff
@@ -67,6 +99,13 @@ def find_gias_school(urn:) = gias_schools.find { |school| school.urn == urn }
 
     def gias_schools = @gias_schools ||= ::GIAS::School.where(urn: self.class.schools.pluck(:urn).sort)
 
+    def migrate_school!(ecf_school)
+      return false if ecf_school.open?
+      return false unless ecf_school.with_induction_records?
```
leoapost, line 104: Something like `return false if ecf_school.induction_records.blank?` feels more intuitive to me, and we could drop the `with_induction_records?` method from the model.

spec/migration/migrators/school_spec.rb
```diff
@@ -41,16 +41,54 @@ def setup_failure_state = create_migration_resource
...
+          expect(rect_school.urn.to_s).to eq(ecf_school.urn)
+          expect(rect_school.website).to eq(ecf_school.school_website)
+        end
       end
 
-      it "adds an error" do
-        expect(data_migration.reload.failure_count).to eq(1)
-        expect(data_migration.migration_failures.count).to eq(1)
-        expect(data_migration.migration_failures.first.failure_message).to eq(":school_missing - School #{ecf_school.urn} (#{ecf_school.name}) missing on RECT!")
+      context "otherwise" do
```
leoapost, line 80: How about changing this to something like
```suggestion
      context "when the school is not closed" do
```

app/migration/migrators/school.rb
```diff
@@ -24,12 +24,18 @@ def self.record_count = schools.count
 
     def self.reset! = nil
 
-    def self.schools = ::Migration::School.includes(:local_authority).eligible_or_cip_only_except_welsh
+    def self.schools
+      ::Migration::School.with(eligible_or_cip_or_with_irs:
+                                 [
+                                   ::Migration::School.includes(:local_authority).eligible_or_cip_only_except_welsh.distinct,
+                                   ::Migration::School.not_open.with_induction_records.distinct
```
leoapost, line 31: Just checking if this also needs to exclude Welsh?

app/migration/migrators/school.rb
ltello: Perfect. I created it before `Migration::School#inductions` was defined. 
Now that we have the associations it's easier to call it and remove the method. 
Thanks.

spec/migration/migrators/school_spec.rb
ltello: I'm changing it to `"when the school is open or has no induction records"`

PR #1027, opened 2025-07-28
Add `api_updated_at` to `DeliveryPartner`

ethax-ross:
### Context

We want to surface the delivery partners relative to a given `updated_since` in the API, but only when certain fields are changed (the ones that we return, serialized).

- Add api_updated_at to DeliveryPartner

Add an `api_updated_at` attribute to `DeliveryPartner` that we can touch when a field we expose over the API is changed (this may be on a related model, not just `DeliveryPartner`).

- Touch DeliveryPartner#api_updated_at on relevant changes

We want to update the `api_updated_at` of a `DeliveryPartner` when a relevant attribute changes that would result in the serialized response being different.

Looking at ECF, for a delivery partner this would be when the `name` of the delivery partner changes, or its relevant `cohort(s)`.

Touch `api_updated_at` when the delivery partner `name` changes or when a new `LeadProviderDeliveryPartnership` is added/removed.

### Guidance to review

Assumes lead provider delivery partnerships are not expected to change their `active_lead_provider` or `delivery_partner` and `ActiveLeadProvider` would never change its `contract_period`.


app/models/lead_provider_delivery_partnership.rb
```diff
@@ -1,9 +1,13 @@
 class LeadProviderDeliveryPartnership < ApplicationRecord
+  include DeclarativeTouch
+
   belongs_to :active_lead_provider
   belongs_to :delivery_partner
   has_many :school_partnerships
   has_many :events
 
+  touch -> { delivery_partner }, on_event: %i[create destroy], timestamp_attribute: :api_updated_at
```
leandroalemao, line 9: I couldn't find in ECF1's "touch" model where we touch a delivery partner when a `Partnership` is created or removed. Is it a new behaviour that we're adding in ECF2? If so, it's ok.. 👍

ethax-ross: My understanding is that we want to touch the `api_updated_at` when any of the serialised response changes, so slightly different from ECF. Is that right @cwrw?

cwrw (approved):
> Assumes lead provider delivery partnerships are not expected to change their active_lead_provider or delivery_partner and ActiveLeadProvider would never change its contract_period.

Good assumptions to make :)

leandroalemao (approved):

PR #1028, opened 2025-07-28
Add `DeliveryPartnerSerializer`

ethax-ross:
### Context

We want to add the `DeliveryPartnerSerializer` so that we can return delivery partners from the appropriate APIs.

- Add DeliveryPartnerSerializer

Add the delivery partner serializer. We return the `name` of the delivery partner, the `created_at` and `api_updated_at` and the `cohort`, which is retrieved via the `lead_provider_delivery_partnerships`.

### Guidance for review

We do not need to `uniq` the cohorts as there is already a unique index on `active_lead_provider` and `delivery_partner` for `lead_provider_delivery_partnerships` and also on `lead_provider_id` and `contract_period_id` for `active_lead_providers`.

### DIFF
--- a/app/serializers/delivery_partner_serializer.rb
+++ b/app/serializers/delivery_partner_serializer.rb
@@ -1,0 +1,26 @@
+class DeliveryPartnerSerializer < Blueprinter::Base
+  class AttributesSerializer < Blueprinter::Base
+    exclude :id
+
+    field :name
+    field(:cohort) do |delivery_partner, options|
+      lead_provider = options[:lead_provider]
+
+      delivery_partner
+        .lead_provider_delivery_partnerships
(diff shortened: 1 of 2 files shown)


app/serializers/delivery_partner_serializer.rb
```diff
@@ -0,0 +1,26 @@
+class DeliveryPartnerSerializer < Blueprinter::Base
+  class AttributesSerializer < Blueprinter::Base
+    exclude :id
+
+    field :name
+    field(:cohort) do |delivery_partner, options|
+      lead_provider = options[:lead_provider]
+
+      delivery_partner
```
leandroalemao, line 9: Will this create N+1 queries?

PR #1030, opened 2025-07-28
Show future ECTs in the school list

peteryates:
### Context

Update the schools ECT index so future joiners are listed too.

To support this there are a few new scopes that are relative to the current date, and a new person in the seed data who has a future joining date so we can see how things look in review apps.

* Add new scopes so we can search for current **or** future teachers
* Update `Teachers::Search#ect_at_school` so it searches for `ECTAtSchoolPeriod` records that are either currently ongoing or start in the future
* Add seed data so this kind of teacher can be previewed

### Preview

| Searching for teacher who hasn't started yet | Seed data with appropriate description |
| ------ | ------ |
| image | image |

### Guidance to review

Log in with Bob Belcher (Abbey Grove School) and you should see Peter Davison at the top of the list. He's from the future.

### Related work

DFE-Digital/register-ects-project-board#2013 (this work supports that)
DFE-Digital/register-ects-project-board#1513 (this work will solve some of it)

### DIFF
--- a/app/models/concerns/interval.rb
+++ b/app/models/concerns/interval.rb
@@ -18,6 +18,11 @@ module Interval
     scope :overlapping_with, ->(period) { where(*overlapping_with_range(period.started_on, period.finished_on)) }
     scope :containing_period, ->(period) { where(*containing_range(period.started_on, period.finished_on)) }
     scope :ongoing_on, ->(date) { where(*date_in_range(date)) }
+
+    # Date relative scopes
+    scope :ongoing_today, -> { ongoing_on(Time.zone.today) }
+    scope :starting_in_the_future, -> { started_on_or_after(Time.zone.tomorrow) }
+    scope :ongoing_today_or_starting_in_the_future, -> { ongoing_today.or(starting_in_the_future) }
   end
 
   # Validations
--- a/app/models/concerns/interval.rb
+++ b/app/models/concerns/interval.rb
@@ -21,8 +21,8 @@ module Interval
 
     # Date relative scopes
     scope :ongoing_today, -> { ongoing_on(Time.zone.today) }
-    scope :starting_in_the_future, -> { started_on_or_after(Time.zone.tomorrow) }
-    scope :ongoing_today_or_starting_in_the_future, -> { ongoing_today.or(starting_in_the_future) }
+    scope :starting_tomorrow_or_after, -> { started_on_or_after(Time.zone.tomorrow) }
+    scope :ongoing_today_or_starting_tomorrow_or_after, -> { ongoing_today.or(starting_tomorrow_or_after) }
   end
 
   # Validations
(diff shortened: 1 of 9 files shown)


spec/services/teachers/search_spec.rb
```diff
@@ -224,10 +224,33 @@
...
+          context 'when ECT has left the school' do
+            let!(:ect_at_school_period) { FactoryBot.create(:ect_at_school_period, :past, school:, teacher:) }
+
+            it 'returns no teachers' do
+              expect(Teachers::Search.new(ect_at_school: school).search).to be_empty
+            end
+          end
+
+          context 'when currently an ECT at the school' do
+            let!(:ect_at_school_period) { FactoryBot.create(:ect_at_school_period, :active, school:, teacher:) }
```
peteryates, line 240: I think we should change `:active` to `:ongoing` in the factory.

Rmcnaney:
@avinhurry @mooktakim @edujackedu  to review pls.

spec/factories/ect_at_school_period_factory.rb
```diff
@@ -12,6 +12,16 @@
     email { Faker::Internet.email }
     working_pattern { WORKING_PATTERNS.keys.sample }
 
+    trait :future do
+      started_on { 2.weeks.from_now }
+      finished_on { nil }
+    end
+
+    trait :past do
```
mooktakim, line 20: `:finished` instead of `:past`?

spec/services/teachers/search_spec.rb
mooktakim: Makes sense to go with `:ongoing`, I think, consistent with scopes.

mooktakim (approved):
Minor comments otherwise looks good

app/models/concerns/interval.rb
```diff
@@ -18,6 +18,11 @@ module Interval
     scope :overlapping_with, ->(period) { where(*overlapping_with_range(period.started_on, period.finished_on)) }
     scope :containing_period, ->(period) { where(*containing_range(period.started_on, period.finished_on)) }
     scope :ongoing_on, ->(date) { where(*date_in_range(date)) }
+
+    # Date relative scopes
+    scope :ongoing_today, -> { ongoing_on(Time.zone.today) }
+    scope :starting_in_the_future, -> { started_on_or_after(Time.zone.tomorrow) }
```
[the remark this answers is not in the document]
peteryates: ah yeah good call

spec/factories/ect_at_school_period_factory.rb
peteryates: I thought about this but went with `past` as it's the opposite of future`, but changing to `started` and `finished` would work too :+1:

spec/services/teachers/search_spec.rb
peteryates: Added to #1057

PR #1031, opened 2025-07-28
[SPIKE][2021] Add db view in order to join schools, lead_providers and contract_periods

leandroalemao:
### Context

Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/2021

After working through the schools endpoint, we realised how complex it is to dynamically calculate values based on a school level, from a model where most things are on a participant level.

- Add db view in order to join schools, lead_providers and contract_periods

Pros:
- Real time data is ready/available to be used;
- No need for a refresh (it's not materialized view);
- The sql is run dynamically for every school × contract_period × lead_provider (using cross joins);
- It's not over complex to understand;

Cons:
- Potentially big result set, as we have around 25k schools in 6 cohorts and 7 lead providers.. so ~1.05M rows;
- Performance can be slow (we can't fully test it right now as altho we have schools, we don't have partnerships neither ect/mentors/training periods in migration env);
- Some data can be useless in the end as some of the rows might be for combinations that are meaningless or unused (ie: a school not working with a lead provider in a specific contract period);
- Materialized view can be also used, but it'll need to be refreshed often in order to have "real" time data;
- Haven't really tested other options (ie: triggers) as it'd add too much complexity in comparison with views and would sit in the database level as well.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -25,6 +25,7 @@ gem "govuk_markdown"
 
 gem "mail-notify"
 
+gem "scenic"
 gem "sentry-rails"
 gem "sentry-ruby"
 gem "stackprof"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -702,6 +702,9 @@ GEM
       nokogiri (>= 1.8.1)
       nori (~> 2.4)
       wasabi (>= 3.7, < 6)
+    scenic (1.9.0)
+      activerecord (>= 4.0.0)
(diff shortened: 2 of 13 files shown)


cwrw (commented):
discussed in person: separate paginated results and run views on them only, benchmark on migration with 100 per page and 3000 per page

leandroalemao:
Closing this as skipe ticket is done.

PR #1044, opened 2025-07-29
Add `DeliveryPartners::Query` service

ethax-ross:
### Context

As part of the delivery partners epic we want to add the query service that can be called from the controller. This should mimic the behaviour of the delivery partners query service in ECF.

- Add DeliveryPartners::Query

Add a query service for retrieving delivery partners, which will be called from the controller. The results will be passed into the `DeliveryPartnerSerializer`.

### Guidance to review

We calculate a `transient_cohort` in the query service for efficiency and fallback to looking it up from sratch if its not present. This will be replaced in the future by a pre-calculated field.

### DIFF
--- a/app/serializers/delivery_partner_serializer.rb
+++ b/app/serializers/delivery_partner_serializer.rb
@@ -4,13 +4,15 @@ class DeliveryPartnerSerializer < Blueprinter::Base
 
     field :name
     field(:cohort) do |delivery_partner, options|
-      lead_provider = options[:lead_provider]
-
-      delivery_partner
-        .lead_provider_delivery_partnerships
-        .joins(:active_lead_provider)
-        .where(active_lead_providers: { lead_provider_id: lead_provider.id })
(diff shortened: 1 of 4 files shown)


app/serializers/delivery_partner_serializer.rb
```diff
@@ -4,13 +4,15 @@ class AttributesSerializer < Blueprinter::Base
...
     field(:cohort) do |delivery_partner, options|
-      lead_provider = options[:lead_provider]
-
-      delivery_partner
-        .lead_provider_delivery_partnerships
-        .joins(:active_lead_provider)
-        .where(active_lead_providers: { lead_provider_id: lead_provider.id })
-        .pluck("active_lead_providers.contract_period_id")
+      if delivery_partner.respond_to?(:transient_cohort)
+        delivery_partner.transient_cohort
```
cwrw, line 8: `transient_cohorts`?

PR #1045, opened 2025-07-29
Allow devs to bypass parameter filtering locally

### LINKED ISSUE #1024: Allow environment specific filtering
Currently when developing, various values are obscured from output/logs. In production where these values might contain sensitive data so it makes sense, but locally it's a bit of a pain:

```ruby
register-early-career-teachers(dev)> Teachers::Search.new(query_string: "Colin")

=>
#<Teachers::Search:0x00007f50be93bac8
 @scope=
  [#<Teacher:0x00007f50bf663318
    id: 5,
    corrected_name: "[FILTERED]",
    trn: "[FILTERED]",
    trs_first_name: "[FILTERED]",
    trs_last_name: "[FILTERED]",
    ecf_user_id: nil,
```

I suggest we add a config flag we can use locally that, when enabled reduces the [filter_parameters](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/config/initializers/filter_parameter_logging.rb) to just `_key`.

**By default it should apply all the filters.**

cpjmcquillan:
### Context

Closes #1024 

In Production we always want to filter sensitive data, but it locally this can be a source of friction -particularly when debugging issues.

This adds a configuration that can be enabled to bypass this filtering in the development environment only.

### Guidance to review

- [ ] Enabling the config locally stops filtering sensitive data
- [ ] Sensitive data is filtered as usual in all other environments


config/environments/development.rb
```diff
@@ -90,4 +90,7 @@
   # Use a real queuing backend for Active Job (and separate queues per environment).
   config.active_job.queue_adapter = :solid_queue
   # config.active_job.queue_name_prefix = "ecf2_development"
+
+  # Custom config
+  config.bypass_filter_parameter_logging = ENV.fetch("ENABLE_BYPASS_FILTER_PARAMETER_LOGGING", false)
```
peteryates, line 95: I wonder if it's worth hardcoding this to `false` in `config/application.rb` but allowing it to be overridden here?

Then we wouldn't need the slightly-complex `if` statement in `config/initializers/filter_parameter_logging.rb`, so it could be:

```ruby
Rails.application.config.filter_parameters += %i[_key]

if Rails.application.config.bypass_filter_parameter_logging
  Rails.application.config.filter_parameters += other_stuff
end
```

peterdavidhamilton: Can't think of anything to add other than maybe sourcing the list from a YAML file.

peterdavidhamilton (approved):

PR #1046, opened 2025-07-29
Fix claim bug party programme type validation

peterdavidhamilton:
### Context

Ensure programme type validation can cope with a blank cell

Change specs to assert missing values using nil which is what the validation receives

### Guidance to review

- Bulk claim with no programme type
- Bulk claim with a file where each row has a blank cell for a different field (extra hardcore test)

### DIFF
--- a/app/services/appropriate_bodies/process_batch/claim.rb
+++ b/app/services/appropriate_bodies/process_batch/claim.rb
@@ -99,7 +99,7 @@ module AppropriateBodies
 
       # @return [Boolean] school-led, provider-led (case-insensitive)
       def invalid_training_programme?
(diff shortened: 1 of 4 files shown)


app/services/appropriate_bodies/process_batch/claim.rb
cpjmcquillan, line 1: Do we need to account for this too? 🤔

https://github.com/DFE-Digital/register-early-career-teachers-public/blob/6487fe43037b86b92a6cd09c31e25e83ba9af834/app/services/appropriate_bodies/process_batch/claim.rb#L34

cpjmcquillan (approved):

PR #1052, opened 2025-07-30
Add delivery partner APIs

ethax-ross:
### Context

Now that we have the query and serializer in place, we can flesh out the `index` and `show` actions of the delivery partner APIs.

- Ignore non-integer cohorts in DeliveryPartners::Query

At the moment it will throw a postgres error when filtering on non-integer contract period values.

Add `integers:` option to `ConditionFormats` so that we can restrict the input.

- Add lead_provider relationship for convenience

To make it a bit cleaner to reference the active lead provider of a delivery partnership.

- Cast serialized cohorts to strings

To be consistent with the lead provider API and the behaviour of the transient cohort logic, we want to cast the cohorts to strings.

- Implement index and show actions for delivery partners

Integrate the query and serializer to flesh out the delivery partner APIs.

Update the shared contexts to allow passing a `lead_provider` to the serializer via `options`.

### DIFF
--- a/app/controllers/api/v3/delivery_partners_controller.rb
+++ b/app/controllers/api/v3/delivery_partners_controller.rb
@@ -1,8 +1,37 @@
 module API
   module V3
     class DeliveryPartnersController < BaseController
-      def show = head(:method_not_allowed)
-      def index = head(:method_not_allowed)
+      def index
+        conditions = { contract_period_years:, sort: }
+        render json: to_json(paginate(delivery_partners_query(conditions:).delivery_partners))
+      end
+
+      def show
+        render json: to_json(delivery_partners_query.delivery_partner_by_api_id(api_id))
+      end
+
+    private
+
+      def delivery_partners_query(conditions: {})
+        conditions[:lead_provider] = current_lead_provider
+        DeliveryPartners::Query.new(**conditions.compact)
+      end
+
+      def delivery_partner_params
+        params.permit(:api_id, :sort)
+      end
+
+      def api_id
+        delivery_partner_params[:api_id]
+      end
+
+      def sort
+        delivery_partner_params[:sort]
+      end
+
+      def to_json(obj)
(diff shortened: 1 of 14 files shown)


app/controllers/api/v3/delivery_partners_controller.rb
```diff
@@ -1,8 +1,43 @@
 module API
   module V3
     class DeliveryPartnersController < BaseController
-      def show = head(:method_not_allowed)
-      def index = head(:method_not_allowed)
+      include ContractPeriodFilterable
```
cwrw, line 4: I think this is already in the base controller

spec/requests/api/v3/delivery_partners_spec.rb
```diff
@@ -1,23 +1,33 @@
...
-      authenticated_api_get path
-      expect(response).to be_method_not_allowed
+    def apply_expected_order(resources)
+      resources.sort_by(&:created_at)
     end
+
+    it_behaves_like "a paginated endpoint"
+    it_behaves_like "a token authenticated endpoint", :get
+    it_behaves_like "a filter by multiple cohorts (contract_period year) endpoint"
+    it_behaves_like "an index endpoint"
```
cwrw, line 22: could we add `it_behaves_like "a sortable endpoint"`?

cwrw (approved):
looks great, some minor comments

spec/requests/api/v3/delivery_partners_spec.rb
```diff
@@ -1,23 +1,33 @@
...
+    let(:path_id) { resource.api_id }
+    let(:path) { api_v3_delivery_partner_path(path_id) }
 
     it_behaves_like "a token authenticated endpoint", :get
-
-    it "returns method not allowed" do
-      authenticated_api_get path
-      expect(response).to be_method_not_allowed
-    end
+    it_behaves_like "a show endpoint"
```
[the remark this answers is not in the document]
ethax-ross: This is already in there?

PR #1055, opened 2025-07-30
Track AB bulk feature activity

peterdavidhamilton:
### Context

DfE Analytics needs to send details of bulk feature activity to BigQuery

Capture and set the web request UUID to associate records created via bulk

### Guidance to review

Review app config has been temporarily set to enable analytics which should be visible in BigQuery


spec/requests/appropriate_bodies/process_batch/actions/create_spec.rb
```diff
@@ -23,6 +23,7 @@
           pending_induction_submission_batch: { csv_file: }
         }
       }.to have_enqueued_job(ProcessBatchActionJob)
+      .with(an_instance_of(PendingInductionSubmissionBatch), user.email, user.name, kind_of(String))
```
peteryates, line 26: Feels a bit odd chaining here without brackets.

peteryates (approved):
This looks fine to me but I think the approach in 1061 is cleaner, especially if we can upstream it.

PR #1056, opened 2025-07-30
Always use `govuk_submit` in forms

cpjmcquillan:
Rather than adding `govuk-button` classes manually, we can use the `govuk_submit` form element from the Form Builder.

In an ideal world, we might introduce a Cop for this and integrate it with a tool like `erblint`, but that's a problem for another day!

Inspired by #1051


peterdavidhamilton (approved):

avinhurry (approved):
Nice

peteryates (approved):

PR #1060, opened 2025-07-30
Always use `secondary: true` for secondary submit buttons

cpjmcquillan:
Similar to #1056 and also inspired by #1051.

Rather than adding the `govuk-button--secondary` class to submit buttons explicitly, we can rely on the FormBuilder to handle it for us if we pass `secondary: true`.


peterdavidhamilton (approved):

avinhurry (approved):
Whoops sorry, I thought I check for others. 

Clearly not very well! 😅

PR #1061, opened 2025-07-30
Set the `DfE::Analytics` request ID in `ActiveJob`

cpjmcquillan:
### Context

https://ukgovernmentdfe.slack.com/archives/C0632U64JHG/p1753796129179769

The `DfE::Analytics::Middleware::RequestIdentity` middleware handles setting the request ID for each request. If this is set, `DfE::Analytics` sends an event for the entity with the request UUID.

Until now, the request UUID have only been set for entity events that occur in the request/response life-cycle.

This works well, but the Analytics team would like bulk actions and outcomes to be correlated to the request that instigated them too.

To facilitate this, we're setting the `dfe_analytics_request_id` when serializing/deserializing arguments in `ActiveJob`.


peterdavidhamilton:
I like the concept. 

1. If you enable the analytics in the review config (temporarily) Emma should be able to see data in BigQuery. I did it on mine but she's not confirmed this yet.
2. They'd start seeing web requests uuids on other entities from other jobs, we should check that's not a bad thing. I imagine more the merrier

lib/dfe_analytics/active_job_extensions.rb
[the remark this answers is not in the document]
cpjmcquillan: Also, if we're in the business of upstreaming things to `DfE::Analytics`, it might be worth removing the `request_store` dependency and using [`CurrentAttributes` ](https://api.rubyonrails.org/classes/ActiveSupport/CurrentAttributes.html) 💭

config/terraform/application/config/review.yml
```diff
@@ -13,9 +13,9 @@ TEST_GUIDANCE: true
 ENABLE_SENTRY: false
 SENTRY_ENVIRONMENT: review
 ENABLE_BLAZER: true
-DFE_ANALYTICS_ENABLED: false
-TRS_API_BASE_URL: 'https://preprod.teacher-qualifications-api.education.gov.uk'
-TRS_API_VERSION: '20250203'
+DFE_ANALYTICS_ENABLED: true
+TRS_API_BASE_URL: "https://preprod.teacher-qualifications-api.education.gov.uk"
+TRS_API_VERSION: "20250203"
```
peteryates, line 18: Were these switched by an auto-linter?

cpjmcquillan: Yes, my overzealous editor 😠. Will revert!

PR #1063, opened 2025-07-31
[2016] - Migrate all ECF schools with partnerships or induction records

ltello:
### Context
[Issue](https://github.com/DFE-Digital/register-ects-project-board/issues/2016)

We are importing only open schools from GIAS into RECT. 
Similarly, we are comparing only open, eligible or cip_only schools on ECF1 with those in RECT.

However, other not open schools have partnerships associated and no induction records.
As those partnerships will need to be migrated into RECT, we need their schools to be migrated as well.

Extend the schools comparison migrator to migrate all ECF schools that have partnerships or induction records.

### Guidance to review

### DIFF
--- a/app/migration/migrators/school.rb
+++ b/app/migration/migrators/school.rb
@@ -25,12 +25,14 @@ module Migrators
     def self.reset! = nil
 
     def self.schools
-      ::Migration::School.with(eligible_or_cip_or_with_irs:
+      ::Migration::School.with(eligible_or_cip_or_with_irs_or_partnerships:
                                  [
                                    ::Migration::School.includes(:local_authority).eligible_or_cip_only_except_welsh.distinct,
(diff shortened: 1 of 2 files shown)


leoapost (approved):

app/migration/migrators/school.rb
```diff
@@ -25,12 +25,14 @@ def self.record_count = schools.count
     def self.reset! = nil
 
     def self.schools
-      ::Migration::School.with(eligible_or_cip_or_with_irs:
+      ::Migration::School.with(eligible_or_cip_or_with_irs_or_partnerships:
```
peteryates, line 28: `irs` made me think of the [Inland Revenue Service](https://en.wikipedia.org/wiki/Inland_Revenue) 😂

PR #1064, opened 2025-07-31
Add support for different parity check URLs

ethax-ross:
### Context

On some of the new RECT endpoints we are omitting `ecf` from the path of the endpoint (schools, for example). This means we need to call different endpoints depending on the application.

- Add support for different parity check URLs

Add support to parity check endpoint options for `ecf_url` or `rect_url` - this will be used with preference over the default URL provided as the key to the check.

### Guidance to review

[Example run](https://cpd-ec2-migration-web.teacherservices.cloud/migration/parity-checks/8)

### DIFF
--- a/app/migration/parity_check/request_builder.rb
+++ b/app/migration/parity_check/request_builder.rb
@@ -16,7 +16,7 @@ module ParityCheck
     delegate :method, :path, :options, to: :endpoint
 
     def url(app:)
-      parity_check_url(app:) + formatted_path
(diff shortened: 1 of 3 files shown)


ethax-ross:
> looks good, do you mind running schools before we merge?

I ran one of the /schools endpoints and one /schools/ecf as above in the example links, they both seemed to work fine

leandroalemao (approved):

PR #1066, opened 2025-07-31
2051: Move school list to own separate page

ghost:
### Context
Move school list to own separate page. Update its route and add a link to it to the navbar.

Screenshot 2025-08-01 at 11 16 40
Screenshot 2025-08-01 at 11 17 04

### DIFF
--- a/spec/views/admin/schools/index.html.erb_spec.rb
+++ b/spec/views/admin/schools/index.html.erb_spec.rb
@@ -1,19 +1,13 @@
 RSpec.describe 'admin/schools/index.html.erb' do
   include Pagy::Backend
 
-  let(:number_of_schools) { 2 }
+  let(:number_of_schools) { 10 }
   let(:schools) { FactoryBot.create_list(:school, number_of_schools) }
   let(:pagy) { Pagy.new(count: number_of_schools, limit: 5, page: 1) }
 
   before do
     assign(:schools, schools)
     assign(:pagy, pagy)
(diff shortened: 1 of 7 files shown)


spec/views/admin/schools/index.html.erb_spec.rb
```diff
@@ -42,7 +35,59 @@
...
+
+  it 'renders pagination and results summary when schools are present' do
+    render
+
+    expect(rendered).to have_css('.govuk-pagination')
+    expect(rendered).to have_css('div.govuk-body', text: "Showing 1 to 5 of 10 results")
+  end
+
+  context 'when there are no schools' do
+    let(:schools) { [] }
```
avinhurry, line 58: I think you should be able to remove this now

spec/views/admin/schools/index.html.erb_spec.rb
```diff
@@ -42,7 +35,59 @@
...
+    it 'shows no schools found message' do
+      render
+
+      expect(rendered).to have_css('p', text: 'No schools found.')
+      expect(rendered).not_to have_css('table.govuk-table')
+    end
+  end
+
+  context 'when search query returns no results' do
+    let(:schools) { [] }
```
avinhurry, line 70: and this one too

avinhurry (approved):
I think you should be able to remove those `let(:schools) { [] }` now (left a couple of comments in line)

Other than that looks good, thanks!

PR #1074, opened 2025-08-01
Bulk action error logic

peterdavidhamilton:
### Context

Nest error hierarchy block similar to claims

### Guidance to review

A teacher who has already completed an induction with an outcome of pass or fail should see an error message in a bulk actions upload.

> "#{name} has already completed their induction"

### DIFF
--- a/app/services/appropriate_bodies/process_batch/action.rb
+++ b/app/services/appropriate_bodies/process_batch/action.rb
@@ -54,17 +54,21 @@ module AppropriateBodies
         elsif (trs_error = fetch_trs_details!)
           capture_error(trs_error)
           true
+        elsif teacher
+          if passed? || failed?
+            capture_error("#{name} has already completed their induction")
+            true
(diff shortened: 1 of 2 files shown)


joe-harrison-dfe:
Looks like it all works to me, it doesn't read 'already completed' but rather 'already passed' or 'already failed' is that correct?

peterdavidhamilton:
> Looks like it all works to me, it doesn't read 'already completed' but rather 'already passed' or 'already failed' is that correct?

- When claiming you see 'already passed' or 'already failed'
- When actioning you see 'already completed' 

These are the error messages decided upon earlier and haven't changed. I have done a file and recreated it with Adam to demonstrate
https://cpd-ec2-review-1074-web.test.teacherservices.cloud/appropriate-body/bulk/actions/5

(valid until new commits are pushed)

cpjmcquillan (approved):

PR #1075, opened 2025-08-01
[2116] Refactor schools query service

leandroalemao:
### Context

Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/2116

Query service select all schools with all transient fields.. which is taking a long time. 

- Schools records are selected by the query service and then a much smaller result set of schools to be paginated will be sent back to the query service to have transient fields added to it.
- Query service now selects plain schools records in separate then send it back to the query service to enrich the much smaller result set with transient fields required for `SchoolSerializer`.
- School endpoint response time tested with this branch deployed to the migration env and comparing main branch with this branch:
  - For 10 records - down from ~1.8s to ~110ms.
  - For 100 records - down from ~1.7s to ~120ms.
  - For 500 records - down from ~1.9s to ~240ms.
  - For 1000 records - down from ~2s to ~390ms
  - For 3000 records - down from ~2.4s to ~900ms
- Still far from ideal response times, but numbers have improved considerably.
- Benchmark results using a view instead and same logic as this branch ([PR#1031](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/1031))
  - For 10 records - ~100ms.
  - For 100 records - ~130ms.
  - For 500 records - ~200ms.
  - For 1000 records - ~320ms
  - For 3000 records - ~720ms

### Guidance to review

Review app

### DIFF
--- a/app/controllers/api/v3/schools_controller.rb
+++ b/app/controllers/api/v3/schools_controller.rb
@@ -4,14 +4,26 @@ module API
       filter_validation required_filters: %i[cohort]
 
       def index
-        render json: to_json(paginate(schools_query.schools))
+        render json: to_json(schools)
       end
 
       def show
-        render json: to_json(schools_query.school_by_api_id(api_id))
+        render json: to_json(school)
       end
 
     private
+
+      def paginated_results
+        paginate(schools_query.schools_for_pagination)
+      end
+
+      def schools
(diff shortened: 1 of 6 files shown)


cwrw:
discussing in person: can we run this after deploying to migration esp the 3000 records to benchmark those on the server to get comparable results to other spike?

PR #1081, opened 2025-08-04
Align API endpoint implementations

ethax-ross:
### Context

Our implementation of API endpoints has diverged a little; this PR aims to bring them back to being more consistent.

- Switch Schools::Query to accept lead_provider

This brings it inline with the `Statements::Query` and `DeliveryPartners::Query`.

Remove unused `transient_lead_provider_id`.

- Switch Schools::Query to accept contract_period_years

This brings it inline with the `Statements::Query` and using `years` instead of `id` makes it clear that the value is a `year`.

- Remove blank? check from Schools::Query

The `ignore?` method will already check and ignore blank values.

- Switch Schools::Query to accept contract_period_years

This brings it inline with the `Statements::Query` and using `years` instead of `id` makes it clear that the value is a `year`.

- Remove blank? check from Schools::Query

The `ignore?` method will already check and ignore blank values.

- Make singular query methods consistent

Make the methods to retrieve a single object from a query service by its `id` consistent and inline with the method to retrieve a single object by its `api_id`.

- Clean up query concerns

Move query-related concerns so they are all under the `Queries` namespace.

Remove the `FilterValidatableEndpoint` spec as this is in `/services/concerns` when it is a controller concern and we don't explicitly test other controller concerns; they are tested via the request specs and shared contexts. Update
shared context to cover the tests we are removing.

- Make sorting in Statements::Query consistent

We have a concern for sorting in our query services. We implemented statements before this was used (I think because we don't allow end-user sorting), but now it makes sense to bring it in line.

- Remove options from serialization

As we pass in the result of the `Query` it will always have the transient attributes, so we don't need to worry about the fallback scenario.


app/controllers/api/v3/schools_controller.rb
```diff
@@ -16,7 +16,7 @@ def show
 
       def schools_query(conditions: {})
         conditions[:lead_provider] = current_lead_provider
-        conditions[:contract_period_id] = contract_period&.id
+        conditions[:contract_period_years] = contract_period&.id
```
cwrw, line 19: should we use `contract_period_year` rather than `contract_period_years` as schools only accepts a single contract year 🤔 ?

app/services/schools/query.rb
```diff
@@ -74,7 +74,7 @@ def transient_in_partnership(contract_period_years)
     end
 
     def transient_mentors_at_school(contract_period_years)
-      return School.none if ignore?(filter: contract_period_years) || contract_period_years.blank?
```
cwrw: I think the blank here is because contract period year is required and not optional, maybe we need to make that clearer with a test?

spec/services/concerns/filter_validatable_spec.rb
```diff
@@ -1,42 +0,0 @@
-class FilterValidatableTestsController < API::BaseController
```
cwrw, line 1: @leandroalemao discussed this with @ethax-ross as I asked for this spec in your pr, we agreed in the end that controller concerns can be tested as part of the shared contexts in request specs, happy to chat later today to agree

app/controllers/admin/finance/statements_controller.rb
```diff
@@ -46,7 +46,7 @@ def statements_query
         contract_period_years:,
         statement_date:,
         fee_type:,
-        order_by: :statement_date,
+        sort: "+year,+month",
```
cwrw, line 49: slightly confused by this, we don't allow sorting at all in statements as you mention, why would we add it 🤔

ethax-ross: We hard code this in to give the statements the correct/matching ordering with ECF; we need to support a `sort` attribute on the query as the admin area also pulls in statements and sorts by `payment_date`

leandroalemao (approved):
Hi @ethax-ross looks good tks 👍

cwrw (approved):
thanks @ethax-ross

PR #1082, opened 2025-08-04
Scope routes instead of specifying controllers

cpjmcquillan:
This is a large change (in terms of LOC), so feel free to push back. There should be no impact on end users, though.

This tidies up our routes and follows more Rails naming conventions.

The intention is to simplify things and reduce the cognitive overhead when trying to grok our routes.


app/controllers/admin/import_ect/find_ects_controller.rb
```diff
@@ -1,6 +1,6 @@
 module Admin
   module ImportECT
-    class FindECTController < AdminController
+    class FindECTsController < AdminController
```
peteryates, line 3: We're only trying to find one ECT here so to me the singular makes more sense.

cpjmcquillan: That's fair enough.

My approach was "rip out all the stuff and see what sticks".

I'll leave the changes that necessitated any controller renaming 👍🏼

config/routes.rb
cpjmcquillan, line 1: Do we think this is better, or worse? 💭

peteryates (approved):
This is much neater, thanks for picking this up.

PR #1072, opened 2025-08-01
Rework listing card component to accept teacher and training period

peteryates:
The `Schools::ECTs::ListingCardComponent` currently recives one object, an `ECTAtSchoolPeriod` when being initialised and then uses `ECTatSchoolPeriods::Training` to determine the training information. This made sense when we weren't concerned about future periods but now we need to take that into consideration.

To hopefully simplify things, there are now `has_one` relationships that retrieve the `current_ect_at_school_period` from the `Teacher`, and the `current_training_period` from the `ECTatSchoolPeriod`.

This means we can pass into the listing component all of the objects we will need, and use them directly.

## Notes

The moving of `training_programme` from `ECTAtSchoolPeriod` to `TrainingPeriod` means there's a chance we'll have teachers who are neither `school-led` or `provider-led`. We'll need to address this in a later change.

## Changes

- **Rename ect variable to ect_at_school_period**
- **Add has_one relationships for current period**
- **Pass relevant objects into listing card component**
- **Use training period as source for training programme**

PR #1084, opened 2025-08-04
Update training service for current training period only

peteryates:
> [!NOTE]
> Follows on from #1072, don't review until it's merged and this is rebased.

  This changes the `ECTAtSchoolPeriods::Training` class so it doesn't differentiate between the latest and current training period. It's needed because we now allow future periods to be recorded and we're look at training from a school's point of
  view, via the ECTAtSchoolPeriod. This means we'll be looking at:

  * either the current (ongoing now) training period
  * a training period due to start in the future

  There's no need (at the moment) for future training periods at the same school, so we only want to retrieve one.


avinhurry (approved):
Makes sense

PR #1088, opened 2025-08-05
Add delivery partners APIs to parity check

ethax-ross:
### Context

We want to run the delivery partner APIs on the parity check now that they are complete to find any differences/issues.

- Add delivery partner API requests to the parity check.

Update `DynamicRequestContent` to retrieve a `delivery_partner_id`.

- Support response bodies with arrays of values

Currently we expect all arrays to contain objects (which can be mapped into a hash). When we return `cohort` for delivery partners, however, it is an array of strings and trips up the `ParityCheck::Filter::ResponseBody` logic.

Ignore nested items in the key hash mapping unless they are also a hash.

### Guidance to review

[Run in migration](https://cpd-ec2-migration-web.teacherservices.cloud/migration/parity-checks/31)

From the most part it looks like the updated_at/created_at are different and a handful of `cohort` entries (some are just ordered differently but there are some that are a small handful that have different content).

### DIFF
--- a/app/migration/parity_check/dynamic_request_content.rb
+++ b/app/migration/parity_check/dynamic_request_content.rb
@@ -34,6 +34,14 @@ module ParityCheck
         .pick(gias_school: :api_id)
     end
 
+    def delivery_partner_id
+      DeliveryPartners::Query.new(lead_provider:)
+        .delivery_partners
+        .distinct(false)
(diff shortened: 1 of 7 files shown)


leandroalemao (approved):
Hi @ethax-ross Looks good tks

PR #1099, opened 2025-08-06
2049 allow admins to add lead providers to delivery partners  

ghost:
### Context
This PR is to create a page for each delivery partner in the admin console where users can see and edit the LP(s) that the DP is working with per contract year.

Screenshot 2025-08-05 at 16 19 44

Screenshot 2025-08-04 at 14 06 58

### DIFF
--- a/app/components/admin/lead_provider_partnerships_table_component.html.erb
+++ b/app/components/admin/lead_provider_partnerships_table_component.html.erb
@@ -9,7 +9,10 @@
   end
 
   table.with_body do |body|
-    grouped_partnerships.each do |contract_period, partnerships|
+    contract_period_partnerships.each do |item|
+      contract_period = item[:contract_period]
+      partnerships = item[:partnerships]
+
       body.with_row do |row|
         row.with_cell(text: contract_period.year)
         row.with_cell do
--- a/app/components/admin/lead_provider_partnerships_table_component.rb
+++ b/app/components/admin/lead_provider_partnerships_table_component.rb
@@ -1,23 +1,19 @@
 module Admin
   class LeadProviderPartnershipsTableComponent < ViewComponent::Base
-    attr_reader :lead_provider_partnerships, :delivery_partner, :page, :q
+    attr_reader :contract_period_partnerships, :delivery_partner, :page, :q
 
-    def initialize(lead_provider_partnerships:, delivery_partner:, page: nil, q: nil)
-      @lead_provider_partnerships = lead_provider_partnerships
+    def initialize(contract_period_partnerships:, delivery_partner:, page: nil, q: nil)
+      @contract_period_partnerships = contract_period_partnerships
       @delivery_partner = delivery_partner
       @page = page
       @q = q
     end
 
     def render?
-      lead_provider_partnerships.any?
+      contract_period_partnerships.any?
     end
 
   private
-
-    def grouped_partnerships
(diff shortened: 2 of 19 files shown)


config/routes.rb
```diff
@@ -57,7 +57,7 @@
         end
 
         resources :lead_providers, only: %i[index], path: 'lead-providers'
-        resources :delivery_partners, only: %i[index show], path: 'delivery-partners'
+        resources :delivery_partners, only: %i[index show edit update], path: 'delivery-partners'
```
peteryates, line 60: This feels a tiny bit odd as we're creating new delivery partnerships - and needing the year as a query param doesn't feel right.

Maybe we can adjust the routes a little so they're more like:

```
/admin/organisations/delivery-partners/1/2025
```

Something along the lines of this?

```
resources :delivery_partners, only: %i[index show], path: 'delivery-partners' do
  resource :lead_provider_delivery_partnerships, only: %i[new create], path: ':year', as: :delivery_partnership
end
```

That would give us this which feels nicer:

```
❯ be rails routes -g 'admin/organisations/delivery-partners'
                                         Prefix Verb URI Pattern                                                                     Controller#Action
new_admin_delivery_partner_delivery_partnership GET  /admin/organisations/delivery-partners/:delivery_partner_id/:year/new(.:format) admin/lead_provider_delivery_partnerships#new
    admin_delivery_partner_delivery_partnership POST /admin/organisations/delivery-partners/:delivery_partner_id/:year(.:format)     admin/lead_provider_delivery_partnerships#create
                        admin_delivery_partners GET  /admin/organisations/delivery-partners(.:format)                                admin/delivery_partners#index
                         admin_delivery_partner GET  /admin/organisations/delivery-partners/:id(.:format)                            admin/delivery_partners#show
```

app/components/admin/add_lead_providers_form_component.html.erb
```diff
@@ -0,0 +1,30 @@
...
+  <%= f.hidden_field :q, value: q %>
+
+  <% if current_partnerships.any? %>
+    <div class="govuk-inset-text">
+      <h3 class="govuk-heading-s">Currently working with:</h3>
+      <ul class="govuk-list govuk-list--bullet">
+        <% current_partnership_names.each do |name| %>
+          <li><%= name %></li>
+        <% end %>
+      </ul>
```
peteryates, line 13: We can use [the list helper](https://govuk-components.netlify.app/helpers/list/#bulleted-lists) here:

```suggestion
      <%= govuk_list(current_partnership_names.map(&:name), type: :bullet) %>
```

app/models/lead_provider_delivery_partnership.rb
```diff
@@ -19,4 +19,15 @@ class LeadProviderDeliveryPartnership < ApplicationRecord
...
+    joins(:active_lead_provider)
+      .where(active_lead_providers: { contract_period_year: contract_period.year })
+      .includes(active_lead_provider: :lead_provider)
+  }
+  scope :active_lead_provider_ids_for, ->(delivery_partner, contract_period) {
+    where(delivery_partner:)
+      .joins(:active_lead_provider)
+      .where(active_lead_providers: { contract_period_year: contract_period.year })
+      .select(:active_lead_provider_id)
+  }
```
peteryates, line 32: I think this should be a service rather than a scope - I always think of scopes as things that filter a recordset but don't change the shape. That can go in as a housekeeping issue to tidy up though, no need for it to be a stopper.

PR #1106, opened 2025-08-07
Add metadata tables for schools

ethax-ross:
### Context

In an effort to clean up our queries and serializers for schools we are going to store some hard/complex look ups as school-level values in metadata tables. This will make the underlying queries and serializers much simpler and faster.

- Add metadata models for schools

Add the `SchoolLeadProviderContractPeriod` metadata for storing the `expression_of_interest` for a school/lead provider/contract period.

Add `SchoolContractPeriod` metadata for storing the `in_partnership` and `induction_programme_choice` of a school/contract period.

Ensure only services in the `Metadata` namespace can persist updates to these models.

- Ignore Rails/SaveBang on FactoryBot usage

Rubocop is failing due to the `FactoryBot.create` call in `metadata_support.rb`. I'm not sure why this is being picked up as an issue in this file but not elsewhere; maybe its to do with it not being in a `_spec.rb` file, but either way we should ignore it.

- Update Mermaid diagram

### Guidance to review

In the ticket it mentions `induction_programme_choice` should be in the `SchoolLeadProviderContractPeriod` metadata, but I think this is a mistake as it doesn't appear to rely on the lead provider to determine the value - I've put it in the `SchoolContractPeriod` metadata table instead.

### DIFF
--- a/documentation/domain-model.md
+++ b/documentation/domain-model.md
@@ -284,10 +284,33 @@ erDiagram
   ActiveLeadProvider {
     integer id
     integer lead_provider_id
-    integer contract_period_id
+    integer contract_period_year
     datetime created_at
     datetime updated_at
   }
   ActiveLeadProvider }o--|| ContractPeriod : belongs_to
   ActiveLeadProvider }o--|| LeadProvider : belongs_to
+  Metadata_SchoolLeadProviderContractPeriod {
+    integer id
+    integer school_id
+    integer lead_provider_id
+    integer contract_period_year
+    boolean expression_of_interest
+    datetime created_at
+    datetime updated_at
+  }
+  Metadata_SchoolLeadProviderContractPeriod }o--|| School : belongs_to
+  Metadata_SchoolLeadProviderContractPeriod }o--|| LeadProvider : belongs_to
+  Metadata_SchoolLeadProviderContractPeriod }o--|| ContractPeriod : belongs_to
+  Metadata_SchoolContractPeriod {
+    integer id
+    integer school_id
+    integer contract_period_year
+    boolean in_partnership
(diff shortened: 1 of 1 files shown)


db/migrate/20250807063339_create_join_table_schools_lead_providers_contract_periods_metadata.rb
```diff
@@ -0,0 +1,14 @@
+class CreateJoinTableSchoolsLeadProvidersContractPeriodsMetadata < ActiveRecord::Migration[8.0]
+  def change
+    create_table :schools_lead_providers_contract_periods_metadata do |t|
```
peteryates, line 3: Might it be clearer from a database point of view if they were prefixed with `metadata` rather than suffixed?

ethax-ross: Yeah that's a good shout, it follows the namespace that way as well. I'll update it 👍

app/models/metadata/school_contract_period.rb
```diff
@@ -0,0 +1,20 @@
...
+      school_led: "school_led"
+    }
+
+    belongs_to :school
+    belongs_to :contract_period, foreign_key: :contract_period_year
+
+    validates :school, presence: true
+    validates :contract_period, presence: true
+    validates :in_partnership, inclusion: { in: [true, false] }
+    validates :induction_programme_choice, inclusion: { in: induction_programme_choices.keys }
```
cwrw, line 17: you're right `induction_programme_choice` is not lead provider specific 👍

mooktakim (approved):

PR #522, opened 2025-04-25
Enable admin user to reopen a completed induction

tonyheadford:
Enable admins to reopen an induction, either with an outcome or one that was released and is on hold.  Typically this arises when a user passes or fails instead of releasing a teacher.

The "Reopen induction" button is only present if the latest induction period has a `finished_on` date.

This action will:
 - only work when the latest induction period has a `finished_on` date
 - remove the `finished_on`, `number_of_terms` and `outcome` on the latest induction period
 - adds an "induction period reopened" event
 - if an outcome is present it will update TRS with a `status` of "InProgress" and clear the `completedDate` for the teacher
 
image

image

PR #1108, opened 2025-08-07
Admins can only "Reopen induction" when there is an outcome

cpjmcquillan:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/2145

#### Use Teacher last_induction_period association
In https://github.com/DFE-Digital/register-early-career-teachers-public/pull/858 we introduced a `last_induction_period` association to the `Teacher`
model to fix some N+1 queries.

Now this association exists, it felt simpler to lean on this rather than
accessing it through the service object.

This moves the `last_induction_period` method out of the public API for the
`Teachers::InductionPeriod` service and updates usage in a few places.

#### Admins can only "Reopen induction" when there is an outcome
In https://github.com/DFE-Digital/register-early-career-teachers-public/pull/522 we introduced functionality that allowed Admin users to reopen a
completed induction.

This works well, but it also allows Admin users to "reopen" a completed
induction without an outcome (i.e the induction is "paused"). This was not the
original intention.

Even though this works without issue, it adds a level of complexity for Admin
users that isn't necessary. Since they can achieve the desired outcome already
by "Edit"ing a paused induction, we can safely remove this.

Now, Admin users can only "reopen induction" when the latest induction period is
completed **and** has an outcome (i.e "pass" or "fail").

### Guidance to review

- [ ] View a teacher where the latest induction period is complete without an outcome (i.e "Induction paused")
- [ ] There is no "Reopen induction" button

### DIFF
--- a/app/controllers/admin/teachers/reopen_induction_controller.rb
+++ b/app/controllers/admin/teachers/reopen_induction_controller.rb
@@ -8,8 +8,10 @@ module Admin
 
         last_induction_period = @teacher.last_induction_period
 
-        if last_induction_period.blank? || last_induction_period.ongoing?
-          redirect_to admin_teacher_path(@teacher), notice: "No completed induction period found"
+        if last_induction_period.blank? ||
+            last_induction_period.ongoing? ||
+            last_induction_period.outcome.blank?
+          return redirect_to admin_teacher_path(@teacher), notice: "No completed induction period found"
         end
 
         Admin::ReopenInductionPeriod.new(author: current_user, induction_period: last_induction_period).reopen_induction_period!
--- a/app/services/admin/reopen_induction_period.rb
+++ b/app/services/admin/reopen_induction_period.rb
@@ -31,11 +31,16 @@ module Admin
 
     def check_can_reopen_period!
       check_induction_period_is_complete!
+      check_induction_period_has_outcome!
       check_induction_period_is_the_latest!
     end
 
     def check_induction_period_is_complete!
       raise ReopenInductionError, "Cannot reopen an ongoing induction" unless induction_period.complete?
+    end
+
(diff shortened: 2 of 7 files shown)


app/controllers/admin/teachers/reopen_induction_controller.rb
```diff
@@ -2,7 +2,11 @@ module Admin
   module Teachers
     class ReopenInductionController < AdminController
       def update
-        @teacher = Teacher.find(params[:teacher_id])
+        @teacher = Teacher
+          .includes(:last_induction_period)
```
cpjmcquillan, line 6: We're always using the `last_induction_period`, so I thought it made sense to eager load it 👌🏼

app/services/teachers/induction_period.rb
```diff
@@ -43,5 +43,9 @@ def overlapping_with?(date)
     teacher.induction_periods.ongoing_on(date).exists?
   end
 
-  delegate :last_induction_period, :first_induction_period, to: :teacher
+  delegate :first_induction_period, to: :teacher
+
+private
+
+  def last_induction_period = teacher.last_induction_period
```
cpjmcquillan, line 50: This might look kind of weird, but I was running into Rubocop issue trying to delegate a method "privately". (Is that even possible? 🤔)

app/services/admin/reopen_induction_period.rb
```diff
@@ -31,13 +31,18 @@ def reopen_induction_period!
 
     def check_can_reopen_period!
       check_induction_period_is_complete!
+      check_induction_period_has_outcome!
```
cpjmcquillan, line 34: This wasn't technically in-scope, but it felt odd to restrict things in the UI without making the same changes here.

PR #1111, opened 2025-08-07
[2030] Sub navigation for guidance pages

mooktakim:
### Context

Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/2030

* Added sidebar navigation with `Navigation::SubNavigationComponent`
* Only show sidebar if "Guidance" pages

### Guidance to review

* https://cpd-ec2-review-1111-web.test.teacherservices.cloud/api/guidance/guidance-for-lead-providers

### DIFF
--- a/documentation/content-editors/api-guidance-pages.md
+++ b/documentation/content-editors/api-guidance-pages.md
@@ -99,14 +99,12 @@ To add a new page to the **Guidance** section:
    `app/components/api/guidance/sidebar_component.rb`
 2. Update the `GUIDANCE_PAGES` constant to include the new page:
 
-<pre><code>```ruby
-  GUIDANCE_PAGES = [
+<pre><code>GUIDANCE_PAGES = [
   { title: "API IDs explained", path: "api-ids-explained" },
   { title: "API data states", path: "api-data-states" },
   { title: "Syncing data best practice", path: "data-syncing" },
   { title: "New API guidance page", path: "new-api-guidance-page" }, # <-- New entry
-].freeze
-```</code></pre>
+].freeze</code></pre>
 
 Field definitions:
 


app/controllers/api/guidance_controller.rb
```diff
@@ -16,18 +17,32 @@ def release_notes
...
-        "guidance-for-lead-providers/ids-explained" => "guidance_for_lead_providers/ids_explained",
-      }.fetch(path)
-
-      render "api/guidance/" + template
+      template = "api/guidance/#{params[:page].underscore}"
+
+      if template_exists?(template)
+        render template
+      else
+        render 'errors/not_found', status: :not_found
```
cwrw, line 25: can we have request specs for this?

app/controllers/api/guidance_controller.rb
```diff
@@ -16,18 +17,32 @@ def release_notes
...
+      end
+    end
+
+  private
+
+    def guidance_pages
+      [
+        { title: "API IDs explained", path: "guidance-for-lead-providers/ids-explained" },
```
cwrw, line 33: should we update IDs explained title to include API everywhere so it's consistent, one to ask Mark

app/controllers/api/guidance_controller.rb
```diff
@@ -16,18 +17,32 @@ def release_notes
...
+      if template_exists?(template)
+        render template
+      else
+        render 'errors/not_found', status: :not_found
+      end
+    end
+
+  private
+
+    def guidance_pages
```
cwrw, line 31: maybe this logic is better suited to be in a component such as guidance sidebar navigation component? We can pass in the current path and it decides if we should show the sidenav, and the guidance pages are defined there?
Happy to chat through

cwrw (commented):
a few comments, happy to chat. Also I think there was a doc I think we need to write to explain how to add new subnavs if required for Mark?

app/controllers/api/guidance_controller.rb
mooktakim: confirmed with mark will change it

app/controllers/api/guidance_controller.rb
mooktakim: moved into a component

app/components/api/guidance/sidebar_component.rb
```diff
@@ -0,0 +1,29 @@
...
+      end
+
+      def render?
+        page.to_s.starts_with?("guidance-for-lead-providers")
+      end
+
+      def structure
+        GUIDANCE_PAGES.map do |p|
+          path = api_guidance_page_path(p[:path])
+          Struct.new(:name, :href, :prefix, :nodes).new(p[:title], path, path, [])
```
peteryates, line 24: Maybe define the struct once and reuse it? Seems wasteful to create new ones each time and use them once.

leandroalemao (approved):
@mooktakim tested all good tks 👍

cwrw (commented):
looks great, thanks Mook, just waiting on docs update as agreed, will approve after

PR #1112, opened 2025-08-07
Use AB name in error messages when validating claims and outcomes

peterdavidhamilton:

Replace

- `#{name} is already claimed by another appropriate body`
- `#{name} is completing their induction with another appropriate body`

With 

- `#{name} is already claimed by another appropriate body (#{ab.name})`
- `#{name} is completing their induction with another appropriate body (#{ab.name})`


cpjmcquillan (approved):
🚀

PR #1118, opened 2025-08-08
Retry background jobs when the TRS API errors

cpjmcquillan:
As part of [this investigation] into mismatched TRS induction statuses, we identified that our background jobs are not retrying by default.

That meant that any intermittent API error (e.g service unavailability) would result in some jobs failing, and the corresponding updates *never happening*.

Now, in jobs that use the TRS API, we retry on API request errors. As part of this change, the `Teachers::RefreshTRSAttributes` service has been refactored so we can use a single `api_client` in a job, rather than instantiating several.

[this investigation]: https://github.com/DFE-Digital/register-ects-project-board/issues/2134

### DIFF
--- a/app/models/concerns/trs/retryable_client.rb
+++ b/app/models/concerns/trs/retryable_client.rb
@@ -1,0 +1,17 @@
+module TRS
+  module RetryableClient
+    extend ActiveSupport::Concern
+
+    included do
+      retry_on Errors::APIRequestError,
+               wait: ->(executions) { 2**executions },
+               attempts: 15
+    end
+
+  private
+
+    def api_client
+      @api_client ||= TRS::APIClient.build
+    end
+  end
+end
--- a/spec/jobs/fail_ect_induction_job_spec.rb
+++ b/spec/jobs/fail_ect_induction_job_spec.rb
@@ -1,8 +1,8 @@
 RSpec.describe FailECTInductionJob, type: :job do
   let(:teacher) { FactoryBot.create(:teacher) }
   let(:trn) { teacher.trn }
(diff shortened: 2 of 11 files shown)


app/models/concerns/trs/retryable_client.rb
```diff
@@ -0,0 +1,17 @@
+module TRS
+  module RetryableClient
+    extend ActiveSupport::Concern
+
+    included do
+      retry_on Errors::APIRequestError,
+               wait: ->(executions) { 2**executions },
+               attempts: 15
```
cpjmcquillan, line 8: I just picked something that seemed vaguely sensible. 

Open to other suggestions, and also happy to iterate later if we want to see how this goes ♻️

spec/jobs/fail_ect_induction_job_spec.rb
```diff
@@ -1,16 +1,19 @@
 RSpec.describe FailECTInductionJob, type: :job do
   let(:teacher) { FactoryBot.create(:teacher) }
   let(:trn) { teacher.trn }
-  let(:start_date) { "2023-11-13" }
-  let(:completed_date) { "2024-01-13" }
+  let(:start_date) { Date.parse("2023-11-13") }
+  let(:completed_date) { Date.parse("2024-01-13") }
```
cpjmcquillan, line 5: The API client expects date objects, not strings

peteryates: Not sure how this was passing earlier if it was passing in strings 🤔

cpjmcquillan: Stubbing!

PR #1123, opened 2025-08-11
Integrate metadata into schools endpoint

ethax-ross:
### Context

Now that we have metadata models for the hard to query school details we want to update the query and serializer for schools in order to return the relevant information from the metadata.

- Update Schools::Query to use metadata

Add metadata relationships on `School`.

Update `Schools::Query` to include the metadata and remove transient attributes.

Remove `lead_provider_id` from the `Schools::Query` as this is not used for the base query; we will filter by lead provider in the serializer instead.

- Update SchoolSerializer to use metadata

Update the `SchoolSerializer` to retrieve the `in_partnership`, `expression_of_interest` and `induction_programme_choice` from the associated metadata.

Pass in the `lead_provider_id` and `contract_period_year` as serializer options so that we can determine the correct metadata.

- Fix request specs

Update the schools request specs to work with the updated query and serializer.

- Update school touch model for metadata

Now that we have a metadata table to store `in_partnership`, `expression_of_interest` and `induction_programme_choice` we can simplify the touch model for schools.

Remove `DeclarativeTouch` from models that would feed into these attribute values and add `DeclarativeTouch` to the metadata models.

- Filter metadata in Schools::Query

We only care about the metadata associated with the given `contract_period_year` and `lead_provider_id` passed into
the schools query. 

Filter the to-many relationship to only return relevant metadata. Using `references` provides an `LEFT OUTER JOIN`
on the metadata tables, which we can then filter. We include `nil` to ensure schools are returned even if there are no
metadata records (which should never be the case, but feels more robust in case it ever happens and also means the other test setup is leaner).

### Guidance to review

We could update the serializer to not pass in the `lead_provider_id` and `contract_period_year`, but it feels
safer to do this so that the serializer isn't tightly coupled to the query.

### DIFF
--- a/spec/requests/api/v3/schools_spec.rb
+++ b/spec/requests/api/v3/schools_spec.rb
@@ -7,12 +7,7 @@ RSpec.describe "Schools API", type: :request do
 
   def create_resource(active_lead_provider:)
     lead_provider_delivery_partnership = FactoryBot.create(:lead_provider_delivery_partnership, active_lead_provider:)
-    school_partnership = FactoryBot.create(:school_partnership, lead_provider_delivery_partnership:)
-
-    school_partnership.school.tap do |school|
-      FactoryBot.create(:school_lead_provider_contract_period_metadata, school:, contract_period:, lead_provider:)
-      FactoryBot.create(:school_contract_period_metadata, school:, contract_period:)
-    end
+    FactoryBot.create(:school_partnership, :with_school_metadata, lead_provider_delivery_partnership:).school
   end
 
   describe "#index" do
--- a/app/models/lead_provider_delivery_partnership.rb
+++ b/app/models/lead_provider_delivery_partnership.rb
@@ -6,6 +6,7 @@ class LeadProviderDeliveryPartnership < ApplicationRecord
   has_many :school_partnerships
   has_many :events
   has_one :lead_provider, through: :active_lead_provider
+  has_one :contract_period, through: :active_lead_provider
 
   touch -> { delivery_partner }, on_event: %i[create destroy], timestamp_attribute: :api_updated_at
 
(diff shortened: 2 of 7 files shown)


app/serializers/school_serializer.rb
```diff
@@ -3,47 +3,36 @@ class AttributesSerializer < Blueprinter::Base
...
-      end
-
-      def in_partnership?(school, options)
-        if school.respond_to?(:transient_in_partnership)
-          school.transient_in_partnership
-        else
-          school.school_partnerships.for_contract_period(options[:contract_period_year]).exists?
-        end
+      def contract_period_metadata(school:, options:)
+        school.contract_period_metadata.select(
```
leandroalemao, line 26: I'm getting an `ActiveRecord::SoleRecordExceeded` error here for some reason.. I've changed these 2 methods as per below to fix it:

```
def contract_period_metadata(school:, options:)
  school.contract_period_metadata.select { |cpmt|
    cpmt.contract_period_year = options[:contract_period_year]
  }.sole
end

def lead_provider_contract_period_metadata(school:, options:)
  school.lead_provider_contract_period_metadata.select { |cpmt|
    cpmt.lead_provider_id = options[:lead_provider_id] &&
      cpmt.contract_period_year = options[:contract_period_year]
  }.sole
end
```

spec/requests/api/v3/schools_spec.rb
```diff
@@ -7,7 +7,12 @@
 
   def create_resource(active_lead_provider:)
     lead_provider_delivery_partnership = FactoryBot.create(:lead_provider_delivery_partnership, active_lead_provider:)
-    FactoryBot.create(:school_partnership, lead_provider_delivery_partnership:).school
+    school_partnership = FactoryBot.create(:school_partnership, lead_provider_delivery_partnership:)
```
cwrw, line 10: maybe we can add metadata to traits in the school factory?

cwrw (approved):
nicely done! tested on migration and went through different iterations. Only thing unrelated to this ticket is passing in an incorrect cohort to show schools, I'll add a ticket for that

leandroalemao (approved):

PR #1127, opened 2025-08-11
Add recurring job to clear finished solid queue jobs

cwrw:
### Context
We would like to clear jobs from the database when they are successfully finished. The number currently is at 7million in production as there has been no recurring job in place to cull those jobs.

Add job that clears finished jobs every hour on the 12th minute in batches.
This job was taken from solid queue: https://github.com/rails/solid_queue?tab=readme-ov-file#other-configuration-settings

I've changed the defaults for `clear_finished_jobs_after` from 1 day to a week instead in production.

### Guidance to review
I'm going to test this in migration env, even though the job runs deletions in batches it's worth monitoring deleting 7 million jobs at once and might require an initial manual task first

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -65,6 +65,8 @@ Rails.application.configure do
   # Use a real queuing backend for Active Job (and separate queues per environment).
   config.active_job.queue_adapter = :solid_queue
   # config.active_job.queue_name_prefix = "ecf2_production"
+  # Remove finished solid queue jobs from database after a week
+  config.solid_queue.clear_finished_jobs_after = 1.week
 
   # GOVUK Notify
   config.action_mailer.delivery_method = :notify
--- a/config/recurring.yml
+++ b/config/recurring.yml
@@ -14,11 +14,24 @@ production:
     queue: default
     schedule: Every hour
 
+  clear_solid_queue_finished_jobs:
+    command: "SolidQueue::Job.clear_finished_in_batches(sleep_between_batches: 0.3)"
+    schedule: every hour at minute 12
+
+sandbox:
+  clear_solid_queue_finished_jobs:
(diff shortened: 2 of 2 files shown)


config/environments/production.rb
```diff
@@ -65,6 +65,8 @@
   # Use a real queuing backend for Active Job (and separate queues per environment).
   config.active_job.queue_adapter = :solid_queue
   # config.active_job.queue_name_prefix = "ecf2_production"
+  # Remove finished solid queue jobs from database after a week
+  config.solid_queue.clear_finished_jobs_after = 1.week
```
peteryates, line 69: Any reason not to set it across all envs?

peteryates (approved):

config/environments/production.rb
cwrw: everything inherits from `config/environments/production` bar dev/test so I think it's ok to leave those to the default

leandroalemao: Following the documentation.. this seems to be configured to be 1 day by default? So we'll be increasing the time instead? Think we need to add the job to clear finished jobs to `recurring.yml` instead or maybe I'm missing something?

https://github.com/rails/solid_queue?tab=readme-ov-file#other-configuration-settings

```
- preserve_finished_jobs: 
  whether to keep finished jobs in the solid_queue_jobs table—defaults to true.
- clear_finished_jobs_after: 
  period to keep finished jobs around, in case preserve_finished_jobs is true — defaults to 1 day. When installing Solid Queue, [a recurring job](https://github.com/rails/solid_queue#recurring-tasks) is automatically configured to clear finished jobs every hour on the 12th minute in batches. You can edit the recurring.yml configuration to change this as you see fit.
```

cwrw:
I've cleared manually all clearable jobs on production

PR #1133, opened 2025-08-12
Admins must confirm before reopening an induction

cpjmcquillan:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/2142

Before, reopening a Teacher's induction was a one-click action.

We've seen some mistakes since launching this feature, so we think we've made it too easy to use accidentally. Especially considering the action is destructive and irreversible.

This adds some friction to the journey by introducing a confirmation screen.

Now, admins must confirm they are sure before then can reopen a Teacher's induction.

### Guidance to review

https://github.com/user-attachments/assets/55445094-5c7e-43b1-afde-f099e3ba546e

### DIFF
--- a/app/controllers/admin/teachers/reopen_induction_controller.rb
+++ b/app/controllers/admin/teachers/reopen_induction_controller.rb
@@ -1,22 +1,34 @@
 module Admin
   module Teachers
     class ReopenInductionController < AdminController
+      before_action :set_teacher
+
+      before_action -> do
+        redirect_to admin_teacher_path(@teacher),
+                    notice: "No completed induction period found"
+      end, unless: :induction_complete_with_outcome?
+
       def update
+        Admin::ReopenInductionPeriod.new(
+          author: current_user,
+          induction_period: @teacher.last_induction_period
+        ).reopen_induction_period!
+
+        redirect_to admin_teacher_path(@teacher),
+                    alert: "Induction was successfully reopened"
+      end
+
+    private
+
+      def set_teacher
         @teacher = Teacher
           .includes(:last_induction_period)
           .find(params[:teacher_id])
+      end
 
-        last_induction_period = @teacher.last_induction_period
-
-        if last_induction_period.blank? ||
-            last_induction_period.ongoing? ||
-            last_induction_period.outcome.blank?
-          return redirect_to admin_teacher_path(@teacher), notice: "No completed induction period found"
-        end
-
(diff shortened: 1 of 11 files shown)


app/controllers/admin/teachers/reopen_induction_controller.rb
```diff
@@ -1,22 +1,38 @@
 module Admin
   module Teachers
     class ReopenInductionController < AdminController
+      before_action :set_teacher,
+                    :set_last_induction_period,
+                    :ensure_induction_is_complete!
```
cpjmcquillan, line 6: What about? 🤔

```ruby
                    :redirect_unless_induction_is_complete
```

spec/features/admin/teachers/reopen_induction_spec.rb
```diff
@@ -0,0 +1,44 @@
+describe "Admins can reopen a teacher's closed induction" do
+  let(:teacher) { FactoryBot.create(:teacher) }
+  let!(:induction_period) do
```
peterdavidhamilton, line 3: You might have noticed I'm in the habit of favouring the before block; unless something needs to be accessed in an assertion, I don't name a variable.

peterdavidhamilton (approved):

spec/features/admin/teachers/reopen_induction_spec.rb
cpjmcquillan: Tbh I loathe `let!` anyway, I'm just trying to fit in 😆

app/views/admin/teachers/reopen_induction/confirm.html.erb
```diff
@@ -0,0 +1,53 @@
...
+        admin_teacher_reopen_induction_path(@teacher),
+        method: :patch,
+        warning: true
+      ) %>
+  <%= govuk_button_to(
+        "Cancel",
+        admin_teacher_path(@teacher),
+        method: :get,
+        secondary: true
+      ) %>
```
peteryates, line 52: Should this be a [govuk_button_link_to](https://govuk-components.netlify.app/helpers/button/#link-styled-like-a-button) instead of a button with `method: :get`?

cpjmcquillan:
> Looks good. I was expecting to see [a success banner](https://govuk-components.netlify.app/components/notification-banner/#notification-banner-with-success) when redirected back to the profile page.

Good question, I've asked the team 🧠

PR #1138, opened 2025-08-12
[2181] Add api_updated_at for partnerships table

mooktakim:
### Context

Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/2181

* Added new column `api_updated_at` to `school_partnerships` table
* Add touch update for `School`, `DeliveryPartner`, `SchoolPartnership`

### DIFF
--- a/app/models/lead_provider_delivery_partnership.rb
+++ b/app/models/lead_provider_delivery_partnership.rb
@@ -9,7 +9,6 @@ class LeadProviderDeliveryPartnership < ApplicationRecord
   has_one :contract_period, through: :active_lead_provider
 
   touch -> { delivery_partner }, on_event: %i[create destroy], timestamp_attribute: :api_updated_at
-  touch -> { school_partnerships }, when_changing: %i[delivery_partner_id], timestamp_attribute: :api_updated_at
 
   delegate :lead_provider, :contract_period, to: :active_lead_provider
 
--- a/spec/models/lead_provider_delivery_partnership_spec.rb
+++ b/spec/models/lead_provider_delivery_partnership_spec.rb
@@ -139,19 +139,5 @@ describe LeadProviderDeliveryPartnership do
 
       it_behaves_like "a declarative touch model", on_event: %i[create destroy], timestamp_attribute: :api_updated_at
     end
-
-    context "target school_partnerships" do
-      let!(:school_partnership) { FactoryBot.create(:school_partnership, lead_provider_delivery_partnership: instance) }
-      let(:target) { school_partnership }
-
-      def generate_new_value(attribute_to_change:)
-        case attribute_to_change
-        when :delivery_partner_id
-          FactoryBot.create(:delivery_partner).id
-        end
-      end
-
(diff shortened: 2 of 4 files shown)


app/models/lead_provider_delivery_partnership.rb
```diff
@@ -8,6 +8,7 @@ class LeadProviderDeliveryPartnership < ApplicationRecord
   has_one :lead_provider, through: :active_lead_provider
 
   touch -> { delivery_partner }, on_event: %i[create destroy], timestamp_attribute: :api_updated_at
+  touch -> { school_partnerships }, when_changing: %i[delivery_partner_id], timestamp_attribute: :api_updated_at
```
ethax-ross, line 11: Would this ever change? 🤔

spec/models/lead_provider_delivery_partnership_spec.rb
```diff
@@ -51,9 +51,21 @@
...
+    context "target delivery_partner" do
+      let(:target) { delivery_partner }
+
+      it_behaves_like "a declarative touch model", on_event: %i[create destroy], timestamp_attribute: :api_updated_at
+    end
+
+    context "target school_partnerships" do
+      let!(:school_partnership) { FactoryBot.create(:school_partnership, lead_provider_delivery_partnership: instance) }
+      let(:target) { school_partnership }
+      let(:new_value) { FactoryBot.create(:delivery_partner).id }
```
ethax-ross, line 66: Are we able to keep this explicit as with how we do it in schools? I'm just thinking if we added any other attribute to the touch model it'll not work as expected and might be confusing? I think this is a bit clearer (but maybe you can't pass an `id` into a factory? never tried...):

```
def will_change_attribute(attribute_to_change:, new_value:)
  FactoryBot.create(:delivery_partner, id: new_value) if attribute_to_change == :delivery_partner_id
end
```

app/models/lead_provider_delivery_partnership.rb
cwrw: actually my bad, this was on the ticket, the change would be on `lead_provider_delivery_partnership_id` on school partnership intead rather than here

spec/models/lead_provider_delivery_partnership_spec.rb
mooktakim: It would be better to allow us to explicitly create the new value on the spec as needed, instead of always randomly generating. Added this:

```ruby
      def generate_new_value(attribute_to_change:)
        case attribute_to_change
        when :lead_provider_delivery_partnership_id
          FactoryBot.create(:lead_provider_delivery_partnership).id
        end
      end
```
Should maintain the existing behaviour.

app/models/lead_provider_delivery_partnership.rb
mooktakim: Moved to `SchoolPartnership.lead_provider_delivery_partnership_id`

cwrw (approved):

leandroalemao (approved):

ethax-ross (approved):

PR #1143, opened 2025-08-13
Add input to ignore jobs tables during backups

ethax-ross:
### Context

We want to have the option to exclude jobs tables when doing backups of the production database for the snapshot and migration environments so that the backups run quicker.

Add inputs to the migration/snapshot database refresh jobs so that we can ignore job tables and backup more quickly.

Include the jobs tables by default in snapshot.

Exclude the jobs tables by default in migration.

Tables excluded:

solid_queue_jobs – the master record of every enqueued job (active, scheduled, or finished) 
solid_queue_ready_executions – jobs ready to run now 
solid_queue_claimed_executions – jobs actively being worked on by a worker 
solid_queue_blocked_executions – jobs blocked due to concurrency limits 
solid_queue_failed_executions – failed job logs
solid_queue_processes – background process monitoring

Tables retained:

solid_queue_scheduled_executions – jobs waiting until a scheduled time before execution
solid_queue_recurring_tasks – definitions of recurring jobs (their key, schedule, arguments, etc.) 
solid_queue_recurring_executions – history/future runs of recurring jobs (can be rebuilt, but useful to keep) 
solid_queue_semaphores – concurrency and resource control settings solid_queue_pauses – paused queue definitions

### Guidance to review

[Snapshot refresh - with jobs](https://github.com/DFE-Digital/register-early-career-teachers-public/actions/runs/16937646823)
[Snapshot refresh - without jobs](https://github.com/DFE-Digital/register-early-career-teachers-public/actions/runs/16958201793)

[Migration refresh - without jobs](https://github.com/DFE-Digital/register-early-career-teachers-public/actions/runs/16938073608)
[Migration refresh - with jobs](https://github.com/DFE-Digital/register-early-career-teachers-public/actions/runs/16938255279)

Without jobs:

```
register-early-career-teachers(migration)> SolidQueue::Job.count
=> 0
```

With jobs:

```
register-early-career-teachers(migration)> SolidQueue::Job.count
register-early-career-teachers(migration)> 
=> 460838
```

I haven't changed the backup/restore to/from azure actions as they appear to use a generic DfE action for doing the backup so we can't hook into it (the snapshot refresh above is the direct copy workflow).

### DIFF
--- a/.github/actions/directly-backup-and-restore-snapshot-database/action.yml
+++ b/.github/actions/directly-backup-and-restore-snapshot-database/action.yml
@@ -5,6 +5,10 @@ inputs:
   environment:
     description: The name of the environment
     required: true
+  exclude-jobs:
+    description: Exclude jobs tables
+    type: boolean
+    default: false
   azure-client-id:
     description: Azure service principal or managed identity client ID when using OIDC
     required: true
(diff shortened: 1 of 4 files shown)


.github/actions/directly-backup-and-restore-snapshot-database/action.yml
```diff
@@ -61,7 +65,22 @@ runs:
     - name: Backup database
       shell: bash
       run: |
-        bin/konduit.sh -n ${{ env.NAMESPACE }} cpd-ec2-${{ inputs.environment }}-web -- pg_dump -E utf8 --compress=1 --clean --if-exists --no-privileges --no-owner --verbose -f backup-${{ inputs.environment }}.sql.gz
+        # Build up exclude arguments based on flags
+        EXCLUDE_ARGS=""
+
+        if [ "${{ inputs.exclude-jobs }}" = "true" ]; then
+          EXCLUDE_ARGS="$EXCLUDE_ARGS \
```
leandroalemao, line 72: Have we got `pg_dump: error: too many command-line arguments` error in the [run without the tables](https://github.com/DFE-Digital/register-early-career-teachers-public/actions/runs/16937822324)? Why can't we add all solid queue tables instead?

ethax-ross: Ah I linked to the wrong run in the PR description; I've updated the link now. I hit that error at one point but it was because I was passing multiple tables to `--exclude-table-data` whereas you need to pass the option multiple times, once for each table.

We could exclude all solid queue tables but some of them contain information about scheduled/recurring jobs; it felt safer to keep these, though I suspect they would be re-created on boot.

leandroalemao: It'd be ok to remove all solid queue tables as we don't need scheduled or recurring jobs running in the new migration env either I reckon? But let's see how it goes.. happy to approve it 👍 tks

PR #1150, opened 2025-08-13
Fix deprecation warning

cpjmcquillan:
In https://github.com/rack/rack/pull/2137, Rack updated the `422` code to `unprocessable_content`.

Since we've updated `rspec-rails`, we've seen a bunch of deprecation warnings.

This addresses those.


spec/requests/admin/induction_periods_spec.rb
```diff
@@ -134,7 +134,7 @@
 
       it "renders the new template with unprocessable_entity status" do
```
avinhurry, line 135:
```suggestion
      it "renders the new template with unprocessable_content status" do
```

spec/requests/admin/induction_periods_spec.rb
```diff
@@ -96,7 +96,7 @@
...
```
avinhurry, line 97:
```suggestion
      it "renders the new template with unprocessable_content status" do
```

spec/requests/admin/induction_periods_spec.rb
```diff
@@ -171,7 +171,7 @@
 
       it "returns unprocessable_entity status" do
```
avinhurry, line 172:
```suggestion
      it "returns unprocessable_content status" do
```

spec/requests/admin/induction_periods_spec.rb
```diff
@@ -208,7 +208,7 @@
...
```
avinhurry, line 209:
```suggestion
      it "returns unprocessable_content status" do
```

spec/requests/admin/induction_periods_spec.rb
```diff
@@ -245,7 +245,7 @@
...
```
avinhurry, line 246:
```suggestion
      it "returns unprocessable_content status" do
```

avinhurry (commented):
Nice, thanks for doing this. 

I was just about to do the same then remembered this PR 😃

PR #1155, opened 2025-08-14
Fix N+1 queries in parity check

ethax-ross:
### Context

The parity check views have a few N+1 queries where we need to preload associations of the runs.

There was also a view that was re-querying grouped requests instead of operating on the already in-memory results.

Update `dispatch_requests` to batch-queue jobs to avoid an N+1 in SolidQueue.

- Fix N+1 queries in parity check


ethax-ross:
Closing as CI doesn't like the '+' in the branch name, I think.

PR #1157, opened 2025-08-14
Fix N+1 queries in parity check

ethax-ross:
### Context

The parity check views have a few N+1 queries where we need to preload associations of the runs.

There was also a view that was re-querying grouped requests instead of operating on the already in-memory results.

Update `dispatch_requests` to batch-queue jobs to avoid an N+1 in SolidQueue.

- Fix N+1 queries in parity check

### DIFF
--- a/app/controllers/migration/parity_checks_controller.rb
+++ b/app/controllers/migration/parity_checks_controller.rb
@@ -26,7 +26,7 @@ class Migration::ParityChecksController < ::AdminController
   end
 
   def show
-    @run = ParityCheck::Run.completed.find(params[:run_id])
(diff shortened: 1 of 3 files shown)


leandroalemao (commented):
Hi @ethax-ross I'm still seeing lots of N+1 warnings in the log when I browse the pages `/migration/parity-checks/completed` and `/migration/parity-checks/{id}`.. could you pls double check? tks

Something in:

`app/models/parity_check/request.rb:70`
and
`app/models/parity_check/run.rb:49`

cwrw (approved):

PR #1161, opened 2025-08-14
Switch to using real lead provider names

ethax-ross:
### Context

Originally we used similar but different lead provider names in development and review. This was in an effort to reduce the risk of changing something in an environment accidentally, however its caused more issues and confusion than we expected so we are going to use the normal/real lead provider names everywhere.

- Switch to using real lead provider names

### Guidance to review

I've also added some missing lead providers from the review/development seeds.

### DIFF
--- a/db/seeds/lead_providers.rb
+++ b/db/seeds/lead_providers.rb
@@ -9,11 +9,13 @@ def describe_lead_provider(lead_provider, years)
 end
 
 lead_providers_data = [
-  { name: 'Ambitious Institute', years: [2022, 2023, 2024, 2025, 2026] },
-  { name: 'Capitan', years: [2021, 2022, 2023] },
-  { name: 'Teach Fast', years: [2022, 2023, 2024, 2025] },
-  { name: 'International Institute of Teaching', years: [2021] },
-  { name: 'Better Practice Network', years: [2022, 2023, 2024, 2025] },
+  { name: 'Ambition Institute', years: [2022, 2023, 2024, 2025, 2026] },
+  { name: 'Best Practice Network', years: [2022, 2023, 2024, 2025] },
+  { name: 'Capita', years: [2021, 2022, 2023] },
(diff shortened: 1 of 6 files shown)


db/seeds/lead_providers.rb
```diff
@@ -9,11 +9,13 @@ def describe_lead_provider(lead_provider, years)
...
-  { name: 'Teach Fast', years: [2022, 2023, 2024, 2025] },
-  { name: 'International Institute of Teaching', years: [2021] },
-  { name: 'Better Practice Network', years: [2022, 2023, 2024, 2025] },
+  { name: 'Ambition Institute', years: [2022, 2023, 2024, 2025, 2026] },
+  { name: 'Best Practice Network', years: [2022, 2023, 2024, 2025] },
+  { name: 'Capita', years: [2021, 2022, 2023] },
+  { name: 'Education Development Trust', years: [2021, 2022, 2023, 2024, 2025] },
+  { name: 'National Institute of Teaching', years: [2021] },
+  { name: 'Teach First', years: [2022, 2023, 2024, 2025] },
+  { name: 'UCL Institute of Education', years: [2021, 2022, 2023, 2024, 2025] },
```
leandroalemao, line 18: Hi @ethax-ross maybe we can have random `ecf_id` added to lead providers in the seeds as well? It's a requirement for running a parity check

ethax-ross: The parity check won't work with random IDs though, so I'm not sure if that would be useful and it might lead to more confusion? 🤔

cwrw: I think the parity check keys in the ENV variables are based on exact IDs from production mapped to API keys so not keen on changing those, unless I misunderstood

documentation/api-tokens.md
cwrw, line 1: Can we add staging as well?

leandroalemao (approved):

PR #1164, opened 2025-08-14
[2058] Add partnership endpoints to parity check

leandroalemao:
### Context

Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/2058

- We'd like to add partnership endpoints to parity check so we can test the results of the data compared to ECF1.

- Add partnership endpoints to parity check

### Guidance to review

Successfully tested in [migration env](https://cpd-ec2-migration-web.teacherservices.cloud/migration/parity-checks/6).

### DIFF
--- a/app/migration/parity_check/request_builder.rb
+++ b/app/migration/parity_check/request_builder.rb
@@ -7,7 +7,7 @@ module ParityCheck
     class IDOptionMissingError < RuntimeError; end
     class UnrecognizedQueryError < RuntimeError; end
 
-    ID_PLACEHOLDER = ":id".freeze
+    PLACEHOLDER_PATTERN = /:[a-z_]+/
     PAGINATION_PER_PAGE = 100
 
     attribute :request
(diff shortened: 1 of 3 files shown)


app/migration/parity_check/request_builder.rb
```diff
@@ -82,6 +82,16 @@ def options_query
 
       raise UnrecognizedQueryError, "Query must be a Hash: #{options_query}" unless options_query.is_a?(Hash)
 
+      # If the query contains a filter with symbol values, replace them with dynamic request content.
+      # This allows us to use dynamic values in the query.
+      options_query_filter_symbol_values = options_query[:filter]&.select { |_k, v| v.to_s.match?(/^:.+$/) }
```
ethax-ross, line 87: Could this be cleaned up, made generic (so not tied to `filter`) and combined with the `ID_PLACEHOLDER` stuff for a general solution to deal with placeholders? I'm thinking along the lines of:

```
PLACEHOLDER_PATTERN = /:[a-z_]+/.freeze

def populate_placeholder(value)
  return value unless value =~ PLACEHOLDER_PATTERN

  dynamic_request_content.fetch(value.delete_prefix(":"))
end
```

Then the path stuff becomes simpler:

```
def formatted_path(app:)
  app_specific_path = app_specific_path(app:)

  populate_placeholder(app_specific_path)
end
```

We can then drop the `IDOptionMissingError` error and it'll be raised as a `UnrecognizedIdentifierError` from the `DynamicRequestContent` class.

Then your `options_query` can be simpler:

```
def options_query
  options_query = options[:query]

  return {} unless options_query

  raise UnrecognizedQueryError, "Query must be a Hash: #{options_query}" unless options_query.is_a?(Hash)

  options_query.deep_transform_values(&method(:populate_placeholder))
end
```

^ I've tested none of this 🙂

ethax-ross (commented):
Looks good, I think the placeholder stuff could be improved though - left some suggestions 👍

mooktakim (approved):
All good, just Ross's comment about redundant code

PR #1178, opened 2025-08-18
Admins must add context when deleting an induction period

cpjmcquillan:
Building on #1176, this ensures Admin users provide extra information (in the form of a support ticket URL or a note) when they delete an induction period.

### DIFF
--- a/app/controllers/admin/induction_periods_controller.rb
+++ b/app/controllers/admin/induction_periods_controller.rb
@@ -1,14 +1,19 @@
 module Admin
   class InductionPeriodsController < AdminController
+    include AuditableParams
+
+    before_action :set_teacher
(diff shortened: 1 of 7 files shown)


app/controllers/admin/induction_periods_controller.rb
```diff
@@ -35,20 +36,31 @@ def update
...
 
   private
 
+    def set_teacher
+      @teacher = Teacher.includes(:induction_periods).find(params[:teacher_id])
+    end
+
+    def set_induction_period
+      @induction_period = @teacher.induction_periods.find(params[:id])
+    end
```
cpjmcquillan, line 62: Before, the induction period wasn't scoped to the current Teacher —which seemed like an oversight

PR #1183, opened 2025-08-19
[2057] create seed data for sandbox

mooktakim:
### Context

Rebases: https://github.com/DFE-Digital/register-early-career-teachers-public/pull/1148
Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/2057

* Updated `SchoolPartnerships` seed file to create second school partnership with same school 10% of the time

### DIFF
--- a/app/services/sandbox_seed_data/school_partnerships.rb
+++ b/app/services/sandbox_seed_data/school_partnerships.rb
@@ -9,21 +9,25 @@ module SandboxSeedData
       log_plant_info("school partnerships")
 
       ActiveLeadProvider.find_each do |active_lead_provider|
-        SCHOOL_PARTNERSHIPS_PER_ACTIVE_LEAD_PROVIDER.times do |n|
-          school = create_school_partnerships(active_lead_provider)
+        SCHOOL_PARTNERSHIPS_PER_ACTIVE_LEAD_PROVIDER.times do
+          create_school_partnership(active_lead_provider)
+        end
 
-          if n < SAME_SCHOOL_DIFFERENT_DELIVERY_PARTNER_PER_ACTIVE_LEAD_PROVIDER
-            create_school_partnerships(active_lead_provider, school:)
-          end
+        schools = select_existing_schools(active_lead_provider)
+        schools.each do |school|
+          create_school_partnership(active_lead_provider, school:)
         end
       end
     end
 
   private
 
-    def create_school_partnerships(active_lead_provider, school: nil)
+    def create_school_partnership(active_lead_provider, school: nil)
       school ||= find_available_school(active_lead_provider)
+      return unless school
+
       delivery_partner = find_available_delivery_partner(active_lead_provider)
+      return unless delivery_partner
 
       lead_provider_delivery_partnership = find_or_create_lead_provider_delivery_partnership(active_lead_provider:, delivery_partner:)
       school_partnership = FactoryBot.create(:school_partnership, school:, lead_provider_delivery_partnership:)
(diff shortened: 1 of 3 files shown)


app/services/sandbox_seed_data/school_partnerships.rb
```diff
@@ -15,14 +15,30 @@ def plant
   private
 
     def create_school_partnerships
-      LeadProviderDeliveryPartnership.find_each do |lead_provider_delivery_partnership|
-        school = School.order("RANDOM()").first
-
-        next if school.school_partnerships.map(&:lead_provider_delivery_partnership).map(&:id).include?(lead_provider_delivery_partnership.id)
+      ids = LeadProviderDeliveryPartnership.pluck(:id)
```
cwrw, line 18: not sure why we need ids here, can't we just loop through `LeadProviderDeliveryPartnership` as it did already?

app/services/sandbox_seed_data/school_partnerships.rb
```diff
@@ -15,14 +15,30 @@ def plant
...
+        existing_school_ids = lead_provider_delivery_partnership.school_partnerships.pluck(:school_id)
+        school = School.where.not(id: existing_school_ids).order("RANDOM()").first
+        next unless school
 
         school_partnership = FactoryBot.create(:school_partnership, school:, lead_provider_delivery_partnership:)
 
         log_seed_info("#{school_partnership.school.name} -> #{school_partnership.lead_provider.name}", colour: Colourize::COLOURS.keys.sample)
+
+        # Create another school partnership with the same school but different delivery_partner
+        if rand < 0.1
```
cwrw, line 29: instead of relying on randomness can we ensure there are at least 2 school partnerships for lp/contract period/school with different dps?

app/services/sandbox_seed_data/school_partnerships.rb
```diff
@@ -15,14 +15,30 @@ def plant
   private
 
     def create_school_partnerships
-      LeadProviderDeliveryPartnership.find_each do |lead_provider_delivery_partnership|
```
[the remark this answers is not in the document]
mooktakim: We create over 300 partnerships.

`app/services/sandbox_seed_data/lead_provider_delivery_partnerships.rb` is where we create `LeadProviderDeliveryPartnership` for every lead provider and contract period. This file is looping through them and creating the school partnerships

app/services/sandbox_seed_data/school_partnerships.rb
mooktakim: ah its not needed, will undo

app/services/sandbox_seed_data/school_partnerships.rb
mooktakim: Refactored to allow specifying number of records

app/services/sandbox_seed_data/school_partnerships.rb
```diff
@@ -1,29 +1,51 @@
...
+    end
+
+    def find_available_school(active_lead_provider)
+      existing_school_ids = LeadProviderDeliveryPartnership.find_by(active_lead_provider:).school_partnerships.pluck(:school_id)
+      School.where.not(id: existing_school_ids).order("RANDOM()").first!
+    end
+
+    def find_available_delivery_partner(active_lead_provider)
+      existing_delivery_partner_ids = active_lead_provider.lead_provider_delivery_partnerships.pluck(:delivery_partner_id)
+      DeliveryPartner.where.not(id: existing_delivery_partner_ids).order("RANDOM()").first!
```
cwrw, line 48: I think although intentional this is falling over midway, can we maybe instead skip on and go to other active lead providers/contract periods to avoid the script not completing, same with school?

cwrw (commented):
looking at the results I think we'll need to backfill as a one off induction tutor/email manually in sandbox, we can run something like this:
```
School.where(induction_tutor_name: nil).each do |school|
school.update!(induction_tutor_name: Faker::Name.name, induction_tutor_email: Faker::Internet.email)
end
```

and can we update the school sandbox script to include `with_induction_tutor` in https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/services/sandbox_seed_data/schools.rb#L20

app/services/sandbox_seed_data/school_partnerships.rb
```diff
@@ -1,29 +1,51 @@
...
+        end
       end
     end
 
   private
 
-    def create_school_partnerships
-      LeadProviderDeliveryPartnership.find_each do |lead_provider_delivery_partnership|
-        school = School.order("RANDOM()").first
+    def create_school_partnerships(active_lead_provider, school: nil)
```
ethax-ross, line 24: `create_school_partnership` (singular)

spec/services/sandbox_seed_data/school_partnerships_spec.rb
```diff
@@ -30,6 +33,17 @@
       expect(SchoolPartnership.all.map(&:contract_period).uniq).to eq(ContractPeriod.all)
     end
 
+    it "creates school partnership with same school but different delivery partner" do
+      instance.plant
+
+      count = SchoolPartnership.all.each_with_object({}) do |sp, sum|
```
ethax-ross, line 39: I think you could verify this with a query, something like:

```
schools_with_multiple_delivery_partners = SchoolPartnership
  .joins(:lead_provider_delivery_partnership)
  .group(:school_id)
  .having("COUNT(DISTINCT lead_provider_delivery_partnerships.delivery_partner_id) > 1")
  
expect(schools_with_multiple_delivery_partners).to be_present
```

cwrw (approved):
thanks Mook

PR #1188, opened 2025-08-19
[1185]  Extract school partnership setup to shared helper

avinhurry:
### Context

Several specs were repeating the same setup logic to create a full school partnership structure. Since this setup is expected to be reused, it made sense to centralise it.

### What this PR does

- Extracts the duplicated setup into a shared helper: `create_school_partnership_with_lead_provider`
- Places the helper in `spec/support/schools/partnership_helper.rb` under a namespaced module
- Replaces inline setup code with calls to the helper in affected specs

This reduces duplication, improves maintainability, and makes future test setups more consistent and reusable.


spec/support/schools/partnership_helper.rb
```diff
@@ -0,0 +1,11 @@
+module Schools
+  module PartnershipHelper
+    def create_school_partnership_with_lead_provider
+      @contract_period = FactoryBot.create(:contract_period, :current)
+      @lead_provider = FactoryBot.create(:lead_provider, name: "Xavier's School for Gifted Youngsters")
+      @active_lead_provider = FactoryBot.create(:active_lead_provider, contract_period: @contract_period, lead_provider: @lead_provider)
+      @lead_provider_delivery_partnership = FactoryBot.create(:lead_provider_delivery_partnership, active_lead_provider: @active_lead_provider)
+      @school_partnership = FactoryBot.create(:school_partnership, school: @school, lead_provider_delivery_partnership: @lead_provider_delivery_partnership)
```
peteryates, line 8: Think we should pass `school` in instead of relying on the instance variable to be set?

PR #1194, opened 2025-08-20
Add `Partnerships::Create` service

ethax-ross:
### Context

We want to support creating partnerships in RECT.

- Add `Partnerships::Create` service

Add a service to create partnerships, inline with the behaviour in ECF and our update rules for RECT:

- `contract_period`, `school`, `lead_provider` and `delivery_partner` must all be present.
- `contract_period` must be `enabled`.
- a `LeadProviderDeliveryPartnership` exists for the `lead_provider` and `delivery_partner` (for the school).
- the school is not CIP-only.
- the school is eligible.
- the school does not already have a partnership with the lead provider for the given `contract_period`.
- the school is not `school_led`.

Does not deal with expressions of interest at the moment; that will come later.

### Guidance to review

We will look into replacing the error messages so that they are closer to the RECT domain in the service and have a way of returning API terminology (i.e. `cohort` instead of `contract period` in a follow on ticket).

### DIFF
--- a/app/services/events/record.rb
+++ b/app/services/events/record.rb
@@ -499,8 +499,8 @@ module Events
         author.event_author_params
       when Events::SystemAuthor
         author.system_author_params
-      when Events::LeadProviderAuthor
-        author.lead_provider_author_params
+      when Events::LeadProviderAPIAuthor
+        author.lead_provider_api_author_params
       when Events::AppropriateBodyBatchAuthor
         author.event_author_params
       else
--- a/app/services/events/record.rb
+++ b/app/services/events/record.rb
@@ -343,6 +343,32 @@ module Events
       new(event_type:, author:, heading:, lead_provider:, happened_at: Time.zone.now, metadata:).record_event!
     end
 
+    # School Partnership Events
+
+    def self.record_school_partnership_created_event!(author:, school_partnership:)
+      event_type = :school_partnership_created
+      school = school_partnership.school
+      delivery_partner = school_partnership.delivery_partner
+      lead_provider = school_partnership.lead_provider
+      contract_period = school_partnership.contract_period
+      heading = "#{school.name} partnered with #{delivery_partner.name} (via #{lead_provider.name}) for #{contract_period.year}"
+      metadata = {
+        contract_period_year: contract_period.year,
+      }
+
+      new(
+        event_type:,
+        author:,
+        heading:,
+        school_partnership:,
+        delivery_partner:,
+        school:,
+        lead_provider:,
+        happened_at: Time.zone.now,
+        metadata:
+      ).record_event!
+    end
+
     # Statement Adjustment Events
 
     def self.record_statement_adjustment_added_event!(author:, statement_adjustment:)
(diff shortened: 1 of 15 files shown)


app/services/school_partnerships/create.rb
```diff
@@ -0,0 +1,119 @@
...
+    validate :school_is_eligible
+    validate :delivery_partner_exists
+    validate :lead_provider_delivery_partnership_exists
+    validate :school_partnership_does_not_already_exists
+    validate :not_school_led
+
+    def create
+      return false unless valid?
+
+      SchoolPartnership.create!(
```
peteryates, line 30: I think we'll want to [record an event here](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/services/events/record.rb) too.

ethax-ross: Added 👍 we don't have any LP events yet so I added a new user type for them.

cwrw: Ooo I need to start adding those to action API tickets 👍

app/services/events/record.rb
```diff
@@ -343,6 +343,32 @@ def self.record_lead_provider_api_token_revoked_event!(author:, api_token:)
...
+      metadata = {
+        contract_period_year: contract_period.year,
+      }
+
+      new(
+        event_type:,
+        author:,
+        heading:,
+        school_partnership:,
+        delivery_partner:,
```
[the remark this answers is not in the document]
ethax-ross: I went with "if its not an option in the initializer it goes in as metadata" - so DP/LP/School are in the initializer but contract period is not. @peteryates is that the right approach?

app/services/school_partnerships/create.rb
```diff
@@ -0,0 +1,121 @@
...
+    def metadata
+      return unless school && contract_period
+
+      @metadata ||= school.contract_period_metadata.find_by(contract_period:)
+    end
+
+    def not_school_led
+      return unless metadata&.induction_programme_choice == "school_led"
+
+      errors.add(:induction_programme_choice, "The school you have entered has not yet confirmed they will deliver DfE-funded training. Contact the school for more information.")
```
cwrw, line 118: new message: `This school has only registered school-led participants. Contact the school for more information.`

cwrw (commented):
very nicely done, just some minor comments

app/services/events/record.rb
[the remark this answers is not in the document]
cwrw: caught up with @peteryates : @peteryates kindly going to add contract period to events, we'll keep as is for now in this pr :)

app/services/school_partnerships/create.rb
```diff
@@ -0,0 +1,121 @@
+# frozen_string_literal: true
```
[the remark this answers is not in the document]
avinhurry: https://github.com/DFE-Digital/register-early-career-teachers-public/pull/1204

cwrw (approved):
past Pete's comment about adding the new lead provider value to the `event_author_types` enum, looks great 🙌

cwrw (approved):

peteryates (approved):

PR #1202, opened 2025-08-21
Add swagger docs for POST /api/v3/partnerships

ethax-ross:
### Context

We want to add the swagger docs for the POST /api/v3/partnerships endpoint.

- Add API documentations for POST /api/v3/partnerships

Add the swagger documentation for POST /api/v3/partnerships.

Use shared context so that we can re-use it for other POST requests.

Add needed response schemas.

### Guidance to review

[Swagger docs](https://cpd-ec2-review-1202-web.test.teacherservices.cloud/api/docs/v3)

Note the attribute names in the error response will be incorrect for the time being; these will be fixed when we address the general API/RECT error message/terminology consistency issues.

### DIFF
--- a/spec/support/shared_contexts/api/docs/api_create_endpoint_documentation_support.rb
+++ b/spec/support/shared_contexts/api/docs/api_create_endpoint_documentation_support.rb
@@ -1,0 +1,48 @@
+RSpec.shared_context "an API create endpoint documentation", :exceptions_app do |options = {}|
+  path options[:url] do
+    post "Creates a #{options[:resource_description]}" do
+      tags options[:tag]
+      consumes "application/json"
+      produces "application/json"
+      security [api_key: []]
+
+      parameter name: :params,
+                in: :body,
+                style: :deepObject,
+                required: false,
+                schema: {
+                  "$ref": options[:request_schema_ref],
+                }
+
+      response "200", "The created #{options[:resource_description]}" do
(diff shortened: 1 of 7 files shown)


spec/support/shared_contexts/api/docs/api_create_endpoint_documentation_support.rb
```diff
@@ -0,0 +1,63 @@
...
+                style: :deepObject,
+                required: false,
+                schema: {
+                  "$ref": options[:request_schema_ref],
+                }
+
+      response "200", "The created #{options[:resource_description]}" do
+        schema({ "$ref": options[:response_schema_ref] })
+
+        # after do |example|
```
cwrw, line 20: not sure why this is commented out?

cwrw (approved):

spec/support/shared_contexts/api/docs/api_create_endpoint_documentation_support.rb
ethax-ross: ah I had brought in [this helper](https://github.com/DFE-Digital/npq-registration/blob/main/spec/support/helpers/swagger_example_parser.rb) that lets us modify examples as I thought it was needed but its not yet, so I pulled it back out and forgot to remove this 🙂 I've removed it now

mooktakim (approved):

leandroalemao (approved):

PR #1205, opened 2025-08-22
CSV upload statuses

peterdavidhamilton:
### Context

Following user testing with admins we need clarification on what status indicators mean.

Update documentation and explain the batch statuses for uploaded CSV files.


cpjmcquillan (approved):
LGTM!

PR #1208, opened 2025-08-22
Speed up migrations with in-memory records caching

leoapost:
### Context
This change introduces a simple in-memory caching layer to improve migration performance by storing frequently accessed data and reducing database round-trips during the migration process.

The migrations run on isolated Kubernetes pods with [4Gi](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/dfc383c4fb7720a6b261dfdc946e65ec494c75a6/config/terraform/application/config/migration.tfvars.json#L6) of memory and a [single thread](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/dfc383c4fb7720a6b261dfdc946e65ec494c75a6/config/queue.yml#L25). Currently, the workers use up to 1.5Gi at most, so there’s ample headroom for in-memory caching. Additionally, thread safety isn’t a concern at this stage due to the single-threaded execution.

- Add a basic CacheManager singleton for centralized data caching
- Add batch caching methods for schools, teachers, lead providers, etc.
- Integrate the cache across all migrators and builders


tonyheadford (approved):
Looks good!

PR #1209, opened 2025-08-22
Add `Partnerships::Update` service

ethax-ross:
### Context

We want to add a service that will enable school partnerships to change delivery partner in order to support the `PUT /api/v3/partnerships/:id` endpoint.

- Add `Partnerships::Update` service

Add service to update the delivery partner for a given partnership.

Send an update event that tracks the change (including the previous delivery partner in the metadata).

In order to perform an update, there must: 

- be a school partnership and delivery partner matching the provided ids.
- be a lead provider delivery partnership for the lead provider and delivery 
partner.
- not already be a partnership with the same lead provider and delivery partner.

### Guidance to review

We don't appear to need to refresh any metadata off the back of a partnership update.

### DIFF
--- a/app/services/events/record.rb
+++ b/app/services/events/record.rb
@@ -369,7 +369,7 @@ module Events
       ).record_event!
     end
 
-    def self.record_school_partnership_updated_event!(author:, school_partnership:, previous_delivery_partner:)
+    def self.record_school_partnership_updated_event!(author:, school_partnership:, previous_delivery_partner:, modifications:)
       event_type = :school_partnership_updated
       school = school_partnership.school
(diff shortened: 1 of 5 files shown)


app/services/events/record.rb
```diff
@@ -369,6 +369,31 @@ def self.record_school_partnership_created_event!(author:, school_partnership:)
       ).record_event!
     end
 
+    def self.record_school_partnership_updated_event!(author:, school_partnership:, previous_delivery_partner:)
```
cwrw, line 372: can we include `modifications` here as well, as that is expected in "update" events?

Issue #1008, opened 2025-07-24
Validate absence of EOI and school partnership when TrainingPeriod.school_lead_training_programme?


ltello:
## Why?
After moving `training_programme` to `TrainingPeriod` from `ECTAtSchoolPeriod` it would be good to add a validation to error in case `expresion_of_interest` or `school_partnership` are not absent when `school_lead_training_programme?` is true.

This way we make sure they are only set for `provider_let_training_programme?` training periods.

## What?
Extend `TrainingPeriod` with a validation to meet the above criteria.

Rmcnaney:
@JSJohal / @StephenC-V1 to review pls

PR #1211, opened 2025-08-22
1008: Validate absence of expression of interest

ghost:
Add validation to ensure `expression_of_interest` and `school_partnership` are absent for `school-led` training programmes, enforcing the business rule that these associations should only exist for provider-led programmes.

## Changes Made:
- TrainingPeriod Model: Added expression_of_interest_and_school_partnership_absent validation

- Schools::TrainingProgramme Service: Extended to detect school-led periods without associations
- Test Factories: Fixed invalid school-led training period creation across multiple specs
- Bug Fix: Corrected contract_period_year usage (was using ID instead of year)

Why These Changes Were Needed
After adding the validation, the Schools::TrainingProgramme service could no longer find school-led training periods because it only looked for periods WITH associations (EOI/partnerships). This would cause school-led training to show as "not_yet_known" in the UI.
Solution: Extended the service's discovery logic to include a third pathway that specifically finds school-led periods without associations.

Fixes: https://github.com/DFE-Digital/register-early-career-teachers-public/issues/1008

### DIFF
--- a/db/seeds/teacher_histories.rb
+++ b/db/seeds/teacher_histories.rb
@@ -141,11 +141,6 @@ ambition_artisan_partnership_2022 = find_school_partnership(
   delivery_partner: artisan_education_group,
   contract_period: ContractPeriod.find_by!(year: 2022)
 )
-ambition_artisan_partnership_2023 = find_school_partnership(
-  lead_provider: ambition_institute,
-  delivery_partner: artisan_education_group,
-  contract_period: ContractPeriod.find_by!(year: 2023)
-)
 teach_first_grain_partnership_2022 = find_school_partnership(
   contract_period: ContractPeriod.find_by!(year: 2022),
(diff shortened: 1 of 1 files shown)


app/models/training_period.rb
```diff
@@ -101,4 +102,14 @@ def only_provider_led_mentor_training
...
     end
   end
+
+  def expression_of_interest_and_school_partnership_absent
+    if expression_of_interest.present?
+      errors.add(:expression_of_interest, 'must be absent for school-led training programmes')
+    end
+
+    if school_partnership.present?
+      errors.add(:school_partnership, 'must be absent for school-led training programmes')
```
peteryates, line 112: Won't this mean the error message we display in the error summary doesn't have the field prefix? I think this should be `'School partnership must be absent for school-led training programme'`.

PR #1217, opened 2025-08-26
Refactor lead_provider association in TrainingPeriod model

### LINKED ISSUE #1009: TrainingPeriod model: association `has_one :lead_provider, through :active_lead_provider` is not taking into account the `lead_provider` through `eoi`
## Why?
A training period can have a `lead_provider` via their `expression_of_interest` and another via its `school_partnership` -->>`active_lead_provider`

Currently the code only defines the later.

## What?
Adjust the code so that the `lead_provider` of a training period can come from any of those associations.

## Helpful links or other information

Refs DFE-Digital/register-ects-project-board#2213

ghost:
Removed the direct association to lead_provider through active_lead_provider and added a custom lead_provider method to determine the lead provider based on the presence of expression_of_interest and school_partnership. Updated tests to cover the new logic for lead_provider retrieval.

fixes: https://github.com/DFE-Digital/register-early-career-teachers-public/issues/1009

### DIFF
--- a/app/models/training_period.rb
+++ b/app/models/training_period.rb
@@ -15,7 +15,6 @@ class TrainingPeriod < ApplicationRecord
   belongs_to :expression_of_interest, class_name: 'ActiveLeadProvider'
   has_one :lead_provider_delivery_partnership, through: :school_partnership
   has_one :active_lead_provider, through: :lead_provider_delivery_partnership
-  has_one :lead_provider, through: :active_lead_provider
   has_one :delivery_partner, through: :lead_provider_delivery_partnership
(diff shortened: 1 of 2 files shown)


peteryates (commented):
Looking at this, I'm not sure it's a good idea.

We had [a chat](https://ukgovernmentdfe.slack.com/archives/C01F7SE8TLY/p1755772119808529) about it and it became clear that there are very few places where we want either, in most of the app we want the _confirmed_ one (via school partnership).

My fault really, I didn't update the issue on the back of the discussion, but I think we're best without the combined `lead_provider`.

PR #1218, opened 2025-08-26
Use `params.expect` to permit and require params

cpjmcquillan:
Since https://github.com/rails/rails/pull/51674, the [`expect`](https://api.rubyonrails.org/classes/ActionController/Parameters.html#method-i-expect) method allows us to `require` and `permit` params in one step.

According to the guides, this is the preferred way to to require and permit parameters.

This updates our controllers to follow this pattern.

What do people think?


peteryates (approved):
TIL.

peteryates:
This totally flew under my radar. I don't think it's common that a suggestion like this is not only accepted, but goes straight to being the recommended pattern in such a short period of time.

avinhurry (commented):
Nice! 

Do you think it's worth enforcing it with :

```ruby
Rails/StrongParametersExpect:
  Enabled: true
```

To prevent the old style being used again?

avinhurry (approved):
Nice, thank you!

PR #1221, opened 2025-08-26
Add swagger docs for PUT /api/v3/partnerships/:id

ethax-ross:
### Context

We want to add the swagger docs for the PUT /api/v3/partnerships/:id endpoint.

- Add API documentations for PUT /api/v3/partnerships/:id

Add the swagger documentation for PUT api/v3/partnerships/:id.

Use shared context so that we can re-use it for other PUT requests.

### Guidance to review

[Swagger docs](https://cpd-ec2-review-1221-web.test.teacherservices.cloud/api/docs/v3)

Note the attribute names in the error response will be incorrect for the time being; these will be fixed when we address the general API/RECT error message/terminology consistency issues.

### DIFF
--- a/spec/support/shared_contexts/api/docs/api_update_endpoint_documentation_support.rb
+++ b/spec/support/shared_contexts/api/docs/api_update_endpoint_documentation_support.rb
@@ -1,10 +1,17 @@
-RSpec.shared_context "an API create endpoint documentation", :exceptions_app do |options = {}|
+RSpec.shared_context "an API update endpoint documentation", :exceptions_app do |options = {}|
   path options[:url] do
-    post "Create a #{options[:resource_description]}" do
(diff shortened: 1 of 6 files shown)


spec/support/shared_contexts/api/docs/api_update_endpoint_documentation_support.rb
```diff
@@ -0,0 +1,65 @@
+RSpec.shared_context "an API update endpoint documentation", :exceptions_app do |options = {}|
+  path options[:url] do
+    put "Updates a #{options[:resource_description]}" do
```
leandroalemao, line 3: Hi @ethax-ross for consistency.. should we use 3rd person singular here or as just as the other endpoints that use "Retrieve" instead?

PR #1236, opened 2025-08-27
[1249] Don't show lead provider row  if `!mentoring_at_new_school_only?` (#MentorRegistrationFlow) 

avinhurry:
### Changes

- Removed the `change_mentoring_at_new_school` step:
  - Changing `mentoring_at_new_school` answer from **Yes -> No** now redirects to the additional question instead of CYA.
  - Changing answer from **No -> Yes** returns directly to the CYA page.

- Updated CYA behaviour:
  - The **Lead provider** row is hidden when the answer to *"Mentoring only at your school?"* is **No**, since the lead provider may not be accurate in this case.


Rmcnaney:
Ready for review plssss

PR #1237, opened 2025-08-28
Update individual school page in the admin console

ghost:
## Description
This story is to update the admin console page for an individual school's details & teachers at the school. 

This updates the existing placeholder page when you click into a school (e.g. [this one](https://cpd-ec2-staging-web.test.teacherservices.cloud/admin/schools/1759427))

## User story

**As** an internal DfE user
**I need** to be able to view an individual school's details
**So that** I can find and check the details or teachers I'm looking for and make changes as needed

## Prototype, screenshots & content
[Link to prototype](https://ecf2-prototype-e3c7aba62d82.herokuapp.com/v14/admin/organisations/school-details?section=overview)

Screenshots of relevant screens:

Image

### DIFF
--- a/app/controllers/admin/schools/partnerships_controller.rb
+++ b/app/controllers/admin/schools/partnerships_controller.rb
@@ -3,23 +3,12 @@ module Admin
     class PartnershipsController < AdminController
       layout "full"
 
-      before_action :setup_school_page
-      before_action :setup_navigation_items
-
       def show
-      end
-
-    private
-
-      def setup_school_page
         @school = School.includes(:gias_school).find_by!(urn: params[:school_urn])
         @breadcrumbs = {
           "Schools" => admin_schools_path(page: params[:page], q: params[:q]),
           @school.name => nil
         }
-      end
-
-      def setup_navigation_items
         @navigation_items = helpers.admin_school_navigation_items(params[:school_urn], request.fullpath)
       end
     end
--- a/app/controllers/admin/schools/overviews_controller.rb
+++ b/app/controllers/admin/schools/overviews_controller.rb
@@ -3,23 +3,12 @@ module Admin
     class OverviewsController < AdminController
       layout "full"
 
-      before_action :setup_school_page
-      before_action :setup_navigation_items
-
       def show
-      end
-
-    private
-
-      def setup_school_page
         @school = School.includes(:gias_school).find_by!(urn: params[:school_urn])
         @breadcrumbs = {
           "Schools" => admin_schools_path(page: params[:page], q: params[:q]),
           @school.name => nil
         }
-      end
-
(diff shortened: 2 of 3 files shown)


spec/features/admin/schools_spec.rb
peteryates, line 1: As this page doesn't really _do_ anything, could these be tested in the view spec?

I think we might be best saving features for multi-page journeys.

app/views/admin/schools/show.html.erb
```diff
@@ -1,169 +1,21 @@
...
-end %>
+<% page_data(
+  title: @school.name,
+  caption: "URN: #{@school.urn}",
+  caption_size: "xl",
+  header_size: "xl",
+  breadcrumbs: @breadcrumbs
+) %>
+
+<%= govuk_tabs(title: nil) do |tabs| %>
```
peteryates, line 9: Actually I just realised that the design doesn't have tabs here, it uses the secondary nav pattern as per this screenshot. I think we're best avoiding tabs if pos and using it (splitting the below bits onto three separate pages)

image

app/components/admin/schools/overview_component.rb
```diff
@@ -0,0 +1,30 @@
...
+    if address_lines.any?
+      safe_join(address_lines, tag.br)
+    else
+      "Not available"
+    end
+  end
+
+private
+
+  attr_reader :school
```
peteryates, line 29: Does it need to be private and tucked away at the bottom? I don't think there's any danger of it being misused.

app/assets/stylesheets/admin.scss
```diff
@@ -118,3 +118,63 @@
   margin: 0;
   padding: govuk-spacing(2) govuk-spacing(3) govuk-spacing(2) 0;
 }
+
+// X-GOVUK Secondary Navigation
+// Based on https://x-govuk.github.io/govuk-prototype-components/secondary-navigation/
+.x-govuk-secondary-navigation {
```
peteryates, line 124: I think we should be using this bit, the other visual difference I mentioned works when we use this directly:

https://github.com/x-govuk/govuk-prototype-components/blob/283481392bd4cd6e39d1a4b10058f25336917e52/dist/govuk-prototype-components.scss#L173C1-L238

app/controllers/admin/schools/partnerships_controller.rb
```diff
@@ -0,0 +1,27 @@
+module Admin
+  module Schools
+    class PartnershipsController < AdminController
+      layout "full"
+
+      before_action :setup_school_page
+      before_action :setup_navigation_items
```
peteryates, line 7: Is there any advantage in using `before_action` when there's only one method? :sweat_smile:

PR #1239, opened 2025-08-28
Optimise query services and serialisation

ethax-ross:
> ⚠️ WIP

### Context

Currently our query services are responsible for preloading the relevant relationships that we reference during serialisation. This has two issues:

1. It's slow to preload and paginate
2. Queries should be generic/not coupled to a particular use case

We want to address this to make the queries faster and the more generic/less coupled to the serialisation.

- Optimise query services and serialization

Pull out any unnecessary `includes` from the query services and instead passes the output from a query to a class method on the relevant serializer to perform specific optimisations.

This means we only do the heavy lifting on the result of the pagination and also places the optimisations relevant to the serializer in the serializer itself, so its clearer when it should need to change.

### Guidance to review

As part of this change we need to move back to `pagy` instead of `pagy_countless`. This is because `pagy_countless` internally calls `to_a`, unlike `pagy` which returns the ActiveRecord collection. We could get around this by plucking the `id` in the `pagy_countless` result and then re-querying, but in my initial testing this doesn't make much of a performance difference.


app/services/delivery_partners/api/query.rb
```diff
@@ -0,0 +1,19 @@
+module DeliveryPartners::API
```
cwrw, line 1: I wonder if it might be better to move this subclass and others to `api/#model_name/query` as that would sit nicely with the existing structure?

leandroalemao (approved):

app/services/api/school_partnerships/query.rb
```diff
@@ -20,72 +20,69 @@ def initialize(school_id: :ignore, contract_period_years: :ignore, lead_provider
...
 
       fail(ArgumentError, "id needed")
     end
 
   private
 
+    def preload_associations(results)
+      results
+        .strict_loading
+        .includes(:delivery_partner, :active_lead_provider, school: :gias_school)
```
cwrw, line 43: do we not need `lead_provider`?

leandroalemao (approved):
All good @ethax-ross 👍 PR needs rebasing I guess.. tks

PR #1240, opened 2025-08-28
Custom bulk upload event for DfE Analytics

peterdavidhamilton:
### Context

The Data team require a way to connect `Teacher` and `InductionPeriod` records edited via the RIAB CSV bulk upload feature in BigQuery.

Use a `DfE::Analytics` custom event to send a subset of `Event` record attributes.

### Guidance to review

image

Streaming to BigQuery has been turned on temporarily. Use an Appropriate Body persona and complete a CSV upload.

### DIFF
--- a/app/jobs/analytics_batch_job.rb
+++ b/app/jobs/analytics_batch_job.rb
@@ -1,0 +1,35 @@
+# Send selected bulk upload Event records to DfE Analytics that connect Teachers to their Induction Periods.
+class AnalyticsBatchJob < ApplicationJob
+  queue_as :dfe_analytics
+
+  # @return [Array<Symbol>]
+  EVENT_FIELDS = %i[
+    author_type
+    created_at
+    event_type
+    happened_at
+    id
+    induction_period_id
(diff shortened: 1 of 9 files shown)


peterdavidhamilton:
@mooktakim how does this look to you?

app/jobs/analytics_batch_job.rb
```diff
@@ -0,0 +1,35 @@
...
+      .select(*EVENT_FIELDS)
+
+    events.map do |db_event|
+      data = db_event.attributes
+      event = DfE::Analytics::Event.new
+                                   .with_type(:bulk_upload_action)
+                                   .with_entity_table_name(:bulk_upload_actions)
+                                   .with_data(data:)
+
+      DfE::Analytics::SendEvents.do(Array.wrap(event.as_json))
```
mooktakim, line 32: I recommend moving out the send event parts into a service class similar to:
https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/services/api_request.rb

That way its easier to keep track of all the dfe analytics events we send.

PR #1242, opened 2025-08-28
Add parity-check environment

ethax-ross:
> ⚠️ WIP

### Context

We often want to use the migration environment for running parity checks as part of the API development. This migration environment is becoming quite contested, so we want to add a dedicated parity check environment.

- Add parity-check environment

Add a new Rails environment for running parity checks, along with the necessary terraform/infrastructure.

This will be the same as the migration environment in all but name.


.github/workflows/refresh_parity_check_database.yml
```diff
@@ -0,0 +1,71 @@
+name: Refresh parity check DB from production DB
```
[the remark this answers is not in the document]
ethax-ross: Updated and tested here https://github.com/DFE-Digital/register-early-career-teachers-public/actions/runs/17576713145

leandroalemao (approved):
All done 👏 tksss

.github/actions/refresh-database/action.yml
```diff
@@ -77,6 +80,6 @@ runs:
 
         bin/konduit.sh -n ${{ env.NAMESPACE }} cpd-ec2-${{ inputs.environment }}-web -- pg_dump -E utf8 --compress=1 --clean --if-exists $EXCLUDE_ARGS --no-privileges --no-owner --verbose -f backup-${{ inputs.environment }}.sql.gz
 
-    - name: Restore to migration DB
+    - name: Restore to ${{ inputs.target_environment }} DB
       shell: bash
-      run: bin/konduit.sh -n ${{ env.NAMESPACE }} -d s189p01-cpdec2-mg-pg -k s189p01-cpdec2-mg-app-kv -i backup-${{ inputs.environment }}.sql.gz -c -t 7200 cpd-ec2-migration-web -- psql
+      run: bin/konduit.sh -n ${{ env.NAMESPACE }} -d s189p01-cpdec2-mg-pg -k s189p01-cpdec2-mg-app-kv -i backup-${{ inputs.environment }}.sql.gz -c -t 7200 cpd-ec2-${{ inputs.target_environment }}-web -- psql
```
cwrw, line 85: I think we need to switch up mg with pc here depending on env

cwrw: I think what is there now should be ok, let's keep as is :)

PR #1250, opened 2025-08-29
Split `SchoolPartnerships::Query` to `SchoolPartnerships::Search`

peteryates:
We originally wanted to keep searching/querying generic and have a single set of services for everything, but the API queries differ so much from the rest of the app it makes sense to separate them.

We unintentionally created a divide in naming with the app mainly using `Search` and the API using `Query` - the one place where there were both was `SchoolPartnerships` - hence this clash.

This change copies `SchoolPartnerships::Query` and to `SchoolPartnerships::Search` and simpifies the `Search` variant by removing the API-specific stuff (and reinstating the `.earliest_first` that was dropped in bc5da359d1e4aaa29ed51d41447607e0f9a72adc).

The `Query` objects will move under the API namespace which should hopefully avoid confusion.

### DIFF
--- a/app/models/concerns/training_period_sources.rb
+++ b/app/models/concerns/training_period_sources.rb
@@ -10,12 +10,7 @@ module TrainingPeriodSources
   end
 
   def earliest_matching_school_partnership
-    SchoolPartnerships::Search.new(
-      school:,
-      lead_provider:,
-      contract_period:,
-      sort: "created_at"
-    ).school_partnerships.first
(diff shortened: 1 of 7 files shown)


leandroalemao:
Hi @peteryates Think we've got a pattern where we send ids to the query services as per the other query services used in API? Any special reason to move it to also allow objects in the params? should we make changes to all other query services for this new pattern for consistency? tksss

PR #1256, opened 2025-09-01
Add statement search service

cwrw:
### Context
After chatting to @peteryates we've agreed to move the query services to only be used by the API. We however use statements query for the admin currently, so we need to create its own `search` service instead

- Remove admin only logic from `Statement::Query` as its only for the API
- Add a new `Statement::Search` service which is targeted only towards the admin interface, used https://github.com/DFE-Digital/register-early-career-teachers-public/pull/777/files#diff-6444f0bb86900e702fad0a69e08e91a93df5dca42676b515bb8be929bf1fd29f for guidance
- Use the new service in admin interface

### Guidance to review
Can be tested [here](https://cpd-ec2-review-1256-web.test.teacherservices.cloud/admin/finance/statements)

### DIFF
--- a/app/controllers/admin/finance/statements_controller.rb
+++ b/app/controllers/admin/finance/statements_controller.rb
@@ -46,7 +46,7 @@ module Admin::Finance
         contract_period_years:,
         statement_date:,
         fee_type:,
-        order_by: :statement_date,
+        order: :statement_date,
       }
       Statements::Search.new(**opts.compact)
     end
--- a/app/services/statements/search.rb
+++ b/app/services/statements/search.rb
@@ -7,14 +7,14 @@ module Statements
 
     attr_reader :scope
 
-    def initialize(lead_provider_id: :ignore, contract_period_years: :ignore, fee_type: 'output', statement_date: :ignore, order_by: :payment_date)
+    def initialize(lead_provider_id: :ignore, contract_period_years: :ignore, fee_type: 'output', statement_date: :ignore, order: :payment_date)
       @scope = Statement.distinct.includes(active_lead_provider: %i[lead_provider contract_period])
 
       where_lead_provider_is(lead_provider_id)
       where_contract_period_year_in(contract_period_years)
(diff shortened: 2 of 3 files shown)


cwrw:
@leandroalemao planning on doing that in a subsequent pr :)

app/services/statements/search.rb
```diff
@@ -0,0 +1,65 @@
...
+    def where_lead_provider_is(lead_provider_id)
+      return if ignore?(filter: lead_provider_id)
+
+      scope.merge!(Statement.joins(:lead_provider).where(lead_providers: { id: lead_provider_id }))
+    end
+
+    def where_contract_period_year_in(contract_period_years)
+      return if ignore?(filter: contract_period_years)
+
+      scope.merge!(Statement.joins(:contract_period).where(contract_periods: { year: extract_conditions(contract_period_years) }))
```
ethax-ross, line 35: I don't feel that strongly about it, but I find `merge!` less natural than just re-assigning the scope:

```
@scope = scope.joins(:contract_period).where(contract_periods: { year: extract_conditions(contract_period_years) }
```

I may have changed all these in the API queries on my other PR 🙄

app/controllers/admin/finance/statements_controller.rb
```diff
@@ -46,9 +46,9 @@ def statements_query
         contract_period_years:,
         statement_date:,
         fee_type:,
-        sort: "+year,+month",
+        order_by: :statement_date,
```
ethax-ross, line 49: Is it worth passing the ordering as well here to be consistent with the route we're taking in the API? so `order_by: { statement_date: :asc }`

ethax-ross (approved):
Looks good, two take it or leave it comments

app/services/statements/search.rb
cwrw: updated :)

ethax-ross (approved):

PR #1257, opened 2025-09-02
bugfix: do not call undefined method on training period

ghost:
This broke the ECT details page:
https://dfe-teacher-services.sentry.io/issues/6851355260/?project=4508369974788096&query=is%3Aunresolved&referrer=issue-stream

### DIFF
--- a/app/models/training_period.rb
+++ b/app/models/training_period.rb
@@ -46,7 +46,6 @@ class TrainingPeriod < ApplicationRecord
   # Delegations
   delegate :name, to: :delivery_partner, prefix: true, allow_nil: true
   delegate :name, to: :lead_provider, prefix: true, allow_nil: true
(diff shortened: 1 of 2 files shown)


app/models/training_period.rb
```diff
@@ -46,6 +46,7 @@ class TrainingPeriod < ApplicationRecord
   # Delegations
   delegate :name, to: :delivery_partner, prefix: true, allow_nil: true
   delegate :name, to: :lead_provider, prefix: true, allow_nil: true
+  delegate :name, to: :expression_of_interest_lead_provider, prefix: :expression_of_interest_lead_provider, allow_nil: true
```
peteryates, line 49: Would it be more straightforward to just call `training_period.expression_of_interest_lead_provider.name` in the component, than to add another delegation?

mason-emily (approved):

PR #1260, opened 2025-09-02
Switch metadata refreshing to callbacks

ethax-ross:
### Context

We are finding that remembering to update the metadata is difficult and often makes things like PO review (where we manually create data in certain states) a pain.

We are switching to use active record callbacks in order to trigger the metadata refreshes, which should be a more robust solution.

- Add DeclarativeMetadata concern

Similar to the `DeclarativeTouch` concern, this will allow us to trigger a metadata refresh for a given object(s) off the back of model changes.

-  Filter changed files

Add DeclarativeMetadata to models Trigger metadata refreshes when models change that feed into the metadata
attribute values.

Casting a wide net initially, if we find that we are issuing too many refreshes we can be more granular with the refreshing (off specific attribute changes).

We don't need to worry about over-refreshing metadata resulting in the `api_updated_at` being touched as the service only touches the underlying data if it has changed.

- Remove metadata refresh from SchoolPartnerships::Create

We don't need this any longer as the callbacks will take care of it. 

- Remove with_metadata factory traits

We no longer need these as the metadata is created via callbacks.

You may run into an issue where the metadata is not created if you create the object under test before the contract period and/or lead provider as we currently do not have callbacks in these models. In this scenario you either need to be explicit in the order you create your fixtures or call `Metadata::Manager.refresh_all_metadata!` in a `before` block.

- Remove transient references from TrainingProgramme

We no longer provide these values from the query so they are redundant.

### Guidance for review

The ticket mentions about skipping metadata for migration, but I'm going to look at that properly in the follow up ticket that focuses on it.


app/models/concerns/declarative_metadata.rb
```diff
@@ -0,0 +1,15 @@
...
+  class_methods do
+    def refresh_metadata(target, when_changing: [], on_event: %i[update])
+      after_commit(on: on_event) do
+        should_touch = when_changing.blank? || when_changing.any? do |attr|
+          saved_change_to_attribute?(attr)
+        end
+
+        Metadata::Manager.new.refresh_metadata!(instance_exec(&target)) if should_touch
+      end
+    end
```
mooktakim, line 13: Based on: https://github.com/DFE-Digital/register-early-career-teachers-public/pull/1260/files#diff-07f753084c94c0d0d5823e7662a1400fd534a80bcac415fb6a1c0c60c3573fdaR30

`instance_exec(&target) ` could potentially return nil, i think we will need to check for that here before running `refresh_metadata!`

app/services/schools/training_programme.rb
```diff
@@ -32,14 +32,10 @@ def mentors_at_school_periods
     end
 
     def mentors_at_school?
-      return school.transient_mentors_at_school if school.respond_to?(:transient_mentors_at_school)
```
cwrw, line 35: I thought they were just in the tests by accident, I missed the service itself oops

app/models/concerns/declarative_metadata.rb
```diff
@@ -0,0 +1,15 @@
...
+  extend ActiveSupport::Concern
+
+  class_methods do
+    def refresh_metadata(target, when_changing: [], on_event: %i[update])
+      after_commit(on: on_event) do
+        should_touch = when_changing.blank? || when_changing.any? do |attr|
+          saved_change_to_attribute?(attr)
+        end
+
+        Metadata::Manager.new.refresh_metadata!(instance_exec(&target)) if should_touch
```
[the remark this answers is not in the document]
cwrw: good question, I don't think this will affect anything live right now so let's keep as is and monitor, and possibly go with option 3 if its an issue. Good shout

cwrw (approved):
nicely done

leandroalemao (approved):

tonyheadford (approved):
Had a chat about naming and namespacing with @ethax-ross and happy that this has been thought about

PR #1264, opened 2025-09-02
Schools can change an ECT's email address

cpjmcquillan:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/2098

This introduces a wizard for updating an ECT's email address.

This wizard leans on the wizards that came before it, but tries to be as lightweight as possible.

Updates are stored in the session until the User confirms the change (after which the ECT record is updated).

This is one of many attributes schools will be able to be update about ECTs, which will ensure any errors or changes can be made autonomously by end users via the Teacher details page.

### Guidance to review

- [ ] Acceptance criteria satisfied
- [ ] Content/designs match prototype


config/routes.rb
```diff
@@ -272,6 +272,16 @@
 
       get "confirmation", action: :new
     end
+
+    scope module: :ects, path: "/ects/:ect_id", as: :ects do
```
peterdavidhamilton, line 276: I think I broke with convention and that's the cause of my `step_path` issue. I like how you've done it, and if you're including the new module, I should too.

cpjmcquillan: I vaguely remember running into path issues and having to tweak something and then it worked. I can't remember exactly what I did though, I think it was adding `as: :ects` rather than `as: :ect`

app/controllers/schools/ects/change_email_address_wizard_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+          step_params: params,
+          store:,
+          school: @school,
+          ect: @ect
+        )
+      end
+
+      def current_step
+        @current_step ||= request.path.split("/").last.underscore.to_sym.tap do |step_from_path|
+          redirect_to "/404" unless WIZARD_CLASS.step?(step_from_path)
```
peteryates, line 41: Worth adding an `as: :not_found` to the route so we can `redirect_to not_found_path`?

config/routes.rb
avinhurry: Could we keep school singular in the path? The rest of our routes (like `/school/ects/:id`) use the singular form, so `/schools` here feels inconsistent and a bit misleading.

cpjmcquillan: I put this here because it's where the other ECT wizard routes lived.

Is there a reason we have some `/schools` routes and some `/school` routes in the `Schools` namespace?

Could we align them all into `/school`? 🤔

avinhurry: Good question! I think singular makes sense when it's referring to a single school. Looks like [schools was renamed to school ](https://github.com/DFE-Digital/register-early-career-teachers-public/commit/7b7ae154b13615e58959e4c19d0d34a8524687ac)a while back to avoid clashes with the model, but we can still set the path manually.

FWIW, the `:register_mentor_wizard` journey already uses `/school` in the path, e.g `/school/register-mentor/email-address`, even though it’s in the `Schools::` namespace. So there's precedent for this approach. It's the same for `/school/ects/:id`.  

This might be beyond the scope of this PR but the reason I noticed is because the URL goes from `/school/ects/:id` to `schools/ects/:id/change-email-address/edit`.

Do you know any reason this change wouldn't be sensible @Peter?

spec/features/schools/ects/change_email_address_spec.rb
```diff
@@ -0,0 +1,125 @@
+describe "School user can change ECTs email address" do
+  before do
+    allow(Rails.application.config)
+      .to receive(:enable_schools_interface)
+      .and_return(true)
+  end
```
avinhurry, line 6: I don't this is needed, the helper should handle it

spec/requests/schools/ects/change_email_address_wizard_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+      school:,
+      email: "ect@example.com"
+    )
+  end
+
+  before do
+    allow(Rails.application.config)
+      .to receive(:enable_schools_interface)
+      .and_return(true)
+  end
```
avinhurry, line 18: Don't think we need this?

spec/features/schools/ects/change_email_address_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+    and_i_continue
+    then_i_see_an_error("The email must be different from the current email")
+
+    when_i_enter_an_invalid_email_address
+    and_i_continue
+    then_i_see_an_error("Enter an email address in the correct format, like name@example.com")
+
+    when_i_enter_an_email_address_that_is_too_long
+    and_i_continue
+    then_i_see_an_error("Enter an email address that is less than 254 characters long")
```
avinhurry, line 31: Our specs are getting slower and slower every day, could we create step specs and test the validations in there please? It should help keep our feature tests focused and faster 🙏🏽

app/wizards/schools/ects/change_email_address_wizard/check_answers_step.rb
```diff
@@ -0,0 +1,29 @@
...
+        def previous_step = :edit
+        def next_step = :confirmation
+
+        def current_email = ect_at_school_period.email
+        def new_email = store.email
+
+        def save!
+          ApplicationRecord.transaction do
+            ect_at_school_period.update!(email: new_email)
+            Events::Record.record_teacher_email_changed_event!(
```
avinhurry, line 14: The timeline wording looks a bit off. It says "unchanged" even though I changed it 🤔

image

app/views/schools/ects/change_email_address_wizard/edit.html.erb
```diff
@@ -0,0 +1,33 @@
...
+  headteacher@school.com
+</p>
+
+<%= form_with model: @wizard.current_step, url: @wizard.current_step_path do |form| %>
+  <%= content_for(:error_summary) { form.govuk_error_summary } %>
+
+  <%= form.govuk_email_field(
+        :email,
+        label: {text: "What’s the correct email address for #{@wizard.teacher_full_name}"},
+        value: @wizard.current_step.email
```
avinhurry, line 21: If we go back from the check your answers page, should the email address not be the new one that it was changed to?

This may be one that we need to double check with Jas and card up because I think the name one behaves the same way.

config/routes.rb
```diff
@@ -294,6 +294,16 @@
       end
     end
 
+    scope module: :ects, path: "/ects/:ect_id", as: :ects do
+      namespace :change_email_address_wizard, path: "change-email-address" do
```
peterdavidhamilton, line 298: We should consolidate the change wizard namespaces into a single module scope. The first one is defined above this block.

spec/features/schools/ects/change_email_address_spec.rb
peterdavidhamilton: @cpjmcquillan you might be able to copy the ones for "change name" or turn them into shared bits-n-bobs.

spec/requests/schools/ects/change_email_address_wizard_spec.rb
peterdavidhamilton: The flag is used in a guard clause now we have established multiple policies and a school user role.

app/wizards/schools/ects/change_email_address_wizard/check_answers_step.rb
cpjmcquillan: This happened because the old email was actually the new email (we record the event after updating).

I've added a unit test and ensured we pass the correct email 👍🏼

spec/features/schools/ects/change_email_address_spec.rb
cpjmcquillan: I've added some unit tests for the steps.

Personally I think feature tests that touch the unhappy paths are useful, but I've removed them 👌🏼

avinhurry (approved):
Looks really clean, nice work!

PR #1265, opened 2025-09-02
Upgrade to support ViewComponent v4

peteryates:
### Context

ViewComponent now mandates that helpers are included. So we don't need to do this across all the components I introduced a base class, the `RECTComponent`, which includes the helpers from `govuk-components` we use throughout.

- **Add ApplicationComponent**
- **Make all components inherit from RECTComponent**
- **Test with prerelease govuk-components**


avinhurry (commented):
There are a few components that include `GovukLinkHelper`, shall we remove those?

peteryates:
Ah interest, yeah that'd make sense.

avinhurry (approved):
Happy to clear those out if you like so we can get this moving

peteryates:
@dependabot rebase

Issue #273, opened 2025-02-27
Add feasible steps functionality to the Register ECT journey


ltello:
Even though each step of the Register ECT journey has a previous_step and next_step logic included to redirect the user to the appropriate next step, nothing prevents them from redirect their browser to any random step at any time or navigate though them via the back and forward buttons of the browser.

We need therefore a way to be able to redirect the user's browser to the appropriate next step based on the overall state in the journey and the currently visited step.

The suggestion in this ticket is to extend the journey with functionality so that it can compute the list of allowed steps (steps we consider 'visitable') based on the current overall state in the journey, the actually visited step and automatically redirect the user to one of those allowed steps if the current one is not included in that list.

mason-emily:
Need to see the other PR to be able to size this - but it is ready to pick up as is

claire-hughez:
Jack is busy on support but addressing comments where he can.

mason-emily:
Reviewed by Lorenzo - Jack is addressing comments

Rmcnaney:
Chase Lorenzo today. Review in stand up tomorrow if he has reviewed or not. If not Check in with Pete to see if another Dev can PR this as it's currently blocked.

Rmcnaney:
Have chased Lorenzo/Gen to understand when this could be picked up.

Rmcnaney:
Few things to finish off and will be ready for dev review for another dev/PR review.

PR #1271, opened 2025-09-03
Add feasible steps functionality to register ECT wizard

ghost:
### Context
Even though each step of the Register ECT journey has a previous_step and next_step logic included to redirect the user to the appropriate next step, nothing prevents them from redirect their browser to any random step at any time or navigate though them via the back and forward buttons of the browser.

We need therefore a way to be able to redirect the user's browser to the appropriate next step based on the overall state in the journey and the currently visited step.

Extended the journey with functionality so that it can compute the list of allowed steps (steps we consider 'visitable') based on the current overall state in the journey, the actually visited step and automatically redirect the user to one of those allowed steps if the current one is not included in that list.

### Fixes
https://github.com/DFE-Digital/register-early-career-teachers-public/issues/273

### DIFF
--- a/app/wizards/schools/register_ect_wizard/wizard.rb
+++ b/app/wizards/schools/register_ect_wizard/wizard.rb
@@ -117,20 +117,19 @@ module Schools
         end
 
         unless school.last_programme_choices? && ect.use_previous_ect_choices
-          if school.independent?
-            steps << :independent_school_appropriate_body
-            return steps unless [ect.appropriate_body_id, ect.appropriate_body_type].all?
-          else
-            steps << :state_school_appropriate_body
-            return steps unless ect.appropriate_body_id
-          end
+          steps << if school.independent?
+                     :independent_school_appropriate_body
+                   else
+                     :state_school_appropriate_body
+                   end
+          return steps unless ect.appropriate_body_id
 
           steps << :training_programme
           return steps unless ect.training_programme
 
           if ect.provider_led?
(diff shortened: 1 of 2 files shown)


peteryates:
This PR is starting to feel like it's nearing completion. @ltello, @edujackedu maybe we can get it product reviewed/merged and if there are any outstanding improvements we can make maybe add them to the backlog?

ltello (commented):
Added commit with suggested changes. Please review

app/wizards/schools/register_ect_wizard/wizard.rb
```diff
@@ -54,9 +55,137 @@ def self.step?(step_name)
...
+        allowed_steps.include?(step_name)
+      end
+
       def ect
         @ect ||= ECT.new(store)
       end
+
+    private
+
+      def calculate_allowed_steps
```
peteryates, line 72: This method is quite heavy and unless I'm mistaken, it's recreating logic from the `Step` classes?

Clearly we're too invested to do anything now but there _must_ be a better way of structuring this. If you have any ideas on a better approach @edujackedu, @ltello, please raise something for the backlog.

peteryates (approved):
Code-wise this looks fine to me but I think `#calculate_allowed_steps` way too heavy and we should have a look at improving this post MVP. Perhaps it's a limitation of this style of wizard. It's probably not helped by treating the insert/change steps separately.

PR #1283, opened 2025-09-04
Allow ECT transfers between schools

ghost:
and Automatically close periods when an ECT transfers

- Updated the `already_registered_as_an_ect?` method to check for ongoing ECT periods at the current school.
- Add Teachers::CloseOldSchoolPeriods service to handle period closure
- Update Schools::RegisterECT to close old periods before creating new ones
- Modify ECTAtSchoolPeriod validation to allow school transfers
- Add test coverage for transfer scenarios
- Ensure proper event recording

This enables ECTs to transfer between schools by automatically closing
ongoing periods at previous schools while preventing duplicate
registrations at the same school.

### DIFF
--- a/app/services/schools/register_ect.rb
+++ b/app/services/schools/register_ect.rb
@@ -117,7 +117,7 @@ module Schools
     def close_ongoing_ect_period!
       return unless teacher
 
-      ongoing_period = teacher.ect_at_school_periods.ongoing.started_before(started_on).first
+      ongoing_period = teacher.ect_at_school_periods.ongoing.started_on_or_before(started_on).first
       return unless ongoing_period
 
       ECTAtSchoolPeriods::Finish.new(
--- a/app/models/concerns/interval.rb
+++ b/app/models/concerns/interval.rb
@@ -12,6 +12,7 @@ module Interval
     scope :earliest_first, -> { order(started_on: 'asc') }
     scope :latest_first, -> { order(started_on: 'desc') }
     scope :started_before, ->(date) { where(started_on: ...date) }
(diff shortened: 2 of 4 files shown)


app/services/schools/register_ect.rb
```diff
@@ -60,7 +61,12 @@ def register!
   private
 
     def already_registered_as_an_ect?
-      ::Teacher.find_by_trn(trn)&.ect_at_school_periods&.exists?
+      # Check if ECT is already registered at THIS school to prevent duplicates
+      # School transfers to different schools are allowed
+      teacher = ::Teacher.find_by_trn(trn)
+      return false unless teacher
```
peteryates, line 67: I don't know if it makes sense to do this here, if there's no `teacher`, `#create_teacher!` should `fail`.

peteryates: Yeah, I think it can go. If we get here with no `teacher` we have bigger problems! :smile:

app/services/schools/register_ect.rb
```diff
@@ -112,6 +114,19 @@ def update_school_last_choices!
       school.update!(**choices.compact)
     end
 
+    def close_ongoing_ect_period!
+      return unless teacher
+
+      ongoing_period = teacher.ect_at_school_periods.ongoing.started_before(started_on).first
+      return unless ongoing_period
```
peteryates, line 121: I think we can make this more concise using a `has_one`, like the [has one ongoing_induction_period](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/models/teacher.rb#L20) relationship.

For now though let's get this merged and then increment.

PR #1285, opened 2025-09-04
Refresh metadata after a data migration

ethax-ross:
### Context

We want to prevent any metadata from being created during a data migration, as it will lead to a slower migration time and multiple changes to the same object over the duration of a run. It's going to be more performant to instead trigger a 'refresh all metadata' once the data migration has completed.

-  Clear metadata on `LegacyDataImporter.reset!`

To be consistent with the other migration data, we want to clear out the metadata when the migration is reset.

- Kck off metadata refresh after migration finishes

We want to populate the metadata once a migration has finished, but we don't want to block/slow the migration process down.

Kick off an async metadata refresh after migrating all data.

- Prevent metadata from being created during data migration

We want to stop metadata from being created during a migration as we refresh all the metadata at the end (which is more efficient/happens in a background job).

Add a method to `Metadata::Manager` that accepts a block; for code executed in the block no metadata will be created.

Wrap the contents of the `migrate` method in the `Migrators::Base` to ensure no metadata is created.

### Guidance to review

Tested by running on migration and checking metadata tables are empty after the delivery partner and school migrators have ran. Checked at the end to ensure the refresh all metadata is kicked off.

### DIFF
--- a/app/migration/legacy_data_importer.rb
+++ b/app/migration/legacy_data_importer.rb
@@ -14,6 +14,8 @@ class LegacyDataImporter
     # we want to do this?
     DataMigration.all.find_each(&:destroy!)
 
+    Metadata::Manager.destroy_all_metadata!
+
     Migrators::Base.migrators_in_dependency_order.reverse.each(&:reset!)
   end
 end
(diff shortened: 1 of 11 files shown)


mooktakim (approved):
Other than comment from abeer, looks good 👍

PR #1310, opened 2025-09-08
[1249] Hide lead provider row if ineligible for funding (#MentorRegistrationFlow)

avinhurry:
### Context

Hide the lead provider row on the check your answers page when ineligible for funding.  

This question is not asked in this pathway, so the row should not be displayed.

| Before | After |
|--------|-------|
|image|image|

### DIFF
--- a/app/views/schools/register_mentor_wizard/check_answers.html.erb
+++ b/app/views/schools/register_mentor_wizard/check_answers.html.erb
@@ -52,7 +52,7 @@
     end
   end
 
-  if @mentor.provider_led_ect? && @mentor.mentoring_at_new_school_only?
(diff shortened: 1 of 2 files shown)


app/views/schools/register_mentor_wizard/check_answers.html.erb
```diff
@@ -52,7 +52,7 @@
     end
   end
 
-  if @mentor.provider_led_ect? && @mentor.mentoring_at_new_school_only?
+  if @mentor.provider_led_ect? && @mentor.mentoring_at_new_school_only? && @mentor.funding_available?
```
peteryates, line 55: It feels like there's quite a lot of business logic wrapped up in this template, perhaps once we're done with this journey we can look at reorganising `mentor.rb` a bit (perhaps splitting it up) so the logic is easy to get at from a testing POV?

PR #1319, opened 2025-09-09
Render array columns explicitly as `array[<type>]` in Mermaid ERD

avinhurry:
### Context

Fix Mermaid ERD generator so that array columns are rendered explicitly as `array[<type>]`.

- Continue to use Rails abstract types via `column.type`.
- Add explicit handling for array columns so they render as `array[<type>]`.
  - Example: `array[integer] contract_period_years`
  - Example: `array[string] tags`

### Why are these changes being made

Previously, array columns were rendered with only their base type (e.g. `integer`), which lost the fact that the column was an array. This made the generated ERD misleading for models that use Postgres array types.


cpjmcquillan (approved):

spec/lib/mermaid_erd/generator_spec.rb
```diff
@@ -73,6 +71,47 @@
...
+        define_test_model(model_name: 'TestVarcharArrayModel', table_name: 'test_varchar_array_models')
+      end
+
+      after { ActiveRecord::Base.connection.drop_table(:test_varchar_array_models, if_exists: true) }
+
+      it 'renders a varchar array as array[string]' do
+        generator.generate
+        expect(File.read(output_path)).to include('array[string] tags')
+      end
+    end
```
leandroalemao, line 114: Hi @avinhurry In npq-reg app we have [array of enums](https://github.com/DFE-Digital/npq-registration/blob/main/db/schema.rb#L482) and we might need to create one in ecf2 as well for declarations types.. what do you think about adding a new context here in the specs for it? tkss

```

    context 'when the model has a enum array column' do
      before do
        ActiveRecord::Base.connection.create_enum("declaration_types", %w[started retained-1 retained-2 completed])
        ActiveRecord::Base.connection.create_table(:test_enum_array_models, force: true) do |t|
          t.enum "allowed_declaration_types", default: %w[started retained-1 retained-2 completed], array: true, enum_type: "declaration_types"
        end
        define_test_model(model_name: 'TestEnumArrayModel', table_name: 'test_enum_array_models')
      end

      after { ActiveRecord::Base.connection.drop_table(:test_enum_array_models, if_exists: true) }

      it 'renders a enum array as array[enum]' do
        generator.generate
        expect(File.read(output_path)).to include('array[enum] allowed_declaration_types')
      end
    end
```

avinhurry: Good shout. I checked the schema in ECF2 but not the other apps. 

I've updated, thanks!

cwrw (commented):
thanks @avinhurry!

PR #1320, opened 2025-09-09
Fix schools endpoint behaviour on cohort filtering

ethax-ross:
### Context

The schools endpoint is not behaving consistently with ECF, and the query service is also not consistent with the other query services.

We want the schools endpoint to return no schools if the `cohort` cannot be matched to a `ContractPeriod`.

We want the `contract_period_year` to be mandatory when querying a `School` (which makes sense now that the query service is only used by the API).

- Update `API::Schools::Query` to make `contract_period_year` mandatory.

Return no schools if `contract_period_year` is `nil`, blank or not found.

Update `a filter by a single cohort (contract_period year) endpoint` to use schools behaviour and remove from statements endpoint specs (as this is covered by the multiple cohorts filtering context, which has different behaviour for `nil`/blank/not found).

### DIFF
--- a/app/services/api/schools/query.rb
+++ b/app/services/api/schools/query.rb
@@ -4,13 +4,13 @@ module API::Schools
 
     attr_reader :scope, :sort, :lead_provider_id, :contract_period_year
 
-    def initialize(lead_provider_id: :ignore, urn: :ignore, updated_since: :ignore, contract_period_year: :ignore, sort: { created_at: :asc })
+    def initialize(contract_period_year:, lead_provider_id: :ignore, urn: :ignore, updated_since: :ignore, sort: { created_at: :asc })
       @lead_provider_id = lead_provider_id
       @contract_period_year = contract_period_year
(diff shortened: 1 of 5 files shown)


app/services/api/schools/query.rb
```diff
@@ -54,11 +54,19 @@ def preload_associations(results)
...
-      School
-        .where(id: School.select("schools.id")
-        .joins(school_partnerships: { lead_provider_delivery_partnership: { active_lead_provider: :contract_period } })
-        .where(contract_periods: { year: contract_period_year }))
+    def where_contract_period_exists(contract_period_year)
+      @scope = School.none unless ContractPeriod.exists?(year: contract_period_year)
+    end
+
+    def or_where_contract_period_year_is(contract_period_year)
+      return if ignore?(filter: contract_period_year)
```
cwrw, line 62: do we still need this if the field is required?

leandroalemao (approved):

mooktakim (approved):

PR #1323, opened 2025-09-09
[2177] SPIKE: data model for participant_id_changes

leandroalemao:
### Context

**_This is just a spike PR.. not to be merged at all._**

Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/2177

- Creates a new table/model to store participant id changes;
- It's an existing table in ECF1, so we need to migrate ecf1 ids to `api_id` column;
- `from_participant` and `to_participant` fields are keys to `teachers`.`ecf_user_id` which is not a required field at the moment;
- In order to have the above link, we need to create a unique index in `teachers`.`ecf_user_id`;
- Created a [draft participant serializer](https://github.com/DFE-Digital/register-early-career-teachers-public/compare/2177-spike-participant-id-changes-model?expand=1#diff-9b6be1e63afc85dbb8ed4a743f9e4c6cf249de7de998883d954457176e2022a6), where it calls the [participant id change serializer](https://github.com/DFE-Digital/register-early-career-teachers-public/compare/2177-spike-participant-id-changes-model?expand=1#diff-638d34a27efc44259cfee56061819cb18aa7d958b624f9b2ec35dab7f56cd30f);
- If needed, same sort of validation which is done in [npq-reg app](https://github.com/DFE-Digital/npq-registration/commit/424f2acf1bebbb191cc8b36b9e6dac670fc5f12d) can be used in the participants endpoint so when a lead provider attempts to use/fetch a transferred `participant_id` (i.e. a teacher who has been archived & had their declarations moved to a different teacher), so instead of getting an error response that says the `participant_id` does not exist... a message informing providers that the `participant_id` has changed is displayed, so that they can understand how to rectify the situation.

### Guidance to review

Review all bullets above with the team.

### DIFF
--- a/app/serializers/api/participant_serializer.rb
+++ b/app/serializers/api/participant_serializer.rb
@@ -1,0 +1,26 @@
+class API::ParticipantSerializer < Blueprinter::Base
+  class AttributesSerializer < Blueprinter::Base
+    exclude :id
+
+    field(:full_name) { |teacher, _options| Teachers::Name.new(teacher).full_name }
+    field(:trn, name: :teacher_reference_number)
+    field(:api_updated_at, name: :updated_at)
+
+    field(:ecf_enrolments) do |_teacher, _options|
+      []
+    end
+
+    field(:participant_id_changes) do |teacher, _options|
+      (teacher.participant_id_changes || []).map do |participant_id_change|
+        API::ParticipantIdChangeSerializer.render_as_hash(participant_id_change)
+      end
+    end
+  end
+
+  identifier :ecf_user_id, name: :id
+  field(:type) { "participant" }
+
+  association :attributes, blueprint: AttributesSerializer do |participant|
+    participant
+  end
+end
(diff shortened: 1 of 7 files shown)


app/serializers/api/participant_serializer.rb
```diff
@@ -0,0 +1,26 @@
...
+
+    field(:ecf_enrolments) do |_teacher, _options|
+      []
+    end
+
+    field(:participant_id_changes) do |teacher, _options|
+      (teacher.participant_id_changes || []).map do |participant_id_change|
+        API::ParticipantIdChangeSerializer.render_as_hash(participant_id_change)
+      end
+    end
```
mooktakim, line 17: I think with blueprinter we can just do:
```ruby
association :participant_id_changes, blueprint: ParticipantIdChangeSerializer
```

mooktakim (approved):
just a comment to discuss around `ecf_user_id`

PR #1328, opened 2025-09-10
create training period when assigning mentor to provider-led ECT

ghost:
- Add Schools::CreateMentorTrainingPeriod service to handle training period creation
- Update assign existing mentor wizard to create training periods on lead provider selection
- Training period links to school partnership if confirmed, otherwise creates EOI
- Record teacher_starts_training_period events for audit trail
- 
https://github.com/DFE-Digital/register-ects-project-board/issues/2306


app/wizards/schools/assign_existing_mentor_wizard/lead_provider_step.rb
```diff
@@ -24,7 +26,48 @@ def persist
           author: wizard.author
         ).assign!
 
-        # TODO: Update the LP if there is a confirmed partnership, if not add an EOI
+        create_mentor_training_period!
+      end
+
+      def create_mentor_training_period!
```
avinhurry, line 32: Suggestion to rename this to something like: `create_mentor_training_period_and_add_event!`

To make it clear an event is being added too?

avinhurry (commented):
I'm getting an error when going through the journey in the UI when I submit

image

I think [this validation](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/models/concerns/interval.rb#L36-L44) might be kicking in. It could be just the example data but happy to chat through it!

PR #1332, opened 2025-09-10
Don't reset the ECT wizard when navigating "back"

cpjmcquillan:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/2395#issuecomment-3274504568

Before, we reset the ECT wizard every time a User visited the `/edit` step.

It turns out this wasn't the intention! Instead, design want the User's edit to be "remembered" when they navigate back from the `/check-answers` page.

Now, in the `before_action` callback that resets the wizard, we check whether the referer is the `check_answers` step. If it is (i.e the User clicked the "Back" link), then we **don't** reset the wizard and the User sees the value they previously entered.

### Guidance to review

- Go back and see the value you entered
- Cancel and restart the form and see the original (unedited) value


peterdavidhamilton (approved):
`alias_method` gets used in other specs 🤷🏼‍♂️

spec/features/schools/ects/change/email_address_spec.rb
```diff
@@ -69,15 +74,27 @@ def and_i_continue
     page.get_by_role("button", name: "Continue").click
   end
 
+  def when_i_continue = and_i_continue
```
avinhurry, line 77: Any reason not to use `alias_method`?

avinhurry (approved):
Nice one

PR #1334, opened 2025-09-11
Split routes into multiple files

peterdavidhamilton:
### Context

- Our routes file now exceeds 400 lines
~- Our new mini-wizard routes are identical~ I have replaced the proposed static route file for inclusion into the blocks with a solution that works for all wizards.

Split routes into multiple files to help keep things manageable and DRY

~We might consider going a step further and give big wizards their own route file~

Have gone a step further and addressed big wizard routes using a routing concern

Also suggest RECT addresses the mix of singular and plural paths in the `schools` namespace;
`register_ect_wizard` and `register_mentor_wizard` use different paths for some reason.


config/routes.rb
```diff
@@ -1,12 +1,21 @@
 Rails.application.routes.draw do
+  # DFE-Digital/dfe-wizard auto-routing
+  concern :wizardable do |options = {}|
```
cpjmcquillan, line 3: I'd never come across [`concern`](https://api.rubyonrails.org/classes/ActionDispatch/Routing/Mapper/Concerns.html#method-i-concern) before, but this looks useful!

cpjmcquillan (approved):
I like it. I think it's a nice balance of magic/convention, and it ties in nicely with the `Wizardable` controller concern.

Nice one 👌🏼

PR #1335, opened 2025-09-11
Add `Schedule` and `Milestone` models

peteryates:
### Context

Schedules and milestones are used to control when the different kinds of declarations can be made by lead providers.

The long term goal in ECF2 is to replace these predefined schedules and use term-based validation to control when declarations are made, but those changes require a lot of work so we're going to import and use the ECF1 style ones initially.

A teacher's training is tracked by lead providers _declaring_ that training has

* started via a `started` declaration
* reached a milestone via a `retained-1|2|3|4` declaration
* completed with a `completed` declaration 
* on extended schedules, `extended-1|2|3` are submitted post-completion

### Changes

This PR introduces a slimmed down version of the schedules and milestones models from ECF1, some changes are:

* general
  - enumerated types are used for `schedule_identifiers` and `declaration_types` - they provide an extra
    level of validation and they support ordering

* schedules
  - are linked to a `contract_period` instead of a `cohort`
  - doesn't have a name field, just an identifier (in ECF1 the name and identifier match but the name is
    more readable 'ECF Extended January' vs 'ecf-extended-january')
  - `schedule_identifier` is renamed to `identifier` and uses the `schedule_identifiers` type instead of a `varchar`
  
* milestones
  - don't have a `name` column either, just a `declaration_type`. I _think_ the name was used for ordering
    and displaying in the UI but we can use the enum ordering for that
  - `declaration_type` now uses the `declaration_types` type instead of a `varchar`
 

### Remaining tasks

* [x] chat with @cwrw to make sure any assumptions I've made above are correct
* [x] add a spec for ordering based on the enum if that will be sufficient, if not add a `name` column
* [x] data seeds

### Screenshots

Seeds :man_farmer: 

image

### References

DFE-Digital/register-ects-project-board#2167


db/schema.rb
```diff
@@ -418,6 +420,18 @@
     t.index ["parent_id"], name: "index_migration_failures_on_parent_id"
   end
 
+  create_table "milestones", force: :cascade do |t|
+    t.bigint "schedule_id"
+    t.enum "declaration_type", null: false, enum_type: "declaration_types"
+    t.date "start_date", null: false
+    t.date "payment_date", null: false
```
cwrw, line 427: I don't think we use `payment_date` in ECF1, I've double checked, so I think we can drop :)

db/seeds/schedules_and_milestones.rb
```diff
@@ -0,0 +1,817 @@
...
+  first_char = declaration_type[0]
+  last_char = declaration_type[-1]
+
+  "#{first_char}#{last_char if last_char.match?(/\A\d+\Z/)}"
+end
+
+def schedule_and_milestone_data
+  [
+    {
+      identifier: "ecf-extended-september",
```
cwrw, line 34: there's a bit of duplication with the schedules especially extended/reduced/replacement. Maybe we can set one date for each of those types and reuse to avoid having large list of schedules. I understand though might be easier to read and change with duplication, happy to chat

cwrw (commented):
looks great, just a few minor comments otherwise 🌟

db/seeds/schedules_and_milestones.rb
peteryates: Yeah that sounds like a good idea. I built this directly from an export of ECF1 data converted to a big hash. I'll take a look tomorrow.

cwrw (approved):
Looks great 🙌 , only final comment is would be nice to order the the `schedule_and_milestone_data` according to cohort so its easier to visually parse the hash (for instance start with 2021 and move up to 2025), but otherwise 👍

PR #1352, opened 2025-09-12
Strip extra whitespace on names

peteryates:
Rails' [normalizes](https://api.rubyonrails.org/v7.1/classes/ActiveRecord/Normalization/ClassMethods.html) method allows us to apply formatting, in this case the removal of leading/trailing whitespace via ActiveSupport's `String#squish`.

Squishing removes:

* leading whitespace `" Appropriate body"` becomes `"Appropriate body"`
* trailing whitespace `"Appropriate body "` becomes `"Appropriate body"`
* contiguous whitespace `"Appropriate  body"` becomes `"Appropriate body"`

These often creep in when people are editing and paste in a name from another source. We never want them and they're likely to make formatting look wonky if we ever end up with them.

### DIFF
--- a/spec/models/lead_provider_spec.rb
+++ b/spec/models/lead_provider_spec.rb
@@ -23,4 +23,12 @@ describe LeadProvider do
       end
     end
   end
+
+  describe "normalizing" do
+    subject { FactoryBot.build(:lead_provider, name: " Some lead provider ") }
+
+    it "removes leading and trailing spaces from the name" do
+      expect(subject.name).to eql("Some lead provider")
+    end
+  end
 end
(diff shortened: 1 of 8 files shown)


cwrw (approved):

spec/models/lead_provider_spec.rb
```diff
@@ -23,4 +23,12 @@
       end
     end
   end
+
+  describe "normalizing" do
+    subject { FactoryBot.build(:lead_provider, name: " Some lead provider ") }
+
+    it "removes leading and trailing spaces from the name" do
```
avinhurry, line 30: Do you think it could be worth adding another spec for multiple spaces? 

Something like: 

> it "collapses multiple spaces into one"

avinhurry (approved):

PR #1353, opened 2025-09-15
[2384] Fix parity check POST/PUT endpoints

leandroalemao:
### Context

Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/2384

Currently, the parity check is sending different PUT/POST requests to ECF and RECT. Instead we'd like them to be similar

- Memoise calls to `dynamic_request_content` service class so same endpoints calls to both ecf & rect will use the same params in the body.

### DIFF
--- a/spec/migration/parity_check/client_spec.rb
+++ b/spec/migration/parity_check/client_spec.rb
@@ -1,12 +1,15 @@
 RSpec.describe ParityCheck::Client do
   let(:ecf_url) { "https://ecf.example.com" }
   let(:rect_url) { "https://rect.example.com" }
-  let(:endpoint) { FactoryBot.build(:parity_check_endpoint) }
-  let(:request) { FactoryBot.build(:parity_check_request, endpoint:) }
+  let(:endpoint) { FactoryBot.create(:parity_check_endpoint) }
+  let(:lead_provider) { FactoryBot.create(:lead_provider) }
+  let(:active_lead_provider) { FactoryBot.create(:active_lead_provider, lead_provider:) }
+  let(:lead_provider_delivery_partnership) { FactoryBot.create(:lead_provider_delivery_partnership, active_lead_provider:) }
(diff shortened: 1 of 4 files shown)


app/migration/parity_check/request_builder.rb
```diff
@@ -115,5 +115,10 @@ def pagination_enabled?
     def dynamic_request_content
       @dynamic_request_content ||= DynamicRequestContent.new(lead_provider:)
     end
+
+    def dynamic_request_contents(value)
+      @dynamic_request_contents ||= {}
```
ethax-ross, line 120: We're not testing this anywhere I don't think 🤔 I think it will be hard to test as well; would it be cleaner/easier to test if we just memoize the methods in `DynamicRequestContent` so it consistently returns the same values for a given lead provider/instance?

FWIW I'm adding some request consistency checks to the `Client` that should cover it more generally as part of the PR to expose the request body/URIs in the interface: https://github.com/DFE-Digital/register-early-career-teachers-public/commit/e4d6c3f50aae43b992344f4743a26dbace6b33d6

leandroalemao: Hi @ethax-ross makes sense.. I've moved the memoization to `DynamicRequestContent` instead. Added a spec to test it. I hope it's all good. 👍

spec/migration/parity_check/client_spec.rb
```diff
@@ -97,18 +100,27 @@
       end
     end
 
-    context "when performing a POST request" do
-      let(:endpoint) { FactoryBot.build(:parity_check_endpoint, :post) }
+    context "with body" do
+      before do
+        # Create some data to be used in the request body
+        FactoryBot.create_list(:lead_provider_delivery_partnership, 3, active_lead_provider:)
+        FactoryBot.create_list(:school, 10, :eligible)
```
ethax-ross, line 107: Is it worth just creating 2 of each of the test data here instead of 10?

ethax-ross (approved):
Looks good 👍

PR #1354, opened 2025-09-15
Show request body in parity check

ethax-ross:
### Context

We want to store and surface the request URIs and body from parity check requests in the UI to help us notice/debug potential issues with the check/requests.

- Ensure ParityCheck::Client makes consistent requests

We ran into an issue with the parity check where the request bodies were being uniquely generated for a particular `ParityCheck::Request` (so one sent to ECF and another to RECT).

This is hard to debug in the user interface; we have since fixed the underlying issue, but this puts a safety net in place to ensure at the `Client` level it is verifying the requests are being consistently made by comparing `method`, `request_body`, `request_headers` and `query`.

- Retain request URIs and body in ParityCheck::Response

We want to surface the request body and URIs in the parity check interface so that we can easily double check the requests that were made.

Add request URIs to the `Response` model; we need to store both separately as on occasion the paths will differ.

Add request body to the `Response` model; we only need to persist this once for both ECF and RECT as we have tests that ensure the same request body is used for both requests.

- Fix heading sizes on parity check pages

The heading sizes were a bit mismatched between pages and even on the same page.

Retain one `l` heading and then use `m` for sub headings.

- Show request details in parity check UI

Surface the request URIs and the request body when viewing a parity check response. This will help us to debug any issues with requests going forward.

### Guidance to review

It's hard to generate decent seed data with paths and request bodies due to not all LPs having, for example, school partnerships. I've left the seed data generic for now, so it won't match the rest of the response details.

screencapture-0-0-0-0-3000-migration-parity-checks-7-responses-404-2025-09-15-13_23_27

Example failure on migration (as this was without the incoming fix being applied):

Screenshot 2025-09-16 at 08 48 54

### DIFF
--- a/db/migrate/20250915065439_add_request_details_to_parity_check_responses.rb
+++ b/db/migrate/20250915065439_add_request_details_to_parity_check_responses.rb
@@ -1,0 +1,9 @@
+class AddRequestDetailsToParityCheckResponses < ActiveRecord::Migration[8.0]
+  def change
+    change_table :parity_check_responses, bulk: true do |t|
+      t.string :ecf_request_uri
+      t.string :rect_request_uri
+      t.jsonb :request_body
+    end
+  end
+end
--- a/app/assets/stylesheets/migration/parity_check.scss
+++ b/app/assets/stylesheets/migration/parity_check.scss
@@ -127,3 +127,11 @@
   margin-top: 100px;
   text-align: center;
 }
+
+.response-details code {
+  padding: 2px 5px;
+  font-family: monospace;
+  background-color: $govuk-template-background-colour;
+  white-space: pre-wrap;
+  font-size: 0.8em;
+}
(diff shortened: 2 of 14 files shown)


db/migrate/20250915065439_add_request_details_to_parity_check_responses.rb
```diff
@@ -0,0 +1,9 @@
+class AddRequestDetailsToParityCheckResponses < ActiveRecord::Migration[8.0]
+  def change
+    change_table :parity_check_responses, bulk: true do |t|
+      t.string :ecf_request_uri
+      t.string :rect_request_uri
+      t.jsonb :request_body
```
leandroalemao, line 6: Hi @ethax-ross Shouldn't these be in `parity_check_requests` instead?

leandroalemao (approved):

PR #1358, opened 2025-09-16
Add payments frozen attributes to `Teacher`

ethax-ross:
### Context

We need to store if a participant has migrated from a payments frozen contract period so that we can surface this information in the API. In ECF we store a `boolean` flag to indicate this, but it makes determining the particular year they migrated from more complicated as we have to dive into their declarations. In RECT we are opting to store the contract period year on the `Teacher` model; we need this for both ECTs and mentors, so are adding attributes for each.

We will revisit this approach down the line when we are building the change schedule service and have a better idea of how all these ECT/mentor specific attributes could hang together.

- Add ect/mentor_payments_frozen_year to Teacher

We want to track the payments frozen year for teachers, so we can expose this information in the API and lead providers can determine if a participant has migrated from a payments frozen cohort.

- Add payments frozen year to teacher seeds

Generate `Teacher` records with `ect/mentor_payments_frozen_year` values.

- Add with_payments_frozen_years trait to Teacher factory

### DIFF
--- a/db/seeds/teachers.rb
+++ b/db/seeds/teachers.rb
@@ -7,7 +7,18 @@ def describe_teacher(teacher)
                  Colourize.text('no', :red)
                end
 
-  print_seed_info("#{teacher_name} (early roll out mentor: #{ero_status})", indent: 2)
+  payments_frozen =
+    if teacher.ect_payments_frozen_year.present? && teacher.mentor_payments_frozen_year.present?
+      Colourize.text('ECT/mentor', :green)
+    elsif teacher.ect_payments_frozen_year.present?
+      Colourize.text('ECT', :green)
+    elsif teacher.mentor_payments_frozen_year.present?
+      Colourize.text('mentor', :green)
+    else
+      Colourize.text('no', :red)
+    end
+
+  print_seed_info("#{teacher_name} (early roll out mentor: #{ero_status}, payments frozen: #{payments_frozen})", indent: 2)
 end
 
 early_roll_out_mentor_attrs = {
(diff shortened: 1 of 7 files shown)


db/seeds/teachers.rb
peteryates, line 1: Would it be useful to indicate the frozen payments in the seed output? Currently we do for ERO but nothing else.

```
🪴 planting db/seeds/teachers.rb
🌱   Emma Thompson (early roll out mentor: no)
🌱   Kate Winslet (early roll out mentor: no)
🌱   Alan Rickman (early roll out mentor: no)
```

peteryates (approved):
Looks good, left a couple of minor comments.

db/seeds/teachers.rb
ethax-ross: Updated to include the frozen payment state:

Screenshot 2025-09-16 at 11 28 06

spec/factories/teacher_factory.rb
```diff
@@ -4,6 +4,11 @@
     sequence(:trs_first_name) { |n| "First name #{n}" }
     sequence(:trs_last_name) { |n| "Last name #{n}" }
 
+    trait :with_payments_frozen_years do
+      ect_payments_frozen_year { FactoryBot.create(:contract_period, year: [2021, 2022].sample).year }
+      mentor_payments_frozen_year { FactoryBot.create(:contract_period, year: [2021, 2022].sample).year }
```
cwrw, line 9: can we only set those values if there is an ect/mentor at school period present?

leandroalemao (approved):

PR #1364, opened 2025-09-16
Finish periods on new start date instead of previous day

avinhurry:
## Context

Previously we used `.prev_day`, which introduced an unnecessary gap and prevented re-registering a mentor the day after their previous period.

Update `finish_existing_at_school_periods!` to pass `finished_on: started_on` instead of `started_on.prev_day`.

## Guidance to review

- Register a mentor at one school, then register the same mentor at another school using tomorrow's date - it should work.

- Try the same flow on `main` or `staging` - it should blow up.

### DIFF
--- a/app/services/schools/register_mentor.rb
+++ b/app/services/schools/register_mentor.rb
@@ -82,7 +82,7 @@ module Schools
     end
 
     def finish_existing_at_school_periods!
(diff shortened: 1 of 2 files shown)


peteryates (approved):
:ok_hand:

PR #1370, opened 2025-09-17
[2416] Skip metadata callbacks on gias importers - run only in the end

leandroalemao:
### Context

Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/2416

The metadata callbacks are slowing down the gias importers, we need to ensure metadata refresh runs after gias importers run.

- Use `skip_metadata_updates` to skip metadata updates during the importer process;
- Trigger metadata refresh async true after gias imports finished running;

### Guidance to review

Locally think it's the best way of testing it, running gias importer locally and checking how long it's taking compared with main branch. See if `Metadata::SchoolContractPeriod.all.size` is updated after the importer is finished.


app/services/gias/importer.rb
```diff
@@ -7,7 +7,10 @@ class Importer
     SCHOOL_LINKS_FILENAME = "links.csv"
 
     def fetch
-      import_only? ? fetch_and_import_only : fetch_and_update
+      Metadata::Manager.skip_metadata_updates do
+        import_only? ? fetch_and_import_only : fetch_and_update
+      end
+      Metadata::Handlers::School.refresh_all_metadata!(async: true, track_changes: true)
```
mooktakim, line 13: I dont think we should track changes here, as we expect some changes.
We should track changes for the nightly runs so we receive sentry events for any changes

PR #1371, opened 2025-09-17
[2178] spike latest training period - Option 2

mooktakim:
### Context

Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/2178

* Participants query service `Participants::Query`
* Serialize `API::ParticipantSerializer`
* Latest training period by `started_on DESC`
* Query is union of ECT teachers and Mentor teachers
* Serializer chooses the latest training period by choosing the newest `started_on` training period
* No N+1 queries
* Using `has_one` association on ECT/Mentor At School Periods to filter training periods and only return that one needed

### DIFF
--- a/app/services/participants/query.rb
+++ b/app/services/participants/query.rb
@@ -1,0 +1,45 @@
+module Participants
+  class Query
+    attr_reader :scope, :lead_provider
+
+    def initialize(lead_provider:)
+      @lead_provider = lead_provider
+      @scope = all_participants
+    end
+
+    def participants
+      scope.order(created_at: :asc)
+    end
+
+    def participant(id:)
+      scope.where(id:)
+    end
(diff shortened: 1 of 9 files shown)


app/services/participants/query.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  private
+
+    def all_participants
+      # Union both queries and preload associations for serialization
+      Teacher
+        .select("teachers.*")
+        .from("(#{ect_teachers.to_sql} UNION #{mentor_teachers.to_sql}) as teachers")
+        .includes(
+          ect_at_school_periods: { latest_training_period: :lead_provider },
```
cwrw, line 26: we should probably use the partnership lead provider not the EOI

PR #1379, opened 2025-09-18
Reduce DB writes by batching counter updates in migrators

leoapost:
### Context

Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/2447

This PR is part of ongoing migrator optimisations aimed at reducing overall runtime. It focuses on cutting down unnecessary database writes triggered by the Migrator::Base class when updating migration counters for each processed record.

Currently, these actions generate ~900k DB writes. This change reduces that to less than 1k writes by batching the operations.

- Batch migration counter updates: now written every 1,000 processed records

### DIFF
--- a/app/migration/migrators/base.rb
+++ b/app/migration/migrators/base.rb
@@ -12,6 +12,8 @@ module Migrators
     include ActiveModel::Model
     include ActiveModel::Attributes
 
+    COUNTER_UPDATE_BATCH_SIZE = 1_000
+
     attribute :worker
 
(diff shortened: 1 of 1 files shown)


leoapost:
> Thinking about this, would this mean potentially no updates when a migration is in progress? Would a migrator worker go from 0 to 100 only with nothing in between?

Each migrator worker processes 3,000 records. By updating the counters every 1,000 records, it will give 3 updates per migrator worker.

PR #1382, opened 2025-09-18
Introduce `ECTAtSchoolPeriods::SwitchTraining` service object

cpjmcquillan:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/2100

This was pulled out as part of the work to facilitate schools changing an ECT's training programme.

There is some relatively complex business logic to consider when changing an ECT's training programme from school-led to provider-led (and vice versa), so it seemed sensible to abstract this into a service object we can test rigorously in isolation.

This has been designed and implemented based on point 6. of the acceptance criteria outlined [here]. Lots of this is new ground for me, so it's possible I might have missed something!

The intention is to call this service object as part of the "Check answers" step in the new "Change training programme" wizard, which will follow in a subsequent PR.

[here]: https://github.com/DFE-Digital/register-ects-project-board/issues/2100

### Guidance to review

- [ ] 6. The new training programme (and lead provider if applicable) will be recorded for the ECT, so that it is updated when the user returns to the ECT details page.
    - [ ] 6a. If the training has been changed from school-led to provider-led, the existing training_period will be closed and a new provider-led training_period will be created. If the school already has a partnership with the new selected LP in the ECT's contract year, the training_period will be confirmed straight away. If there is no partnership, this will be recorded as a new EOI for the LP for the ECT's contract period.
    - [ ] 6b. If the training has been changed from provider-led to school-led:
        - [ ] If there is a confirmed partnership for the existing training_period, we will close the existing training_period and create a new school-led training_period.
        - [ ] If there is no confirmed partnership for the existing training_period, we will remove the existing training_period and create a new school-led training_period.

### DIFF
--- a/app/models/concerns/interval.rb
+++ b/app/models/concerns/interval.rb
@@ -62,9 +62,7 @@ module Interval
     started_on <= finished_on
   end
 
-  def invalid_date_order?
-    finished_on <= started_on
-  end
+  def invalid_date_order? = !valid_date_order?
 
   def ongoing? = finished_on.nil?
 
--- a/spec/models/concerns/interval_spec.rb
+++ b/spec/models/concerns/interval_spec.rb
@@ -15,16 +15,6 @@ describe Interval do
           expect(subject.errors.messages).to include(finished_on: ["The end date must be later than the start date (#{Date.yesterday.to_fs(:govuk)})"])
         end
       end
-
-      context "when finished_on matches started_on" do
-        subject { DummyMentor.new(started_on: Date.yesterday, finished_on: Date.yesterday) }
-
-        before { subject.valid? }
-
-        it "adds an error" do
-          expect(subject.errors.messages).to include(finished_on: ["The end date must be later than the start date (#{Date.yesterday.to_fs(:govuk)})"])
-        end
-      end
     end
   end
 
--- a/spec/services/ect_at_school_periods/switch_training_spec.rb
+++ b/spec/services/ect_at_school_periods/switch_training_spec.rb
@@ -99,6 +99,52 @@ module ECTAtSchoolPeriods
             SwitchTraining.to_school_led(ect_at_school_period, author:)
           end
         end
+
+        context "when the switch happens on the same day the training period started" do
+          let!(:training_period) do
+            FactoryBot.create(
+              :training_period,
+              :for_ect,
+              :ongoing,
+              :provider_led,
+              :with_school_partnership,
+              ect_at_school_period:,
+              started_on: Date.current
+            )
+          end
+
+          it "finishes the existing training period" do
+            freeze_time
+
+            SwitchTraining.to_school_led(ect_at_school_period, author:)
+
+            expect { training_period.reload }.not_to raise_error
+            expect(training_period.finished_on).to eq(Date.current)
+          end
+
+          it "creates a new school-led training period" do
+            SwitchTraining.to_school_led(ect_at_school_period, author:)
+
+            expect(ect_at_school_period.reload).to be_school_led_training_programme
+          end
+
+          it "records an event" do
+            freeze_time
+
+            expect(Events::Record)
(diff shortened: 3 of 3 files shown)


app/services/ect_at_school_periods/switch_training.rb
```diff
@@ -0,0 +1,85 @@
...
+      end
+    end
+
+    def finish_training_period!
+      TrainingPeriods::Finish.ect_training(
+        training_period: @training_period,
+        ect_at_school_period: @ect_at_school_period,
+        finished_on:,
+        author: @author
+      ).finish!
```
avinhurry, line 63: This is blowing up for me when I change from school-led to provider-led

image

app/services/ect_at_school_periods/switch_training.rb
```diff
@@ -0,0 +1,85 @@
...
+      @author = author
+    end
+
+    def to_school_led
+      raise IncorrectTrainingProgrammeError if
+        @ect_at_school_period.school_led_training_programme?
+
+      ActiveRecord::Base.transaction do
+        finish_or_destroy_existing_training_period!
+        create_school_led_training_period!
```
avinhurry, line 31: Do we want to record an event when changing training programme?

avinhurry: I agree yeah 👍🏽 

https://github.com/DFE-Digital/register-early-career-teachers-public/pull/1398#issuecomment-3318701449

app/services/ect_at_school_periods/switch_training.rb
avinhurry: I fixed a similar bug for mentors in #1364 , where we allowed same day handovers. I think it makes sense to align ECTs with that behaviour, but worth running it past @mason-emily just to be sure.

cpjmcquillan: We agreed to allow same day handovers. This has been fixed in https://github.com/DFE-Digital/register-early-career-teachers-public/pull/1382/commits/7a0c227907655b0323a3d889ad482a5bc9a99ec8 👌🏼

app/services/ect_at_school_periods/switch_training.rb
```diff
@@ -0,0 +1,109 @@
...
+
+    def record_training_period_switched_to_provider_led_event!
+      Events::Record.record_teacher_switches_to_provider_led_training!(
+        author: @author,
+        ect_at_school_period: @ect_at_school_period,
+        teacher: @ect_at_school_period.teacher,
+        training_period: @new_training_period,
+        school:,
+        happened_at: Time.current
+      )
```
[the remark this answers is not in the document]
cpjmcquillan: Good question. I thought about using 1 event with some `from`, `to` metadata, but I figured the distinct event types would be helpful.

app/services/ect_at_school_periods/switch_training.rb
```diff
@@ -0,0 +1,109 @@
...
+
+  private
+
+    attr_reader :school, :lead_provider
+
+    def finish_or_destroy_existing_training_period!
+      if @training_period.school_partnership.present?
+        finish_training_period!
+      else
+        @training_period.destroy!
```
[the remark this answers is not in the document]
cpjmcquillan: No, we don't define anything on the relationship (e.g `dependent: :destroy`).

This could leave us with some "broken" data, though (i.e Events that belong to Training Periods that don't exist).

The acceptance criteria said we should "remove" these training periods, but I wonder whether we could close them instead?

@peteryates @mason-emily do you have any thoughts?

app/services/ect_at_school_periods/switch_training.rb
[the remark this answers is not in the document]
cpjmcquillan: To be honest, maybe the event should be `teacher_training_programme_updated` and a heading that outlines the from/to. That's what we're doing for other things (e.g working pattern).

PR #1383, opened 2025-09-18
Spike into teacher induction start/end dates

ethax-ross:
> ⚠️ WIP

### Context

We want to spike how we could model the induction period start/end dates for a teacher. We are looking into two options in this spike:

- Add `induction_started_on` and `induction_finished_on` to `Teacher`
- Add `Metadata::Teacher` for storing induction start/end dates

- Add `induction_started_on` and `induction_finished_on` to `Teacher`

Add two helper methods to return the `induction_started_on` and `induction_finished_on` for a `Teacher`.

`induction_started_on` is the `induction_period` with the earliest `started_on`.

`induction_finished_on` is the `induction_period` with the latest `finished_on` where the `outcome` is present.

- Add `Metadata::Teacher` for storing induction start/end dates

Add a migration to introduce a `Metadata::Teacher` model with a `has_one` association on `Teacher` (as it won't be scoped to any other models like we do with other metadata).

Refresh the metadata when creating a `Teacher` or updating an `InductionPeriod`.

Add a handler to refresh `Teacher` metadata.

#### Understanding `InductionPeriod` finished state with respect to `finished_on` and `outcome`

Based on the below, we should be fine to take the `finished_on` for any induction period with an outcome, as there will be no later induction periods with a later `finished_on` date.

We don't seem to enforce all of these scenarios in validation on `InductionPeriod`, but I think we should.

##### No teachers with multiple outcomes in their induction periods

```
Teacher
  .joins(:induction_periods)
  .where.not(induction_periods: { outcome: nil })
  .group("teachers.id")
  .having("COUNT(induction_periods.id) > 1")

=> []
```

##### Teachers with multiple, unfinished induction periods - should this be possible?

```
Teacher
  .joins(:induction_periods)
  .where(induction_periods: { finished_on: nil })
  .group("teachers.id")
  .having("COUNT(induction_periods.id) > 1")

=> 
{46092 => 2,
 53501 => 2,
 62682 => 2,
 62096 => 2,
 40855 => 2,
 65740 => 2,
 69751 => 2,
 28840 => 2,
 56788 => 2}
```

##### No induction periods with an outcome but no finished on date

```
InductionPeriod
  .where.not(outcome: nil)
  .where(finished_on: nil)

=> []
```

##### No teachers that have an outcome and another period with later finished on

```
Teacher
  .joins("INNER JOIN induction_periods ip1 ON ip1.teacher_id = teachers.id")
  .joins("INNER JOIN induction_periods ip2 ON ip2.teacher_id = teachers.id")
  .where.not("ip1.outcome IS NULL")
  .where("ip2.finished_on IS NOT NULL")
  .where("ip2.finished_on > ip1.finished_on")
  .distinct

=> []
```

#### Optimising a query for multiple teachers

If we opted to put a method on the `Teacher` model to retrieve the dates, we would need an optimal query to avoid n+1s -- I'm not a fan of this, but it is an option and doesn't have much of a performance hit:

```
def avg_benchmark(runs = 10, &block)
  # warm up once
  block.call

  # now do the actual timed runs
  times = runs.times.map { Benchmark.measure(&block).real }
  times.sum / times.size
end

benchmark1 = avg_benchmark(10) do
  started_on_subquery = InductionPeriod
    .select("MIN(started_on)")
    .where("induction_periods.teacher_id = teachers.id")

  finished_on_subquery = InductionPeriod
    .select("MAX(finished_on)")
    .where("induction_periods.teacher_id = teachers.id")
    .where.not(outcome: nil)

  Teacher.select(
    "teachers.*",
    "(#{started_on_subquery.to_sql}) AS induction_started_on",
    "(#{finished_on_subquery.to_sql}) AS induction_finished_on"
  ).to_a
end

benchmark2 = avg_benchmark(10) do
  Teacher.all.to_a
end

puts "Benchmark 1 (avg): #{benchmark1.round(4)}s"
puts "Benchmark 2 (avg): #{benchmark2.round(4)}s"

if benchmark2 > 0
  ratio = benchmark1 / benchmark2
  percentage = ((benchmark1 - benchmark2) / benchmark2) * 100
  puts "Benchmark 1 is #{ratio.round(2)}x slower (≈ #{percentage.round(1)}% slower)"
else
  puts "Benchmark 2 was too fast to compare reliably."
end

Benchmark 1 (avg): 3.3189s
Benchmark 2 (avg): 2.9632s
Benchmark 1 is 1.12x slower (≈ 12.0% slower)
```


leandroalemao (commented):
Option 3 seems to be the simplest and the way to go for, we can move to the metadata idea if needed in the future.

peteryates:
A couple of things to note:

* we have a `Teachers::Induction` service that [finds the earliest induction start date already](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/services/teachers/induction.rb#L18), if we're introducing a `has_one` to get it, it'd make sense to update that service so it's using the relationship
* there's also a `Teachers::RefreshMetadata` class that [syncs data](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/8d28e2c30d8ab469715e39ae40d8caa0c2e6c30b/app/services/teachers/refresh_trs_attributes.rb#L21) from TRS. It's automatically triggered when an induction is completed so it'll be up to date. The API client [currently retrieves the start date](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/8d28e2c30d8ab469715e39ae40d8caa0c2e6c30b/lib/trs/teacher.rb#L46) but it'd be really straightforward to add the end date to `TRS::Teacher`.

  I think the TRS end date is going to be added to auto-sync before long anyway by RIAB (cc @emily-prudence-dfe) because it'll provide a good sense check to make sure the two systems are properly talking to each other, especially since we will soon have [blazer alerts](#1350)

PR #1386, opened 2025-09-18
[2409] Implement uplift fields in RECT

mooktakim:
### Context

Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/2409

* New columns on `Teacher`
  - `ect_pupil_premium_uplift`
  - `ect_sparsity_uplift`
* `default: false, null: false`
* Seed updated for `Teacher`
* Factory trait added
* mermaid regenerated

### DIFF
--- a/db/migrate/20250918151250_add_uplift_fields_to_teachers.rb
+++ b/db/migrate/20250918151250_add_uplift_fields_to_teachers.rb
@@ -1,0 +1,8 @@
+class AddUpliftFieldsToTeachers < ActiveRecord::Migration[8.0]
+  def change
+    change_table :teachers, bulk: true do |t|
+      t.boolean :ect_pupil_premium_uplift, default: false, null: false
+      t.boolean :ect_sparsity_uplift, default: false, null: false
+    end
+  end
+end
--- a/spec/factories/teacher_factory.rb
+++ b/spec/factories/teacher_factory.rb
@@ -8,6 +8,19 @@ FactoryBot.define do
       corrected_name { [trs_first_name, Faker::Name.middle_name, trs_last_name].join(' ') }
     end
 
+    trait :with_sparsity_uplift do
+      ect_sparsity_uplift { true }
+    end
+
+    trait :with_pupil_premium_uplift do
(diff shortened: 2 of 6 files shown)


spec/factories/teacher_factory.rb
```diff
@@ -8,6 +8,19 @@
...
 
+    trait :with_sparsity_uplift do
+      sparsity_uplift { true }
+    end
+
+    trait :with_pupil_premium_uplift do
+      pupil_premium_uplift { true }
+    end
+
+    trait :with_uplift do
```
leandroalemao, line 19: Hi @mooktakim do we need these traits as we're still not using anywhere? or do you think it's good to add anyway? tks

mooktakim: Good question. I suppose because its related to Teacher, its good that its there as it'll show as one of the options.

ethax-ross: We've been adding the factories/traits for other new fields, FWIW

leandroalemao (approved):

db/migrate/20250918151250_add_uplift_fields_to_teachers.rb
```diff
@@ -0,0 +1,8 @@
+class AddUpliftFieldsToTeachers < ActiveRecord::Migration[8.0]
+  def change
+    change_table :teachers, bulk: true do |t|
+      t.boolean :pupil_premium_uplift, default: false, null: false
```
cwrw, line 4: can those both be prefixed by `ect` as we aren't adding for mentors?

peteryates:
I think the names could be a little more descriptive.

I guess they're stamped once at the time of first registration (as if it was per school it'd be on `ECTAtSchoolPeriod`).

e.g., `ect_pupil_premium_uplift_when_registered`?

db/migrate/20250918151250_add_uplift_fields_to_teachers.rb
mooktakim: prefixed columns with `ect`

PR #1395, opened 2025-09-19
Fix `ErrorsController` for unauthenticated users

ethax-ross:
### Context

At the moment, if we hit a 404 error it ends up being swallowed by the `authenticate` `before_action` and instead results in a redirect to the `pre_login_redirect_path`.

We want to surface errors correctly for unauthenticated users.

- Skip authenticate in ErrorsController

By skipping `authenticate` in `ErrorsController` we will correctly pick up and respond with the correct status code/view.

- Add test coverage to ErrorsController

Add tests that will ensure our `ErrorsController` is correctly handling errors.

### Guidance to review

Initially we wanted to restrict the changes to the API, thinking this was API-specific, however it turned out to be a general issue with error handling in the app.

The API errors already have test coverage in the various shared contexts for testing endpoints.

### DIFF
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -1,4 +1,6 @@
 class AdminController < ApplicationController
+  before_action :authenticate
+
   include Authorisation
 
   def index
--- a/app/controllers/api/documentation_controller.rb
+++ b/app/controllers/api/documentation_controller.rb
@@ -2,8 +2,6 @@
 
 module API
   class DocumentationController < ApplicationController
-    skip_before_action :authenticate
-
     layout "api_docs"
 
     def index
--- a/app/controllers/api/guidance_controller.rb
+++ b/app/controllers/api/guidance_controller.rb
@@ -2,8 +2,6 @@ module API
   class GuidanceController < ApplicationController
     include ReleaseNotes
 
-    skip_before_action :authenticate
-
     layout 'api_guidance'
 
     def show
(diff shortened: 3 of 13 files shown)


peteryates:
Approved despite this still being in draft, but it looks spot on.

I do wonder if it might be easier to remove the authorisation from `ApplicationController` and just have these 'top level' controllers open by default, but to start authorising at the next level (`SchoolController`, `AdminController`, `APIController` etc).

leandroalemao (commented):
Hi @ethax-ross For the API calls ie: calls to `/api/v3/blabla`, should we see a more user friendly message in JSON? rather than getting the html code of the 404 page? or it doesn't matter the body content as usually it just cares about response code anyway?

cwrw (dismissed):

cpjmcquillan (approved):

ethax-ross:
> Approved despite this still being in draft, but it looks spot on.
> 
> I do wonder if it might be easier to remove the authorisation from `ApplicationController` and just have these 'top level' controllers open by default, but to start authorising at the next level (`SchoolController`, `AdminController`, `APIController` etc).

I've added a commit to do this if you could cast your eye over it in case I've missed anything 👍

ethax-ross:
> Hi @ethax-ross For the API calls ie: calls to `/api/v3/blabla`, should we see a more user friendly message in JSON? rather than getting the html code of the 404 page? or it doesn't matter the body content as usually it just cares about response code anyway?

Yeah, so. you should get something like this from API endpoints:

```
{
    "errors": [
        {
            "title": "Resource not found",
            "detail": "Nothing could be found for the provided details"
        }
    ]
}
```

PR #1402, opened 2025-09-22
Rework previous choices page

### LINKED ISSUE #1393: Stop defaulting to the `first_school`
In the `SchoolsController` we're falling back to the `#first_school` when there's no school in the session.

https://github.com/DFE-Digital/register-early-career-teachers-public/blob/8d28e2c30d8ab469715e39ae40d8caa0c2e6c30b/app/controllers/schools_controller.rb#L22-L25

We should be raising an error and redirecting somewhere sensible (back to the login page with an alert/notice) if there's no school.

peteryates:
### Context

This branch attempts to make the ECT previous choices page simpler and ensures the previous choice information is accurate.

It does this by using the `Schools::LatestRegistrationChoices` service (DFE-Digital/register-early-career-teachers-public#856). Because this is very much a school level thing, rather than pulling it via the wizard's `ECT` class I added a new `DecoratedSchool` instance variable which has a couple of school-specific methods.

I think this is a sensible first move when it comes to simplifying the `ECT` class.

### Final tasks

* [x] ensure the choices made are stored correctly in the wizard store

### Links

Fixes #1393, DFE-Digital/register-ects-project-board#2309

### DIFF
--- a/app/views/schools/register_ect_wizard/_use_previous_ect_choices.html.erb
+++ b/app/views/schools/register_ect_wizard/_use_previous_ect_choices.html.erb
@@ -32,7 +32,7 @@
   end
 end %>
 
-<% if @decorated_school.provider_led_training_programme_chosen? && !@decorated_school.has_partnership_with?(lead_provider: @decorated_school.latest_registration_choices.lead_provider.name, contract_period: @ect.contract_start_date) %>
+<% if @decorated_school.provider_led_training_programme_chosen? && !@decorated_school.has_partnership_with?(lead_provider: @decorated_school.latest_registration_choices.lead_provider, contract_period: @ect.contract_start_date) %>
   <p class="govuk-body">
     <%= @decorated_school.latest_registration_choices.lead_provider.name %> will confirm if they’ll be working with your school and which delivery partner will deliver training events.
   </p>
--- a/app/views/schools/register_ect_wizard/_use_previous_ect_choices.html.erb
+++ b/app/views/schools/register_ect_wizard/_use_previous_ect_choices.html.erb
@@ -5,39 +5,36 @@
 <p class="govuk-body">You used these programme choices the last time you registered an ECT.</p>
 
 <%= govuk_summary_list do |summary_list|
-  summary_list.with_row do |row|
-    row.with_key(text: 'Appropriate body')
-    row.with_value(text: @school.last_chosen_appropriate_body_name)
+  if @decorated_school.latest_registration_choices.appropriate_body.present?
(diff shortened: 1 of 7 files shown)


mason-emily (approved):

app/views/schools/register_ect_wizard/_use_previous_ect_choices.html.erb
```diff
@@ -5,39 +5,36 @@
...
-        row.with_key(text: 'Lead provider')
-        row.with_value(text: @ect.previous_eoi_lead_provider_name || @school.last_chosen_lead_provider&.name)
+        row.with_value(text: @decorated_school.latest_registration_choices.delivery_partner.name)
       end
     end
   end
 end %>
 
-<% if @school.provider_led_training_programme_chosen? && !@ect.lead_provider_has_confirmed_partnership_for_contract_period?(@school) %>
+<% if @decorated_school.provider_led_training_programme_chosen? && !@decorated_school.has_partnership_with?(lead_provider: @decorated_school.latest_registration_choices.lead_provider.name, contract_period: @ect.contract_start_date) %>
```
avinhurry, line 35: Would this need the object rather than name?

```suggestion
<% if @decorated_school.provider_led_training_programme_chosen? && !@decorated_school.has_partnership_with?(lead_provider: @decorated_school.latest_registration_choices.lead_provider, contract_period: @ect.contract_start_date) %>
```

From what I can tell, `SchoolPartnerships::Search` expects an object, so passing the name wouldn't match anything.

app/wizards/schools/decorated_school.rb
```diff
@@ -0,0 +1,11 @@
+module Schools
+  class DecoratedSchool < SimpleDelegator
+    def latest_registration_choices(contract_period: ContractPeriod.containing_date(Time.zone.today))
+      @latest_registration_choices ||= Schools::LatestRegistrationChoices.new(school: self, contract_period:)
```
avinhurry, line 4: Will the memoisation cause `@latest_registration_choices` to  return stale results if a different `contract_period` is passed in? 

I mean it's not much of an issue since contract period doesn't change much but wanted to flag.

avinhurry (approved):
Couple of comments/questions in line but otherwise a very nice improvement!

app/views/schools/register_ect_wizard/_use_previous_ect_choices.html.erb
peteryates: Ah yeah well spotted. Pushed a fix and a spec covering it!

PR #1421, opened 2025-09-24
Update show.html.erb_spec.rb

StephenC-V1:
2300. Update assign mentor text (summary and details page)

### Context
Update the CTA link for assigning a mentor on ECT summary and detail pages.

Single line change, reflected on ECT summary and detail page. Shorten text, remove 'must' but keep warning pattern.

### DIFF
--- a/spec/views/schools/ects/show.html.erb_spec.rb
+++ b/spec/views/schools/ects/show.html.erb_spec.rb
@@ -85,7 +85,7 @@ RSpec.describe 'schools/ects/show.html.erb' do
       context 'when unassigned' do
         it 'has instruction to assign' do
           expect(rendered).to have_css('dt.govuk-summary-list__key', text: 'Mentor')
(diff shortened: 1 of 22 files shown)


spec/views/schools/ects/show.html.erb_spec.rb
```diff
@@ -85,7 +85,7 @@
       context 'when unassigned' do
         it 'has instruction to assign' do
           expect(rendered).to have_css('dt.govuk-summary-list__key', text: 'Mentor')
-          expect(rendered).to have_css('dd.govuk-summary-list__value', text: 'You must assign a mentor or register a new one for this ECT.')
+          expect(rendered).to have_css('dd.govuk-summary-list__value', text: 'Assign a mentor to this ECT.')
```
avinhurry, line 88: You've only updated the content in a test here? Did you mean to update the content too?

StephenC-V1: Most definitely. I searched for 'must assign' in the repo and this is the only file that was returned. I see now that it's a test file.

PR #1436, opened 2025-09-26
Fix AppropriateBodies::RecordOutcome for old induction period programme types

peterdavidhamilton:
### Context

@cpjmcquillan spotted this validation failure which is caused by old records having no `training_programme` when they come to be closed by the AB.

https://dfe-teacher-services.sentry.io/issues/6900212741/?environment=production&project=4508369974788096&query=is%3Aunresolved&referrer=issue-stream

Insert new programme type when recording an outcome.


cpjmcquillan (approved):
🗺️ 

I wonder whether it's worth us running a migration that maps all of the data, then making the column nullable and removing this workout at some point?

PR #1438, opened 2025-09-26
Update error message for past start dates

StephenC-V1:
2494. Update error message to remove reference to 'registration periods'.

### Context
The message currently uses the term '2 registration periods'. We think this is language that won't be fully understood by our users.

Update error message for past start dates to remove reference to 'registration periods'.


avinhurry (approved):
Looks good to me. 

Looks like your branch might need a rebase, shout if you want a hand with that.

peteryates (approved):

PR #1441, opened 2025-09-26
Prevent duplicate induction periods from being created via bulk claims

peterdavidhamilton:
### Context

Bit of a head scratcher because only 9 ECTs have been impacted. After confirming there were not duplicate rows in the data and that any would prevent submission; my initial assumption was that a bulk job had been run twice, perhaps submitted twice or had resumed after a worker restart/crash.

https://register-early-career-teachers.education.gov.uk/admin/blazer/queries/140-duplicate-open-induction-periods

However the TRNs are not sequential nor do they appear at the start of the processed data.

By tackling the planned move to use more delayed jobs we can also ensure the job is only queued once.

1. Extract completion of bulk jobs into separate jobs. 
2. Ensure claim jobs guard against duplication, this is the goal of the PR. Do not enqueue the `RegisterECTJob` if the ECT has an ongoing induction. 
3. Namespace all AB bulk jobs to keep this organised.

### Guidance to review

Testing this should be difficult seeing as it has not occurred consistently and can't be recreated.

### DIFF
--- a/spec/features/appropriate_bodies/process_batch_combo_spec.rb
+++ b/spec/features/appropriate_bodies/process_batch_combo_spec.rb
@@ -37,10 +37,13 @@ RSpec.describe 'Process bulk claims then actions events' do
       "The Appropriate Body completed a bulk claim"
     )
 
-    expect(AnalyticsBatchJob).to have_been_enqueued
-      .once.with(pending_induction_submission_batch_id: PendingInductionSubmissionBatch.last.id)
-
+    expect(AnalyticsBatchJob).to have_been_enqueued.once
+      .with(pending_induction_submission_batch_id: PendingInductionSubmissionBatch.last.id)
+    expect(AppropriateBodies::ProcessBatch::RegisterECTJob).to have_been_enqueued.twice
     expect(perform_enqueued_jobs).to be(3)
+    expect(RecordEventJob).to have_been_enqueued.exactly(6).times
+    expect(BeginECTInductionJob).to have_been_enqueued.twice
+
     expect(perform_enqueued_jobs).to be(8)
     expect(Event.all.map(&:heading)).to contain_exactly(
       "The Appropriate Body started a bulk claim",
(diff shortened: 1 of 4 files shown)


app/jobs/appropriate_bodies/process_batch/record_outcome_job.rb
```diff
@@ -0,0 +1,33 @@
...
+        appropriate_body = pending_induction_submission_batch.appropriate_body
+
+        author = Events::AppropriateBodyBatchAuthor.new(
+          email: author_email,
+          name: author_name,
+          appropriate_body_id: appropriate_body.id,
+          batch_id: pending_induction_submission_batch.id
+        )
+
+        teacher = Teacher.find_by(trn: pending_induction_submission.trn)
```
cpjmcquillan, line 19: Do we want to raise an error here if the teacher record is missing (i.e `find_by!`), or will the `RecordOutcome` blow up without a teacher anyway?

cpjmcquillan (approved):
Splitting into smaller jobs that do an explicit action for a given row is much easier to follow IMO! Nice one 👌🏼

app/jobs/appropriate_bodies/process_batch/record_outcome_job.rb
peterdavidhamilton: It will blow but I want to do some follow up work and make `RecordOutcome` and `ReleaseECT` behave the same (whilst also doing the model association. They should look very similar IMO but both services are written slightly different.

spec/features/appropriate_bodies/process_batch_combo_spec.rb
```diff
@@ -40,7 +40,8 @@
     expect(AnalyticsBatchJob).to have_been_enqueued
       .once.with(pending_induction_submission_batch_id: PendingInductionSubmissionBatch.last.id)
 
-    expect(perform_enqueued_jobs).to be(9)
+    expect(perform_enqueued_jobs).to be(3)
+    expect(perform_enqueued_jobs).to be(8)
```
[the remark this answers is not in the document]
peterdavidhamilton: I agree. Is there a nice way of querying queued jobs, otherwise I can resort to comments? I do like having the expectation though because at least it reveals the delayed work going on.

PR #1452, opened 2025-09-29
Convert session debugger to a view component

peterdavidhamilton:
### Context

Role info is not currently surfaced when debugging.

- Refactor into a component ready for more complex session debugging
- Relocate debugger to the footer
- Display the role name for Admin users
- Control rendering using the `TEST_GUIDANCE ` flag
- Redefine the flag on app config as `enable_test_guidance` to align with other feature flags and update specs

### Guidance to review

Screenshot 2025-09-29 at 14 33 29


peteryates:
What do we think about putting it in the footer? Then it's still available but doesn't affect the look of the page by default.

PR #1461, opened 2025-09-30
[2477] Implement latest training period metadata

mooktakim:
### Context

Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/2477

* New model `Metadata::TeacherLeadProvider`
* With columns `latest_ect_training_period` and `latest_mentor_training_period`
* Metadata refresh on `ECTAtSchoolPeriod` and `MentorAtSchoolPeriod`


leandroalemao (approved):
Hi @mooktakim all good 👍 👏  would just add more edge cases to spec/services/metadata/handlers/teacher_spec.rb maybe? ie: for deleting the most recent period, switching lead providers, no periods ..  also how long a metadata refresh would take to run in sandbox for example?

spec/factories/metadata/teacher_lead_provider_factory.rb
```diff
@@ -0,0 +1,6 @@
+FactoryBot.define do
+  factory(:teacher_lead_provider_metadata, class: "Metadata::TeacherLeadProvider") do
+    association :teacher
+    association :lead_provider
```
cwrw, line 4: can we add a trait for including the latest training periods?

app/models/ect_at_school_period.rb
```diff
@@ -16,6 +16,7 @@ class ECTAtSchoolPeriod < ApplicationRecord
   has_one :current_or_next_mentorship_period, -> { current_or_future.earliest_first }, class_name: 'MentorshipPeriod'
 
   refresh_metadata -> { school }, on_event: %i[create destroy update]
+  refresh_metadata -> { teacher }, on_event: %i[create destroy update]
```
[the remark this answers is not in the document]
cwrw: agree with those being on training period and not on school periods 👍

spec/services/metadata/handlers/teacher_spec.rb
```diff
@@ -0,0 +1,202 @@
+RSpec.describe Metadata::Handlers::Teacher do
```
cwrw, line 1: can we add specs for `nil` values, or when someone is an ect and not a mentor, or a teacher without any training periods etc.

app/models/training_period.rb
```diff
@@ -49,6 +50,18 @@ class TrainingPeriod < ApplicationRecord
     left_outer_joins(:ect_at_school_period, :mentor_at_school_period)
       .merge(ECTAtSchoolPeriod.for_school(school).or(MentorAtSchoolPeriod.for_school(school)))
   }
+
+  scope :latest_ect_training_period, ->(teacher:, lead_provider:) {
```
ethax-ross, line 54: Would this make more sense as something like `ect_training_periods_latest_first` so its clear it returns multiple?

app/models/teacher.rb
```diff
@@ -33,6 +36,8 @@ class Teacher < ApplicationRecord
   # TODO: remove after migration complete
   has_many :teacher_migration_failures
 
+  refresh_metadata -> { self }, on_event: %i[create update]
```
cwrw, line 39: will this cause issues with prod data as this is populated, I suppose there are no lead providers but worth confirming

mooktakim: if there's no lead providers the call back won't update anything

app/models/training_period.rb
mooktakim: @peteryates we can't make these `has_one` as we need to filter by lead provider

spec/factories/metadata/teacher_lead_provider_factory.rb
```diff
@@ -0,0 +1,53 @@
...
+      lead_provider = metadata.lead_provider
+
+      contract_period = FactoryBot.create(:contract_period, :current)
+      active_lead_provider = FactoryBot.create(:active_lead_provider, contract_period:, lead_provider:)
+      lead_provider_delivery_partnership = FactoryBot.create(:lead_provider_delivery_partnership, active_lead_provider:)
+      school_partnership = FactoryBot.create(:school_partnership, lead_provider_delivery_partnership:)
+      school = school_partnership.school
+
+      [
+        (:ect if evaluator.ect_teacher_type),
```
ethax-ross, line 22: Would it be easier to pass in this as the value to `ect_teacher_type`? so `ect_teacher_type { :ect }` - I think we will want some teachers with both types as well so maybe supporting an array, `ect_teacher_types { %i[ect mentor] }` could be useful

ethax-ross (commented):
Just a couple of minor suggestions but I think its about there 👍

spec/factories/metadata/teacher_lead_provider_factory.rb
mooktakim: you can't pass in as you might want to do something like this:
```ruby
FactoryBot.create :teacher_lead_provider_metadata, :with_latest_ect_training_period, :with_latest_mentor_training_period
```

You want to use the trait to enable ect or mentor

cwrw (approved):

PR #1462, opened 2025-10-01
Spike participant actions

ethax-ross:
A look into participant actions following the logic defined around withdrawals, deferrals and resuming a participant.

[Discussion on processes](https://github.com/DFE-Digital/register-ects-project-board/discussions/2425)

# Key Concepts
- **Participant status** → determined by **training status** (school view).  
- **Training status** → determined by **training periods** (lead provider view).  

---

# Scenarios

## 1. New school registers an ECT already training elsewhere

- **School A**
  - School period has `finished_on` set.  
    - If `finished_on` is **in the future** → `LEAVING`  
    - If `finished_on` is **in the past** → `LEFT`  
  - Training status is `ACTIVE` until LP withdraws/defers
- **School B**
  - School period has `started_on` set.  
    - If `started_on` is **in the future** → `JOINING`  
    - If `started_on` is **in the past** → `ACTIVE`  
  - Training status is:
    - `JOINING` if `started_on` is in the future
    - `ACTIVE` if `started_on` is in the past

---

## 2. Lead provider submits withdrawal/deferral

- Training period `finished_on` is set.  
- Until `finished_on` is reached:  
  - Training status = `ACTIVE`  
  - Participant status = `ACTIVE`  
- After `finished_on` is reached:  
  - Training status = `WITHDRAWN` (or `DEFERRED`)  
  - Participant status = `LEFT`

---

## 3. Lead provider submits resume

- New training period created:  
  - `started_on` = today  
  - `finished_on` = nil  
- Status updates:  
  - Training status = `ACTIVE`  
  - Participant status = `ACTIVE`  
- Note: there may be a **gap** between training periods

---

# Status Definitions

## Participant status
- **JOINING** → latest training period exists with `started_on` in the future.  
- **LEAVING** → latest training period exists with `finished_on` in the future.  
- **LEFT** →  latest training period has `finished_on` in the past.
- **ACTIVE** → latest school period exists with no `finished_on`.  

---

## Training status
- **N/A** → latest training period exists with `started_on` in the future.  
- **ACTIVE** → latest training period exists with `started_on` in the past and `withdrawn_at`/`deferred_at` are both `nil` or future dated.
- **WITHDRAWN/DEFERRED** → latest training period exists and `withdrawn_at`/`deferred_at` set and in the past.

### DIFF
--- a/app/services/teachers/defer.rb
+++ b/app/services/teachers/defer.rb
@@ -1,0 +1,70 @@
+module Teachers
+  class Defer
+    attr_reader :teacher, :lead_provider, :course_identifier, :deferral_reason
+
+    def initialize(teacher:, lead_provider:, course_identifier:, deferral_reason:)
+      @teacher = teacher
+      @lead_provider = lead_provider
+      @course_identifier = course_identifier
+      @deferral_reason = deferral_reason
+    end
+
+    def defer!
+      finish_training_period! if training_period.ongoing?
+      mark_training_period_as_deferred!
+    end
+
+  private
+
+    def finish_training_period!
+      options = {
+        training_period:,
+        finished_on:,
+        teacher:,
+        author: Events::LeadProviderAPIAuthor.new(lead_provider:),
+      }
+
+      service = if mentor_course_identifier?
+                  TrainingPeriods::Finish.mentor_training(**options.merge(ect_at_school_period:))
+                else
+                  TrainingPeriods::Finish.ect_training(**options.merge(mentor_at_school_period:))
+                end
+
+      service.finish!
+    end
+
+    def mark_training_period_as_deferred!
+      training_period.update!(api_deferred_at: finished_on, deferral_reason:)
+    end
+
+    def finished_on
+      @finished_on ||= Time.zone.now.to_date
+    end
+
+    def ect_at_school_period
+      @ect_at_school_period ||= training_period.ect_at_school_period
+    end
+
+    def mentor_at_school_period
+      @mentor_at_school_period ||= training_period.mentor_at_school_period
+    end
+
+    def training_period
+      @training_period ||= if mentor_course_identifier?
+                             metadata.latest_mentor_training_period
+                           else
+                             metadata.latest_ect_training_period
+                           end
+    end
+
+    def mentor_course_identifier?
+      course_identifier == "ecf-mentor"
+    end
+
+    def metadata
+      @metadata ||= teacher
+        .lead_provider_metadata
+        .find_by(lead_provider:)
+    end
+  end
+end
--- a/app/services/teachers/resume.rb
+++ b/app/services/teachers/resume.rb
@@ -1,0 +1,57 @@
+module Teachers
+  class Resume
+    attr_reader :teacher, :lead_provider, :course_identifier
+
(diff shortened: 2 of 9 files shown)


app/services/teachers/withdraw.rb
```diff
@@ -0,0 +1,70 @@
...
+
+    def initialize(teacher:, lead_provider:, course_identifier:, withdrawal_reason:)
+      @teacher = teacher
+      @lead_provider = lead_provider
+      @course_identifier = course_identifier
+      @withdrawal_reason = withdrawal_reason
+    end
+
+    def withdraw!
+      finish_training_period! if training_period.ongoing?
```
cwrw, line 13: same as above :)

app/services/training_periods/finish.rb
```diff
@@ -51,5 +51,11 @@ def finish!
         )
       end
     end
+
+  private
+
+    def earliest_finished_on
+      [finished_on, training_period.api_withdrawn_at, training_period.api_deferred_at].compact.min
```
cwrw, line 58: ah its here 👍

ethax-ross:
> Just a few thoughts on:
> 
> > Participant status → determined by school periods (school view).
> 
> I don't think that holds, as a provider can withdraw a participant and this will change to left, same with deferral
> 
> > Training status
> > N/A → latest training period exists with started_on in the future.
> 
> This will be active for training status

Ah yes you're right, if a lead provider withdraws the participant status becomes `left` even if the period is ongoing

app/services/teachers/resume.rb
```diff
@@ -0,0 +1,57 @@
...
+    attr_reader :teacher, :lead_provider, :course_identifier
+
+    def initialize(teacher:, lead_provider:, course_identifier:)
+      @teacher = teacher
+      @lead_provider = lead_provider
+      @course_identifier = course_identifier
+    end
+
+    def resume!
+      create_training_period!
```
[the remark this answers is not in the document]
ethax-ross: Is this in terms of validation rules?

app/services/teachers/defer.rb
```diff
@@ -0,0 +1,70 @@
...
+
+    def initialize(teacher:, lead_provider:, course_identifier:, deferral_reason:)
+      @teacher = teacher
+      @lead_provider = lead_provider
+      @course_identifier = course_identifier
+      @deferral_reason = deferral_reason
+    end
+
+    def defer!
+      finish_training_period! if training_period.ongoing?
```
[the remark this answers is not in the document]
ethax-ross: I'm not sure I follow 🤔 if the school period is finished then the training period will also be finished at the same time won't it? so we don't need to include the school period `finished_on` when setting the training period `finished_on` as it can never be later?

I think a bit that's missing/I didn't include in the spike is if a training period has been deferred and then the school period is finished but with an earlier date; in that case we need to back-date the training period `finished_on`?

app/services/training_periods/finish.rb
ethax-ross: Ah yes, but there's a bug in my spike as this would always be called before the training period being withdrawn/deferred how I've currently got it. I need to switch the order to be:

```
mark_training_period_as_withdrawn!
finish_training_period! if training_period.ongoing?
```

PR #1464, opened 2025-10-01
Destroy batch dependents (submissions and events)

peterdavidhamilton:
### Context

There is a `failed` bulk upload, a state we should never see, because our Redis cache was down.

We want to delete it because we have a Blazer email check being triggered by this false positive. 

The script to delete failed batches will not work unless we also delete dependent records.


app/models/pending_induction_submission_batch.rb
```diff
@@ -28,7 +28,8 @@ def self.new_action_for(appropriate_body:, **)
 
   # Associations
   belongs_to :appropriate_body
-  has_many :pending_induction_submissions
+  has_many :pending_induction_submissions, dependent: :destroy
+  has_many :events, dependent: :destroy
```
cpjmcquillan, line 32: Does it **ever** make sense to destroy Events? 🤔

Philosophical!

I wonder whether we could use something like `dependent: :nullify` so that we can safely delete the `PendingInductionSubmissionBatch` without losing potentially useful information in the form of the `Event` record.

What do you think?

cpjmcquillan (approved):

app/models/pending_induction_submission_batch.rb
peterdavidhamilton: It'll be a single event most of the time, for example `AB started a bulk claim`. Good idea to err on the side of caution though, will tweak.

As for the principal of deleting events linked to teachers and inductions that the upcoming work to purge will remove, would you destroy or nullify?

PR #1474, opened 2025-10-03
Fix some broken bits of HTML

peteryates:
Namely:

- using `</br>` instead of `<br>`. MDN says:
  > Must have a start tag and must not have an end tag. In XHTML
  > documents, write this element as `<br />`.
- closing opened `<p>` paragraphs
- fixing mismatching headers

Thanks @rjlynch for suggesting [herb](https://github.com/marcoroth/herb)!

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -84,8 +84,8 @@ jobs:
         run: |-
           npm run lint:scss
 
-  erb_linting:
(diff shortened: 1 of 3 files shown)


cpjmcquillan (approved):
I've been running `herb` for a little while with my LSP but not auto-correcting anything. No more squiggly lines for me 🎉

PR #1476, opened 2025-10-03
Fix a failing spec by hardcoding the period dates

peteryates:
We have a few instances of these specs failing because the sequence in the training period factory appears to push the start date beyond the finish date.

```
ActiveRecord::StatementInvalid:
  PG::DataException: ERROR:··range lower bound must be less than or equal to range upper bound
```

In the longer term I think we ought to stop using the base dates (or restrict them to traits) and hardcode them where possible, as in the long run it'll be more robust (at the expense of being a bit more verbose).


ethax-ross (approved):
Thanks Pete 👍

PR #1477, opened 2025-10-03
Add deferral/withdrawal attribute to `TrainingPeriod`

ethax-ross:
### Context

In order to support deferrals and withdrawals we need to stamp the date and a reason for the action on the teacher's `TrainingPeriod`.

Add `withdrawn_at`, `withdrawal_reason`, `deferred_at` and `deferral_reason` to `TrainingPeriod`.

### DIFF
--- a/app/models/training_period.rb
+++ b/app/models/training_period.rb
@@ -41,6 +41,11 @@ class TrainingPeriod < ApplicationRecord
   validate :trainee_distinct_period
   validate :enveloped_by_trainee_at_school_period
   validate :only_provider_led_mentor_training
+  validate :withdrawn_deferred_are_mutually_exclusive
+  validates :withdrawn_at, presence: true, if: -> { withdrawal_reason.present? }
+  validates :withdrawal_reason, presence: true, if: -> { withdrawn_at.present? }
+  validates :deferred_at, presence: true, if: -> { deferral_reason.present? }
+  validates :deferral_reason, presence: true, if: -> { deferred_at.present? }
 
   # Scopes
   scope :for_ect, ->(ect_at_school_period_id) { where(ect_at_school_period_id:) }
@@ -137,5 +142,11 @@ private
     return if school_partnership.blank?
 
     errors.add(:school_partnership, 'School partnership must be absent for school-led training programmes')
+  end
+
+  def withdrawn_deferred_are_mutually_exclusive
+    return unless withdrawn_at.present? && deferred_at.present?
+
+    errors.add(:base, "A training period cannot be both withdrawn and deferred")
   end
 end
--- a/config/analytics_blocklist.yml
+++ b/config/analytics_blocklist.yml
@@ -191,6 +191,10 @@
   - ecf_end_induction_record_id
   - expression_of_interest_id
   - training_programme
+  - deferred_at
+  - deferral_reason
+  - withdrawn_at
+  - withdrawal_reason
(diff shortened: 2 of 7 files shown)


app/models/training_period.rb
```diff
@@ -41,6 +41,10 @@ class TrainingPeriod < ApplicationRecord
   validate :trainee_distinct_period
   validate :enveloped_by_trainee_at_school_period
   validate :only_provider_led_mentor_training
+  validates :api_withdrawn_at, presence: true, if: -> { api_withdrawal_reason.present? }
+  validates :api_withdrawal_reason, presence: true, if: -> { api_withdrawn_at.present? }
+  validates :api_deferred_at, presence: true, if: -> { api_deferral_reason.present? }
+  validates :api_deferral_reason, presence: true, if: -> { api_deferred_at.present? }
```
peteryates, line 47: I know this is in draft so just an initial though, but I assume someone's training can either be withdrawn or deferred, not both. In some of the earlier discussions we had this as `closure_reason` with a single enum that had a prefix, so `withdrawn-left-teaching-profession` or `deferred-bereavement`. Then there's no need to add validation to make sure there's only one present.

ethax-ross: So I think you're right in that a `TrainingPeriod` could only be withdrawn or deferred but not both. The only knock-on effect I can see with combining them is that we would need to format the value when we accept/return it from the API, so map to/from:

`withdraw-left-teaching-profession`/`left-teaching-profession`

That shouldn't be a big deal though and we do it already for withdrawal reasons after a change to some of the programme type namings (see [this in ECF](https://github.com/DFE-Digital/early-careers-framework/blob/main/app/services/programme_type_mappings.rb#L9-L18)). I'm going to hold off for Abeer's thoughts before I rework it though in case I'm missing anything; I'll add the mutually exclusive validation for now to the existing attributes 👍

peteryates: Ah, yeah - that makes sense!

ethax-ross: Hey @peteryates, I caught up with @cwrw this morning and we discussed the idea of combining the timestamp and reason fields. 

It feels like combining them would make it clearer that there’s only one reason to close off a training period and would reduce validation. However, it will also complicate API slightly with mapping to/from and it feels like it makes the overall deferred/withdrawn state less clear. It will also add a bit of complexity to querying and conditional logic (since we’d have to inspect the reason to determine state). For these reasons, we think it’s going to make more sense to keep them separate. Does that sound reasonable?

db/migrate/20251003110349_add_withdrawn_deferred_to_training_periods.rb
```diff
@@ -0,0 +1,26 @@
...
+      long-term-sickness
+      parental-leave
+      career-break
+      other
+    ]
+
+    change_table :training_periods, bulk: true do |t|
+      t.date :api_deferred_at
+      t.enum :api_deferral_reason, enum_type: :api_deferral_reasons
+      t.date :api_withdrawn_at
```
mooktakim, line 22: Should we make the type `datetime`? especially as the columns end in `_at` which usually means datetime.

ethax-ross: yeah looking at the API spec this should be a datetime, I'll update 👍

mooktakim (approved):

PR #1479, opened 2025-10-03
Fix flaky test

cpjmcquillan:
Before, this test could fail due to some ambigious matching!!!

If the "Lead Provider" ID clashed with the request IDs created in the test (i.e `123` or `456` were a subset of the lead provider ID (e.g `1237`).

Here is an example [failure].

This refines the expectation so we're looking specifically at the relevant part of the parity check diff.

Now, the test always passes!

[failure]: https://github.com/DFE-Digital/register-early-career-teachers-public/actions/runs/18224131789/job/51890828513


cwrw:
Oops, thanks @cpjmcquillan!

PR #1484, opened 2025-10-05
Schools can change an ECT's mentor to another registered mentor

cpjmcquillan:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/2102

This builds on some similar journeys and introduces a "Change mentor" journey
for schools.

Even though this might seem straightforward, there is a bunch of logic and a few
different branches to cover.

This change covers:

- Changing to a registered mentor who is not eligible for training
- Changing to a registered mentor who is eligible for training and using the
  existing provider
- Changing to a registered mentor who is eligible for training and choosing a
  different provider
- Changing to a registered mentor when the ECT is undergoing school-led training

Lots of this is tests and boilerplate.

Some steps share quite a few methods. Previous iterations of this change
experimented with a few different abstractions, but none quite felt right.

This keeps things explicit and verbose (if a little repetitive) for now. We can
always circle back later and make improvements if things become clearer.

### DIFF
--- a/app/views/schools/ects/change_mentor_wizard/check_answers.html.erb
+++ b/app/views/schools/ects/change_mentor_wizard/check_answers.html.erb
@@ -1,0 +1,31 @@
+<% page_data(
+    title: "Check and confirm change",
+    backlink_href: @wizard.previous_step_path
+) %>
+
+<%= govuk_summary_list do |list| %>
+  <% list.with_row do |row| %>
+    <%= row.with_key { "Early career teacher" } %>
+    <%= row.with_value { @wizard.teacher_full_name } %>
+  <% end %>
+
+  <% list.with_row do |row| %>
+    <%= row.with_key { "Current mentor" } %>
+    <%= row.with_value { @wizard.current_step.current_mentor_name } %>
+  <% end %>
+
+  <% list.with_row do |row| %>
+    <%= row.with_key { "New mentor" } %>
+    <%= row.with_value { @wizard.current_step.new_mentor_name } %>
+  <% end %>
+<% end %>
+
+<%= govuk_button_to "Confirm change", @wizard.current_step_path %>
+
+<p class="govuk-body">
+  <%= govuk_link_to(
+        "Cancel and go back to #{@wizard.teacher_full_name}’s details",
+        schools_ect_path(@wizard.ect_at_school_period),
+        no_visited_state: true
+      ) %>
+</p>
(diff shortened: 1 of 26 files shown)


cpjmcquillan:
I've managed to fix the flaky tests (base date related), so this is ready for review now.

I'm not overly happy with the abstractions at the minute. I've got a WIP branch that tries to integrate the `MentorAssignmentContext` and reuse some new partials etc, but I'm going to make an issue to pick this up later 👍🏼

app/wizards/schools/ects/change_mentor_wizard/training_step.rb
```diff
@@ -0,0 +1,36 @@
+module Schools
+  module ECTs
+    module ChangeMentorWizard
+      class TrainingStep < Step
```
avinhurry: Could we call this `ReviewMentorEligibilityStep` for consistency with the others that are the same? 

It might even be worth extracting a shared partial for the view since this is repeated in three places?

- https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/views/schools/register_mentor_wizard/review_mentor_eligibility.html.erb

- https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/views/schools/assign_existing_mentor_wizard/review_mentor_eligibility.html.erb

app/views/schools/ects/change_mentor_wizard/check_answers.html.erb
```diff
@@ -0,0 +1,31 @@
...
+    <%= row.with_value { @wizard.teacher_full_name } %>
+  <% end %>
+
+  <% list.with_row do |row| %>
+    <%= row.with_key { "Current mentor" } %>
+    <%= row.with_value { @wizard.current_step.current_mentor_name } %>
+  <% end %>
+
+  <% list.with_row do |row| %>
+    <%= row.with_key { "New mentor" } %>
```
avinhurry: Do we need to render a lead provider row depending on if the user goes through that path?

avinhurry (approved):
Going to queue this up to merge because it's quite big already and @thenapking is waiting for some of it to progress his card.

I'm going to make the changes in follow up PRs.

app/wizards/schools/ects/wizard.rb
```diff
@@ -7,9 +7,12 @@ def self.step?(step_name)
         Array(steps).first[step_name].present?
       end
 
+      # @return [String]
+      def name_for(...) = ::Teachers::Name.new(...).full_name
```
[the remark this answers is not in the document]
avinhurry: I was wrong, didn't realise it was used in the steps too. Sorry!

PR #1491, opened 2025-10-07
Add `Teachers::Query` service

ethax-ross:
### Context

We are adding the `Teachers::Query` service to allow us to query teachers according to the API specification.

- Make api_from_teacher_id optionally set

If we want to specify our own `api_from_teacher_id` to avoid extra creations (which throws off our tests of the `Teachers::Query`, for example) its useful to be able to pass in the `api_from_teacher_id`.

- Add Teachers::Query

### Guidance to review

Note that we have to `eager_load` instead of `includes` when preloading the associations in order to be able to filter out the metadata not associated with the given lead provider. As `eager_load` can be a little less efficient that `includes` (it creates a single, bloated query and the results are paired back in memory) we may want to drop the metadata filtering and `includes` instead if we find performance is an issue.

Filtering on `contract_period_year` is also non-trivial in this query due to needing to filter across two relationships
(`latest_ect_training_period` OR `latest_mentor_training_period`). The alias names for joined tables Rails generates are not very clear. We will look into adding the contract periods to `Metadata::TeacherLeadProvider` to clean this up and make it more efficient in a follow up ticket.

#### Benchmarking

Running the query on the parity check env yields the following results:

##### 1,000 items per page

```
{
    "no filters": {
        "label": "",
        "real": 2.656694935634732,
        "cstime": 0.0,
        "cutime": 0.0,
        "stime": 0.6337030000000001,
        "utime": 0.9968810000000001,
        "total": 1.630584
    },
    "lead provider filtering": {
        "label": "",
        "real": 0.9812330780550838,
        "cstime": 0.0,
        "cutime": 0.0,
        "stime": 0.24396299999999993,
        "utime": 0.2689579999999996,
        "total": 0.5129209999999995
    },
    "contract period filtering": {
        "label": "",
        "real": 4.378857615403831,
        "cstime": 0.0,
        "cutime": 0.0,
        "stime": 0.5035270000000001,
        "utime": 1.1909360000000007,
        "total": 1.6944630000000007
    },
    "updated since filtering": {
        "label": "",
        "real": 2.023762419819832,
        "cstime": 0.0,
        "cutime": 0.0,
        "stime": 0.1946370000000002,
        "utime": 0.821847,
        "total": 1.016484
    }
}
```

###### 100 items per page

```
{
    "no filters": {
        "label": "",
        "real": 1.234193793963641,
        "cstime": 0.0,
        "cutime": 0.0,
        "stime": 0.42882599999999993,
        "utime": 0.2486369999999996,
        "total": 0.6774629999999995
    },
    "lead provider filtering": {
        "label": "",
        "real": 0.6680415971204638,
        "cstime": 0.0,
        "cutime": 0.0,
        "stime": 0.14522000000000013,
        "utime": 0.07106499999999993,
        "total": 0.21628500000000006
    },
    "contract period filtering": {
        "label": "",
        "real": 2.08220246527344,
        "cstime": 0.0,
        "cutime": 0.0,
        "stime": 0.4018510000000002,
        "utime": 0.1903639999999998,
        "total": 0.5922149999999999
    },
    "updated since filtering": {
        "label": "",
        "real": 0.6533589838072658,
        "cstime": 0.0,
        "cutime": 0.0,
        "stime": 0.04337200000000019,
        "utime": 0.05860899999999969,
        "total": 0.10198099999999988
    }
}
````

### DIFF
--- a/app/services/api/teachers/query.rb
+++ b/app/services/api/teachers/query.rb
@@ -1,0 +1,161 @@
+module API::Teachers
+  class Query
+    include Queries::FilterIgnorable
+
+    attr_reader :scope, :lead_provider_id
+
+    def initialize(
+      lead_provider_id: :ignore,
+      contract_period_years: :ignore,
+      training_status: :ignore,
+      api_from_teacher_id: :ignore,
+      updated_since: :ignore,
+      sort: { created_at: :asc }
+    )
+      @lead_provider_id = lead_provider_id
+      @scope = Teacher.distinct
+
+      where_lead_provider_is(lead_provider_id)
+      where_contract_period_year_in(contract_period_years)
+      where_training_status_is(training_status)
+      where_api_from_teacher_id_is(api_from_teacher_id)
+      where_updated_since(updated_since)
+      set_sort_by(sort)
+    end
+
+    def teachers
+      preload_associations(block_given? ? yield(scope) : scope)
+    end
+
+    def teacher_by_api_id(api_id)
+      return preload_associations(scope).find_by!(api_id:) if api_id.present?
+
+      fail(ArgumentError, "api_id needed")
+    end
+
+    def teacher_by_id(id)
+      return preload_associations(scope).find(id) if id.present?
+
+      fail(ArgumentError, "id needed")
+    end
+
+  private
+
+    def preload_associations(results)
+      preloaded_results = results
+        .strict_loading
+        .eager_load(
+          :earliest_ect_at_school_period,
+          :earliest_mentor_at_school_period,
+          :teacher_id_changes,
+          lead_provider_metadata: {
+            latest_ect_training_period: {
+              school_partnership: %i[
+                school
+                contract_period
+                delivery_partner
+              ],
+              ect_at_school_period: {
+                teacher: %i[
+                  started_induction_period
+                  finished_induction_period
+                ]
+              }
+            },
+            latest_mentor_training_period: {
+              school_partnership: %i[
+                school
+                contract_period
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: app/services/api/teachers/query.rb [hunk 1]
```diff
@@ -0,0 +1,156 @@
...
+    end
+
+  private
+
+    def preload_associations(results)
+      preloaded_results = results
+        .strict_loading
+        .eager_load(
+          :teacher_id_changes,
+          lead_provider_metadata: {
```
cwrw, line 49: can we add `earliest_ect_at_school_period` and `earliest_mentor_at_school_period`

app/services/api/teachers/query.rb
```diff
@@ -0,0 +1,156 @@
...
+      preloaded_results
+    end
+
+    def where_lead_provider_is(lead_provider_id)
+      return if ignore?(filter: lead_provider_id)
+
+      @scope = scope
+        .joins(:lead_provider_metadata)
+        .where(lead_provider_metadata: { lead_provider_id: })
+        .where.not(
```
leandroalemao, line 94: Hi @ethax-ross quick question pls.. why do we need those here? worth adding a comment?

```
latest_ect_training_period: nil,
latest_mentor_training_period: nil
```

leandroalemao (approved):

cwrw (approved):
We'll need to revisit the performance, with cohort on metadata and maybe beefing up the env and spending time on performance tuning

mooktakim (approved):

PR #1493, opened 2025-10-07
[2518] Add `eligibilty_for_funding` to data model

leandroalemao:
### Context

Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/2518

- Add the eligibility fields for ECT/Mentor to teacher
- Create a service to update those values (if the timestamp is populated it will never get unpopulated, or changed)
- Call the teacher funding eligibility service within some other services when registering a teacher as ect or mentor

### DIFF
--- a/spec/views/schools/ects/show.html.erb_spec.rb
+++ b/spec/views/schools/ects/show.html.erb_spec.rb
@@ -66,7 +66,11 @@ RSpec.describe 'schools/ects/show.html.erb' do
       context 'when assigned' do
         before do
           mentor = FactoryBot.create(:teacher, trs_first_name: 'Moby', trs_last_name: 'Dick')
-          mentor_at_school_period = FactoryBot.create(:mentor_at_school_period, :ongoing, school: current_school, teacher: mentor)
+          mentor_at_school_period = FactoryBot.create(:mentor_at_school_period,
+                                                      started_on: current_ect_period.started_on,
+                                                      finished_on: nil,
+                                                      school: current_school,
+                                                      teacher: mentor)
 
           FactoryBot.create(:mentorship_period, :ongoing,
                             started_on: current_ect_period.started_on,


spec/validators/immutable_once_set_validator_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+  end
+
+  context "when attribute is `nil` and changing to `nil`" do
+    let(:current_value) { nil }
+    let(:new_value) { nil }
+
+    it { is_expected.to be_empty }
+  end
+
+  context "when attribute is `\"set\"` and changing to `\"set\"`" do
```
ethax-ross, line 28: Not sure we need double quotations on these 🙂 can we just have one or the other?

app/services/events/record.rb
```diff
@@ -451,6 +451,14 @@ def self.record_training_period_assigned_to_school_partnership_event!(
       ).record_event!
     end
 
+    def self.record_teacher_set_funding_eligibilty_event!(author:, teacher:, happened_at:)
```
ethax-ross, line 454: Do we want to distinguish between ECT/mentor in the events?

spec/services/teachers/set_funding_eligibilty_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+
+  describe "#set!" do
+    context "when teacher is eligible for ECT training" do
+      before do
+        FactoryBot.create(:induction_period, :ongoing, teacher:)
+        FactoryBot.create(:ect_at_school_period, :ongoing, teacher:)
+      end
+
+      it "sets `first_became_eligible_for_ect_training_at` if not already set" do
+        expect { service.set! }.to change(teacher, :first_became_eligible_for_ect_training_at).from(nil)
```
ethax-ross, line 14: Can we expect it to `be_within(5.seconds).of(Time.zone.now)` on the tests that check the attributes are set?

spec/services/teachers/set_funding_eligibilty_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+      end
+
+      it "does not record a teacher set funding eligibility event" do
+        expect(Events::Record).not_to receive(:record_teacher_set_funding_eligibilty_event!)
+
+        service.set!
+      end
+    end
+
+    it "saves the teacher" do
```
ethax-ross, line 80: I think this and the 'runs inside a transaction' are already covered by your other tests

ethax-ross (commented):
Looks good just some minor suggestions 👍

cwrw (commented):
looks great thanks @leandroalemao just one question on naming, but happy to keep as is as well, @peteryates just an FYI we've included setting eligibility into services

spec/services/events/record_spec.rb
```diff
@@ -1526,4 +1526,27 @@
...
+        Events::Record.record_teacher_set_funding_eligibilty_event!(author:, teacher:, happened_at:, modifications: raw_modifications)
+
+        expect(RecordEventJob).to have_received(:perform_later).with(
+          hash_including(
+            **author_params,
+            event_type: :teacher_funding_eligibilty_set,
+            happened_at:,
+            heading: "Rhys Ifans funding eligibilty was set",
+            metadata: raw_modifications,
+            modifications: anything,
```
ethax-ross, line 1545: Can we assert on the changes here rather than `anything`?

ethax-ross (approved):
Looks good past Abeer's comment on naming.

app/services/events/record.rb
```diff
@@ -451,6 +451,14 @@ def self.record_training_period_assigned_to_school_partnership_event!(
       ).record_event!
     end
 
+    def self.record_teacher_set_funding_eligibilty_event!(author:, teacher:, happened_at:, modifications:)
+      event_type = :teacher_funding_eligibilty_set
+      teacher_name = Teachers::Name.new(teacher).full_name
+      heading = "#{teacher_name} funding eligibilty was set"
```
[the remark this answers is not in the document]
peteryates: Ah it's in the method name too! I do this all the time too.

app/models/teacher.rb
```diff
@@ -56,6 +56,8 @@ class Teacher < ApplicationRecord
   validates :api_id, uniqueness: { case_sensitive: false, message: "API id already exists for another teacher" }
   validates :api_ect_training_record_id, uniqueness: { case_sensitive: false, message: "API ect training record id already exists for another teacher" }
   validates :api_mentor_training_record_id, uniqueness: { case_sensitive: false, message: "API mentor training record id already exists for another teacher" }
+  validates :first_became_eligible_for_ect_training_at, immutable_once_set: true
+  validates :first_became_eligible_for_mentor_training_at, immutable_once_set: true
```
peteryates, line 60: It feels to me like we're bringing in quite a big dependency here and only using a tiny bit of its functionality.

I think we can _probably_ achieve this with something like:

```ruby
validate :first_became_eligible_for_ect_training_at_does_not_change, on: :update

def first_became_eligible_for_ect_training_at_does_not_change
  if first_became_eligible_for_ect_training_at_changed?
    errors.add(:first_became_eligible_for_ect_training_at, "Not allowed ...")
  end
end
```

PR #1495, opened 2025-10-07
[2239] Spike on nuclear withdrawals - part 1

leandroalemao:
### Context

Spike ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/2239

- add a `withdrawn_by_error` flag: Introduce a boolean column, withdrawn_by_error, to the ect/mentor_at_school_period model. This flag will explicitly mark records corresponding to accidental registrations

- modify Interval concern: Update the overlapping validation logic to ignore ect/mentor_at_school_period records where `withdrawn_by_error` is true when checking for overlaps. This will prevent validation issues when a participant is legitimately claimed or transferred by/to another school after a mistaken registration.

- use `EctAtSchoolPeriod.active` where needed in API related queries / metadata refresh methods

- decided not to go for `default_scope` to ect/mentor at school periods, as it can hide data unexpectedly data in other places and adding `unscoped` to these places is often forgotten

- I'm suggesting the name of the field to be `withdrawn_by_error` but this could be only a boolean field called `withdrawn` also with `closed_at` field with to stamp when the teacher was withdrawn and a `withdrawn_reason= 'legacy_ecf1_withdrawn'` to mark the reason for that.

### DIFF
--- a/app/models/ect_at_school_period.rb
+++ b/app/models/ect_at_school_period.rb
@@ -59,6 +59,7 @@ class ECTAtSchoolPeriod < ApplicationRecord
     with_expressions_of_interest_for_contract_period(year)
     .where(expression_of_interest: { lead_provider_id: })
   }
+  scope :active, -> { where(withdrawn_by_error: false) }
 
   def school_reported_appropriate_body_name = school_reported_appropriate_body&.name
 
@@ -67,7 +68,7 @@ class ECTAtSchoolPeriod < ApplicationRecord
   def siblings
     return ECTAtSchoolPeriod.none unless teacher
 
-    teacher.ect_at_school_periods.excluding(self)
+    teacher.ect_at_school_periods.active.excluding(self)
   end
 
   delegate :trn, to: :teacher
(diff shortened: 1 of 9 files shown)


app/models/ect_at_school_period.rb
```diff
@@ -59,6 +59,7 @@ class ECTAtSchoolPeriod < ApplicationRecord
     with_expressions_of_interest_for_contract_period(year)
     .where(expression_of_interest: { lead_provider_id: })
   }
+  scope :active, -> { where(withdrawn_by_error: false) }
```
cwrw, line 62: Thanks Leandro, I guess my concern with only including this in siblings and thus the overlap validation is ignored, is that the rest of the scopes will return this record especially where we do not want it to be returned or included.

Maybe it would be good to try the default scope, where it is ignored and since we have to include it in the API how that would look like in terms of metadata and other queries that matter, happy to chat

cwrw:
thanks Leandro, closing as spike is done :)

PR #1500, opened 2025-10-08
Prevent school periods that end before they start

leoapost:
### Context
Some induction records report that a participant left the school before their start date. These incorrect records cause issues when extracting school periods.

To address this, such induction records should be ignored. Instead, the extractor should use the most recent induction record where the school period has a valid date range (end date after the start date).

- Update the SchoolPeriodExtractor to ignore induction records where the end date is before the period's start date.


leoapost:
Closing this PR for now. A wider investigation is ongoing and we will revisit the migration code changes once that's complete.

PR #1501, opened 2025-10-08
[2199] - set first mentor_at_school_period created_at time to that of its participant profile imported from ECF

ltello:
[Ticket](https://github.com/DFE-Digital/register-ects-project-board/issues/2199)

### Context
When we migrate a ParticipantProfile::Mentor record we store the ECF ID on the Teacher record in the attribute ecf_mentor_profile_id. We also need to store the ParticipantProfile.created_at attribute in RECT alongside this.

We should store the ParticipantProfile::Mentor.created_at timestamp in the earliest/first MentorAtSchoolPeriod.created_at

Updates the Migrators::MentorAtSchoolPeriod to store the original ECF ParticipantProfile::Mentor.created_at in the first MentorAtSchoolPeriod record's created_at attribute.

### DIFF
--- a/app/migration/builders/mentor/school_periods.rb
+++ b/app/migration/builders/mentor/school_periods.rb
@@ -1,21 +1,23 @@
 module Builders
   module Mentor
     class SchoolPeriods
-      attr_reader :teacher, :school_periods
+      attr_reader :teacher, :school_periods, :created_at
 
-      def initialize(teacher:, school_periods:)
+      def initialize(teacher:, school_periods:, created_at:)
         @teacher = teacher
         @school_periods = school_periods
+        @created_at = created_at
       end
 
       def build
         success = true
-        school_periods.each do |period|
+        school_periods.each_with_index do |period, idx|
           school = find_school_by_urn!(period.urn)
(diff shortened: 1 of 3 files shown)


app/migration/builders/mentor/school_periods.rb
```diff
@@ -1,21 +1,23 @@
...
       def build
         success = true
-        school_periods.each do |period|
+        school_periods.each_with_index do |period, idx|
           school = find_school_by_urn!(period.urn)
           school_period = ::MentorAtSchoolPeriod.find_or_initialize_by(teacher:, school:, started_on: period.start_date)
           school_period.finished_on = period.end_date
           school_period.ecf_start_induction_record_id = period.start_source_id
           school_period.ecf_end_induction_record_id = period.end_source_id
+          school_period.created_at = created_at if idx.zero?
```
peteryates, line 20: Is this going to set `created_at` to `nil` when it's not zero?

```ruby
register-early-career-teachers(dev)> a = 'test' if false
=> nil
register-early-career-teachers(dev)> a
=> nil
```

PR #1502, opened 2025-10-08
Add `mentor_api_id` to `Metadata::TeacherLeadProvider`

ethax-ross:
### Context

We want to stamp the `api_mentor_id` on the `Metadata::TeacherLeadProvider` so that we can easily include it in the `TeacherSerializer`. This will always be the mentor of the latest `MentorshipPeriod` relative to the `LeadProvider`.

- Add `api_mentor_id` to `Metadata::TeacherLeadProvider`

This will be used to store the mentor for a given teacher relative to each lead provider based on their latest ECT training/mentorship period.

- Populate `api_mentor_id` on `Metadata::TeacherLeadProvider`

Populate the `api_mentor_id` on `Metadata::TeacherLeadProvider` from the `latest_ect_training_period` and `latest_mentorship_period`.

Add `latest_mentorship_period` scope to find the latest mentor for a teacher.

- Refresh `Metadata::TeacherLeadProvider` for `api_mentor_id`

We need to refresh the `Metadata::TeacherLeadProvider` to update `api_mentor_id` when any of the following occur:

`ECTAtSchoolPeriod` created/destroyed
`MentorshipPeriod` created/destroyed

This assumes the `mentor` will not change on a `MentorshipPeriod`.

The `teacher` metadata is already refreshed when a `TrainingPeriod` is created/destroyed/updated.

Minor update to the declarative updates shared context as we now have scenarios where the `instance` and `target` refresh the metadata on `create`, so we have multiple calls to `refresh_metadata!`.

- Add `api_mentor_id` to `TeacherSerializer`

We can now populate the `mentor_api_id` from the value we stamp on the `Metadata::TeacherLeadProvider`.

Inline the metadata lookup as we can't call the class method from the nested classes/serializers.

### DIFF
--- a/app/models/ect_at_school_period.rb
+++ b/app/models/ect_at_school_period.rb
@@ -17,6 +17,7 @@ class ECTAtSchoolPeriod < ApplicationRecord
   has_one :latest_mentorship_period, -> { latest_first }, class_name: 'MentorshipPeriod'
 
   refresh_metadata -> { school }, on_event: %i[create destroy update]
+  refresh_metadata -> { teacher }, on_event: %i[create destroy]
 
   # Validations
   validate :appropriate_body_for_independent_school,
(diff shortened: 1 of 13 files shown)


mooktakim (approved):
One comment, but otherwise looks good

leandroalemao (approved):

cwrw (approved):

PR #1527, opened 2025-10-10
[2561] - migrate frozen cohort attributes

ltello:
### Context
[Issue](https://github.com/DFE-Digital/register-ects-project-board/issues/2561)

We need to reflect in RECT when an ECT or Mentor has been moved from a frozen cohort in ECF.

This will require us to store the cohort year on either `Teacher.ect_payments_frozen_year` or `Teacher.mentor_payments_frozen_year` when the ECF `ParticipantProfile#cohort_changed_after_payments_frozen` is true.

`Teacher.ect_payments_frozen_year` is correctly populated for `ParticipantProfile::ECT` records when cohort_changed_after_payments_frozen is set in ECF.

`Teacher.mentor_payments_frozen_year` is correctly populated for `ParticipantProfile::Mentor` records when cohort_changed_after_payments_frozen is set in ECF.

`Teacher.ect_payments_frozen_year` and `Teacher.mentor_payments_frozen_year` are nil when the participant has not moved from a frozen cohort (when cohort_changed_after_payments_frozen is not set in ECF).

We have infered the previous frozen cohort of a participant from their induction records (rather than from their declarations) for performance reasons: we had induction records and most of their relationships already eager loaded in the migrator (as opposed to extend the query to retrieve declarations and their associations)

### DIFF
--- a/spec/models/migration/cohort_spec.rb
+++ b/spec/models/migration/cohort_spec.rb
@@ -14,7 +14,7 @@ describe Migration::Cohort, type: :model do
       it do
         freeze_time do
           subject.payments_frozen_at = 1.second.ago
-          is_expected.to be_payments_frozen
+          expect(subject).to be_payments_frozen
         end
       end
     end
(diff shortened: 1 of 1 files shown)


tonyheadford (approved):
LGTM

PR #1530, opened 2025-10-10
[2554] Link schedule to training period

leandroalemao:
### Context

Ticket: [2554](https://github.com/DFE-Digital/register-ects-project-board/issues/2554)

After the various investigations and data model updates we need to ensure the data model is fully working. The schedules model was added but we aren't linking the schedule to the training period yet.

- Add schedule to training period;
- Validate to make sure the contract period from the schedule is the same as the contract period from the school partnership;
- Seeds also creates `provider-led` training periods with schedules now;
- Factories also comply with the validation;


db/migrate/20251013110710_backfill_training_periods_schedule_ids.rb
```diff
@@ -0,0 +1,17 @@
+class BackfillTrainingPeriodsScheduleIds < ActiveRecord::Migration[8.0]
+  def up
+    TrainingPeriod.find_each do |training_period|
+      contract_period = training_period.contract_period || training_period.expression_of_interest_contract_period
+
+      default_schedule = Schedule.find_by(contract_period:, identifier: "ecf-standard-september")
+
+      if default_schedule
```
ethax-ross: Would we not want to error out if the default schedule is not found to avoid this failing silently?

spec/factories/training_period_factory.rb
```diff
@@ -37,17 +38,36 @@
...
     end
 
+    trait :with_schedule do
+      transient do
+        schedule { FactoryBot.build(:schedule, contract_period: contract_period || expression_of_interest_contract_period) }
+      end
+
+      after(:build) do |training_period, evaluator|
+        # Only `provider_led` training periods should have a schedule
+        if training_period.provider_led_training_programme?
```
ethax-ross: I think if we added validation to the training period we could drop this check?

spec/wizards/schools/register_ect_wizard/ect_spec.rb
```diff
@@ -640,6 +640,7 @@
           FactoryBot.create(
             :training_period,
             :for_ect,
+            :with_only_expression_of_interest,
             ect_at_school_period: ect_period,
             started_on: Date.new(2025, 3, 10),
             expression_of_interest:
```
cwrw: do we need to remove this as we're setting the trait?

cwrw (approved):
looks good just one final minor comment, thanks Leandro, also looks like seed data failing 🤔

PR #1536, opened 2025-10-13
Implement GET participants show/index actions

ethax-ross:
### Context

Now that we have the query and serializer we can implement the `ParticipantsController` `index` and `show` actions.

- Implement GET participants show/index actions

Fix up the routes so the correct `api_id` parameter is used (as it was previously the id parameter is inferred as `participants_api_id`).

Fix eager loading in the `Participants::Query` and use `include` on `teacher_id_changes` for better performance.

Implement `ParticipantsController`, calling the existing query and serializer.

- Add training_status filtering to participants endpoint

We need to allow lead providers to filter participants by a given `training_status`. As we don't store the training status like we did in ECF we have to calculate this state on the fly.

We have a service to calculate it for a given `TrainingPeriod`, but in the query we need to replicate the logic to determine the state efficiently.

We surface the teacher if either of their training periods match the given state.

### Guidance to review

Note that the for consistency with other endpoints and to make the implementation more straight forward we can accept multiple contract periods when filtering. We will need to reflect this in the API documentation but it should be a backwards compatible change.

### DIFF
--- a/app/services/api/teachers/query.rb
+++ b/app/services/api/teachers/query.rb
@@ -126,25 +126,27 @@ module API::Teachers
       # The latest_ect_training_period is joined as training_period and
       # latest_mentor_training_period as latest_mentor_training_periods_metadata_teachers_lead_providers
       # (using the alias names Rails creates for the join tables).
-      case training_status.to_sym
-      when :withdrawn
-        @scope = scope.where(
-          "training_periods.withdrawn_at IS NOT NULL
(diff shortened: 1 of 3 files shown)


cwrw:
Thanks @ethax-ross this looks great, and tested with Andy and tests are passing after multiple runs, there is only one issue with the training status filter but after that this can go on, nicely done!

PR #1557, opened 2025-10-15
Add swagger docs for GET participants API

cpjmcquillan:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/2366

This adds swagger documentation for the GET participants API endpoints (`index` and `show` actions).

The documentation has been autogenerated based on the defined schema specs.

Documentation follows ECF as much as possible, but references to "ECF" have been removed.

### Guidance to review

- [ ] Check swagger docs
- [ ] Check schema content (e.g descriptions)

### DIFF
--- a/spec/requests/api/docs/v3/participants_spec.rb
+++ b/spec/requests/api/docs/v3/participants_spec.rb
@@ -1,0 +1,54 @@
+require "swagger_helper"
+
+describe "Participants endpoint", :with_metadata, openapi_spec: "v3/swagger.yaml" do
+  include_context "with authorization for api doc request"
+
+  let(:lead_provider_delivery_partnership) do
+    FactoryBot.create(
+      :lead_provider_delivery_partnership,
+      active_lead_provider:
+    )
+  end
+  let(:school_partnership) do
+    FactoryBot.create(:school_partnership, lead_provider_delivery_partnership:)
+  end
+  let(:ect_at_school_period) do
+    FactoryBot.create(
+      :ect_at_school_period,
+      :ongoing,
+      school: school_partnership.school
+    )
+  end
+  let!(:training_period) do
+    FactoryBot.create(
+      :training_period,
+      :for_ect,
+      :ongoing,
+      :with_school_partnership,
+      school_partnership:,
+      ect_at_school_period:,
+      started_on: ect_at_school_period.started_on + 1.week
+    )
+  end
+  let(:teacher) { ect_at_school_period.teacher }
+
+  let(:resource) { teacher }
+
+  it_behaves_like "an API index endpoint documentation",
+                  {
+                    url: "/api/v3/participants",
+                    tag: "Participants",
+                    resource_description: "participants",
+                    response_schema_ref: "#/components/schemas/ParticipantsResponse",
+                    filter_schema_ref: "#/components/schemas/ParticipantsFilter",
+                    sorting_schema_ref: "#/components/schemas/SortingTimestamps",
+                  }
+
+  it_behaves_like "an API show endpoint documentation",
+                  {
+                    url: "/api/v3/participants/{id}",
+                    tag: "Participants",
+                    resource_description: "participant",
+                    response_schema_ref: "#/components/schemas/ParticipantResponse",
+                  }
+end
--- a/spec/swagger_schemas/models/participant.rb
+++ b/spec/swagger_schemas/models/participant.rb
@@ -1,0 +1,54 @@
+PARTICIPANT = {
+  description: "A participant.",
+  type: :object,
+  required: %i[id type attributes],
+  properties: {
+    id: {
+      "$ref": "#/components/schemas/IDAttribute",
+    },
+    type: {
+      description: "The data type.",
+      type: :string,
+      example: "participant",
+      enum: %w[
+        participant
+      ],
+    },
+    attributes: {
+      properties: {
+        full_name: {
+          description: "The full name of the participant",
+          type: :string,
+          nullable: false,
+          example: "John Doe",
+        },
+        teacher_reference_number: {
+          description: "The Teacher Reference Number (TRN) for this participant",
+          type: :string,
+          nullable: true,
+          example: "1234567",
+        },
+        enrolments: {
+          type: :array,
+          nullable: false,
+          items: {
+            "$ref": "#/components/schemas/ParticipantEnrolment"
+          }
+        },
+        participant_id_changes: {
+          type: :array,
+          nullable: false,
+          items: {
+            "$ref": "#/components/schemas/ParticipantIDChange"
+          }
+        },
+        updated_at: {
+          description: "The date and time the participant was last updated.",
+          type: :string,
+          format: :"date-time",
+          example: "2021-05-31T02:22:32.000Z",
+        },
+      }
+    }
+  }
+}.freeze
--- a/spec/swagger_schemas/models/participants/enrolment.rb
+++ b/spec/swagger_schemas/models/participants/enrolment.rb
@@ -1,0 +1,173 @@
+PARTICIPANT_ENROLMENT = {
+  description: "The details of a participant enrolment",
+  type: :object,
+  required: %i[
+    training_record_id
+    email
+    school_urn
+    participant_type
+    cohort
+    training_status
+    participant_status
+    eligible_for_funding
+    pupil_premium_uplift
+    sparsity_uplift
+    schedule_identifier
+    delivery_partner_id
+    created_at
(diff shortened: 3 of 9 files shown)


spec/swagger_schemas/models/participants/enrolment.rb
```diff
@@ -0,0 +1,158 @@
...
+    delivery_partner_id: {
+      description: "Unique ID of the delivery partner associated with the participant",
+      type: :string,
+      format: :uuid,
+      example: "42a9ef2f-9059-400a-92ff-4830a629d0c5"
+    },
+    withdrawal: {
+      description: "todo",
+      type: :string
+    },
```
cpjmcquillan: Do we want to define another model for this, or can we define it inline?

```ruby
withdrawal: {
  description: "The details of a participant withdrawal",
  type: :object,
  required: %I[reason date],
  properties: {
    reason: {
      description: "The reason a participant was withdrawn",
      type: :string,
      enum: %w[
        left-teaching-profession
        moved-school
        mentor-no-longer-being-mentor
        switched-to-school-led
        other
      ],
      example: "moved-school"
    },
    date: {
      description: "The date and time the participant was withdrawn",
      type: :string,
      format: "date-time",
      example: "2021-05-31T02:22:32.000Z"
    }
  }
}
```

spec/swagger_schemas/models/participants/enrolment.rb
```diff
@@ -0,0 +1,158 @@
...
+        ecf-standard-april
+        ecf-reduced-september
+        ecf-reduced-january
+        ecf-reduced-april
+        ecf-extended-september
+        ecf-extended-january
+        ecf-extended-april
+        ecf-replacement-september
+        ecf-replacement-january
+        ecf-replacement-april
```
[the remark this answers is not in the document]
leandroalemao: Hi @cpjmcquillan Think we can use `enum: Schedule.identifiers.keys` instead like we did in [npq-reg](https://github.com/DFE-Digital/npq-registration/blob/main/spec/swagger_schemas/models/participant.rb#L116)? same for other enums or constants.

leandroalemao (commented):
PR still in draft but well done @cpjmcquillan! Left a few comments anyway tkss 👍

spec/requests/api/docs/v3/participants_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+      school_partnership:,
+      ect_at_school_period:,
+      started_on: ect_at_school_period.started_on + 1.week
+    )
+  end
+  let(:teacher) { ect_at_school_period.teacher }
+
+  let(:resource) { teacher }
+
+  before { Metadata::Handlers::Teacher.new(teacher).refresh_metadata! }
```
cpjmcquillan: There is a lot of setup here!

Things we need to ensure (otherwise tests fail because we can't retrieve a Teacher in `API::Teachers::Query` https://github.com/DFE-Digital/register-early-career-teachers-public/blob/bdef1b32de1e033734bf367fb78c8bb1e6292fd3/app/services/api/teachers/query.rb#L31 or because of dates chaos):

- Training period lead provider matches the authenticated lead provider
- Teacher lead provider metadata matches the authenticated lead provider
- Teacher school has a partnership with the delivery partner (via the active lead provider)
- Training period is enveloped by the teacher's ECT at school period

Are we comfortable with this? I like that it's explicit, and the complexity of the setup reflects the complexity of the relationships. But it is pretty verbose. I could imagine it might get frustrating if we're doing it lots.

Also, maybe there is something I have missed and this isn't needed!

mooktakim: Unfortunately, the data model is a bit annoying to set up. Maybe we can create some basic traits that build all the related, eg for Teacher we add `:with_ect, lead_provider: lead_provider1`, so we can pass in LP.

spec/swagger_schemas/models/participants/enrolment.rb
```diff
@@ -0,0 +1,164 @@
...
+    },
+    cohort_changed_after_payments_frozen: {
+      description: "Identify participants that migrated to a new cohort as payments were frozen on their original cohort",
+      type: :boolean,
+      example: true
+    },
+    mentor_ineligible_for_funding_reason: {
+      description: "The reason why funding for a mentor's training has ended",
+      type: :string,
+      enum: Teacher.mentor_became_ineligible_for_funding_reasons.keys + %w[mentor_not_completed],
```
cpjmcquillan: Are we supporting the `mentor_not_completed` reason?

I can't find any references to it in this codebase 🤔

cwrw: the only ones we support are:
```
  - completed_declaration_received
  - completed_during_early_roll_out
  - started_not_completed
```
not sure where `mentor_not_completed` came from?

spec/swagger_schemas/models/participants/enrolment.rb
```diff
@@ -0,0 +1,164 @@
...
+        }
+      }
+    },
+    created_at: {
+      description: "The date and time the participant was created",
+      type: :string,
+      format: "date-time",
+      example: "2023-01-01T00:00:00Z"
+    },
+    induction_end_date: {
```
cwrw: we have a new field in RECT (not in ECF1) called `overall_induction_start_date`, can we add it please?

spec/swagger_schemas/models/participant.rb
```diff
@@ -0,0 +1,54 @@
...
+          example: "123456789",
+        },
+        enrolments: {
+          type: :array,
+          nullable: false,
+          items: {
+            "$ref": "#/components/schemas/ParticipantEnrolment"
+          }
+        },
+        teacher_id_changes: {
```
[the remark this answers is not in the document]
cpjmcquillan: Good catch!

spec/swagger_schemas/models/participants/enrolment.rb
```diff
@@ -0,0 +1,164 @@
...
+      enum: Schedule.identifiers.keys,
+      example: "ecf-standard-january"
+    },
+    delivery_partner_id: {
+      description: "Unique ID of the delivery partner associated with the participant",
+      type: :string,
+      format: :uuid,
+      example: "42a9ef2f-9059-400a-92ff-4830a629d0c5"
+    },
+    withdrawal: {
```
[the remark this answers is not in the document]
ethax-ross: So I think what Abeer means is that in the ECF participants example request in the current documentation we show participants in the 'active' state, which means `withdrawal` and `deferral` are `null`:

Screenshot 2025-10-21 at 08 06 35

But as you currently have it (and what rswag does by default) is renders the whole schema as the example, so you get:

Screenshot 2025-10-21 at 08 07 03

Overriding the response examples in rswag is a pain, in NPQ reg we added a [helper for this](https://github.com/DFE-Digital/npq-registration/blob/cb6d7775ced9c885289e4ea5cdc634dbe492132b/spec/support/helpers/swagger_example_parser.rb#L20) that then let us [set specific overrides](https://github.com/DFE-Digital/npq-registration/blob/cb6d7775ced9c885289e4ea5cdc634dbe492132b/spec/requests/api/docs/v3/participants_spec.rb#L66-L70) from the default example.

I think we should probably do this as a follow-up to keep the changes here to a minimum; I'll add a ticket for it.

spec/swagger_schemas/models/participants/enrolment.rb
```diff
@@ -0,0 +1,173 @@
...
+      enum: %w[active deferred withdrawn]
+    },
+    participant_status: {
+      description: "Replaces the old status field. Indicates if a SIT has advised DfE of a transfer or a withdrawal of the participant from the school",
+      type: :string,
+      enum: %w[active joining leaving left],
+      example: "active"
+    },
+    eligible_for_funding: {
+      description: "Indicates whether this participant is eligible to receive DfE funded induction",
```
claire-hughez:
```suggestion
      description: "Indicates whether this participant has become eligible to receive DfE funded induction",
```

spec/swagger_schemas/models/participants/enrolment.rb
```diff
@@ -0,0 +1,173 @@
...
+      nullable: true
+    },
+    created_at: {
+      description: "The date and time the participant was created",
+      type: :string,
+      format: "date-time",
+      example: "2023-01-01T00:00:00Z"
+    },
+    induction_end_date: {
+      description: "The induction period end date",
```
claire-hughez:
```suggestion
      description: "The date when an ECT has passed or failed their induction.",
```

PR #1560, opened 2025-10-15
[2553] Participants seed data review apps

mooktakim:
### Context

Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/2553

* 5x participants for every lead provider and contract period
* Each one set to different variables.
* Schedule is commented out for now as currently rebased from `2365-get-participants-index-show`, which needs to be rebased from `main`. Will uncomment schedule.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -34,4 +34,4 @@ DeclarativeUpdates.skip(:metadata) do
 end
 
 # Refresh all metadata
-Metadata::Manager.refresh_all_metadata!(async: true, track_changes: false)
+Metadata::Manager.refresh_all_metadata!(async: false, track_changes: false)


ethax-ross (commented):
Looks good, can we move this out of `teacher_histories` though, maybe into `api_teachers_with_histories` if it makes sense to do them all in the same seed file?

I think if we are going to need to have school partnerships for more combinations of LPs/DPs we should move that logic into the relevant seed files as well, so create them in `lead_provider_delivery_partnerships` and `school_partnerships` seed files to avoid it being confusing to other people; then just do a random lookup in the `api_teachers_with_histories`

db/seeds/school_partnerships.rb
```diff
@@ -61,3 +61,21 @@ def find_lead_provider_delivery_partnership(lead_provider:, delivery_partner:, c
   { school: brookfield_school, lead_provider_delivery_partnership: teach_first__grain__2022 },
   { school: brookfield_school, lead_provider_delivery_partnership: capita__delivery_partner__2022 },
 ].each { |kwargs| FactoryBot.create(:school_partnership, **kwargs).tap { |sp| describe_school_partnership(sp) } }
+
+# Create 3 school partnerships for each lead provider and contract period
+ActiveLeadProvider.find_each do |active_lead_provider|
+  3.times do
+    delivery_partner = DeliveryPartner.order("RANDOM()").first
+    lead_provider_delivery_partnership = LeadProviderDeliveryPartnership.where(active_lead_provider:, delivery_partner:).first
```
ethax-ross:
```
LeadProviderDeliveryPartnership.find_by(active_lead_provider:, delivery_partner:)
```

db/seeds.rb
```diff
@@ -25,8 +25,13 @@ def print_seed_info(text, indent: 0, colour: nil, blank_lines_before: 0)
   priority_seeds.index(filename.chomp(".rb")) || Float::INFINITY
 end
 
-seed_files.each do |seed_file|
-  puts "\r\n🪴 planting #{seed_file}"
+DeclarativeUpdates.skip do
```
ethax-ross: I think we'll just want to `skip(:metadata)` here and include `touch` updates (as we have no way to refresh these outside of the callbacks as it stands)

db/seeds.rb
```diff
@@ -25,8 +25,13 @@ def print_seed_info(text, indent: 0, colour: nil, blank_lines_before: 0)
...
+  seed_files.each do |seed_file|
+    puts "\r\n🪴 planting #{seed_file}"
 
-  ApplicationRecord.transaction { load(seed_file) }
+    ApplicationRecord.transaction { load(seed_file) }
+  end
 end
+
+# Refresh all metadata
+Metadata::Manager.refresh_all_metadata!(async: true, track_changes: false)
```
ethax-ross: Out of interest; what was it in the metadata that was causing a hang? was it just volume and if you wait long enough it completes or did it hit some kind of death loop we should investigate?

My only concern is having it `async: true` will trip people up; I know I don't run jobs in development usually. It looks like it takes a while for me with `async: false` though when I tested; maybe we stick with `async: true` for now and switch it back as part of the handler optimisations?

PR #1562, opened 2025-10-15
[2367] Add participants to sandbox seed data task

leandroalemao:
### Context

Ticket: [2367](https://github.com/DFE-Digital/register-ects-project-board/issues/2367)

Lead providers need to be able to test the APIs in sandbox before using them in a live environment.

- Add participants to `sandbox_seed_data.rake`
- Created a file called `api_teachers` in order to create teachers in specific scenarios for the API (so we can run in separate and not mix up with existing teachers / teacher histories that it was already run)
- Make sure we disable metadata refreshes during the run, running it in the end to speed up the process
- Added a `TRNGenerator`, so we make sure we don't clash with existing TRNs (as the factories use sequences for TRN, I had issues with it)
- Mix of participants in:
  - Ensure there are participants in every cohort
  - set pupil_premium_uplift and sparsity_uplift only on ECTs as true/false
  - add different schedule_identifier values
  - ensure dates are different and make sense
  - ensure when setting cohort_changed_after_payments_frozen to true that participants come from cohort 2021 or 2022 and are in 2024
  - Participants with teacher id changes
  - Participants with two enrolments

### Guidance to review

I'll run it in review app for testing. After reviewed, we can run the same in sandbox.

```
Rails.logger.silence do
    DeclarativeUpdates.skip do
      SandboxSeedData::APITeachersWithHistories.new.plant
    end
  end
end
```

### DIFF
--- a/app/services/sandbox_seed_data/api_teachers_with_histories.rb
+++ b/app/services/sandbox_seed_data/api_teachers_with_histories.rb
@@ -1,0 +1,383 @@
+module SandboxSeedData
+  class APITeachersWithHistories < Base
+    NUMBER_OF_RECORDS = 500
+
+    TRAINING_STATUS_COLOURS = {
+      active: :green,
+      withdrawn: :red,
+      deferred: :yellow,
+    }.freeze
+
+    def plant
+      return unless plantable?
+
+      log_plant_info("api teachers with histories")
+
+      groups_of_active_lead_providers.each_value do |grouped_active_lead_providers|
+        grouped_active_lead_providers.each do |active_lead_provider|
+          (NUMBER_OF_RECORDS / grouped_active_lead_providers.size).times do
+            create_api_teachers_records_for(active_lead_provider)
+          end
+        end
+      end
+    end
+
+  private
+
+    def create_teacher
+      teacher = FactoryBot.create(
+        :teacher,
+        :with_realistic_name,
+        trn: Helpers::TRNGenerator.next
+      ).tap do |t|
+        random_date = rand(1..100).days.ago
+        t.update!(created_at: random_date, updated_at: random_date)
+      end
+
+      log_seed_info(::Teachers::Name.new(teacher).full_name, indent: 2)
+      teacher
+    end
+
+    def groups_of_active_lead_providers
+      ActiveLeadProvider.all.group_by(&:lead_provider_id)
+    end
+
+    def create_api_teachers_records_for(active_lead_provider)
+      school_partnership = find_school_partnership(active_lead_provider)
+      return if school_partnership.blank?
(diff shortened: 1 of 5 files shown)


lib/tasks/sandbox_seed_data.rake
```diff
@@ -11,11 +11,16 @@ namespace :sandbox_seed_data do
       SandboxSeedData::SchoolPartnerships,
       SandboxSeedData::Teachers,
       SandboxSeedData::TeacherHistories,
+      SandboxSeedData::APITeachers,
```
ethax-ross: `APITeachersWithHistories`?

ethax-ross (commented):
Looks good just a few questions/suggestions

app/services/sandbox_seed_data/api_teachers.rb
```diff
@@ -0,0 +1,353 @@
...
+      school_reported_appropriate_body = random_appropriate_body
+
+      FactoryBot.create(
+        :ect_at_school_period,
+        *traits.compact,
+        teacher:,
+        school:,
+        email:,
+        school_reported_appropriate_body:,
+        **school_period
```
cwrw: we need to set the `created_at` to the same time we expect this participant to have been created (maybe started on) as that feeds the created at timestamp in the response, and that can't be after the updated at. Similar in mentors as well

app/services/sandbox_seed_data/api_teachers_with_histories.rb
```diff
@@ -0,0 +1,383 @@
...
+
+    def assign_ect_to_mentor(teacher:, school:, mentor_at_school_period_record:)
+      return unless Faker::Boolean.boolean(true_ratio: 0.20)
+
+      # Assign an ECT to the mentor for the same period excluding ECTs who are already mentees
+      ect_at_school_period = school.ect_at_school_periods.where.not(teacher_id: teacher.id).where.not(id: MentorshipPeriod.distinct.pluck(:ect_at_school_period_id)).started_on_or_after(mentor_at_school_period_record.started_on).finished_before(mentor_at_school_period_record.finished_on).last
+
+      return if ect_at_school_period.blank?
+
+      FactoryBot.create(:mentorship_period,
```
ethax-ross: I think you're missing a test for the `MentorshipPeriod` path?

ethax-ross (approved):
Looks good, thanks Leandro! Can we make sure to run this past @andycrawshaw ahead of running it in sandbox, and we need to make sure we only run the new seed file on sandbox and not all of them (I tend to run them in a transaction as well incase they blow up halfway through).

mooktakim (approved):

PR #1565, opened 2025-10-16
Access denied page

### LINKED ISSUE #1432: Improve UX around Sessions::Users::Builder::UnknownOrganisation error
Similar to https://github.com/DFE-Digital/register-early-career-teachers-public/issues/1431 

We often see this error when a check is made against an authenticating user to find their organisation 
- ABs using the UUID (currently 82 on production)
- Schools using the URN (currently 0 on production)

It is also possible for an authenticating user to mistakenly attempt login after choosing the wrong organisation if the work across multiple establishments.

We could surface the current organisation name and:

- If it is a school, and RECT has not gone live, give a useful message 
- If it is an AB, say that this AB needs to contact support to gain access (to have their UUID added to the DB).

Sentry Issue: [REGISTER-EARLY-CAREER-TEACHERS-8F](https://dfe-teacher-services.sentry.io/issues/6849961547/?referrer=github_integration)

```
Sessions::Users::Builder::UnknownOrganisation: #<OmniAuth::AuthHash DistrictAdministrativeCode=nil DistrictAdministrativeName=nil DistrictAdministrative_code=nil GIASProviderType=nil GIASStatus=nil GIASStatusName=nil IsOnAPAR=nil LegalName=nil MasterProviderStatusCode=nil MasterProviderStatusName=nil OpenedOn=nil PIMSProviderType=nil PIMSProviderTypeCode=nil PIMSStatusName=nil ProviderTypeCode=nil SourceSystem=nil address=nil category=#<OmniAuth::AuthHash id="002" name="Local Authority"> closedOn=nil companyRegistrationNumber=nil establishmentNumber="001" id="7D48D8FF-A322-4F22-8A79-C0425AC69749" legacyId="1027142" name="Department for Education" pimsStatus=nil providerTypeName=nil status=#<OmniAuth::AuthHash id=1 name="Open" tagColor="green"> statutoryHighAge=nil statutoryLowAge=nil telephone=nil uid=nil ukprn=nil upin=nil urn=nil> (Sessions::Users::Builder::UnknownOrganisation)
  app/services/sessions/users/builder.rb:42:in `session_user'
    raise(UnknownOrganisation, organisation)
  app/controllers/sessions_controller.rb:36:in `session_user'
    delegate :session_user, :id_token, to: :user_builder
  app/controllers/sessions_controller.rb:11:in `create'
    session_manager.begin_session!(session_user, id_token:)
  app/middlewares/api/request_middleware.rb:25:in `_call'
    status, @headers, @response = @app.call(env)
  app/middlewares/api/request_middleware.rb:20:in `call'
    dup._call(env)
...
(90 additional frame(s) were not displayed)
```

peterdavidhamilton:
### Context

When authenticating via DfE Sign In we have a few criteria.

1. We need the user to have a role to determine which parts of the service they can access (RIAB vs. RECT vs. Admin console)
2. And we need the unique identifier of their organisation to be in the database. (URN for RECT school users and a UUID for ABs)

When either of these criteria is not met, we do not fail gracefully, rather one of three errors is raised which ultimately provides no useful information to the user.

1. `Sessions::Users::Builder::UnknownOrganisation`
2. `Sessions::Users::Builder::UnknownPersonaType`
3. `Sessions::Users::Builder::UnknownProvider`

Rescue from the errors and redirect to a page `/access-denied` that provides guidance.

The current proposal is generic but this spike includes surfacing the authenticating user's organisation name.

The value here is if a user (most likely a RIAB user) works across multiple orgs, but only one has RIAB access, it helps them identify that they need to change their active organisation.

Potentially, a user authenticated against DSI but who does not have access will need to:

- Use DfE Sign In to request access to the service, prompting their org's approver to grant them a role. Either `SchoolUser` for RECT or `AppropriateBodyUser` for RIAB.
- Or, if they have already completed this step but work across multiple orgs and have selected the invalid one, will need to use DfE Sign In to log out and back in again.

We can link to the correct DfE Sign In environment's "My Services" page so that they can do both, but this is deployment specific and requires additional environment variables.

Closes #1432 

### Guidance to review

Unfortunately DfE Sign In is not able to function with wildcard domains so cannot be used in a review app. A product review can be demonstrated by a developer locally against the DSI `test` environment using the available "internal" org to demonstrate an authenticated user who is not recognised by the service.

Caveat: ensure the authenticating user hasn't been added to the `users` table. This is not likely as it would require a dev to do it manually and reseed will ensure this is not the case.

http://localhost:3000/access-denied

localhost_3000_access-denied


RachF-content:
@peterdavidhamilton as per my Teams message, I've updated the design:
image
Lucid link: https://lucid.app/lucidspark/01200372-0c29-400c-a15b-69db42505ba8/edit?view_items=UkxZV4.-HnN-&page=0_0

Please can you confirm that this works and [[org name]] will always be available so it works for all scenarios and doesn't create an error on the page. Thanks

PR #1569, opened 2025-10-17
[2522] Teacher withdraw service object

mooktakim:
### Context

Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/2522

### Changes proposed in this pull request

* New `API::Teachers::Withdraw` service
* New event `teacher_withdraws_training_period`

PR #1570, opened 2025-10-17
[2525] Add defer teachers/participants service

leandroalemao:
### Context

Ticket: [2525](https://github.com/DFE-Digital/register-ects-project-board/issues/2525)

If a LP can view a participant over the GET Participants endpoint and their training_status is 'active' or 'deferred' they should be able to update it to 'Deferred'.

- Add defer teachers/participants service


app/services/api/teachers/action.rb
```diff
@@ -0,0 +1,63 @@
...
+    validates :course_identifier,
+              inclusion: { in: %w[ecf-mentor ecf-induction],
+                           message: "The entered '#/course_identifier' is not recognised for the given participant. Check details and try again." },
+              presence: { message: "Enter a '#/course_identifier' value for this participant." },
+              allow_blank: false
+    validate :teacher_exists
+    validate :training_period_exists
+
+    def teacher
+      @teacher ||= Query.new(lead_provider_id:).teacher_by_api_id(participant_id)
```
ethax-ross: The controller will be doing this lookup/safety check anyway so I think we can keep it to a simple `Teacher.find_by` as in the `Withdraw` service https://github.com/DFE-Digital/register-early-career-teachers-public/pull/1569/files#diff-3066a5a6fe15f8cfbb6c56873b9f49a1d01e990a38564cde7f4e14dc7576fb93R51

app/services/api/teachers/action.rb
```diff
@@ -0,0 +1,63 @@
...
+    rescue ActiveRecord::RecordNotFound, ArgumentError
+      nil
+    end
+
+    def latest_ect_training_period
+      TrainingPeriod.ect_training_periods_latest_first(teacher:, lead_provider: lead_provider_id).first
```
ethax-ross: Can we keep this consistent with the other service as well, I don't mind if we use `metadata` or this lookup -- maybe its worth @mooktakim branching off this PR as you've added a base class?

spec/support/shared_examples/teacher_action_support.rb
```diff
@@ -0,0 +1,88 @@
...
+      ect_at_school_period:,
+      started_on: ect_at_school_period.started_on + 1.week,
+      finished_on: nil
+    )
+  end
+  let(:teacher) { ect_at_school_period.teacher }
+  let(:participant_id) { teacher.api_id }
+  let(:course_identifier) { "ecf-induction" }
+
+  before { Metadata::Manager.refresh_all_metadata!(async: false) }
```
ethax-ross: Can we use `with_metadata` and avoid explicit refreshes

app/services/api/teachers/action.rb
```diff
@@ -0,0 +1,63 @@
...
+    end
+
+    def latest_mentor_training_period
+      TrainingPeriod.mentor_training_periods_latest_first(teacher:, lead_provider: lead_provider_id).first
+    end
+
+    def training_period
+      @training_period ||= if course_identifier == "ecf-induction"
```
ethax-ross: We probably need an error message/validation check here in case someone passes in `ecf-induction` for a mentor-only participant (as the training period will be `nil`). Might need to talk to Mark about an error message unless there's an equivalent one in ECF, maybe the course identifier error?

app/services/api/teachers/action.rb
```diff
@@ -0,0 +1,63 @@
+module API::Teachers
+  class Action
```
mooktakim: I don't think we should break this out into `Defer < Action`, `Defer` service class on its own would be better. It'll be simpler without the abstraction.

ethax-ross: I don't mind either way for now; as long as the logic in `Action` will apply to all participant actions I think it makes sense to extract it, but if we want to pull it out into the parent class until we have all services in place in order to align `Defer` and `Withdraw` more easily I'm fine with that

app/services/api/teachers/action.rb
```diff
@@ -0,0 +1,63 @@
...
+    include ActiveModel::Model
+    include ActiveModel::Attributes
+
+    attribute :author
```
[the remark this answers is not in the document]
ethax-ross: Agreed, I've seen this pattern in other services but since these are scoped to `API` I think we can encapsulate the author in the service (we can always pull it out later if we find we call this service from the admin interface or elsewhere)

PR #1576, opened 2025-10-20
Induction record export button

tonyheadford:
### Context

We want to provide a CSV export of all the induction records/associated data without any PII so that it can be used locally by developers and BAs for analysis.

Adds a service that generates the CSV
Adds a Job to perform the generation overnight as it takes a couple of minutes
Adds a controller to handled the download
Updated the main migration view to provide a download button

### DIFF
--- a/config/recurring.yml
+++ b/config/recurring.yml
@@ -60,6 +60,6 @@ review:
 migration:
   build_induction_record_export_csv:
     class: BuildInductionRecordExportJob
-    queue: default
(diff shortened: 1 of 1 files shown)


claire-hughez:
Fixing a thing that crept in yesterday. Needs some specs. 

💅

gen-sole:
This needs a review please @ltello - in the migration env if you want to see it in action

PR #1577, opened 2025-10-20
Standard Ruby

peterdavidhamilton:
### Context

DfE favours Rubocop for linting. See #1318 for some rubocop options, but we could silence the debate by adopting [StandardRB](https://github.com/standardrb/standard).

Replace Rubocop with opinionated alternative


peteryates:
I dunno about this - just about all Ruby projects across gov have some things in common, like `{ spaces_around: 'hashes' }` and keeping `private` and `protected` in line with `class` and `end.

The majority of the changes here seem to be either quotes, indentation or spacing and I don't really see the need.

I think we're best sticking with Rubocop and just disabling the incorrect rules, and either fixing or doing away with the rubocop todo list.

Are there any areas of the codebase we think are messy or unreadable as a result of single vs double quotes or indenting?

PR #1585, opened 2025-10-21
Record induction outcome admin notes 

peterdavidhamilton:
### Context

Giving context and linking to support tickets when an Admin makes changes.

- Fix event recording methods to use the admin version. [When the old services were repurposed from the AB side to the Admin side this was omitted.](https://github.com.mcas.ms/DFE-Digital/register-early-career-teachers-public/pull/1503/files#diff-e4a2f3aa62ad7cb1c731652c6d942912de27d4db3a96b78b197a78dfd1f07859)
- When recording passes and fails as an Admin provide two new text fields `zendesk_ticket_id` and `note` which are saved in the event history.
- Complete unfinished work converting forms for recording induction outcomes to use a component
- Broad refactoring to support making close induction forms extensible and controllers leaner.

In summary:

- Services are initialised with teacher, appropriate body and current user, and then `call` on the form submission `params`
- PendingInductionSubmissions are moved into the service 
- The service is used in the form and determines the outcome value 
- Services and concerns define the params the controller will expect
- Forms and fields are built in view components that have different user modes
- View components are initialised with the service giving access to all objects at once

Pros:

- Simplifies testing and reduces controller logic
- Pattern establishes a way to extend journeys shared by ABs and Admins independently
- The use of the ephemeral submission object becomes a service internal

Cons:

- Bulk uploads generate duplicate PendingInductionSubmissions 

Follow up:

- Apply convention to other RIAB/Admin induction journeys

### Guidance to review

Deliverable:

- Log in as an admin persona
- Pass and fail some ECTs adding context notes or support ticket
- Visit ECTs' timelines to confirm the actions were recorded

Extra checks:

- Log in as an AB persona
- Pass and fail some ECTs
- Release an ECT
- Pass and fail ECTs via a bulk upload
- Log in as an admin persona
- Visit ECTs' timelines to confirm the actions were recorded

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -68,7 +68,6 @@ group :development do
   gem 'binding_of_caller'
   gem "pg_query"
   gem 'prosopite'
-  gem 'yard'
 end
 
 group :test do
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -808,7 +808,6 @@ GEM
       zeitwerk (>= 2.7)
     xpath (3.2.0)
       nokogiri (~> 1.8)
-    yard (0.9.37)
     zeitwerk (2.7.3)
 
 PLATFORMS
@@ -891,7 +890,6 @@ DEPENDENCIES
   webmock
   webrick
   with_advisory_lock
-  yard
 
 RUBY VERSION
    ruby 3.4.7p58


app/components/concerns/user_modes.rb
cpjmcquillan: I wonder if we could lean on `Current.user` or add `Current.user_mode` to determine which components to render?

Gemfile
```diff
@@ -68,6 +68,7 @@ group :development do
   gem 'binding_of_caller'
   gem "pg_query"
   gem 'prosopite'
+  gem 'yard'
```
cpjmcquillan: If we want to bring in this dependency, should we do that in a separate PR?

spec/rails_helper.rb
```diff
@@ -22,6 +22,7 @@
 
   config.include ActiveSupport::Testing::TimeHelpers
   config.include Features::ViewHelpers, type: :feature
+  config.include AuthHelper, type: :request
```
cpjmcquillan: Nice 👌🏼

app/jobs/appropriate_bodies/process_batch/record_fail_job.rb
```diff
@@ -6,21 +6,25 @@ class RecordFailJob < ApplicationJob
...
-          author:
-        ).fail!
+        record_fail = RecordFail.new(teacher:, appropriate_body:, author:)
+
+        record_fail.call(
+          finished_on: pending_induction_submission.finished_on,
+          number_of_terms: pending_induction_submission.number_of_terms
+        )
+
+        pending_induction_submission.update!(delete_at: 24.hours.from_now)
```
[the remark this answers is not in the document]
peterdavidhamilton: I am gonna do something about this

Gemfile
peterdavidhamilton: Will do. I actually thought I'd already added it to codebase; I'll have to see...

PR #1586, opened 2025-10-21
Expose training period schedule in the participants serializer

cpjmcquillan:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/2555

Before, the `schedule_identifier` field in the participants serializer was a fixed value.

This updates the serializer so it retrieves the identifier from the relevant training period's schedule.

This also updates the Teacher history seeds and sandbox seeds to create training periods with schedules.


ethax-ross (approved):
LGTM - let me know if you want to pair on updating the sandbox seed data with schedules

PR #1590, opened 2025-10-21
Visualising induction records

peteryates:
~~WIP - follows on from #1576~~

Fixes #DFE-Digital/register-ects-project-board#2657

## Remaining tasks

* [x] add `plantuml` to the Dockerfile so we can use it in the migration env - this isn't _ideal_ as it's an extra dependency but it's only for the short term and can be removed when we no longer need the migration stuff.

## Preview

Screenshot From 2025-10-21 14-24-21


peteryates:
This is rebased and working with the new changes merged on Friday (#1576)

It needs some tests writing but should be good to go. Tested locally with the anonymised dataset and it produces accurate charts.

image

PR #1598, opened 2025-10-22
Remove unused spec method

avinhurry:
### Context

I added this in #1579 then realised there was an existing method and forgot to remove it.

### Changes proposed

🗑️


peteryates (approved):

spec/features/schools/ects/register/happy_path_spec.rb
```diff
@@ -102,10 +102,6 @@
     and_i_should_see_the_ect_i_registered
   end
 
-  def when_i_click_change_name
-    page.get_by_role('link', name: 'change name').first.click
```
peteryates: Do you think we should have a reusable set of `when_i_click_link(text)` style methods?

PR #1603, opened 2025-10-23
Extract the defer/withdraw base service logics

leandroalemao:
### Context

We need to follow the pattern created for school partnerships, where we have the base service logic extracted to be used for other user cases.. not only for the API.

- Extract the defer/withdraw base service logics following the same pattern used for school partnerships;
- Move shared action concern to the correct folder;


ethax-ross (approved):
Looks great, thanks for picking this up Leandro!

PR #1604, opened 2025-10-23
Optimise metadata handlers

ethax-ross:
> ⚠️ Before this is merged we should rebase to include #1560 and switch `async` to `false`.

### Context

The metadata handlers are becoming slow to run; this is down to executing several queries for each loop performed in a handler (of which there can be many as we join on multiple tables for some metadata). This results in some slow tests (where we create a lot of test objects that have metadata - e.g. `TeachersSerializer`) and it may cause us issues down the line if it keeps getting slower.

- Add missing composit index to Metadata::TeacherLeadProvider

There can be only one entry for each teacher and lead provider; this will allow us to `upsert_all, unique_by:` into the table.

- Optimise handler queries

Instead of running multiple queries as we loop through the products of the join tables we can run a single query to find existing metadata and another single query to update all in one go.

- Pair back with_metadata context for handler tests

We only need to pre-populate metadata for the tests that destroy metadata; it makes the rest of the tests more awkward using `with_metadata` as the handlers themselves create metadata records and we include them in test fixture setup explicitly.

- Optimise DeliveryPartner handler

Instead of querying the contract periods for each lead provider individually we can query them all once up front and index into the in-memory results.

- Optimise Teacher handler

Instead of querying the latest training periods for each lead provider we can instead query them all upfront and index into them in memory.

As these queries are unlikely to be used elsewhere and are optimisations for the handlers I've chosen to encapsulate them there.

- Optimise Schools::TrainingProgramme service

Previously, we instantiated a new service for each training programme calculation/contract period year. This doesn't scale well when we want to know the training programme for all contract period years (as we do for calculating metadata).

Instead, we can create a single instance for a school and query across all contract periods. This allows us to efficiently calculate training periods for all contract periods relative to a school.

- Optimise in_partnership for School handler

Instead of querying the partnership for each contract period year, we can query them all once up front and check if the contract period year is within the list.

### Guidance to review

Best reviewed by-commit.

The `expression_of_interest` isn't optimised here as this has already been picked up as part of [this change](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/1602).

This adds a fair amount of complexity; mainly in the `TrainingProgramme` service, but I think that's unavoidable and worth it in this case.

Between this and the expression of interest update in the other PR, it brings the `TeacherSerializer` spec runtime down from ~65s to ~15s, so ~75% improvement.

### DIFF
--- a/app/services/metadata/handlers/school.rb
+++ b/app/services/metadata/handlers/school.rb
@@ -27,7 +27,7 @@ module Metadata::Handlers
         changes = {
           school_id: school.id,
           contract_period_year:,
-          in_partnership: school.school_partnerships.for_contract_period(contract_period_year).exists?,
+          in_partnership: contract_period_year.in?(school_partnership_contract_period_years),
           induction_programme_choice: school.training_programme_for(contract_period_year)
         }
 
@@ -61,6 +61,13 @@ module Metadata::Handlers
 
     def contract_period_years
       @contract_period_years ||= ContractPeriod.pluck(:year)
+    end
+
+    def school_partnership_contract_period_years
+      @school_partnership_contract_period_years ||= school.school_partnerships
+        .includes(lead_provider_delivery_partnership: :active_lead_provider)
+        .pluck(:contract_period_year)
+        .uniq
     end
 
     def existing_contract_period_metadata


mooktakim (approved):
One comment about the scopes.

Ran this with the new seed with async off, adds only about 15 seconds! Great work!

cpjmcquillan (approved):

PR #1609, opened 2025-10-23
[2556] Implement participant status service

mooktakim:
### Context

Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/2556

* New service `API::TrainingPeriods::TeacherStatus`
* Updated `TeacherSerializer` to use new service for `participant_status`

### DIFF
--- a/spec/serializers/api/teacher_serializer_spec.rb
+++ b/spec/serializers/api/teacher_serializer_spec.rb
@@ -170,7 +170,7 @@ describe API::TeacherSerializer, :with_metadata, type: :serializer do
           end
 
           it "serializes `training_status`" do
-            expect(ect_enrolment["training_status"]).to eq("active")
+            expect(ect_enrolment["training_status"]).to eq(mock_training_status.status)
           end
 
           it "serializes `withdrawal`" do
@@ -182,7 +182,7 @@ describe API::TeacherSerializer, :with_metadata, type: :serializer do
           end
 
           it "serializes `participant_status`" do
-            expect(ect_enrolment["participant_status"]).to eq("active")
+            expect(ect_enrolment["participant_status"]).to eq(mock_teacher_status.status)
           end
 
           it "serializes `eligible_for_funding`" do
@@ -279,7 +279,7 @@ describe API::TeacherSerializer, :with_metadata, type: :serializer do
           end
 
           it "serializes `training_status`" do
-            expect(mentor_enrolment["training_status"]).to eq("active")
+            expect(mentor_enrolment["training_status"]).to eq(mock_training_status.status)
           end
 
           it "serializes `withdrawal`" do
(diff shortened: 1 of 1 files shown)


app/services/api/training_periods/teacher_status.rb
```diff
@@ -0,0 +1,43 @@
+module API::TrainingPeriods
+  class TeacherStatus
+    attr_reader :training_period, :teacher
```
ethax-ross: Is it worth calling the `training_period` out as `latest_training_period`?

app/services/api/training_periods/teacher_status.rb
```diff
@@ -0,0 +1,43 @@
...
+
+    delegate :started_on, :finished_on, :trainee, to: :training_period, private: true
```
ethax-ross: I don't think we're using `trainee` anywhere; maybe we could make the interface simpler by just accepting the `latest_training_period` and accessing `latest_training_period.trainee.teacher` as I think we preload this in the query anyway?

spec/serializers/api/teacher_serializer_spec.rb
```diff
@@ -190,8 +190,32 @@
             end
           end
 
-          it "serializes `participant_status`" do
-            expect(ect_enrolment["participant_status"]).to eq("active")
+          context "serializes `participant_status`" do
```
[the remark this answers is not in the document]
ethax-ross: The `TeacherSerializer` is quite a slow spec as it stands given all the metadata setup; maybe its enough to just mock the service once and expect the serialized response to match as we cover these conditions in the service tests anyway?

spec/services/api/training_periods/teacher_status_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+      let(:training_period) { FactoryBot.build(:training_period, :ongoing) }
+
+      it { is_expected.to eq(:active) }
+      it { expect(service).to be_active }
+      it { expect(service).not_to be_joining }
+      it { expect(service).not_to be_leaving }
+      it { expect(service).not_to be_left }
+    end
+
+    context "when 'teacher.mentor_became_ineligible_for_funding_on' is set" do
```
ethax-ross: Can we maybe call out what this means as I don't think its particularly obvious for this and the `finished_induction_period`, maybe:

`context "when 'teacher.mentor_became_ineligible_for_funding_on' is set (indicating the mentor has completed training)"`

`context "when 'finished_induction_period.finished_on' is set (indicating the ECT has completed training)"`

app/services/api/training_periods/teacher_status.rb
ethax-ross: Ah yes that's right; we used to include `trainee.teacher` but then changed the serializer to pass the `teacher` down as well as `training_period`, fine to keep `teacher` in the parameters, but I think we can still drop the `trainee` delegate here

spec/serializers/api/teacher_serializer_spec.rb
```diff
@@ -166,30 +181,6 @@
             expect(ect_enrolment["deferral"]).to be_nil
           end
 
-          context "when `training_status` is withdrawn" do
```
ethax-ross: Can we keep an assertion for the `training_status` and `participant_status` so we're expecting it to equal the `mock_training_status.status` and `mock_teacher_status.status` in both the ECF and mentor contexts for completeness/clarity?

PR #1610, opened 2025-10-23
[2526] Add defer participants PUT endpoint

leandroalemao:
### Context

Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/2526

* Updated controller `defer` action to use `API::Teachers::Defer` service class
* Updated API request spec for defer

### DIFF
--- a/app/controllers/api/v3/participants_controller.rb
+++ b/app/controllers/api/v3/participants_controller.rb
@@ -19,7 +19,18 @@ module API
       end
 
       def change_schedule = head(:method_not_allowed)
-      def defer = head(:method_not_allowed)
+
+      def defer
+        service = API::Teachers::Defer.new(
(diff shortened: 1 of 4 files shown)


app/controllers/api/v3/participants_controller.rb
```diff
@@ -55,8 +56,15 @@ def participants_params
         params.permit(:api_id, :sort, filter: %i[training_status from_participant_id])
       end
 
-      def withdraw_participant_params
-        params.require(:data).expect({ attributes: %i[course_identifier reason] })
+      def participant_action_params
```
ethax-ross: Happy to discuss/wait for other opinions, but I think this was clearer before; when its like this you have to dig into the service to figure out what it expects

ethax-ross (approved):

PR #1614, opened 2025-10-24
Withdraw participant API documentation

cpjmcquillan:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/2524

This adds swagger documentation for the PUT participant `/withdraw` API endpoint.

The example response has been overriden to return a participant with a training status of "withdrawn", no deferral and a withdrawal.

### Guidance to review

- [ ] Docs appear as expected with appropriate content

### DIFF
--- a/spec/support/shared_contexts/api/docs/api_update_endpoint_documentation_support.rb
+++ b/spec/support/shared_contexts/api/docs/api_update_endpoint_documentation_support.rb
@@ -26,6 +26,8 @@ RSpec.shared_context "an API update endpoint documentation", :exceptions_app do 
       response "200", "The updated #{options[:resource_description]}" do
         schema({ "$ref": options[:response_schema_ref] })
 
+        after { override_response_content!(it) }
+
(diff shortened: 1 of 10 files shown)


spec/support/shared_contexts/api/docs/api_update_endpoint_documentation_support.rb
```diff
@@ -26,6 +26,21 @@
       response "200", "The updated #{options[:resource_description]}" do
         schema({ "$ref": options[:response_schema_ref] })
 
+        after do |example|
+          if defined?(response_example)
```
ethax-ross: I wonder if we should stick this in a helper as we're going to pull it into most/all the shared examples at some point I reckon

ethax-ross (commented):
Looks great, just the `type` needs updating and one thought for DRYing up the example override with a helper but its fine as-is.

leandroalemao (approved):

PR #1641, opened 2025-10-29
[2225] Change mentor lead provider

thenapking:
### Context
Adds a change button next to the mentor's lead provider

Add `MentorAtSchoolPeriods::ChangeLeadProvider` Service
Abstracts the common methods from `ECTAtSchoolPeriods::SwitchLeadProvider` to a new Module (*Please help me name this!*)
Adds a Wizard and controller

### Guidance to review
The card mentions creating more seed data for testing. I think I just need to create a mentor that only has an EOI on their training period, not a school partnership, but I could do with a bit more guidance as I don't quite know what will be required.

### DIFF
--- a/app/views/schools/mentors/change_lead_provider_wizard/confirmation.html.erb
+++ b/app/views/schools/mentors/change_lead_provider_wizard/confirmation.html.erb
@@ -16,7 +16,7 @@
   What happens next?
 </h2>
 
-<p class="govuk-body">We’ll let <%= @wizard.current_step.new_lead_provider_name %> know your school wants to work with them. They'll contact us to confirm the change and tell us the delivery partner they'll be working with.</p>
+<p class="govuk-body">We’ll let <%= @wizard.current_step.new_lead_provider_name %> know your school wants to work with them. They’ll contact us to confirm the change and tell us the delivery partner they'll be working with.</p>
 <p class="govuk-body">We’ll also let the old lead provider know that they’ll no longer be training <%= @wizard.teacher_full_name %>.</p>
 
 <p class="govuk-body">


app/services/mentor_at_school_periods/change_lead_provider.rb
```diff
@@ -0,0 +1,112 @@
...
+    include TrainingPeriodSources
+
+    def initialize(mentor_at_school_period:, lead_provider:, author:)
+      @mentor_at_school_period = mentor_at_school_period
+      @lead_provider = lead_provider
+      @author = author
+    end
+
+    def call
+      raise LeadProviderNotChangedError unless lead_provider_changed?
```
avinhurry: > I am not sure if I am dealing correctly with the situation where the Lead Provider hasn't changed. Currently the service raises an error, which is caught by the wizard. The wizard has validation for lead provider not changing, but I did wonder whether that was sufficient to stop a rogue request going through. Unlike the service to change LP for an ECT, afaik it is not possible for the mentor to have school led training, so I do not raise an error for that situation

I think this is fine, you're right, mentors are always provider-led. The unchanged LP check already covers the only invalid state that might hit. It's nice defensive programming IMO.

app/services/mentor_at_school_periods/change_lead_provider.rb
```diff
@@ -0,0 +1,112 @@
...
+
+      ActiveRecord::Base.transaction do
+        if date_of_transition.future? || training_period_not_confirmed
+          training_period.destroy!
```
avinhurry: > The ticket talks about whether the school partnership is confirmed. I understood this to mean that there was only an EOI if the partnership was not confirmed. The ACs also mention that if the old training period was not confirmed it should be removed. I took this to mean that the actual record was deleted. I am not sure that is correct.

Yeah I think your understanding is spot on.

avinhurry: > I spent a bit of time working out how to deal with the periods of time. I am assuming from the text given to the school user that this change is to be effective on the day they make the change, and there isn't any way for the school user to enter an effective date. That being the case, I thought it was possible that a school user might make two of these changes on the same day, especially if they made a mistake. Hopefully I have implemented this correctly. I have taken my cues for this from the ECTAtSchoolPeriods::SwitchLeadProvider service.

Yes, the changes are effective on the day and there isn't any way for the school to enter a specific date. The users should be able to make changes in the same day, your implementation looks correct to me.

`date_of_transition.future?` ensures that we drop a not yet started period before creating the new one I believe rather than trying to finish something that hasn't begun yet.

avinhurry (commented):
> The card mentions creating more seed data for testing. I think I just need to create a mentor that only has an EOI on their training period, not a school partnership, but I could do with a bit more guidance as I don't quite know what will be required.

If you can add any seed data that makes product review easier in the review app (e.g. a mentor with only an EOI and no school partnership) if there isn't one already, that would be super useful. If it feels too niche or time consuming, just screen share with whoever is doing the product review instead.

`SwitchLeadProvider` and `ChangeLeadProvider` share quite a bit of logic have you given any thought to abstracting the common logic?

It's looking great so far though, nice work!

app/services/mentor_at_school_periods/change_lead_provider.rb
```diff
@@ -0,0 +1,112 @@
...
+
+    include TrainingPeriodSources
+
+    def initialize(mentor_at_school_period:, lead_provider:, author:)
+      @mentor_at_school_period = mentor_at_school_period
+      @lead_provider = lead_provider
+      @author = author
+    end
+
+    def call
```
avinhurry: I like `call` but just FYI, there's a push within the team to stop using generic method names like that and use names that are a bit more descriptive.

app/services/mentor_at_school_periods/change_lead_provider.rb
```diff
@@ -0,0 +1,112 @@
...
+      TrainingPeriods::Create.provider_led(
+        period: mentor_at_school_period,
+        started_on: date_of_transition,
+        school_partnership:,
+        expression_of_interest:
+      ).call
+    end
+
+    def record_lead_provider_updated_event!
+      ::Events::Record.record_mentor_lead_provider_updated_event!(
```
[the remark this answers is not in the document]
thenapking: Have changed this to use the same event, but will confirm this is OK before merge.

app/views/schools/mentors/change_lead_provider_wizard/edit.html.erb
```diff
@@ -0,0 +1,39 @@
...
+    backlink_href: schools_mentor_path(@mentor_at_school_period)
+) %>
+
+<%= form_with(model: @wizard.current_step, url: @wizard.current_step_path) do |form| %>
+  <%= content_for(:error_summary) { form.govuk_error_summary } %>
+
+  <p>We recommend that you keep a mentor's lead provider the same until they complete their training.</p>
+  <p>If you change <%= @wizard.teacher_full_name %>'s lead provider, it means immediately:</p>
+  <ul class="govuk-list govuk-list--bullet">
+    <li>the old lead provider will no longer have access to <%= @wizard.teacher_full_name %>'s training record</li>
```
avinhurry:
```suggestion
    <li>the old lead provider will no longer have access to <%= @wizard.teacher_full_name %>’s training record</li>
```

app/views/schools/mentors/change_lead_provider_wizard/edit.html.erb
```diff
@@ -0,0 +1,39 @@
...
+    <li>training will stop and <%= @wizard.teacher_full_name %> will have to wait until the new lead provider confirms the partnership before they can continue</li>
+  </ul>
+
+  <p>Changing the lead provider might mean <%= @wizard.teacher_full_name %> will have a different delivery partner. Their training might be delivered differently.</p>
+  <p>If you do change lead provider, we'll let the:</p>
```
avinhurry:
```suggestion
  <p>If you do change lead provider, we’ll let the:</p>
```

app/services/concerns/teachers/switch_lead_provider_helper.rb
```diff
@@ -0,0 +1,117 @@
+module Teachers
+  module SwitchLeadProviderHelper
```
thenapking: I wasn't really sure what to name this.  I don't think it should use the word Helper, but I could do with some input on appropriate naming.

app/views/schools/mentors/change_lead_provider_wizard/confirmation.html.erb
```diff
@@ -0,0 +1,24 @@
+<%
+  page_data(
+    title: "You have changed the mentor’s lead provider to #{@wizard.current_step.new_lead_provider.name}",
```
avinhurry:
```suggestion
    title: "You have changed the mentor’s lead provider to #{@wizard.current_step.new_lead_provider.name}",
```

app/views/schools/mentors/change_lead_provider_wizard/confirmation.html.erb
```diff
@@ -0,0 +1,24 @@
...
+    as the new lead provider for #{@wizard.teacher_full_name}
+  TXT
+) %>
+
+<h2 class="govuk-heading-m">
+  What happens next?
+</h2>
+
+<p class="govuk-body">We'll let <%= @wizard.current_step.new_lead_provider.name %> know your school wants to work with them. They'll contact us to confirm the change and tell us the delivery partner they'll be working with.</p>
+<p class="govuk-body">We'll also let the old lead provider know that they'll no longer be training <%= @wizard.teacher_full_name %>.</p>
```
avinhurry:
```suggestion
<p class="govuk-body">We’ll also let the old lead provider know that they’ll no longer be training <%= @wizard.teacher_full_name %>.</p>
```

avinhurry (commented):
Looking better! I've left a few comments.

Nice clean de-dupe of the service without overly abstracting!

app/views/schools/mentors/change_lead_provider_wizard/edit.html.erb
```diff
@@ -0,0 +1,39 @@
...
+  <ul class="govuk-list govuk-list--bullet">
+    <li>new provider know that you want to work with them</li>
+    <li>old provider know they're no longer working with the mentor</li>
+  </ul>
+  <p>The new provider will confirm the delivery partner they're working with.</p>
+  <h2 class="govuk-heading-m">Select the new lead provider for <%= @wizard.teacher_full_name %></h2>
+
+  <%= form.govuk_collection_radio_buttons(
+        :lead_provider_id,
+        @wizard.current_step.lead_providers_for_select,
```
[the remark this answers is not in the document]
thenapking: Let me double check that - I don't think it should be in the list.

avinhurry (approved):
Nice work, looks good!

PR #1650, opened 2025-10-30
[2528] Add PUT participant withdraw to parity check

leandroalemao:
### Context

We'd like to add participant endpoints to parity check so we can test the results of the data compared to ECF1.

- Parity check for successful withdrawing a participant
- Parity check for withdrawing a participant that is already withdrawn

### Guidance to review

Parity check env

### DIFF
--- a/app/migration/parity_check/dynamic_request_content.rb
+++ b/app/migration/parity_check/dynamic_request_content.rb
@@ -52,20 +52,20 @@ module ParityCheck
         .pick(:api_id)
     end
 
-    def teacher_api_id_for_withdraw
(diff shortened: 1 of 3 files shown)


ethax-ross (commented):
Looks good, can we align the `DynamicRequestContent` in your PR @mooktakim with what's here -- I think how @leandroalemao has it here works well as it will test ECTs and mentors instead of just ECTs. 

I think if we changed the methods to:

`active_teacher_api_id_for_participant_action` and `withdrawn_teacher_api_id_for_participant_action` it would mean we could use it for both the error paths on defer and withdraw (as you can't withdraw a deferred participant)?

PR #1674, opened 2025-11-04
Test access to Admin dashboards

peterdavidhamilton:
### Context

Path helpers weren't working when attempting to access mounted apps as an unauthenticated user.
We could use strings for the redirect value but I opted to draw attention to it with a rescue and kick it down the road.

Test and rescue middleware redirects for `/admin/{blazer,jobs}`

### Guidance to review

Visit Mission Control and Blazer without being logged in


peterdavidhamilton:
Fixes #1684

PR #1683, opened 2025-11-05
[2568] Extract registration session facade and collaborators

avinhurry:
### Context

`ect.rb` has become a bit of a dumping ground. It carried a mixture of queries, boolean checks, formatting and previous registration data while still decorating the session store, which can be confusing to new joiners. 

This PR shows the full picture but i'm happy to split if reviewers prefer smaller slices.

- Replace the old `ect` session wrapper with `RegistrationSession`, a thinner facade around the session store.
- Split behaviour into focused collaborators (and inject them via the facade):
  - `RegistrationSession::Queries`: database lookups.
  - `RegistrationSession::Status`:  boolean checks
  - `RegistrationSession::Presenter`: UI formatting
  - `RegistrationSession::PreviousRegistration`:  wraps the previous `ect_at_school_period` and exposes all `previous_*` helpers.
- Rename and update all affected code to use the new `RegistrationSession` namespace.
- Added explicit unit tests for each collaborator and keep the wizards public facing interface unchanged.

### Guidance to review

- Does the organisation make sense?
- Are we happy with the naming?
- Is this level of abstraction appropriate for now?
- Does everything still work as expected?
- If this organisation looks good, i'll mirror it for [mentor.rb](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/wizards/schools/register_mentor_wizard/mentor.rb) next

### DIFF
--- a/spec/wizards/schools/register_ect_wizard/registration_session/queries_spec.rb
+++ b/spec/wizards/schools/register_ect_wizard/registration_session/queries_spec.rb
@@ -69,15 +69,17 @@ RSpec.describe Schools::RegisterECTWizard::RegistrationSession::Queries do
       let(:contract_period) { FactoryBot.create(:contract_period, year: 2025) }
       let(:start_date) { (contract_period.started_on + 2.days).to_s }
       let(:lead_provider) { FactoryBot.create(:lead_provider) }
+      let(:another_lead_provider) { FactoryBot.create(:lead_provider) }
 
       before do
         FactoryBot.create(:active_lead_provider, contract_period:, lead_provider:)
+        FactoryBot.create(:active_lead_provider, contract_period:, lead_provider: another_lead_provider)
       end
 
       it 'returns the active lead providers for the contract period' do
         ids = queries.lead_providers_within_contract_period.map(&:id)
 
-        expect(ids).to eq([lead_provider.id])
+        expect(ids).to contain_exactly(lead_provider.id, another_lead_provider.id)
       end
     end
   end


app/wizards/schools/register_ect_wizard/registration_session.rb
```diff
@@ -0,0 +1,95 @@
+module Schools
+  module RegisterECTWizard
```
thenapking: This feels like a concern to me, which should live elsewhere, especially as there are other subclasses. All the classes in the wizard directories are steps, except for those which extend the wizard base class.  Obviously, this wizard is the chunkiest, and this class is specifically for this wizard.  But it feels like a lib or a concern, especially as it is described as "decorator for the SessionRepository".

spec/wizards/schools/register_ect_wizard/registration_session/queries_spec.rb
```diff
@@ -0,0 +1,136 @@
...
+    context 'when a contract period is present' do
+      let(:contract_period) { FactoryBot.create(:contract_period, year: 2025) }
+      let(:start_date) { (contract_period.started_on + 2.days).to_s }
+      let(:lead_provider) { FactoryBot.create(:lead_provider) }
+
+      before do
+        FactoryBot.create(:active_lead_provider, contract_period:, lead_provider:)
+      end
+
+      it 'returns the active lead providers for the contract period' do
```
thenapking: Minor, but this might be more substantive if there are several lead providers which are returned and we match the array of IDs?

app/wizards/schools/register_ect_wizard/registration_session.rb
avinhurry: I see where you're coming from. There's definitely an argument for extracting it into a shared "decorator"concern. For this refactor I kept it under the wizard name space because:

- It's tightly coupled to the ect registration journey, nothing else consumes it yet, so lifting it into a shared concern feels premature

- Keeping it next to the steps makes the wiring easy to follow while we reshape things

- Once the teams comfortable with the structure (and we've mirrored it for mentors), we can revisit whether it's worth moving it into a shared concern. 

So yeah, we might end up moving it, but i'd like to land the refactor first and then decide if a shared abstraction adds value.

avinhurry:
> This looks like it has refactored and cleaned up the classes as desired. The separation of classes makes a lot of sense. My only thought is that it should live somewhere else in the app, as a service, concern or lib. Other than that I've left a few comments about minor issues.

Thanks @thenapking, I have responded to your comments and made some changes.

PR #1691, opened 2025-11-07
[2167] Assign ECTs to Schedules

thenapking:
### Context

Acceptance Criteria:

- [ ] Create a service that when provided with a date assigns to:
- 1st June - 31st October = standard September schedule
- 1st November - end of February = standard January schedule
- 1st March - 31st May = standard April schedule
- [ ] Ensure that when subsequent training periods are created for an ECT or mentor, they stick to the same schedule as the most recent provider-led training period
- [ ] Enforce the presence of a schedule when the training period is provider led
- [ ] Ensure the absence of a schedule when the training period is school led
- [ ] Add a relationship between schedules and events and update the 'registered as ECT' event to save it

I have added a Schedule::Find service.  This is named on the basis that the schedules already exist and it should not create missing ones.  
The service which creates TraingPeriods calls this to find the appropriate schedule.  
I have removed the logic which passes in a schedule that was added for the Teachers::Resume service, as we should keep the assignment of schedules tied to the creation of TrainingPeriods and not have it scattered throughout the app.
I have added a foreign key relation to the events table for schedule_id as requested, and updated the register as ECT event.

### Guidance to review

I am unsure currently how schedules will be created for future years.  I have not added an error for the possibility that the service does not find a schedule, as I am assuming they will always exist.  In tests however, especially as some of these run with fixed dates in the past this has made things a little tricky.   I have used an approach of creating schedules for a whole contract year (to avoid recreating the logic of creating the schedule for the current date in a test, and allowing tests to have several to select from).  In some tests, where they have hard coded dates which create TrainingPeriods several years in the past it is necessary to use time travel to ensure consistent data in the test DB.

The other issue I need a steer on is migrations.  I do not yet know much about that side of the app.  I have tried having the service assign a schedule when a TrainingPeriod is migrated.  This works for some tests, however the service looks for the first training period for an ECT/mentor and depending on the order the data is migrated will produce different results here.  So I assume we will need some custom logic here.  If this requires a ticket for another team we will need to temporarily disable the validation added to enforce the presence of schedules when the training period is provider led, or allow it to be overridden in the migration.

### DIFF
--- a/spec/services/schedules/find_spec.rb
+++ b/spec/services/schedules/find_spec.rb
@@ -20,8 +20,6 @@ RSpec.describe Schedules::Find do
   let(:lead_provider_delivery_partnership) { FactoryBot.create(:lead_provider_delivery_partnership, active_lead_provider:, delivery_partner:) }
   let(:school_partnership) { FactoryBot.create(:school_partnership, lead_provider_delivery_partnership:, school:) }
 
-  # include_context 'current schedules'
-
   before do
     FactoryBot.create(:schedule, contract_period:, identifier: "ecf-standard-january")
     FactoryBot.create(:schedule, contract_period:, identifier: "ecf-standard-april")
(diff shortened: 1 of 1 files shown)


thenapking:
Replaced with this smaller PR to add the service: https://github.com/DFE-Digital/register-early-career-teachers-public/pull/1701

PR #1694, opened 2025-11-10
[2775] Add a teacher type attribute to Schedule

mooktakim:
### Context

Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/2775

* Hardcode `teacher_type` method on `Schedule` to return `:ect`/`:mentor`


leandroalemao (approved):

ethax-ross (approved):
Looks good. Having these models know about teacher type isn't sitting well with me, but I can't think of a better solution at the moment; it might be something we circle back to at some point though 🤔

PR #1695, opened 2025-11-10
Add `api_updated_at` to training periods

cpjmcquillan:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/2746

#### Add api_updated_at to training periods
We want to return an `updated_at` for "Tranfers" in the API.

Roughly speaking, a transfer represents the transition of a Teacher's training
period **between** schools.

A transfer always has a "leaving" training period, and sometimes has a "joining"
training period.

Since we don't model a transfer directly, we need to determine the time a
transfer was last updated by looking at the relevant training periods.

This introduces a new column to training periods so we can store the
`api_updated_at` timestamp. This follows a similar pattern to Schools, Teachers
etc.

#### Touch training period api_updated_at
So that providers can filter transfers using the "updated_since" parameter, we
need to ensure tranfers have a relevant `updated_at`.

We can use the most recently `updated_at` from a transfer's training periods.

This touches the relevant training period `api_updated_at` values whenever:

- a School's URN changes
- a Lead Provider's name changes
- a Training Period's finished on changes

### DIFF
--- a/app/models/active_lead_provider.rb
+++ b/app/models/active_lead_provider.rb
@@ -1,8 +1,11 @@
 class ActiveLeadProvider < ApplicationRecord
   belongs_to :contract_period, inverse_of: :active_lead_providers, foreign_key: :contract_period_year
   belongs_to :lead_provider, inverse_of: :active_lead_providers
+
   has_many :lead_provider_delivery_partnerships
   has_many :delivery_partners, through: :lead_provider_delivery_partnerships
+  has_many :school_partnerships, through: :lead_provider_delivery_partnerships
+  has_many :training_periods, through: :school_partnerships
   has_many :statements
   has_many :expressions_of_interest, class_name: 'TrainingPeriod', foreign_key: 'expression_of_interest_id', inverse_of: :expression_of_interest
   has_many :events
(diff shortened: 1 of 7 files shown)


db/migrate/20251110162648_add_api_updated_at_to_training_periods.rb
```diff
@@ -0,0 +1,8 @@
+class AddAPIUpdatedAtToTrainingPeriods < ActiveRecord::Migration[8.0]
+  disable_ddl_transaction!
+
+  def change
+    add_column :training_periods, :api_updated_at, :datetime, null: false, default: -> { "CURRENT_TIMESTAMP" }
```
ethax-ross: Would `api_transfer_updated_at` make more sense here? usually we reserve `api_updated_at` for the endpoint that returns the model, but we don't have one for `TrainingPeriod` so it may be a bit confusing as to where its used?

ethax-ross (commented):
Looks good just a couple of minor suggestions but I think this is good to pick up; if you agree with the comments I'll raise a new ticket for the actual implementation 👍

PR #1701, opened 2025-11-11
[2167] Add Find schedule service

thenapking:
### Context
Acceptance Criteria:

Create a service that when provided with a date assigns to:
1st June - 31st October = standard September schedule
1st November - end of February = standard January schedule
1st March - 31st May = standard April schedule
 Ensure that when subsequent training periods are created for an ECT or mentor, they stick to the same schedule as the most recent provider-led training period

I have added a Schedule::Find service. This is named on the basis that the schedules already exist and it should not create missing ones.
The service which creates TraingPeriods calls this to find the appropriate schedule.
I have removed the logic which passes in a schedule that was added for the Teachers::Resume service, as we should keep the assignment of schedules tied to the creation of TrainingPeriods and not have it scattered throughout the app.

### Guidance to review
This is branched off the PR which adds schedules to events.  It covers adding schedules for both ECTs and Mentors

### DIFF
--- a/app/models/teacher.rb
+++ b/app/models/teacher.rb
@@ -107,7 +107,7 @@ class Teacher < ApplicationRecord
   def eligible_for_funding?
     Teachers::MentorFundingEligibility.new(trn:).eligible?
   end
-  
+
   def training_periods
     return ect_training_periods if ect_at_school_periods.exists?
 


app/services/schedules/find.rb
```diff
@@ -0,0 +1,94 @@
...
+      schedule_date.year
+    end
+
+    def june_start
+      Date.new(contract_period_year, 6, 1)
+    end
+
+    def october_end
+      Date.new(contract_period_year, 10, 1).end_of_month
+    end
```
peteryates: Super minor, but the day arg defaults to `1` here, so omitting it might make it clearer that we don't intend on using it.

 
```suggestion
    def october_end
      Date.new(contract_period_year, 10).end_of_month
    end
```

app/services/schedules/find.rb
```diff
@@ -0,0 +1,94 @@
...
+      @period = period
+      @training_programme = training_programme
+      @started_on = started_on
+    end
+
+    def call
+      return unless provider_led?
+      return most_recent_schedule if previous_provider_led_periods.exists?
+
+      Schedule.find_by(contract_period_year:, identifier:)
```
ericaporter: Minor improv - I would make this a private method so I can just easily read what is happening in the call e.g. `find_standard_schedule`

app/services/schedules/find.rb
```diff
@@ -0,0 +1,94 @@
...
+      training_programme == "provider_led"
+    end
+
+    def previous_provider_led_periods
+      period.teacher.training_periods.where(training_programme: 'provider_led')
+    end
+
+    def most_recent_provider_led_period
+      previous_provider_led_periods.latest_first.first
+    end
```
ericaporter: I'm just wondering with this set of private methods - would they be better centralised? I would be surprised if they didn't already exist elsewhere but if they don't (and I did a quick grep and couldn't find them...) they seem like ones that we might end up wanting to use elsewhere and so might end up defining in multiple places

app/services/schedules/find.rb
```diff
@@ -0,0 +1,94 @@
...
+      return unless provider_led?
+      return most_recent_schedule if previous_provider_led_periods.exists?
+
+      Schedule.find_by(contract_period_year:, identifier:)
+    end
+
+  private
+
+    def provider_led?
+      training_programme == "provider_led"
```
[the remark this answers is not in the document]
ericaporter: Ah, in that case then this is probably the simplest!

PR #1704, opened 2025-11-11
[2735] Add unfunded mentors query service

leandroalemao:
### Context

Ticket: [2735](https://github.com/DFE-Digital/register-ects-project-board/issues/2735)

We need to ensure a Lead Provider only views details for unfunded mentors which are linked to ECTs being trained by the provider AND who are not previously known to the Lead Provider

- Add unfunded mentors query service;

### DIFF
--- a/app/serializers/api/unfunded_mentor_serializer.rb
+++ b/app/serializers/api/unfunded_mentor_serializer.rb
@@ -1,0 +1,22 @@
+class API::UnfundedMentorSerializer < Blueprinter::Base
+  class AttributesSerializer < Blueprinter::Base
+    exclude :id
+
+    field(:full_name) { |teacher| Teachers::Name.new(teacher).full_name_in_trs }
+    field(:email) do |teacher, _options|
+      teacher.latest_mentor_at_school_period.email
+    end
+    field(:trn, name: :teacher_reference_number)
+    field :created_at
+
+    # TODO: use `api_unfunded_mentor_updated_at` once we have added it
+    field(:api_updated_at, name: :updated_at)
+  end
+
+  identifier :api_id, name: :id
+  field(:type) { "unfunded-mentor" }
+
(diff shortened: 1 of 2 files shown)


app/services/api/unfunded_mentors/query.rb
```diff
@@ -0,0 +1,78 @@
...
+      # We still want to return only mentors
+      results
+        .strict_loading
+        .joins(:mentor_at_school_periods)
+        .includes(
+          :latest_mentor_at_school_period
+        )
+    end
+
+    def where_lead_provider_is(lead_provider_id)
```
ethax-ross: Did you look at doing it in a single query? I'm wondering how performant this is; can we maybe test it on the parity check environment? if its fast as-is then great, but if not we may need to look into combining into a single query.

ethax-ross: Ok happy to try it as-is as I think it would be harder to follow in a single query. The parity check will give us a comparison with ECF as well 👍

spec/services/api/teachers/unfunded_mentors/query_spec.rb
```diff
@@ -0,0 +1,223 @@
+RSpec.describe API::Teachers::UnfundedMentors::Query, :with_metadata do
+  # Helper method to create a full ECT/Mentor setup
+  def create_ect_and_unfunded_mentor_setup(lead_provider_id: nil)
```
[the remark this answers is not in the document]
ethax-ross: So this is more what I had in mind - we only really care about the mentor, the rest of the setup can stay encapsulated in the method https://github.com/DFE-Digital/register-early-career-teachers-public/commit/c57c2c15fe8f515cb3509dac89f629ac6d89997d

ethax-ross (approved):

PR #1705, opened 2025-11-11
[2637] Participant change schedule service

mooktakim:
### Context

Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/2637

* New service `API::Teachers::ChangeSchedule`
* New service `Teachers::ChangeSchedule`
* New event `Events::Record.record_teacher_training_period_change_schedule_event!`

#### Ross' updates

- Update for new schedule_identifier logic

We only want to prevent ECTs from moving onto replacement schedules.

We don't surface an error in ECF if they give an invalid cohort, we just fallback to the default cohort so updating for consistency.

- Minor error message updates for parity with ECF

- Tweak LP currently training validation

Talking with Claire and Andy its going to be enough to make sure that the training period is 'ongoing today'. We will allow the current lead provider to change schedule even if they are moving to a new lead provider in the future.

We want to ensure that if they are moving to a new school or LP in the future that we cap the newly created training period `finished_on` appropriately to avoid overlaps.

- Change schedule behaviour changes

We only want to allow a lead provider to change schedule if they are currently training the participant **and no future training periods exist**. This is due to future training periods potentially being incompatible with the contract period they are changing to (if a `SchoolPartnership` does not exist for the future lead provider, for example).

We may revisit this in the future as there will be a period of time when a participant is scheduled to start at a new school with a new LP where neither LP will be able to change their schedule, but we would expect this to be an edge case.

We also need to determine the correct `finished_on` behaviour for the existing and new training period. If the existing training period has finished already, we retain that `finished_on`. If it finishes in the future we can truncate the `finished_on` to today in order to be able to slot the new training period in. The new training period will match the `finished_on` of the training period we are closing.

### DIFF
--- a/app/services/api/teachers/change_schedule.rb
+++ b/app/services/api/teachers/change_schedule.rb
@@ -19,6 +19,7 @@ module API::Teachers
       return false unless valid?
 
       Teachers::ChangeSchedule.new(
+        author: Events::LeadProviderAPIAuthor.new(lead_provider:),
         lead_provider:,
         teacher:,
         training_period:,
--- a/app/services/teachers/change_schedule.rb
+++ b/app/services/teachers/change_schedule.rb
@@ -1,13 +1,14 @@
 module Teachers
   class ChangeSchedule
-    attr_reader :lead_provider, :teacher, :training_period, :schedule, :school_partnership
+    attr_reader :lead_provider, :teacher, :training_period, :schedule, :school_partnership, :author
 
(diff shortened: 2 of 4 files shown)


ethax-ross:
> All good @ethax-ross @mooktakim qq please.. are we adding some logic for changing a schedule or contract period due to payments frozen? and about checking if new schedule milestone is valid against existing declarations? as we have in ECF1? or not yet as we don't have declarations in place? tks

Thanks @leandroalemao - frozen cohorts and validation against declarations will come in as part of the declarations work 👍

PR #1707, opened 2025-11-12
Prevent ECT training periods from being associated with replacement schedules

ethax-ross:
Reverts DFE-Digital/register-early-career-teachers-public#1694

Initially, we thought that mentors could only transfer to replacement schedules however looking at ECF again that's not actually the case. A mentor can transfer to _any_ schedule but an ECT can't transfer to a replacement schedule.

Reverting this as I think it makes more sense to implement this check in the `ChangeSchedule` service and impose validation in the `TrainingPeriod` to ensure it.

- Prevent ECTs from being assigned to replacement schedules

We only allow mentors to be assigned to replacement schedules.

**Note: before this goes out we'll potentially need to fix any bad data in staging/sandbox where an ECT is assigned to a replacement schedule** done ✅ 

Script to update staging/sandbox:

```
  replacement_identifiers = %w[ecf-replacement-april ecf-replacement-january ecf-replacement-september]
  replacement_schedules = Schedule.where(identifier: replacement_identifiers)

  ect_training_periods = []
  replacement_schedules.each do |schedule|
    ect_training_periods += schedule.training_periods.select(&:for_ect?)
  end

  ect_training_periods.each do |tp|
    contract_period = tp.schedule.contract_period
    schedule = Schedule.find_by(
      contract_period:,
      identifier: "ecf-standard-september"
    )

    if Faker::Boolean.boolean(true_ratio: 0.2)
      schedule = Schedule
        .where.not(identifier: replacement_identifiers)
        .where(contract_period:)
        .order(Arel.sql("RANDOM()"))
        .first
    end

    tp.update!(schedule:)
  end
end
```

### DIFF
--- a/app/models/schedule.rb
+++ b/app/models/schedule.rb
@@ -1,4 +1,10 @@
 class Schedule < ApplicationRecord
+  REPLACEMENT_SCHEDULE_IDENTIFIERS = %w[
+    ecf-replacement-april
+    ecf-replacement-january
+    ecf-replacement-september
+  ].freeze
+
   enum :identifier,
        {
          'ecf-extended-april' => 'ecf-extended-april',
@@ -28,4 +34,12 @@ class Schedule < ApplicationRecord
               message: 'Can be used once per contract period',
               scope: :contract_period_year
             }
+
+  scope :excluding_replacement_schedules, -> {
+    where.not(identifier: REPLACEMENT_SCHEDULE_IDENTIFIERS)
+  }
+
+  def replacement_schedule?
+    identifier.in?(REPLACEMENT_SCHEDULE_IDENTIFIERS)
+  end
 end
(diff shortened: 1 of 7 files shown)


thenapking (commented):
Just for context please also check this PR: https://github.com.mcas.ms/DFE-Digital/register-early-career-teachers-public/pull/1701.  We have a subsequent one in our sprint to add replacement schedules for mentors which we will get to pretty soon.

app/models/schedule.rb
```diff
@@ -35,7 +35,11 @@ class Schedule < ApplicationRecord
               scope: :contract_period_year
             }
 
-  def teacher_type
-    @teacher_type ||= MENTOR_TEACHER_TYPE_IDENTIFIERS.include?(identifier) ? :mentor : :ect
+  scope :excluding_replacement_schedules, -> {
+    where.not("identifier IN (?)", REPLACEMENT_SCHEDULE_IDENTIFIERS)
```
cpjmcquillan: Can we let ActiveRecord handle this for us?

```suggestion
    where.not(identifier: REPLACEMENT_SCHEDULE_IDENTIFIERS)
```

cpjmcquillan (approved):

PR #1708, opened 2025-11-12
Refactor teacher migrator

tonyheadford:
### Context

Refactoring the logic from `ECTAtSchoolPeriod`, `MentorAtSchoolPeriod` and `TrainingPeriod` migrators into `Teacher` migrator.  This should:

- enable collating the periods together at once
- allow us to apply fixes from the tennis matches to the periods more simply and with less repetition
- reduce the number of queries and therefore speed up the overall migration a bit (be interesting to see how much)

- Logic from `ECTAtSchoolPeriod`, `MentorAtSchoolPeriod` and `TrainingPeriod` migrators moved to `Teacher` migrator
- Removed `ECTAtSchoolPeriod`, `MentorAtSchoolPeriod` and `TrainingPeriod` migrators and specs.
- Incorporated `start_date` fix for teachers created on 09/02/2022 ([ticket](https://github.com/DFE-Digital/register-ects-project-board/issues/2782))

Also limited the teachers to migrate to those with a single IR.  This means that mentorships take a hit as we are not migrating all the mentors.

### Guidance to review

Migrating with this code and 1 IR limit - it is not surfacing errors in the at school periods and training periods in the dashboard at the moment but it is recording any failures in the DB.  We can address this in a future PR.

image

### DIFF
--- a/spec/migration/migrators/teacher_spec.rb
+++ b/spec/migration/migrators/teacher_spec.rb
@@ -30,7 +30,8 @@ RSpec.describe Migrators::Teacher do
     def setup_failure_state
       invalid_trn = '123'
       teacher_profile = FactoryBot.create(:migration_teacher_profile, trn: invalid_trn)
-      FactoryBot.create(:migration_participant_profile, :ect, teacher_profile:, user: teacher_profile.user)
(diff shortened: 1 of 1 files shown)


ltello:
LGTM, much simpler this way with no code duplication. Also, surely the performance has improved?

Let's see if it gets no too complicated when we process pps with more than 1 IR. 🤞

PR #1724, opened 2025-11-14
Transfer history for a given Teacher and Lead Provider

cpjmcquillan:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/2747

We want to compile all the transfers for a given Teacher relative to a lead
provider, so we can return a list of transfers in the API.

Each transfer has a leaving `TrainingPeriod` and an optional joining
`TrainingPeriod`. There are some rules that dictate what status and type a
transfer is, and these are documented in the tests.

We've also introduced a small `Teachers::SchoolTransfers::Transfer` object to
model a transfer.

### Guidance to review

image

### DIFF
--- a/spec/services/teachers/school_transfers/history_spec.rb
+++ b/spec/services/teachers/school_transfers/history_spec.rb
@@ -1,3 +1,4 @@
+# rubocop:disable RSpec/InstanceVariable
 RSpec.describe Teachers::SchoolTransfers::History do
   describe "#transfers" do
     let(:teacher) { FactoryBot.create(:teacher) }
@@ -345,3 +346,4 @@ private
     )
   end
 end
+# rubocop:enable RSpec/InstanceVariable


app/services/teachers/school_transfers/history.rb
```diff
@@ -0,0 +1,72 @@
...
+
+      def status
+        return :complete unless leaving_training_period.finished_on.future? ||
+          joining_training_period&.started_on&.future?
+
+        :incomplete
+      end
+
+      def teacher_type = leaving_training_period.for_ect? ? :ect : :mentor
+    end
```
cpjmcquillan: I put this in here for now, but we should probably extract it to it's own class and add some tests.

app/services/teachers/school_transfers/history.rb
```diff
@@ -0,0 +1,72 @@
...
+        next unless leaving_training_period.complete?
+
+        joining_training_period = next_school_period&.earliest_training_period
+        next unless leaving_training_period.provider_led_training_programme? ||
+          joining_training_period&.provider_led_training_programme?
+
+        if leaving_training_period.lead_provider == @lead_provider ||
+            joining_training_period&.lead_provider == @lead_provider
+          Transfer.new(leaving_training_period, joining_training_period)
+        end
```
cpjmcquillan: Is it clear what's happening and _why_ it's happening here?

I'm mindful my head has been in the space for a while, so keen to hear what you both think!

app/services/teachers/school_transfers/history.rb
```diff
@@ -0,0 +1,72 @@
...
+      school_periods.map.with_index { |school_period, index|
+        next unless school_period.complete?
+
+        next_school_period = school_periods[index + 1]
+        next if next_school_period &&
+          (school_period.school == next_school_period.school)
+
+        leaving_training_period = school_period.latest_training_period
+        next unless leaving_training_period.complete?
+
```
ethax-ross: I think we're missing a scenario here where - if we're on the _last_ `SchoolPeriod` and `TrainingPeriod` and the `Teacher` has completed their training - it would not be classed as a leaving transfer (as they have finished their training). Sorry I should have had an explicit scenario for this, but its sort of the inverse of scenario 2 in the ticket.

Maybe check with Claire or Andy, but I think a teacher has 'completed training' when the `teacher.finished_induction_period&.complete?`

ethax-ross (dismissed):
This is looking good, just a few questions and suggestions 👍

app/models/ect_at_school_period.rb
```diff
@@ -13,6 +13,8 @@ class ECTAtSchoolPeriod < ApplicationRecord
   has_many :mentor_at_school_periods, through: :teacher
   has_many :events
   has_one :current_or_next_training_period, -> { current_or_future.earliest_first }, class_name: 'TrainingPeriod'
+  has_one :earliest_training_period, -> { earliest_first }, class_name: "TrainingPeriod"
```
leandroalemao: Hi @cpjmcquillan Don't we need to add these to `MentorAtSchoolPeriod` as well?

app/services/teachers/school_transfers/history.rb
```diff
@@ -0,0 +1,72 @@
...
+  class History
+    ALLOWED_TEACHER_TYPES = %i[ect mentor].freeze
+
+    Transfer = Data.define(:leaving_training_period, :joining_training_period) do
+      def type
+        return :unknown unless joining_training_period
+        return :new_provider if leaving_training_period.lead_provider !=
+          joining_training_period.lead_provider
+
+        :new_school
```
leandroalemao: Hi @cpjmcquillan maybe we need to check if the school is different before assuming it's a new_school?

leandroalemao (commented):
Hi @cpjmcquillan Just added a few comments.. not sure if all apply tho.. just thinking out loud in some of them.. tks

spec/services/teachers/school_transfers/history_spec.rb
```diff
@@ -0,0 +1,881 @@
...
+
+      it "returns no transfers for lead provider #1" do
+        history = described_class.new(
+          teacher:,
+          lead_provider: lead_provider1,
+          teacher_type:
+        )
+
+        expect(history.transfers).to be_empty
+      end
```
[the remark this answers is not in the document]
cpjmcquillan: Something like this?

```ruby
context "when a teacher finishes school-led training at a single school" do
  let(:ect_at_school_period1) do
    FactoryBot.create(
      :ect_at_school_period,
      started_on: 3.years.ago,
      finished_on: 1.week.ago,
      teacher:
    )
  end
  let(:lead_provider1) { FactoryBot.create(:lead_provider) }

  before do
    create_training_period_for(
      ect_at_school_period1,
      from: ect_at_school_period1.started_on,
      to: 2.years.ago,
      programme_type: :provider_led,
      with: lead_provider1
    )

    create_training_period_for(
      ect_at_school_period1,
      from: 2.years.ago,
      to: 1.year.ago,
      programme_type: :provider_led,
      with: lead_provider1
    )

    create_training_period_for(
      ect_at_school_period1,
      from: 1.year.ago,
      to: ect_at_school_period1.finished_on,
      programme_type: :school_led
    )
  end

  it "returns no transfers for lead provider #1" do
    history = described_class.new(
      teacher:,
      lead_provider: lead_provider1,
      teacher_type:
    )

    expect(history.transfers).to be_empty
  end
end
```

app/services/teachers/school_transfers/history.rb
```diff
@@ -0,0 +1,72 @@
...
+            joining_training_period&.lead_provider == @lead_provider
+          Transfer.new(leaving_training_period, joining_training_period)
+        end
+      }.compact
+    end
+
+  private
+
+    def school_periods
+      @teacher.public_send("#{@teacher_type}_at_school_periods")
```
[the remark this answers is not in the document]
cpjmcquillan: Yeah I think that's a nicer API 👍🏼

app/services/teachers/school_transfers/history.rb
ethax-ross: I think its clear, but it could maybe be clearer if some of the logic was extracted to methods so:

```
next unless different_school?(school_period, next_school_period)
if relevant_to_lead_provider?(leaving_training_period, joining_training_period)
... etc
```

But I think its probably fine as-is tbh.

app/services/teachers/school_transfers/history.rb
ethax-ross: I'm not sure we need to check if the schools are different, its covered already when building the transfers?

https://github.com/DFE-Digital/register-early-career-teachers-public/pull/1724/files#diff-6071826e221eb1062a0f0f283919315891c8ab28f5a39b90ae1f50160ec93a27R14

spec/services/teachers/school_transfers/history_spec.rb
```diff
@@ -0,0 +1,881 @@
+RSpec.describe Teachers::SchoolTransfers::History do
+  describe "#transfers" do
+    let(:teacher) { FactoryBot.create(:teacher) }
+    let(:teacher_type) { :ect }
+
+    context "when a teacher leaves school-led training at one school " \
+            "and joins no known training at another school" do
```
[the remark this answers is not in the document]
cpjmcquillan: Sorry, it's the first scenario in the diagram.

image

spec/services/teachers/school_transfers/history_spec.rb
```diff
@@ -0,0 +1,881 @@
+RSpec.describe Teachers::SchoolTransfers::History do
+  describe "#transfers" do
+    let(:teacher) { FactoryBot.create(:teacher) }
+    let(:teacher_type) { :ect }
```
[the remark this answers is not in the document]
ethax-ross: I think its probably redundant but if you want to add them for completeness I'm not against it

spec/services/teachers/school_transfers/history_spec.rb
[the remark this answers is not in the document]
ethax-ross: Usually I'd be against instance variables being used like this, but tbh I think its contained enough and in the specs so I'm fine with it if it makes the scenarios easier to read. We could always accept the previous as an optional parameter as well if we wanted to be explicit:

```
school_period = create_school_period(previous_school_period:)
```

spec/services/teachers/school_transfers/history_spec.rb
```diff
@@ -0,0 +1,355 @@
...
+        )
+
+        expect(history.transfers).to be_empty
+      end
+    end
+  end
+
+private
+
+  def create_school_period(from:, to: nil)
```
leandroalemao: Hi @cpjmcquillan how about extracting those methods into helpers? like we did in `spec/support/mentorship_period_helpers.rb` and `spec/support/school_partnership_helpers.rb`

app/services/teachers/school_transfers/transfer.rb
```diff
@@ -0,0 +1,30 @@
...
+    def status
+      if leaving_training_period.finished_on.future? ||
+          joining_training_period&.started_on&.future?
+        :incomplete
+      else
+        :complete
+      end
+    end
+
+    def teacher_type = leaving_training_period.for_ect? ? :ect : :mentor
```
ethax-ross: Maybe instead of `teacher_type` we should just delegate `for_ect?` and `for_mentor?` to the leaving training period?

spec/services/teachers/school_transfers/history_spec.rb
```diff
@@ -0,0 +1,355 @@
...
+        expect(transfer.status).to eq(:complete)
+        expect(transfer.teacher_type).to eq(:ect)
+        expect(transfer.leaving_training_period).to eq(@training_period1)
+        expect(transfer.joining_training_period).to eq(@training_period2)
+      end
+    end
+
+    context "when a teacher leaves school-led training at one school " \
+            "and joins school-led training at another school" do
+      let(:ect_at_school_period1) do
```
ethax-ross: I think these `lets` can be dropped?

spec/services/teachers/school_transfers/history_spec.rb
```diff
@@ -0,0 +1,355 @@
...
+      started_on: from,
+      finished_on: to,
+      teacher:
+    )
+  end
+
+  def add_training_period(school_period, programme_type:, from:, to: nil, with: nil)
+    case programme_type
+    when :provider_led
+      existing_school_partnership = SchoolPartnership
```
ethax-ross: Worth pulling the school partnership find/create into its own method?

ethax-ross (commented):
Nice work! Just some final minor suggestions but this is really clean 👍

spec/services/teachers/school_transfers/history_spec.rb
cpjmcquillan: I think I'm going to leave them defined here for now. 

But if they end up being useful in other transfer-related tests, I'll definitely extract them 👌🏼

PR #1727, opened 2025-11-14
Set mentor end date from completion

tonyheadford:
### Context
From tennis matches, this sets the `end_date` for a single IR mentor when the `end_date` is blank using the following rules:

1. If mentor has a completion date < 1/9/2021, keep end date NULL for at_school period but set the training period end date to the next 31 August that follows their started date

2. If mentor has a completion date > 1/1/2024, keep end date NULL for at_school period but set the training period end date to match the mentor completion date

For all other ECTs and mentors with one induction record and no end date, keep end date NULL.

### DIFF
--- a/app/migration/data_fixes.rb
+++ b/app/migration/data_fixes.rb
@@ -15,7 +15,7 @@ module DataFixes
       induction_record.start_date
     end
   end
- 
+
   # there are some mentors with a single IR record that has no end date - we want to set an end date in the following circumstances:
   #
   #   1. If mentor has a completion date < 1/9/2021, keep end date NULL for at_school period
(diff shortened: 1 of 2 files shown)


app/migration/data_fixes.rb
```diff
@@ -14,4 +15,36 @@ def corrected_start_date(induction_record:, sequence_number:)
...
+  #   2. If mentor has a completion date > 1/1/2024, keep end date NULL for at_school period but set the
+  #   training period end date to match the mentor completion date
+  #
+  # For all other ECTs and mentors with one induction record and no end date, keep end date NULL.
+  #
+  def corrected_training_period_end_date(induction_record:)
+    participant_profile = induction_record.participant_profile
+    return induction_record.end_date if participant_profile.ect? || induction_record.end_date.present?
+
+    corrected_date = nil
```
ltello: I think there is no need for this initialisation. Ruby does it for us automagically 😄

tonyheadford: Thanks, it's just habit, I like being explicit, missing this sort of thing in other languages can cause all kinds of problems :)

PR #1735, opened 2025-11-17
[2256] Update route: `admin/teachers/:id` -> `admin/teachers/:id/induction`

avinhurry:
### Context

We are updating the structure of the admin teachers page and we will have different sections on the show page now, so this separates the induction view and starts to build out the foundations. 

This page will be under an "induction" tab in a following PR.

- Update route: `admin/teachers/:id` -> `admin/teachers/:id/induction`
- Update all usages
- `/admin/teachers/:id` was removed for now to keep this PR clear and confined but will be added back in a subsequent PR.

### Guidance to review

Does everything still work as it did before? There should be no functional changes other than the URL as described above.


app/controllers/admin/import_ect/find_ect_controller.rb
```diff
@@ -29,7 +29,7 @@ def create
       # Already registered
       rescue Admin::Errors::TeacherAlreadyExists
         existing_teacher = Teacher.find_by(trn: @pending_induction_submission.trn)
-        redirect_to admin_teacher_path(existing_teacher), notice: "Teacher #{existing_teacher.trn} already exists in the system"
+        redirect_to admin_teacher_induction_path(existing_teacher), notice: "Teacher #{existing_teacher.trn} already exists in the system"
```
thenapking: I'm just flagging that once this whole change is completed that this would actually redirect to the admin_teacher_path and show the overview tab. I'm assuming that for the moment this is fine and will be addressed in your next PR.

avinhurry: Yup, @peteryates has got us covered. From the ticket: 

> We can line a PR up for post launch that fixes the ordering and links so the overview page becomes the main one.

PR #1738, opened 2025-11-18
Update generic email address example

StephenC-V1:
Update generic email address example from 'admin@example.com' to 'headteacher@school.com'.

### Context
We want to make the example of a generic email address more clear to users

Update existing example of email address.

### DIFF
--- a/app/views/schools/register_mentor_wizard/_cant_use_email.md.erb
+++ b/app/views/schools/register_mentor_wizard/_cant_use_email.md.erb
@@ -5,6 +5,6 @@ backlink_href: <%= @wizard.previous_step_path %>
 
 Make sure you use an email that belongs to the person you're registering.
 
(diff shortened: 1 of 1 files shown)


app/views/schools/register_ect_wizard/cant_use_email.html.erb
```diff
@@ -6,6 +6,6 @@
 
 <p class="govuk-body">Make sure you use an email that belongs to the person you're registering.</p>
 
-<p class="govuk-body">Do not use a generic email like <a href="mailto:admin@example.com">admin@example.com</a>.</p>
+<p class="govuk-body">Do not use a generic email like headteacher@school.com.</p>
```
avinhurry: Just wanted to check if we wanted to preserve the link? Or the intention was to make it plain text.

Looks like we preserved the link on the mentors flow, probably worth being consistent?

# Before

Image

# After

Image

PR #1739, opened 2025-11-18
Make period duration minimum one day

peteryates:
### Context

The interval validation for checking period lengths allowed 0 day periods.

It doesn't really make sense for us to store them because they tell us `John Smith did :school_led training for 0 days`, which is the equivalent of `John Smith didn't do any :school_led training`.

1. **Update the validation so 0 day periods are not allowed.** - this happens both in the ActiveRecord validation (in `#valid_date_order?`) and at the database level with a check constraint
2. **Fix the services that allow the creation of 0 day periods**
3. **Tidy up any failing specs** - most of these were caused by the `finished_on` being the same as the `started_on`, so a `finished_on: started_on.tomorrow` did the trick

### Cleanup tasks

There's 1 induction period record in ECF2's database that is 0 days (ID: `102766`). Confirm with @emily-prudence-dfe how we might want to deal with it (nudge it to 1 day? delete it?)

### Help needed

There are random failures in the `spec/services/sandbox_seed_data` - could do with a hand resolving them cc @ethax-ross 

[e.g.](https://github.com/DFE-Digital/register-early-career-teachers-public/actions/runs/19477679473/job/55741561588?pr=1739),

```
  1) SandboxSeedData::APITeachersWithHistories#plant when creating TrainingPeriod records with a finished_on is expected to change `TrainingPeriod::ActiveRecord_Relation#count` by 20
     Failure/Error: it { expect { plant }.to change(TrainingPeriod.where.not(finished_on: nil), :count).by(20) }
       expected `TrainingPeriod::ActiveRecord_Relation#count` to have changed by 20, but was changed by 18
     # ./spec/services/sandbox_seed_data/api_teachers_with_histories_spec.rb:155:in 'block (4 levels) in <main>'
     # ./spec/rails_helper.rb:49:in 'block (3 levels) in <top (required)>'
     # ./app/models/concerns/declarative_updates.rb:49:in 'DeclarativeUpdates.skip'
     # ./spec/rails_helper.rb:49:in 'block (2 levels) in <top (required)>'
     # ./spec/support/rack_attack.rb:15:in 'block (2 levels) in <main>'
```

It's because 2/20 of these records are going to be created on the first day which is no longer included, but I couldn't see an easy way of fixing it!

### Guidance to review

Go commit by commit with whitespace disabled - I re-nested a `context` block in `spec/services/ect_at_school_periods/switch_training_spec.rb` - cc @cpjmcquillan 

### Pre-merge tasks

* [x] ensure the induction period in production with 0 days has been adjusted/removed
* [x] ensure the sandbox data with 0 day periods has been cleaned up

### DIFF
--- a/spec/services/mentor_at_school_periods/change_lead_provider_spec.rb
+++ b/spec/services/mentor_at_school_periods/change_lead_provider_spec.rb
@@ -82,7 +82,7 @@ RSpec.describe MentorAtSchoolPeriods::ChangeLeadProvider, type: :service do
         context "when the training period started today" do
           let(:started_on) { Date.current }
 
-          it "destroys the training period" do
+          it "destroys the training period", pending: "Work out how to deal with 0 day training periods" do
             subject
 
             expect { training_period.reload }.to raise_error(ActiveRecord::RecordNotFound)


cpjmcquillan (approved):
LGTM! 📆

PR #1743, opened 2025-11-19
Keep wizard status objects pure boolean

avinhurry:
### Context

For better separation of concerns we want to try and keep `status.rb` pure boolean, any heavy query logic should be moved into `queries.rb`.

Move some of the query logic from `status.rb` into `queries.rb`

### Guidance to review

Does this structure make sense to people? Do you agree with it?

### DIFF
--- a/app/wizards/schools/register_ect_wizard/registration_store/queries.rb
+++ b/app/wizards/schools/register_ect_wizard/registration_store/queries.rb
@@ -26,7 +26,7 @@ module Schools
           @contract_start_date ||= ContractPeriod.containing_date(registration_store.start_date&.to_date)
         end
 
-        def lead_provider_partnership_for_contract_period(school:)
+        def lead_provider_partnerships_for_contract_period(school:)
           contract_period = contract_start_date
 
           return SchoolPartnership.none unless previous_lead_provider && contract_period && school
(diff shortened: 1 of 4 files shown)


spec/wizards/schools/register_ect_wizard/registration_store/queries_spec.rb
```diff
@@ -84,6 +84,36 @@
...
+                        school:,
+                        lead_provider:)
+    end
+    let!(:training_period) do
+      FactoryBot.create(:training_period,
+                        :for_ect,
+                        ect_at_school_period: FactoryBot.create(:ect_at_school_period,
+                                                                teacher:,
+                                                                school:),
+                        school_partnership:)
```
ericaporter: Maybe weird indentation?

app/wizards/schools/register_ect_wizard/registration_store/queries.rb
```diff
@@ -26,6 +26,17 @@ def contract_start_date
           @contract_start_date ||= ContractPeriod.containing_date(registration_store.start_date&.to_date)
         end
 
+        def lead_provider_partnership_for_contract_period(school:)
```
ericaporter: Possibly rename this to be partnerships? Just housekeeping - I realise you were refactoring in this PR!

PR #1747, opened 2025-11-19
Review release note PR for /withdraw /defer /resume endpoints

tashonrp:
### Context
My first DFE pr! Thought I'd give it a go following guidance from @claire-hughez on this ticket.

Rewritten to follow the guidance in the [original ticket](https://github.com/DFE-Digital/register-ects-project-board/issues/2612) as discussed.

### DIFF
--- a/app/views/api/release_notes/release_notes.yml
+++ b/app/views/api/release_notes/release_notes.yml
@@ -41,7 +41,7 @@
 
     * their `participant_status` isn't `joining`
     * they've been training for at least one day
-    * they've not already been withdrawn or deferred (unchanged from the 'Manage training for ECTs' API)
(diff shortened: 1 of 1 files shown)


app/views/api/release_notes/release_notes.yml
```diff
@@ -21,6 +21,37 @@
 #
 #     **bold** and _italic_ text
 
+- title: Updated validation rules for participant action endpoints
```
claire-hughez: Following the other release notes, I think we should still be clear that these endpoints are now released on the Register ECTs API, as they haven't been previously. https://sandbox.register-early-career-teachers.education.gov.uk/api/guidance/release-notes

PR #1750, opened 2025-11-19
[2738] Add unfunded mentors to seed data for sandbox

leandroalemao:
### Context

Ticket: [2738](https://github.com.mcas.ms/DFE-Digital/register-ects-project-board/issues/2738)

We need unfunded mentors data to be added in sandbox env to be able to test internally and externally (by lead providers) and ensure functionality is correct.

### Guidance to review

I'll run the script created for sandbox in the review app for testing. After reviewed, we can run the same in sandbox.

```
Rails.logger.silence do
    DeclarativeUpdates.skip do
      SandboxSeedData::UnfundedMentors.new.plant
    end
  end
end
```

### DIFF
--- a/app/services/sandbox_seed_data/unfunded_mentors.rb
+++ b/app/services/sandbox_seed_data/unfunded_mentors.rb
@@ -1,4 +1,4 @@
-require Rails.root.join('spec/support/mentorship_period_helpers')
+require Rails.root.join("spec/support/mentorship_period_helpers")
 
 module SandboxSeedData
   class UnfundedMentors < Base


spec/services/sandbox_seed_data/unfunded_mentors_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+    allow(Logger).to receive(:new).with($stdout) { logger }
+    allow(Rails).to receive(:env) { environment.inquiry }
+
+    stub_const("#{described_class}::MIN_UNFUNDED_MENTORS_PER_LP", 2)
+  end
+
+  describe "#plant" do
+    subject(:plant) { instance.plant }
+
+    it { expect { plant }.to change(Teacher, :count).by(20) }
```
ethax-ross: Is it worth querying the unfunded mentors using the `Query` service for each LP here?

PR #1755, opened 2025-11-19
Enforce double quotes via Rubocop across the codebase

avinhurry:
### Context

We're standardising on **double quotes** because:

- It aligns with conventions on GOV.UK projects
- Interpolation is always available when needed
- It avoids noisy diffs caused by quote style churn

This also aligns with the existing preference across the team.

- **No functional changes, just quotes.**
- Enforce double quotes via Rubocop
- Converts existing single quotes to double quotes using `rubocop -A`
- Excludes historical migrations from this rule (new migrations must follow this style)
- Adds the quote style commit to `.git-blame-ignore-revs` to preserve readable history
- Regenerates `rubocop_todo.yml`

### Guidance to review

If you want to hide the large auto-correct commit from git blame after this PR is merged, run:

```sh
git config blame.ignoreRevsFile .git-blame-ignore-revs
```

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,2 +1,2 @@
 # 2025-11-19: Auto-correct strings to double quotes
-570308d5d505f6969ec995fb51604a4c6326bc1b
+4fe986e451feb2e0943f744d2528c195ad5c6540


thenapking (commented):
I guess there isn't too much risk here, but obviously this is a big change.  Is it worth doing it in a couple of batches - eg apply to specs first, then to classes that might have text for the frontend, like views etc, then to everything else?

avinhurry:
> ~Worth us setting `quote_type = double` in the [editorconfig](https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties)?~
> 
> I posted that before reading the full doc, it's a proposed property so might not work in Ruby. Worth a try though?

@peteryates my editor inherits from rubocop so I can't test that but more than happy for someone to add that if it works and it helps them. 

Looks like `quote_type` is in the ideas list rather than the official spec, so I think it will only help if your editor supports it.

PR #1757, opened 2025-11-20
Add a notes column to test fixtures CSV

peterdavidhamilton:
### Context

The full width page gives us room to restore the notes column

Update test guidance fixtures and component so contextual note for the fixture is in a separate cell

### Guidance to review

cpd-ec2-review-1757-web test teacherservices cloud_school_register-ect_find-ect

### DIFF
--- a/app/components/test_guidance_component.rb
+++ b/app/components/test_guidance_component.rb
@@ -35,13 +35,14 @@ class TestGuidanceComponent < ApplicationComponent
         "Date of birth",
         "Induction status",
         active_period_with_header,
+        "Notes",
         ""
       ]
     end
 
     # @return [Array<Array>]
     def rows
-      teachers.map do |name, trn, dob, ni_number, status, *|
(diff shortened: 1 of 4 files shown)


peterdavidhamilton:
> Thanks for doing this! Maybe we can talk through on the call how I can edit? We accidentally deleted the national insurance number column which we'll need for testing too, I can probably find it in the old files changed, right?

The NI has not been deleted, just hidden. The CSV file is here https://github.com.mcas.ms/DFE-Digital/register-early-career-teachers-public/blob/main/spec/fixtures/seeds_trs.csv

claire-hughez:
Let's try to catch up on this - I'm off Thurs and Friday so feel free to merge for now but struggling to work out how I'd edit myself!

I think the records showing as mentor funding ineligible are incorrect. Not sure how that happened, may have been entirely separate to any of this work on blue box testing!

PR #1760, opened 2025-11-20
WIP - Introduce API::Teachers::SchoolTransfers::Query

cpjmcquillan:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/2642

#### Introduce API::Teachers::SchoolTransfers::Query
This query object returns all the participants with school transfers for a given
lead provider.

It leans on established patterns used in other query objects across the API.

#### Introduce SchoolTransferHelpers
Before, we used these helper methods to create "school transfers" for teachers
in the `SchoolTransfers::History` specs.

We need to do something similar for the query specs, too.

This abstracts the helpers into a module and tweaks them slightly so we can
create `MentorAtSchoolPeriod` records too.

### DIFF
--- a/spec/services/api/teachers/school_transfers/query_spec.rb
+++ b/spec/services/api/teachers/school_transfers/query_spec.rb
@@ -6,12 +6,7 @@ RSpec.describe API::Teachers::SchoolTransfers::Query do
     let(:lead_provider) { FactoryBot.create(:lead_provider) }
     let(:params) { { lead_provider_id: lead_provider.id } }
 
-    before do
-      school_period1 = create_school_period(teacher, from: 2.years.ago, to: 1.year.ago)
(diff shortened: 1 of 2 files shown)


app/services/api/teachers/school_transfers/query.rb
```diff
@@ -0,0 +1,172 @@
...
+            },
+            latest_training_period: {
+              school_partnership: :school,
+              active_lead_provider: :lead_provider
+            }
+          }
+        )
+    end
+
+    def where_lead_provider_is(lead_provider_id)
```
ethax-ross: I appretiate this is still WIP so you may be planning it anyway, but is it worth extracting some of this into methods to DRY it up and make the overall logic easier to follow?

PR #1780, opened 2025-11-24
[2757] Validate start date within 4 months when moving schools

rwrrll:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/2757

- Validate ECT `start_date` is within 4 months when moving schools
- Validate Mentor `started_on` is within 4 months when moving schools
- Fix `Schools::RegisterMentorWizard::Mentor#previous_school_mentor_at_school_periods` and add specs (line 123 was previously unscoped and therefore returned period for every mentor in the database)

### Guidance to review

I used these TRNs for suitable seed users who are currently at a school:

ECT: 1023457
Mentor: 8886654

### DIFF
--- a/app/wizards/schools/register_mentor_wizard/started_on_step.rb
+++ b/app/wizards/schools/register_mentor_wizard/started_on_step.rb
@@ -5,6 +5,7 @@ module Schools
 
       validates :started_on, mentor_start_date: true
       validate :started_on_cannot_be_before_previous_started_and_finished_dates, if: :started_on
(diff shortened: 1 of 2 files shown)


thenapking (approved):
This looks fine to me.  The only thing I might consider is whether there is a feature test we could add this to as an appropriate part of the journey.  That said I think the request specs are fine, and we do want to avoid unnecessary duplication in the features, so I am happy if it is not added.

PR #1783, opened 2025-11-24
RIAB 1 - [2630] new appropriate body data model

peterdavidhamilton:
### Context

Decompose AppropriateBody model and data to better represent reality

image

https://app.excalidraw.com/l/52aBR6Dq1pU/6mbs2A38lcj?element=vqkeULpAju3ZZysKePQTk

https://github.com/DFE-Digital/register-ects-project-board/discussions/2949

#### Goals

1. Uncouple imported legacy data and login data from the `Appropriate Body` role.
2. Model the reality that a `Lead School` is authenticating to perform the AB role on behalf of a `Teaching School Hub`.
    - The DSI UUID for a TSH is a school but national organisations like ISTIP log in directly. We wish to make this relationship clear and unambiguous and enable the data model to capture new attributes moving forward.
3. Make it possible to connect multiple TSHs or regions.
    - We will either want to rename our TSH ABs because we will continue with only one TSH record handling the inductions over multiple regions, or we will add new hubs where Lead school has more than one.
4. Prepare the live service to deliver a new data model incrementally.
    - No visible UI changes until release.
    - No data changes that affect existing RIAB or RECT services.
    - Populate the data in an accurate and reliable way without developer intervention.
5. Fix inaccuracies in existing AB types.

### Prerequisites

Before merging and migrating we require the following `production` environment changes :

- [ ] fix incorrect `body_type` values for active ABs that will log in, namely one, ESP. Authentication is the mechanism that triggers new records to be created, driven by AB type. Old LAs will not log in and can be corrected using the PR's script.
- [ ] seed GIAS data and/or lead school data so that the migration for TSHs has access to the ID

- seed genuine ABs in both lower DSI environments for testing
- add migrations for new models
- populate new data only for successfully authenticated ABs
- connect model associations
- add model factories
- seed TSHs, NationalBodies, DfE Sign-In orgs and legacy ABs
- surface new models in "Admin console"
- disable new pages and routes using a feature flag
- unit and feature specs for DSI data capture and model migration
- add a database script to correct AB types identifying legacy LAs
- seed a second TSH for Bright Futures for demoing
- treat ESP like a national AB
- add new models to analytics blocklist
- add script to fix incorrect types for local authority ABs

### Guidance to review

- A number of national and regional organisations in DfE Sign-In have been prepared for the team to test against. For developers running locally against the DSI test environment and, once merged and deployed staging, against the DSI pre-prod environment.
    | Name | Type | Notes | 
    | --- | --- | --- |
    | ISTIP | `national` | |
    | ESP | `national` | type needs correcting |
    | Bright Futures Teaching School Hub | `teaching_school_hub` | seeded example of a lead school with two hubs |
    | Five Counties Teaching School Hubs | `teaching_school_hub` | |
    | Star Teaching School Hub | `teaching_school_hub` | |
    | STEP Ahead Teaching School Hub | `teaching_school_hub` | the correct name for Angel Oak Academy | 
    
- Use `SEED_NEW_APPROPRIATE_BODY_MODELS=y` to seed additional models
- Alternatively, run `rails gias:import` to populate the `gias_schools` table which equips `AppropriateBodyMigrator` to find the lead school.


db/schema.rb
```diff
@@ -71,7 +71,13 @@
     t.uuid "dfe_sign_in_organisation_id"
     t.uuid "dqt_id"
     t.enum "body_type", default: "teaching_school_hub", enum_type: "appropriate_body_type"
+    t.bigint "teaching_school_hub_id"
+    t.bigint "lead_school_id"
+    t.bigint "national_body_id"
```
peteryates: Am I right in thinking only one of these can be present at once?

db/schema.rb
```diff
@@ -805,6 +849,17 @@
     t.check_constraint "trnless OR trn IS NOT NULL", name: "check_trn_presence"
   end
 
+  create_table "teaching_school_hubs", force: :cascade do |t|
+    t.string "name", null: false
+    t.string "region"
```
peteryates: Do we have a definitive list of regions? I know [Ofsted have a published list](https://www.gov.uk/government/organisations/ofsted/about) - DfE [doesn't appear to have a list](https://github.com/DFE-Digital/dfe-reference-data/tree/main/lib/dfe/reference_data) (at least not in the reference data) but they clearly do have them ([here's a list of **regional** directors](https://www.gov.uk/government/organisations/regional-department-for-education-dfe-directors/about#who-we-are)).

Just wondering if this should be an enumerated type. I don't imagine they'll often change!)

peterdavidhamilton: See https://github.com/DFE-Digital/register-early-career-teachers-public/pull/1906 for progress on regions

app/models/appropriate_body.rb
```diff
@@ -1,20 +1,65 @@
...
+  # include Interval
 
   # Associations
+  has_one :legacy_appropriate_body, inverse_of: :appropriate_body_period
+
+  # TODO: remove UUID once linked to DfESignInOrganisation through TSH or NationalBody
+  belongs_to :dfe_sign_in_organisation, primary_key: :uuid, inverse_of: :appropriate_body_period
+  belongs_to :national_body, optional: true
+  belongs_to :teaching_school_hub, optional: true
+  belongs_to :lead_school, class_name: "School", optional: true
```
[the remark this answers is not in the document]
craigmdavidson:
```mermaid
classDiagram
   class NationalBody

   class TeachingSchoolHub

   class School

   class Organisation

   class AppropriateBodyPeriod

   class InductionPeriod

   NationalBody --|> Organisation
   TeachingSchoolHub --|> Organisation
   School --|> Organisation
   Organisation "1" *-- "*" AppropriateBodyPeriod
   AppropriateBodyPeriod "1" *-- "*" InductionPeriod
```

app/models/legacy_appropriate_body.rb
```diff
@@ -0,0 +1,25 @@
+# Name, DQT ID and type of Appropriate Body from imported legacy data
+class LegacyAppropriateBody < ApplicationRecord
```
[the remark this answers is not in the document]
peterdavidhamilton: I'll summarise some of the considerations and hopefully I might answer some questions at the same time:

- In most cases an "Appropriate Body" is a school, what is known as a "Lead school" working via a "Teaching School Hub".
- DfE Sign In authenticates users from organisations and gives us the org's UUID, but if it is a school the URN too.
- GIAS is our external source of truth for all school data "Get Information About Schools"
- We want to evolve "Appropriate Body" entry orgs into "roles done for time" so that hubs and lead schools can close and open. 
- There are a pair of national orgs who are expected to never close, however they do, so they'd just have one ongoing period we think.
- We need to maintain a papertrail of things imported from the old system, for example "Local authorities", which no longer perform the AB role. 
- This import hasn't finished and we are working on importing more.
- Assuming these three PRs merged and we had new models, a table in the same vein as `GIAS::School` that held DSI org details, that linked TSH type ABs to their school, and they each had a region, and maybe many periods of being ABs. Then after we'd corrected the body_types and migrated data we'd be able to remove these duplicated fields from "AppropriateBodyPeriod" and find the info elsewhere. Then we can start on the next stage making them behave like periods.

app/services/appropriate_body_migrator.rb
```diff
@@ -0,0 +1,79 @@
...
+  def dfe_sign_in_organisation
+    @dfe_sign_in_organisation ||=
+      DfESignInOrganisation.create_with(dfe_sign_in_data)
+        .find_or_create_by!(uuid: organisation.id)
+  end
+
+  # @return [TeachingSchoolHub]
+  def teaching_school_hub
+    @teaching_school_hub ||=
+      TeachingSchoolHub.create_with(dfe_sign_in_organisation:, lead_school:)
```
avinhurry: Will we always have a lead school?

app/models/appropriate_body.rb
```diff
@@ -1,20 +1,65 @@
...
+  validate :only_one_regional_or_national_body
+  validate :national_body_limit
+
+  # National Bodies only have one unending AB period (TSHs can have many)
+  def national_body_limit
+    return if national_body_id.blank?
+    return if NationalBody.find(national_body_id).appropriate_body_period.blank?
+
+    errors.add(:base, "A National Body can only have a single Appropriate Body period")
+  end
```
avinhurry: If national bodies have an unending period won't this blow up for every update. Do we need to do something like this instead: 


```ruby
existing = NationalBody.find(national_body_id).appropriate_body_period

return if existing.blank? || existing.id == id
```

Probably worth adding a spec to confirm the behaviour.

app/services/appropriate_body_migrator.rb
```diff
@@ -0,0 +1,79 @@
...
+        .find_or_create_by(name:)
+  end
```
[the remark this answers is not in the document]
peterdavidhamilton: https://github.com/DFE-Digital/register-ects-project-board/issues/3201

Currently the plan is to run the importer and provided it still creates all the schools here is how to test this locally

https://github.com/DFE-Digital/register-ects-project-board/discussions/2949#discussioncomment-15639231

app/models/appropriate_body.rb
```diff
@@ -1,20 +1,65 @@
...
-         national: "national",
-         teaching_school_hub: "teaching_school_hub" },
-       validate: { message: "Must be local authority, national or teaching school hub" }
+  enum :body_type, {
+    local_authority: "local_authority",
+    national: "national",
+    teaching_school_hub: "teaching_school_hub"
+  }, validate: {
+    message: "Must be local authority, national or teaching school hub"
+  }
```
craigmdavidson: Do we still need this enum? Can we use the `belongs_to` arc? 
Maybe one for a subsequent PR?

PR #1784, opened 2025-11-24
Introduce `API::Teachers::SchoolTransfers::Query`

cpjmcquillan:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/2642

#### Introduce SchoolTransferHelpers
Before, we used these helper methods to create "school transfers" for teachers
in the `SchoolTransfers::History` specs.

We need to do something similar for the query specs, too.

This abstracts the helpers into a module and tweaks them slightly so we can
create `MentorAtSchoolPeriod` records too.

#### Refactor Teachers::SchoolTransfers::History
In the serializer we will have access to the lead provider ID rather than the
lead provider record, but until now the `History` compared lead provider
records.

This refactors the `relevant_to_lead_provider?` check to compare IDs, and
reworks the initializer so it accepts IDs instead of lead providers.

As part of this change we also added tests to ensure the `History` class can
handle scenarios where there are no transfers or a teacher has no training
periods.

---

#### Add involved_in_school_transfer to TeacherLeadProvider metadata

As we pivot away from the approach outlined in #1760 to one that uses metadata,
we want to introduce a boolean attribute to the `Metadata::TeacherLeadProvider`
model.

This will make the query to retrieve all the Teachers with a school transfer for
a given Lead Provider much more straightforward.

This column is nullable for now, but we might consider making it non-nullable
once we have refreshed the metadata for all Teachers.

---

#### Refresh induction period teacher metadata

Whether or not a Teacher is involved in a school transfer for a given Lead
Provider depends partly on whether they have completed their induction.

This means we need to refresh a Teacher's metadata when an `InductionPeriod` is
completed (i.e `finished_on` and `outcome` are not nil).

This uses declarative updates to refresh the metadata when either of those
attributes change.

In order to make this change, we needed to update the `InductionPeriod` model to
define an enum for `outcome` to align it with the DB column type.

---

#### Assign involved_in_school_transfers in Teacher metadata handler

We're refreshing Teacher metadata when training periods and induction periods
change (i.e when their school transfer history might be impacted).

This updates the Teacher metadata handler to update the
`involved_in_school_transfer` boolean if there are any school transfers for a
given Teacher and Lead Provider.

---

#### Introduce API::Teachers::SchoolTransfers::Query

This query object returns all the participants with school transfers for a given
lead provider.

It leans on established patterns used in other query objects in the API.

Since the query to retrieve relevant Teachers proved to be extremely complex
(see #1760), we decided to use a flag on the `TeacherLeadProvider` metadata
instead.

This makes the query much more straightforward, and means all the logic around
what is/isn't a "school transfer" is contained in the
`Teachers::SchoolTransfers::History` class.

### Guidance to review

- Create some "school transfers" locally and see them returned by the query (more extensive testing can be done once we create review app/sandbox seed data)

### DIFF
--- a/app/services/api/teachers/school_transfers/query.rb
+++ b/app/services/api/teachers/school_transfers/query.rb
@@ -1,0 +1,87 @@
+module API::Teachers::SchoolTransfers
+  class Query
+    include Queries::FilterIgnorable
+
+    attr_reader :scope
+
+    def initialize(
+      lead_provider_id:,
+      updated_since: :ignore,
+      sort: { created_at: :asc }
+    )
+      @scope = Teacher.distinct
+
+      where_lead_provider_is(lead_provider_id)
+      where_updated_since(updated_since)
+      set_sort_by(sort)
+    end
+
+    def school_transfers
+      preload_associations(block_given? ? yield(scope) : scope)
+    end
+
+    def school_transfers_by_api_id(api_id)
+      return preload_associations(scope).find_by!(api_id:) if api_id.present?
+
+      fail(ArgumentError, "api_id needed")
+    end
+
+    def school_transfers_by_id(id)
+      return preload_associations(scope).find(id) if id.present?
+
+      fail(ArgumentError, "id needed")
+    end
+
+  private
+
+    def preload_associations(results)
+      results
+        .strict_loading
+        .includes(
+          lead_provider_metadata: [],
+          finished_induction_period: [],
+          ect_at_school_periods: {
+            earliest_training_period: {
+              school_partnership: :school,
+              active_lead_provider: :lead_provider
+            },
+            latest_training_period: {
+              school_partnership: :school,
+              active_lead_provider: :lead_provider
+            }
+          },
+          mentor_at_school_periods: {
+            earliest_training_period: {
+              school_partnership: :school,
+              active_lead_provider: :lead_provider
+            },
+            latest_training_period: {
+              school_partnership: :school,
+              active_lead_provider: :lead_provider
+            }
+          }
+        )
+    end
+
+    def where_lead_provider_is(lead_provider_id)
+      @scope = scope
+        .joins(:lead_provider_metadata)
+        .where(
+          lead_provider_metadata: {
+            lead_provider_id:,
+            involved_in_school_transfer: true
+          }
+        )
+    end
+
+    def where_updated_since(updated_since)
+      return if ignore?(filter: updated_since)
+
+      @scope = scope.where(api_updated_at: updated_since..)
+    end
+
+    def set_sort_by(sort)
+      @scope = scope.order(sort)
+    end
+  end
+end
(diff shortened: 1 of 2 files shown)


app/services/metadata/handlers/teacher.rb
```diff
@@ -68,5 +71,9 @@ def latest_mentor_training_period_by_lead_provider(teacher:)
       .order("lead_provider_id, training_periods.started_on DESC")
       .index_by { it.lead_provider&.id }
     end
+
+    def school_transfers_for(school_periods, lead_provider_id)
+      ::Teachers::SchoolTransfers::History.transfers_for(school_periods:, lead_provider_id:)
```
ethax-ross: I don't think this will be a bottleneck, but if it is we could maybe add a `involved_in_school_tranfer?` method to `History` that is more optimal/returns after finding the first transfer for the LP.

ethax-ross (commented):
This is looking good Connor 👍 can you test refreshing all the metadata on the parity check environment and see how long it takes before/after this change? (refresh it async and monitor the jobs to see how long it takes with `Metadata::Manager.refresh_all_metadata!(async: true)`).

spec/models/induction_period_spec.rb
```diff
@@ -4,6 +4,16 @@
 
   it_behaves_like "an induction period"
 
+  describe "enums" do
+    it "uses the outcome enum" do
+      expect(subject).to define_enum_for(:outcome)
+                          .with_values({ fail: "fail", pass: "pass" })
+                          .validating(allowing_nil: true)
+                          .with_suffix(:outcome)
```
leandroalemao: Hi @cpjmcquillan only `with_suffix` will do I guess

leandroalemao (approved):
Well done @cpjmcquillan just very optional minor changes and a method for a single transfer.. tks

app/services/api/teachers/school_transfers/query.rb
```diff
@@ -0,0 +1,83 @@
...
+
+      where_lead_provider_is(lead_provider_id)
+      where_updated_since(updated_since)
+      set_sort_by(sort)
+    end
+
+    def school_transfers
+      preload_associations(block_given? ? yield(scope) : scope)
+    end
+
```
leandroalemao: @cpjmcquillan Just noticed.. it's missing a method for looking up for single transfer?

leandroalemao (commented):
@cpjmcquillan Just noticed.. it's missing a method for looking up for single transfer?

app/services/api/teachers/school_transfers/query.rb
ethax-ross: Not sure we need it for this query as we don't return one transfer anywhere in the API

leandroalemao: yeah you're right @ethax-ross .. I meant transfers for a single participant `/api/v3/participants/ecf/{id}/transfers`

ethax-ross: Ah sorry I see what you mean, yes we will need it for that you're right; so this would be finding a single participant with a transfer 👍

ethax-ross (approved):
Looks good, just one minor comment around metadata in specs 👍

PR #1800, opened 2025-11-26
Add `api_transfer_updated_at` to `TrainingPeriod`

ethax-ross:
### Context

We need a reliable way of determining when a transfer surfaced in the API has been touched/changed. A transfer is a `TrainingPeriod` that is leaving/joining a school (excluding the first training period and last training period for participants that have completed training).

- Add `api_transfer_updated_at` to `TrainingPeriod`.

Touch the new field when any of the following are changed:

- School#urn
- TrainingPeriod#started_on
- TrainingPeriod#finished_on
- LeadProvider#name

Update the query and serializer to work with the new updated at attribute.

Remove tests around sorting by `updated_at` in the query service; we don't support custom ordering in the transfers endpoint (and sorting on `api_transfer_updated_at` leads to a very complex query, so I don't think we should add it).

### DIFF
--- a/app/serializers/api/teachers/school_transfer_serializer.rb
+++ b/app/serializers/api/teachers/school_transfer_serializer.rb
@@ -1,12 +1,12 @@
 class API::Teachers::SchoolTransferSerializer < Blueprinter::Base
   class TrainingPeriodSerializer < Blueprinter::Base
-    field(:school_urn) do |(data, _options)|
+    field(:school_urn) do |data|
       data[:school].urn
     end
-    field(:provider) do |(data, _options)|
+    field(:provider) do |data|
       data[:training_period].lead_provider&.name
     end
-    field(:date) do |(data, _options)|
+    field(:date) do |data|
       data[:training_period].finished_on&.rfc3339
     end
   end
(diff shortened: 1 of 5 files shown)


app/services/api/teachers/school_transfers/query.rb
```diff
@@ -77,7 +77,18 @@ def where_lead_provider_is(lead_provider_id)
...
+            ect_at_school_periods: %i[earliest_training_period latest_training_period],
+            mentor_at_school_periods: %i[earliest_training_period latest_training_period]
+          )
+          .where(
+            "(training_periods.id IS NOT NULL AND training_periods.api_transfer_updated_at >= :updated_since) OR
+            (latest_training_periods_ect_at_school_periods.id IS NOT NULL AND latest_training_periods_ect_at_school_periods.api_transfer_updated_at >= :updated_since) OR
+            (earliest_training_periods_mentor_at_school_periods.id IS NOT NULL AND earliest_training_periods_mentor_at_school_periods.api_transfer_updated_at >= :updated_since) OR
+            (latest_training_periods_mentor_at_school_periods.id IS NOT NULL AND latest_training_periods_mentor_at_school_periods.api_transfer_updated_at >= :updated_since)",
+            updated_since:
+          )
```
cpjmcquillan: Should we add some tests for the filter applying when the school periods have been updated since a given date?

app/serializers/api/teachers/school_transfer_serializer.rb
```diff
@@ -39,21 +39,28 @@ class TransferSerializer < Blueprinter::Base
...
+          school_periods: teacher.ect_at_school_periods,
+          lead_provider_id: options[:lead_provider_id]
+        )
+        mentor_transfers = ::Teachers::SchoolTransfers::History.transfers_for(
+          school_periods: teacher.mentor_at_school_periods,
+          lead_provider_id: options[:lead_provider_id]
+        )
+
+        (ect_transfers + mentor_transfers).sort_by { |it| it.leaving_training_period.started_on }
+      end
```
cpjmcquillan: Do we need to worry about thread-safety here? 🤔

app/services/api/teachers/school_transfers/query.rb
ethax-ross: Yeah good shout I'll add more explicit tests around this filtering 👍

app/services/api/teachers/school_transfers/query.rb
```diff
@@ -74,10 +74,31 @@ def where_lead_provider_is(lead_provider_id)
...
 
-    def where_updated_since(updated_since)
+    def where_updated_since(updated_since, lead_provider_id)
       return if ignore?(filter: updated_since)
 
-      @scope = scope.where(api_updated_at: updated_since..)
+      # Note that this will include the first training period of a teacher and the last
+      # training period even if they have completed training (which are not classed as transfers).
+      # Excluding these would be very complicated; this is close enough for now and an improvement
+      # over what ECF currently provides.
```
cpjmcquillan: Is it worth highlighting what the impact is on the User here?

I'm assuming this means that the API _might_ return participant transfers that haven't actually been updated since the given date if an excluded training period has been.

PR #1801, opened 2025-11-26
[2468] removes "Fail" as an outcome from the bulk upload process.

abdullaaden77:
Removing "Fail" from bulk upload ensures this outcome is only submitted deliberately through the dedicated process.

### Context
Removing "Fail" from bulk upload ensures this outcome is only submitted deliberately through the dedicated process.
Fail outcomes are rare.

Remove all code that implementing "fail" outcome.


abdullaaden77:
> You mentioned you'd made changes, has a push failed?

Yes some tests failed unfortunately, currently fixing them.

peterdavidhamilton:
Ah, as I wrote that message I couldn't see the second commit. Gimme a shout when it's passing

spec/jobs/appropriate_bodies/process_batch/action_job_spec.rb
```diff
@@ -33,7 +33,7 @@
           perform_action_job
         }.to have_broadcasted_to(
           "batch_progress_stream_#{pending_induction_submission_batch.id}"
-        ).from_channel(pending_induction_submission_batch).exactly(11).times
```
peterdavidhamilton: `./spec/support/shared_contexts/csv_file.rb` contains mock CSV data for some specs including `fail` which it now shouldn't (or at least not claim that is a valid option), so we need to edit that.

peterdavidhamilton (changes requested):

peterdavidhamilton (approved):
Looks like everything is covered

PR #1815, opened 2025-11-27
Finish `Declaration` modelling

ethax-ross:
### Context

The `Declaration` model has been stubbed out up until now; we need to flesh this out ahead of the declarations work for the API.

- Finish `Declaration` modelling

Add missing attributes to `Declaration` and change `declaration_type` to an `enum` to match what we use on `Milestone`.

We are adding a couple of new relationships; `voided_by_user` to track which admin user has manually voided a declaration and `mentor_id` which will stamp the `Declaration` with the `Teacher` that was mentoring the participant at the time it was made. We are stamping this as it is needed in the serializer and would involve an expensive and complex set of joins to pull it dynamically (we stamp it in ECF for this reason as well).

Add appropriate validation logic as per ECF.

Control state with a state machine to be consistent with `Statement` and `Statement::LineItem`. The state machine is mostly [lifted from NPQ registration](https://github.com/DFE-Digital/npq-registration/blob/main/app/models/declaration.rb#L56-L88) as the models are pretty much the same.

- Fix seeded declaration

The declaration date needs to be within the milestone dates.

- Switch mentor_teacher to mentorship_period

This is easier to validate and seems like a more sensible association; it will require a bit more joining when we come to
serialize, but shouldn't pose any major performance issues.

- Improve associations and validation

Add `has_one` associations to make the behaviour clearer.

Move validation rules to `Statement::LineItem` as they make more sense here.

- Fix Statement::LineItem state machine

The statement items work differently to how I originally thought; the first statement item goes through billable states but can also end up as `voided` or `ineligible`.

When a `paid` declaration is clawed back, we create a new statement line item in an `awaiting_clawback` state that then transitions to `clawedback`.

Update the state machine to make this clearer. It still feels a bit off though as you can only specify one initial state and we will never go from `eligible` to `awaiting_clawback`.

### Guidance to review

There are a couple of `DeclarationState` entries in ECF that are `ineligible` without a `reason` provided, but I think these are a mistake and it should always be `duplicate` as the reason, so I've enforced that in our validation here.

I thought all ECT declarations might have a `mentor_user_id` in ECF, but this isn't the case - a lot do not have one, so I've left it as optional here.

There is 1 mentor declaration in ECF that has a `mentor_user_id`, which must be a mistake so I've enforced that in validation for RECT.

State machine:

Screenshot 2025-12-01 at 14 43 18

### DIFF
--- a/app/models/statement/line_item.rb
+++ b/app/models/statement/line_item.rb
@@ -56,10 +56,6 @@ class Statement::LineItem < ApplicationRecord
       transition %i[eligible ineligible payable] => :voided
     end
 
-    event :mark_as_awaiting_clawback do
-      transition [:paid] => :awaiting_clawback
-    end
-
     event :mark_as_clawed_back do
       transition [:awaiting_clawback] => :clawed_back
     end
--- a/documentation/state-machines.md
+++ b/documentation/state-machines.md
@@ -30,14 +30,17 @@ stateDiagram-v2
 
 A line item attached to a statement has more complexity and therefore supports a wider range of states and transitions.
 
-A statement line item has the following possible states:
+A statement line item has the following possible payment states:
 
-- `eligible` is the initial state of a line item – it is valid and can progress toward payment.
+- `eligible` is the initial state of a line item that will progress towards payment.
 - `payable` once it has been confirmed for payment.
 - `paid` when the line item has been paid.
 - `voided` when the line item as been cancelled before being paid.
 - `ineligible` when the line item does not qualify for payment.
-- `awaiting_clawback` when the payment has been issued but is later flagged for recovery.
+
+A statement line item has the following possible refund states:
+
+- `awaiting_clawback` is the initial state of a line item that will progress towards a refund.
 - `clawed_back` once the amount has successfully been reclaimed.
 
 There are seven transitions a line item can go through:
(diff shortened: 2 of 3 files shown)


app/models/declaration.rb
```diff
@@ -1,20 +1,142 @@
...
+
+  enum :declaration_type, {
+    started: "started",
+    "retained-1": "retained-1",
+    "retained-2": "retained-2",
+    "retained-3": "retained-3",
+    "retained-4": "retained-4",
+    "extended-1": "extended-1",
+    "extended-2": "extended-2",
+    "extended-3": "extended-3",
```
peteryates: Doesn't it end up being a bit awkward having symbol hash keys with dashes? i.e., `thing[:"retained-1"]`

app/models/declaration.rb
```diff
@@ -1,20 +1,142 @@
...
-         "completed" => "completed",
-         "extended-1" => "extended-1",
-         "extended-2" => "extended-2",
-         "extended-3" => "extended-3"
-       },
-       validate: { message: "Choose a valid declaration type" }
-
   belongs_to :training_period
+  belongs_to :voided_by_user, class_name: "User", optional: true
+  belongs_to :mentor_teacher, class_name: "Teacher", optional: true
```
ethax-ross: To change to `MentorshipPeriod` TBC

PR #1843, opened 2025-12-02
[2862] add EvidenceHeldValidator

mooktakim:
### Context

Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/2862

* Validator for `evidence_held` for simple and detailed types

### DIFF
--- a/app/validators/evidence_held_validator.rb
+++ b/app/validators/evidence_held_validator.rb
@@ -8,7 +8,7 @@ class EvidenceHeldValidator < ActiveModel::Validator
   def validate(record)
     return if record.errors[:evidence_held].any?
 
-    evidence_held_is_present(record) unless record.declaration_type == "started"
+    evidence_held_is_present(record) if self.class.evidence_held_required?(record)
 
     if validate_detailed_evidence_types?(record)
       evidence_held_is_valid_detailed_evidence_type(record)
@@ -19,6 +19,10 @@ class EvidenceHeldValidator < ActiveModel::Validator
 
   def self.evidence_held_required?(record)
     record.declaration_type.present? && record.declaration_type != "started"
+  end
+
+  def self.evidence_held_allowed?(record)
+    record.training_period.contract_period.detailed_evidence_types_enabled || evidence_held_required?(record)
   end
 
 private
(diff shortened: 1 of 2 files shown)


spec/validators/evidence_held_validator_spec.rb
```diff
@@ -0,0 +1,290 @@
...
+
+        context "when `declaration_type` is other than started" do
+          let(:declaration_type) { "retained-1" }
+
+          context "when `evidence_held` is nil" do
+            let(:evidence_held) { nil }
+
+            it "has a meaningful error", :aggregate_failures do
+              expect(subject).to be_invalid
+              expect(subject.errors.messages_for(:evidence_held)).to include("Enter a '#/evidence_held' value for this participant.")
```
cpjmcquillan: Is it worth using `contain_exactly` so we verify this is the only error message? 💭

app/validators/evidence_held_validator.rb
```diff
@@ -0,0 +1,101 @@
...
+    return if record.errors[:evidence_held].any?
+
+    if validate_detailed_evidence_types?(record)
+      evidence_held_is_present(record) unless record.declaration_type == "started"
+      evidence_held_is_valid_detailed_evidence_type(record)
+    elsif validate_simple_evidence_types?(record)
+      evidence_held_is_present(record)
+      evidence_held_is_valid_simple_evidence_type(record)
+    end
+  end
```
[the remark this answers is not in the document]
mooktakim: I've made those change 👍

spec/validators/evidence_held_validator_spec.rb
```diff
@@ -331,4 +331,67 @@ def initialize(declaration_type:, evidence_held:, training_period:)
...
+    let(:record) { Struct.new(:declaration_type, :training_period).new(declaration_type, training_period) }
+    let(:training_period) { Struct.new(:contract_period).new(contract_period) }
+
+    context "when contract period has simple evidence types" do
+      let(:detailed_evidence_types_enabled) { false }
+
+      context "when `declaration_type` is nil" do
+        let(:declaration_type) { nil }
+
+        it "evidence_held is not required" do
```
ethax-ross: I think all these should be `allowed` rather than `required`

ethax-ross (approved):

PR #1845, opened 2025-12-02
[2690] Add ECT permanently leaving flow

avinhurry:
### Context

We are implementing a way for school induction tutors to tell us that an ECT is leaving.

image

### Guidance to review

Navigate to `school/ects/:id`, scroll down and click the link at the bottom to say the ECT is leaving. Follow the journey.

### DIFF
--- a/spec/wizards/schools/ects/teacher_leaving_wizard/edit_step_spec.rb
+++ b/spec/wizards/schools/ects/teacher_leaving_wizard/edit_step_spec.rb
@@ -22,8 +22,10 @@ RSpec.describe Schools::ECTs::TeacherLeavingWizard::EditStep do
       let(:leaving_on) { {} }
 
       it "returns false and does not write to the store" do
-        expect { step.save! }.not_to(change(store, :leaving_on))
-        expect(step.save!).to be_falsey
+        result = step.save!
+
+        expect(result).to be_falsey
+        expect(store.leaving_on).to be_nil
       end
     end
   end


app/components/schools/teacher_profile_summary_list_component.html.erb
```diff
@@ -38,7 +38,7 @@
 
   <% list.with_row do |row| %>
     <% row.with_key { "Status" } %>
-    <% row.with_value { ect_status_tag } %>
+    <% row.with_value { ect_status(@ect) } %>
   <% end %>
```
ericaporter: Just a thought - are we going to be adding the ability to change this status to be consistent with the rest of the change details view? Not on this ticket I am sure...

app/views/schools/ects/teacher_leaving_wizard/check_answers.html.erb
```diff
@@ -0,0 +1,26 @@
+<% page_data(
+    title: "Confirm that #{@wizard.teacher_full_name} has left your school permanently",
+    backlink_href: @wizard.previous_step_path
+) %>
+
+<p class="govuk-body">
+  You’ve told us that <%= @wizard.teacher_full_name %> 
+  left your school on <%= @wizard.current_step.leaving_date.to_fs(:govuk) %>.
```
ericaporter: Does this copy need to be updated if the leaving date is in the future?

app/models/concerns/interval.rb
```diff
@@ -63,6 +63,10 @@ def valid_date_order?
     started_on <= finished_on
   end
 
+  def leaving?
+    finished_on.present? && finished_on >= Time.zone.today
```
ericaporter: I notice that we have the method `leaving?` defined in both the induction record and in teacher_status so in different contexts. I think this could be confusing / potentially lead to bugs. Perhaps we could rename the method here to `leaving_today_or_in_future ` or similar which might be more descriptive

app/wizards/schools/ects/teacher_leaving_wizard/check_answers_step.rb
```diff
@@ -0,0 +1,42 @@
...
+        def leaving_date
+          return if leaving_on.blank?
+
+          Schools::Validation::LeavingDate
+            .new(date_as_hash: leaving_on)
+            .value_as_date
+        end
+
+        def save!
+          return false unless leaving_date
```
ericaporter: Can avoid the double parsing by `date = leaving_date` then `return false unless date` etc.

app/views/schools/ects/teacher_leaving_wizard/check_answers.html.erb
avinhurry: I've added logic for this in #1884

PR #1864, opened 2025-12-04
[2863] Update `Statements::Search` to support finding the next output fee statement

leandroalemao:
### Context

Ticket: [2863](https://github.com/DFE-Digital/register-ects-project-board/issues/2863)

We have an existing `Statements::Search` service that we can update to support finding the next output fee statement.

The next output fee statement is defined as the Statement where:

- `fee_type` is output
- `deadline_date` is >= the current time
- ordered by `deadline_date` ascending (taking the earliest)

- Update `Statements::Search` service to support filtering by deadline_date
- Update `Statements::Search` service to support ordering by deadline_date

### DIFF
--- a/app/services/statements/search.rb
+++ b/app/services/statements/search.rb
@@ -60,7 +60,7 @@ module Statements
       return if ignore?(filter: deadline_date)
       return if deadline_date.blank?
 
-      @scope = scope.where(deadline_date: deadline_date..)
+      @scope = scope.where(deadline_date:)
     end
 
     def set_order(order)
--- a/spec/services/statements/search_spec.rb
+++ b/spec/services/statements/search_spec.rb
@@ -184,14 +184,26 @@ RSpec.describe Statements::Search do
       describe "by `deadline_date`" do
         let!(:statement1) { FactoryBot.create(:statement, deadline_date: Date.new(2025, 7, 1)) }
         let!(:statement2) { FactoryBot.create(:statement, deadline_date: Date.new(2025, 9, 1)) }
+        let!(:statement3) { FactoryBot.create(:statement, deadline_date: Time.zone.today) }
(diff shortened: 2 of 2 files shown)


app/services/statements/search.rb
```diff
@@ -50,6 +56,13 @@ def where_statement_date(statement_date)
       @scope = scope.with_statement_date(year:, month:)
     end
 
+    def where_deadline_date(deadline_date)
+      return if ignore?(filter: deadline_date)
+      return if deadline_date.blank?
+
+      @scope = scope.where(deadline_date: deadline_date..)
```
ethax-ross: Instead of making this a range here can we accept the range (or exact date) in case other callers want statements between two dates?

app/services/statements/search.rb
```diff
@@ -50,6 +56,13 @@ def where_statement_date(statement_date)
       @scope = scope.with_statement_date(year:, month:)
     end
 
+    def where_deadline_date(deadline_date)
+      return if ignore?(filter: deadline_date)
+      return if deadline_date.blank?
```
ethax-ross: I think this check is redundant as the `FilterIngorable` concern will treat blank strings as ignored

leandroalemao: Hi @ethax-ross For some reason `FilterIgnorable` wasn't ignoring `nil` values.
I've fixed that now. I hope it's all good. tks

ethax-ross: Ah I'm forgetting `blank` checks for `nil` as well -- I think that's why we have `:ignore` so that you _can_ pass in `nil` and have it filter. I wouldn't change `FilterIgnorable` as its used in a bunch of places in the API as well - sorry, lets maybe just leave it as you had it 😅

PR #1866, opened 2025-12-05
[2564] Update (simplify) school eligibility

rwrrll:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/2564

ECF1 FIP criteria are identical to the agreed criteria linked in the ticket, so the change here removes CIP and effectively makes FIP the only set of eligibility criteria.

In the ticket, Pete suggested extracting an Eligibility service, which I initially explored, but in this instance yielded a particularly anemic service class and increased overall complexity with no clear payoff.

### Guidance to review

Since this completely remove all traces of CIP schools, which I don't think were visible in RECT anyway, I think success means there is no noticeable change when using the application.

It'd be useful to get some particular attention on migrators and related code; I'm very unfamiliar with the migration process in practice, so someone close to that should probably check this looks good.

### DIFF
--- a/app/components/schools/ects/listing_card_component.rb
+++ b/app/components/schools/ects/listing_card_component.rb
@@ -4,12 +4,13 @@ module Schools
       include TeacherHelper
       include ECTHelper
 
-      attr_reader :teacher, :ect_at_school_period, :training_period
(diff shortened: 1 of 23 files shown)


avinhurry (approved):
Looks good. 

Apologies for the delay with the re-review

PR #1877, opened 2025-12-08
Ignore school transfers if a mentor has completed training

cpjmcquillan:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/2924

Before, we only ignored transfers if the teacher had completed their ECT
induction.

We want to ignore transfers when mentors have completed their training, too.

This leans on https://github.com/DFE-Digital/register-early-career-teachers-public/pull/1853 and updates `Teachers::SchoolTransfers::History` to ignore
transfers when any teacher has completed their training.


cpjmcquillan:
Following discussions offline, it turns out this isn't something we want to do. Also, see https://github.com/DFE-Digital/register-ects-project-board/issues/2994

PR #1882, opened 2025-12-08
Fix N+1 on transfers query

ethax-ross:
We aren't preloading the right associations for the `History` service in the corresponding `Query`. 

The `History` service was also calling `earliest_first` on the `school_periods`, which returns a new scope and means we don't get all the `includes` from the original query, resulting in N+1s.

Changing to an in-memory `sort_by` retains the preloading.


leandroalemao (approved):
Looking better.. tks @ethax-ross 👍

PR #1886, opened 2025-12-09
[2870] Update zero-day behaviour for changing schedule

leandroalemao:
### Context

Ticket: [2870](https://github.com/DFE-Digital/register-ects-project-board/issues/2870
)

A `TrainingPeriod` should not be zero-day (have the same `started_on` as `finished_on`). At the moment when we change schedule we create a new `TrainingPeriod`, however this implodes if a lead provider tries to change schedule twice in quick succession (on the same day).

- We update training periods in place on change-schedule if the `started_on` has not yet passed (so today)
- We create a new training period if the `started_on` has passed
- If the `started_on` has not passed and there are declarations that would result in validation failures if we changed the schedule then we surface a new error to lead providers (**new error message created by @tashonrp**)
- We have test coverage for all of these scenarios

### Guidance to review

Review app

### DIFF
--- a/app/services/teachers/change_schedule.rb
+++ b/app/services/teachers/change_schedule.rb
@@ -1,6 +1,6 @@
 module Teachers
   class ChangeSchedule
-    attr_reader :lead_provider, :teacher, :training_period, :schedule, :school_partnership, :author
+    attr_reader :lead_provider, :teacher, :training_period, :schedule, :school_partnership, :author, :original_schedule
 
     def initialize(author:, lead_provider:, teacher:, training_period:, schedule:, school_partnership:)
       @lead_provider = lead_provider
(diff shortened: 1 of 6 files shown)


app/services/teachers/change_schedule.rb
```diff
@@ -13,19 +13,31 @@ def initialize(author:, lead_provider:, teacher:, training_period:, schedule:, s
...
-        finished_on = determine_finished_on
-
         track_payments_frozen_year!
-        finish_training_period!
-        new_training_period = create_new_training_period!(finished_on:)
-        record_change_schedule_event!(original_training_period: training_period, new_training_period:)
+
+        original_schedule = training_period.schedule
+
+        if training_period.started_on >= Time.zone.today
```
ethax-ross: `training_period.started_on.past?`

app/services/teachers/change_schedule.rb
```diff
@@ -9,23 +9,40 @@ def initialize(author:, lead_provider:, teacher:, training_period:, schedule:, s
...
+        end
       end
 
       teacher
     end
 
   private
 
+    def update_training_period_in_place!
+      training_period.update!(schedule:, school_partnership:, expression_of_interest: school_partnership.active_lead_provider)
```
ethax-ross: I think we may only want to set `expression_of_interest` if the training period currently has one

ethax-ross (approved):

PR #1894, opened 2025-12-10
Introduce void and clawback declarations services

cpjmcquillan:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/2848

#### Introduce `API::Declarations::Void` service
In ECF1, the service either voided or clawed back the declaration depending on
whether the declaration was `paid` or not.

We've decided to simplify this by splitting the service into two distinct
services for voiding and clawing back declarations respectively.

This introduces the `Void` service.

Similarly to the `API::Teachers` services, we have split the validation logic
and actions into separate services.

An optional `voided_by_user_id` can be provided so we can track when
declarations are voided by Admin users etc.

#### Introduce `API::Declarations::Clawback` service
This introduces the services to clawback declarations. They follow the same
pattern as the services to void declarations.

Error messages match those in ECF1.

### DIFF
--- a/app/services/api/concerns/declarations/shared_action.rb
+++ b/app/services/api/concerns/declarations/shared_action.rb
@@ -7,7 +7,7 @@ module API::Concerns::Declarations
       include ActiveModel::Attributes
 
       attribute :lead_provider_id
-      attribute :declaration_id
+      attribute :declaration_api_id
 
       validates :lead_provider_id, presence: { message: "Enter a '#/lead_provider_id'." }
       validate :lead_provider_exists
@@ -20,7 +20,7 @@ module API::Concerns::Declarations
     end
 
     def declaration
-      @declaration ||= Declaration.find_by!(api_id: declaration_id)
+      @declaration ||= Declaration.find_by!(api_id: declaration_api_id)
     end
 
     def lead_provider_exists
--- a/app/services/api/declarations/clawback.rb
+++ b/app/services/api/declarations/clawback.rb
@@ -19,21 +19,21 @@ module API::Declarations
   private
 
     def paid
-      return if errors[:declaration_id].any?
+      return if errors[:declaration_api_id].any?
       return if declaration.payment_status_paid?
 
-      errors.add(:declaration_id, "The declaration must be paid before it can be clawed back")
+      errors.add(:declaration_api_id, "The declaration must be paid before it can be clawed back")
     end
 
     def not_already_refunded
-      return if errors[:declaration_id].any?
+      return if errors[:declaration_api_id].any?
       return if declaration.clawback_status_no_clawback?
 
-      errors.add(:declaration_id, "The declaration will or has been refunded")
+      errors.add(:declaration_api_id, "The declaration will or has been refunded")
     end
 
     def output_fee_statement_available
-      return if errors[:declaration_id].any?
+      return if errors[:declaration_api_id].any?
       return if next_available_output_fee_statement.present?
 
       no_output_fee_statement_error_message = <<~TXT.squish
@@ -41,7 +41,7 @@ module API::Declarations
         contract period. The funding contract for this contract period has
         ended. Get in touch if you need to discuss this with us
       TXT
-      errors.add(:declaration_id, no_output_fee_statement_error_message)
+      errors.add(:declaration_api_id, no_output_fee_statement_error_message)
     end
 
     def next_available_output_fee_statement
--- a/app/services/api/declarations/void.rb
+++ b/app/services/api/declarations/void.rb
@@ -14,20 +14,20 @@ module API::Declarations
   private
 
     def not_already_voided
-      return if errors[:declaration_id].any?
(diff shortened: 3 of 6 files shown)


app/services/api/declarations/void.rb
```diff
@@ -0,0 +1,75 @@
...
+    def not_voided
+      return if errors[:declaration_id].any?
+      return unless declaration.payment_status_voided?
+
+      errors.add(:declaration_id, "The declaration has already been voided")
+    end
+
+    def voidable
+      return if errors[:declaration_id].any?
+      return if declaration.payment_status.in?(VOIDABLE_PAYMENT_STATUSES)
```
ethax-ross: If we move the constant to `Declaration` we could have `declaration.voidable_payment?`

app/services/api/declarations/void.rb
```diff
@@ -0,0 +1,75 @@
+module API::Declarations
+  class Void
```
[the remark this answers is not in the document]
cpjmcquillan: Just thinking out loud...

Are we set against having 2 services (`Void` and `Clawback`) and calling the relevant one in the controller depending on whether `declaration.payment_status_paid?` or not?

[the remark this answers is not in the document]
cpjmcquillan: Ok I'll take it for a spin and see how it goes. We can always fallback to this if it becomes a pain 👌🏼

app/services/api/declarations/clawback.rb
```diff
@@ -0,0 +1,62 @@
...
+    end
+
+    def next_available_output_fee_statement
+      @next_available_output_fee_statement ||= Statements::Search
+        .new(
+          lead_provider_id: declaration.training_period.lead_provider.id,
+          contract_period_years: declaration.training_period.contract_period.year,
+          fee_type: "output",
+          deadline_date: Date.current..
+        )
```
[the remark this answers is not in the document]
ethax-ross: Yeah that looks good 👍

app/services/api/declarations/clawback.rb
```diff
@@ -0,0 +1,62 @@
...
+
+      errors.add(:declaration_id, "The declaration must be paid before it can be clawed back")
+    end
+
+    def not_already_refunded
+      return if errors[:declaration_id].any?
+      return if declaration.clawback_status_no_clawback?
+
+      # The typo here is copied from ECF.
+      errors.add(:declaration_id, "The declaration will or has been be refunded")
```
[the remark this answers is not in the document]
ethax-ross: Ah, missed that - given its been in ECF for 3 years lets just fix it in RECT 👍

app/services/api/declarations/clawback.rb
```diff
@@ -0,0 +1,62 @@
...
+    def output_fee_statement_available
+      return if errors[:declaration_id].any?
+      return if next_available_output_fee_statement.present?
+
+      no_output_fee_statement_error_message = <<~TXT.squish
+        You cannot submit or void declarations for the #{contract_period_year}
+        cohort. The funding contract for this cohort has ended. Get in touch
+        if you need to discuss this with us
+      TXT
+      errors.add(:declaration_id, no_output_fee_statement_error_message)
```
ethax-ross: I think we may need to remap `declaration_id` to `participant_declaration` to match ECF as well

app/services/declarations/clawback.rb
```diff
@@ -0,0 +1,47 @@
...
+      @declaration = declaration
+      @voided_by_user_id = voided_by_user_id
+      @next_available_output_fee_statement = next_available_output_fee_statement
+    end
+
+    def clawback
+      ActiveRecord::Base.transaction do
+        if voided_by_user_id
+          declaration.voided_by_user_at = Time.current
+          declaration.voided_by_user = User.find(voided_by_user_id)
```
ethax-ross: `User.find!` or we'll end up with a date and no user, which will be a validation error anyway I guess

ethax-ross: Doh, of course it does. Fine as-is 👍

app/services/api/concerns/declarations/shared_action.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+
+  private
+
+    def lead_provider
+      @lead_provider ||= LeadProvider.find_by(id: lead_provider_id) if lead_provider_id
+    end
+
+    def declaration
+      @declaration ||= Declaration.find_by!(api_id: declaration_id)
```
mooktakim: We need to check if the lead provider owns the declaration.

The old ECF1 service we would only pass in `declaration` to the service if the lead provider owns the declaration:
https://github.com/DFE-Digital/early-careers-framework/blob/main/app/controllers/api/v3/participant_declarations_controller.rb#L102

For RECT we either only pass in verified declaration (and maybe dont need to pass in lead provider) or we check in the service whether the lead provider owns the declaration.

leandroalemao (approved):

PR #1905, opened 2025-12-12
Remove ECT completed check from transfer history

ethax-ross:
Originally we thought that if an ECF has completed their training then their last, closed training period shouldn't be classed as a leaving transfer (of type `unknown`).

We don't believe this is the case any longer; an ECT may transfer schools even after they have completed their training. We also don't take into account completion state of ECTs in the ECF service, so dropping this will help us achieve better parity as we migrate from the service.


cpjmcquillan (approved):
🔥

PR #1908, opened 2025-12-12
Fix mentors not appearing in mentor list - add validation on mentorship_period - Refactor Seeds

ericaporter:
### Context
The original bug is that a **MentorshipPeriod must link a mentor-at-school period and an ECT-at-school period belonging to the same school** N.B. this is not the same as a mentor cannot mentor across schools in general

While investigating an issue with mentorship periods being created across different schools, it was identified that some API seed data (notably for unfunded mentors) was creating cross-school mentorships. 

This led to inconsistent data and downstream issues in API queries that assume mentorships are scoped to a single school via the underlying at-school periods.

- Added `mentor_and_mentee_periods_are_at_same_school ` validation to MentorshipPeriod
- Updated API seed logic to:
- Ensure mentor and mentee at-school periods are selected from the same school
- Avoid creating cross-school mentorship periods in seed data
- Added model specs covering 
       - Valid when mentor and mentee periods belong to the same school
       - Invalid when mentor and mentee periods belong to different schools

### DIFF
--- a/db/seeds/teacher_histories.rb
+++ b/db/seeds/teacher_histories.rb
@@ -183,6 +183,7 @@ ambition_artisan_abbey_grove_2025 = find_or_create_school_partnership!(
 )
 
 # Abbey Grove — Teach First / Grain
+
 teach_first_grain_abbey_grove_2025 = find_or_create_school_partnership!(
   school: abbey_grove_school,
   lead_provider: teach_first,
@@ -190,7 +191,15 @@ teach_first_grain_abbey_grove_2025 = find_or_create_school_partnership!(
   contract_period: cp_2025
 )
 
+teach_first_grain_abbey_grove_2024 = find_or_create_school_partnership!(
+  school: abbey_grove_school,
+  lead_provider: teach_first,
+  delivery_partner: grain_teaching_school_hub,
+  contract_period: cp_2024
+)
+
 # Ackley Bridge — Ambition / Artisan
+
 ambition_artisan_brookfield_2023 = find_or_create_school_partnership!(
   school: brookfield_school,
   lead_provider: ambition_institute,
@@ -207,6 +216,7 @@ teach_first_grain_ackley_bridge_2022 = find_or_create_school_partnership!(
 )
 
 # Mallory Towers — Teach First / Grain
+
 teach_first_grain_mallory_towers_2024 = find_or_create_school_partnership!(
   school: mallory_towers,
   lead_provider: teach_first,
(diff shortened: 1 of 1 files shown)


spec/services/ect_at_school_periods/finish_spec.rb
```diff
@@ -150,6 +151,7 @@
         let(:finished_on) { 1.week.from_now.to_date }
 
         it "leaves the training period finished_on untouched" do
+          subject.finish!
```
ericaporter: This was missing

thenapking (approved):
This is great.  Cleans up the seeds and tests, and has got to the bottom of the issue with the seed data.  One minor comment.

spec/factories/teacher_factory.rb
```diff
@@ -1,13 +1,10 @@
 FactoryBot.define do
   factory(:teacher) do
+    api_id { SecureRandom.uuid }
```
ethax-ross: What was this change needed for, out of interest?

ethax-ross: Ok that makes sense, I think we're fine to have it as the factory default, thanks 👍

spec/factories/gias/school_factory.rb
```diff
@@ -80,7 +76,7 @@
     end
 
     trait(:with_school) do
-      school { build(:school, urn:) }
+      association :school
```
avinhurry: Do we need something like:

```ruby
association :school, urn: urn
```

[Since main has been updated now](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/spec/factories/gias/school_factory.rb#L83).

avinhurry (commented):
Tested locally, mentor list looks good. Just a couple of comments.

app/services/api_seed_data/unfunded_mentors.rb
```diff
@@ -59,11 +59,13 @@ def select_mentor_lp(lead_providers, mentee_lp)
 
     def create_unfunded_mentors(mentee_lp, mentor_lp, mentee_partnerships, mentor_partnerships)
       mentee_partnerships_cycle = mentee_partnerships.cycle
-      mentor_partnerships_cycle = mentor_partnerships.cycle
 
       MIN_UNFUNDED_MENTORS_PER_LP.times do |i|
         mentee_school_partnership = mentee_partnerships_cycle.next
-        mentor_school_partnership = mentor_partnerships_cycle.next
+        mentor_lpdp = mentor_partnerships.first.lead_provider_delivery_partnership
```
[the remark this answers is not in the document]
ericaporter: Good catch - now updated!

PR #1913, opened 2025-12-14
Rename declarations `voided_at` column

cpjmcquillan:
### Context

Broken out of #1894 

We only ever want to set `voided_at` when `voided_by_user_id` is set.

This renames the timestamp to `voided_by_user_at`, which conveys the intention more explicitly.

There isn't any declarations data yet, so we can rename the column safely.

### DIFF
--- a/documentation/domain-model.md
+++ b/documentation/domain-model.md
@@ -96,9 +96,11 @@ erDiagram
     string induction_tutor_name
     citext induction_tutor_email
     uuid api_id
+    integer induction_tutor_last_nominated_in
   }
   School }o--|| AppropriateBody : belongs_to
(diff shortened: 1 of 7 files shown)


documentation/domain-model.md
```diff
@@ -96,9 +96,11 @@ erDiagram
     string induction_tutor_name
     citext induction_tutor_email
     uuid api_id
+    integer induction_tutor_last_nominated_in
   }
   School }o--|| AppropriateBody : belongs_to
   School }o--|| LeadProvider : belongs_to
+  School }o--|| ContractPeriod : belongs_to
```
cpjmcquillan: This change was auto-generated

PR #1922, opened 2025-12-16
[2845] Add create declaration service

leandroalemao:
### Context

Ticket: [2845](https://github.com/DFE-Digital/register-ects-project-board/issues/2845)

We need to be able to submit declarations via the API.

- Add create declaration service

### Guidance to review

Review app

### DIFF
--- a/app/services/api_seed_data/declarations.rb
+++ b/app/services/api_seed_data/declarations.rb
@@ -74,7 +74,7 @@ module APISeedData
           training_period:,
           ineligibility_reason:,
           mentorship_period:,
-          **uplifts
+          **uplifts(training_period:)
         )
 
         next if declaration.duplicate_declaration_exists?
@@ -84,7 +84,9 @@ module APISeedData
       end
     end
 
-    def uplifts
+    def uplifts(training_period:)
+      return {} unless training_period.for_ect?
+
       { sparsity_uplift: Faker::Boolean.boolean, pupil_premium_uplift: Faker::Boolean.boolean }
     end
 


app/models/concerns/interval.rb
```diff
@@ -20,6 +20,23 @@ module Interval
     scope :overlapping_with, ->(period) { where(*overlapping_with_range(period.started_on, period.finished_on)) }
     scope :containing_period, ->(period) { where(*containing_range(period.started_on, period.finished_on)) }
     scope :ongoing_on, ->(date) { where(*date_in_range(date)) }
+    scope :ongoing_on_including, ->(date) { started_on_or_before(date).and(finished_on_or_after(date)) }
```
ethax-ross: Do we need this after the catch up with Pete? or can we just change the behaviour of `ongoing_on` to be inclusive?

app/services/api/declarations/create.rb
```diff
@@ -0,0 +1,142 @@
...
+  class Create
+    include API::Concerns::Teachers::SharedAction
+
+    attribute :declaration_date
+    attribute :declaration_type
+    attribute :evidence_type
+
+    validates :declaration_date, presence: { message: "Enter a '#/declaration_date'." }
+    validates :declaration_date,
+              future_date: true,
```
ethax-ross: This validator has caught me out in ECF before - it looks like the declaration date has to be future dated, but it actually needs to be in the past. Can we just inline this validation into the service? I think unless we expect to reuse the validation elsewhere or its suitably complicated (like the evidence held validator) then I would opt to just inline them in the service to make it easier to follow

app/validators/declaration_date_within_milestone_validator.rb
```diff
@@ -0,0 +1,21 @@
+class DeclarationDateWithinMilestoneValidator < ActiveModel::Validator
+  def validate(record)
+    declaration_within_milestone(record)
+  end
+
+private
+
+  def declaration_within_milestone(record)
```
ethax-ross: I would leave this inlined; its not that complex and we aren't going to reuse it anywhere else

ethax-ross (commented):
Thanks Leandro, I've left some initial thoughts/comments; there's a lot going on here though! could you maybe break this PR up into a few separate commits to make it easier to review?

app/services/api/declarations/create.rb
```diff
@@ -0,0 +1,142 @@
...
+      return false unless valid?
+
+      Declarations::Create.new(
+        author: Events::LeadProviderAPIAuthor.new(lead_provider:),
+        lead_provider:,
+        teacher:,
+        training_period:,
+        declaration_date:,
+        declaration_type:,
+        evidence_type:
```
ethax-ross: Can we clear the evidence type if its not applicable for the declaration; in ECF we do this in the validator:

https://github.com/DFE-Digital/early-careers-framework/blob/main/app/validators/evidence_held_validator.rb#L57

But I think we should do it in this service, Mook added a method for checking if its applicable:

https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/validators/evidence_held_validator.rb#L24-L26

app/services/api/declarations/create.rb
```diff
@@ -0,0 +1,142 @@
...
+    def training_period
+      return unless teacher
+
+      @training_period ||= case teacher_type
+                           when :ect
+                             teacher.ect_training_periods
+                             .includes(:lead_provider)
+                             .where(active_lead_providers: { lead_provider_id: })
+                             .ongoing_or_closest_to(declaration_date)
+                             .earliest_first
```
ethax-ross: Not sure we need `earliest_first`? would that not just re-arrange the results again?

app/services/api/declarations/create.rb
```diff
@@ -0,0 +1,142 @@
...
+    def validate_milestone_exists
+      return if errors[:declaration_type].any?
+      return if errors[:teacher_api_id].any?
+
+      if milestone.blank?
+        errors.add(:declaration_type, "The property '#/declaration_type' does not exist for this schedule.")
+      end
+    end
+
+    def teacher_not_withdrawn
```
ethax-ross: Can we make this `teacher_not_withdrawn_before_declaration_date`

app/validators/declaration_date_within_milestone_validator.rb
leandroalemao: Hi @ethax-ross We have the same validator being used in [declaration model](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/models/declaration.rb#L58).. so that's why I extracted it..

ethax-ross: Ah so it is 👍

spec/models/concerns/interval_spec.rb
```diff
@@ -123,6 +123,26 @@
...
       end
     end
+
+    describe ".closest_to" do
+      it "returns a SQL relation selecting the closest record" do
+        date = Date.new(2023, 12, 10)
+        sql = DummyMentor.closest_to(date).to_sql
+
+        expect(sql).to include("ORDER BY LEAST")
+        expect(sql).to include('ABS("mentor_at_school_periods".started_on - DATE \'2023-12-10\')')
```
ethax-ross: Can we change these tests to use actual data/models? I think asserting on the raw SQL for these quite complex scopes is probably a bit error prone 🤔

app/models/teacher.rb
```diff
@@ -23,6 +23,7 @@ class Teacher < ApplicationRecord
   has_many :appropriate_bodies, through: :induction_periods
   has_many :events
   has_many :mentor_declarations, through: :mentor_training_periods, source: :declarations
+  has_many :ect_declarations, through: :ect_training_periods, source: :declarations
```
ethax-ross: Can we add a spec for this, and maybe update the `mentor_declarations` spec with `source(:declarations)`

spec/models/training_period_spec.rb
```diff
@@ -909,6 +909,37 @@ def will_change_attribute(attribute_to_change:, new_value:)
         expect(relation.includes_values).to include(:school_partnership)
       end
     end
+
+    describe ".closest_to" do
```
ethax-ross: Might be worth having some common setup for these tests and just changing the date in the individual specs; can we make sure we test:

- training period covering the date
- closest by started on
- closest by finished on
- closest by started on when finished_on is `nil`

app/services/api/concerns/teachers/shared_action.rb
```diff
@@ -24,6 +24,17 @@ module SharedAction
       }, allow_blank: true
     end
 
+    def training_period
```
ethax-ross: Can we revert this back to `private` now that we're not using this concern for declarations

app/services/api/declarations/create.rb
```diff
@@ -0,0 +1,191 @@
...
+      return false unless valid?
+
+      Declarations::Create.new(
+        author:,
+        lead_provider:,
+        teacher:,
+        training_period:,
+        declaration_date:,
+        declaration_type:,
+        evidence_type: EvidenceTypeValidator.evidence_type_allowed?(self) ? evidence_type : nil,
```
ethax-ross: My OCD wants this in a `def evidence_type` method to keep this all consistent 😅

app/services/api/declarations/create.rb
```diff
@@ -0,0 +1,191 @@
...
+
+    def existing_declarations
+      @existing_declarations ||= if training_period.for_ect?
+                                   teacher.ect_declarations
+                                 else
+                                   teacher.mentor_declarations
+                                 end
+    end
+
+    def existing_declaration
```
ethax-ross: Can we distinguish this from `existing_declarations`, maybe something like `existing_duplicate_declaration`?

app/models/concerns/interval.rb
```diff
@@ -20,6 +20,21 @@ module Interval
...
+    scope :closest_to, ->(date) {
+      return all if date.blank?
+
+      target_date = date.to_date
+      table = connection.quote_table_name(table_name)
+
+      order(
+        Arel.sql(<<~SQL)
+          LEAST(
+            ABS(#{table}.started_on - DATE '#{target_date}'),
```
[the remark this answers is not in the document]
cpjmcquillan: If the `finished_on` is `null` and the period is "ongoing", wouldn't that mean an ongoing period is "closer" to today or a future date than _any_ "finished" period? 🤔

Or is that assumption incorrect?

spec/models/concerns/interval_spec.rb
```diff
@@ -123,6 +123,26 @@
...
+      end
+    end
+
+    describe ".ongoing_on" do
+      it "builds SQL that queries the range contains the date" do
+        sql = DummyMentor.ongoing_on(Date.new(2023, 2, 15)).to_sql
+
+        expect(sql).to end_with(%{"range" @> date('2023-02-15'))})
+      end
+    end
```
cpjmcquillan: I think these tests would be more helpful if they focused on the behaviour rather than implementation details.

Reading a test that echoes the SQL we can see in the concern isn't that helpful to me, but reading a test that tells a story of what the method does (and why) is really helpful.

What do you think?

```ruby
describe ".closest_to" do
  it "returns the period with closest to the given date" do
    close_mentor = DummyMentor.create(
      teacher_id:,
      school_id:,
      started_on: 1.week.ago,
      finished_on: 1.day.ago
    )
    ongoing_mentor = DummyMentor.create(
      teacher_id:,
      school_id:,
      started_on: 2.weeks.ago,
      finished_on: nil
    )
    distant_mentor = DummyMentor.create(
      teacher_id:,
      school_id:,
      started_on: 1.year.ago,
      finished_on: 6.months.ago
    )
    
    closest_mentor = DummyMentor.closest_to(Date.current)

    expect(closest_mentor).to eq(ongoing_mentor)
  end
end
```

(although just realised this might not actually be what we want)

spec/services/declarations/create_spec.rb
```diff
@@ -0,0 +1,216 @@
...
+
+          context "when teacher is eligible for funding" do
+            before do
+              teacher.update!("#{trainee_type}_first_became_eligible_for_training_at": 1.year.ago)
+            end
+
+            it "marks declaration `payment_status` as eligible" do
+              declaration = create_declaration
+
+              expect(declaration.payment_status_eligible?).to be(true)
```
ethax-ross: Just a suggestion but all of these could be changed to this format, which is more readable IMO:

`expect(declaration).to be_payment_status_eligible`

spec/services/declarations/create_spec.rb
```diff
@@ -0,0 +1,216 @@
...
+
+            expect(declaration.payment_statement).to be_nil
+            if trainee_type == :ect
+              expect(declaration.mentorship_period).to eq(mentorship_period)
+            end
+            expect(declaration.evidence_type).to eq(evidence_type)
+            expect(declaration.payment_status_eligible?).to be(false)
+          end
+
+          if trainee_type == :mentor
```
ethax-ross: FWIW the mentor completion service is safe to call for ECTs as well, so we could drop the check in the calling service and here

app/models/concerns/interval.rb
```diff
@@ -20,6 +20,24 @@ module Interval
     scope :overlapping_with, ->(period) { where(*overlapping_with_range(period.started_on, period.finished_on)) }
     scope :containing_period, ->(period) { where(*containing_range(period.started_on, period.finished_on)) }
     scope :ongoing_on, ->(date) { where(*date_in_range(date)) }
+    scope :closest_to, ->(date) {
```
mooktakim: We should add a comment explaining what this scope does for future devs reading this.

app/services/declarations/create.rb
```diff
@@ -0,0 +1,97 @@
...
+
+        update_uplifts!(declaration)
+        set_eligibility!(declaration)
+        set_payment_statement!(declaration)
+        check_mentor_completion!(declaration)
+        record_declaration_created_event!(declaration)
+
+        declaration
+      end
+    end
```
cpjmcquillan: Do we need to write to the database in each step here, or could we assign attributes and save at the end?

It probably doesn't matter too much, but it felt a little unnecessary (to me) to create a declaration then update it twice in quick succession.

What do you think?

PR #1923, opened 2025-12-16
Add declarations seed data

ethax-ross:
> ⚠️ We need to run this on sandbox after merging

Add API seed data for declarations.

This will seed a random number of declarations for up to 50 participants per lead provider/contract period combination.

The declarations should have sensible declaration dates and a mixture of states (some paid, clawed back, etc).

The script creates around 5,000 declarations for lead providers.

This PR also adds some additional validation into `Declaration` model for issues I came across while building the seed data out.

I've intentionally kept the testing light; trying to test all the combinations of declaration types created would be more effort than its worth.

### DIFF
--- a/app/models/declaration.rb
+++ b/app/models/declaration.rb
@@ -67,6 +67,12 @@ class Declaration < ApplicationRecord
   validate :declaration_date_within_milestone
   validate :mentorship_period_belongs_to_teacher
   validate :contract_period_consistent_across_associations
+  validate :declaration_does_not_already_exist
+  validate :declaration_type_started_or_completed_for_mentor_funding_contract_period
+
+  scope :billable_or_changeable, -> {
+    where(payment_status: BILLABLE_OR_CHANGEABLE_PAYMENT_STATUSES, clawback_status: :no_clawback)
+  }
 
   scope :billable_or_changeable, -> {
     where(payment_status: BILLABLE_OR_CHANGEABLE_PAYMENT_STATUSES, clawback_status: :no_clawback)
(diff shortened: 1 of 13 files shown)


app/services/api_seed_data/declarations.rb
```diff
@@ -0,0 +1,175 @@
...
+
+          create_declarations!(active_lead_provider, training_periods)
+        end
+      end
+    end
+
+  protected
+
+    def plantable?
+      super && Declaration.none?
```
leandroalemao: Hi @ethax-ross There's one declaration being created when running the normal seeds locally.. so this is not running.. Is there anyway we can run both if local?

ethax-ross: The two seeds are totally separate now, so to run locally you need to wipe your DB and then just run the API seeds; they don't depend on any of the dev seeds any more so can run from scratch

app/models/declaration.rb
```diff
@@ -136,6 +142,27 @@ def uplift_paid?
...
   def voidable_payment? = payment_status.in?(VOIDABLE_PAYMENT_STATUSES)
 
+  def teacher
+    training_period&.trainee&.teacher
+  end
+
+  def duplicates_existing_declaration?
+    return unless billable_or_changeable?
+
+    teacher = training_period.trainee.teacher
```
mooktakim: you created the `teacher` method, so won't need this line

spec/models/declaration_spec.rb
```diff
@@ -192,6 +192,77 @@ def will_change_attribute(attribute_to_change:, new_value:)
...
+    let!(:paid_declaration) { FactoryBot.create(:declaration, :paid) }
+    let!(:voided_declaration) { FactoryBot.create(:declaration, :voided) }
+    let!(:ineligible_declaration) { FactoryBot.create(:declaration, :ineligible) }
+    let!(:awaiting_clawback_declaration) { FactoryBot.create(:declaration, :awaiting_clawback) }
+    let!(:clawed_back_declaration) { FactoryBot.create(:declaration, :clawed_back) }
+
+    describe ".billable_or_changeable" do
+      subject { described_class.billable_or_changeable }
+
+      it { is_expected.to include(no_payment_declaration, eligible_declaration, payable_declaration, paid_declaration) }
```
cpjmcquillan: Could we use `contain_exactly` here? I think `include` leaves a little ambiguity (i.e are there other records we haven't explicitly expected?).

What do you think?

```suggestion
      it { is_expected.to contain_exactly(no_payment_declaration, eligible_declaration, payable_declaration, paid_declaration) }
```

ethax-ross: Yep good spot 👍

cpjmcquillan (approved):

PR #1927, opened 2025-12-16
[2730] Change school induction tutor details

thenapking:
### Context

This allows school users to edit their induction tutor details from the induction tutor page.  

Add an update induction tutor wizard which has slightly different validations and wording from the new induction tutor wording.

### Guidance to review

I've tried as much as possible to abstract common elements from the other wizards, but I didn't go quite as far as combining this and the new induction tutor wizard.


app/views/schools/induction_tutor/update_induction_tutor_wizard/edit.html.erb
```diff
@@ -0,0 +1,34 @@
+<% page_data(
+    title: "Tell us your induction tutor",
+    error: @wizard.school.errors.present?
```
avinhurry: Do we want a bank link for the change journey?

avinhurry (commented):
Looks good, just a quick 🧹

PR #1934, opened 2025-12-16
[2991] Only show leaving status to the reporting school

avinhurry:
### Context

Record which school reported a leaving event and surface the "Leaving school" tag only for that school, avoiding leaking the tag to other schools.

- `AddReportedLeavingBySchoolToPeriods` migration
- Wire the finishing flows to set the `reported_leaving_by_school_id` when a school reports a leaving date
- Expose predicates on the period models to check whether a given school reported the leaving and only show the "Leaving school" tag for that school.

### Guidance to review

1) Register an ECT at one school
2) Sign in as a different school and register the same ECT with a future school start date
3) Sign back in as the old school and find the ECT to see the leaving status & content which should not be shown

### DIFF
--- a/config/analytics_blocklist.yml
+++ b/config/analytics_blocklist.yml
@@ -327,6 +327,7 @@
   - ecf_start_induction_record_id
   - ecf_end_induction_record_id
   - email
+  - reported_leaving_by_school_id
   :lead_providers:
   - id
   - ecf_id
@@ -383,6 +384,7 @@
   - email
   - school_reported_appropriate_body_id
   - training_programme
+  - reported_leaving_by_school_id
   :delivery_partners:
   - id
   - name
--- a/db/migrate/20251218165603_add_reported_leaving_by_school_to_periods.rb
+++ b/db/migrate/20251218165603_add_reported_leaving_by_school_to_periods.rb
@@ -1,0 +1,11 @@
+class AddReportedLeavingBySchoolToPeriods < ActiveRecord::Migration[8.0]
+  def change
+    add_column :ect_at_school_periods, :reported_leaving_by_school_id, :bigint
+    add_index :ect_at_school_periods, :reported_leaving_by_school_id
+    add_foreign_key :ect_at_school_periods, :schools, column: :reported_leaving_by_school_id
+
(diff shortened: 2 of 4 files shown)


db/migrate/20251211000000_add_reported_leaving_by_school_to_periods.rb
```diff
@@ -0,0 +1,9 @@
+class AddReportedLeavingBySchoolToPeriods < ActiveRecord::Migration[7.1]
+  def change
+    add_column :ect_at_school_periods, :reported_leaving_by_school_id, :bigint
+    add_index :ect_at_school_periods, :reported_leaving_by_school_id
+
+    add_column :mentor_at_school_periods, :reported_leaving_by_school_id, :bigint
+    add_index :mentor_at_school_periods, :reported_leaving_by_school_id
+  end
```
ericaporter: Do we need to add a foreign key on `:reported_leaving_by_school_id`? - I'm thinking it might protect against invalid or orphaned school IDs

spec/helpers/ect_helper_spec.rb
```diff
@@ -10,11 +10,11 @@
...
     context "when the ECT is leaving" do
       before do
-        ect_at_school_period.update!(finished_on: Time.zone.today + 1.day)
+        ect_at_school_period.update!(finished_on: Time.zone.today + 1.day, reported_leaving_by_school_id: school.id)
       end
 
       it "returns a yellow 'Leaving school' tag" do
-        expect(helper.ect_status(ect_at_school_period)).to have_css("strong.govuk-tag.govuk-tag--yellow", text: "Leaving school")
+        expect(helper.ect_status(ect_at_school_period, current_school: school)).to have_css("strong.govuk-tag.govuk-tag--yellow", text: "Leaving school")
       end
```
ericaporter: You could add in specs to cover 1) if `current_school` isn't passed, it should not show leaving and possibly 2) if a different school is passed it must not show leaving. Probably the first is enough

avinhurry: Good shout - done!

ericaporter (approved):
Looks great! 🎄

PR #1964, opened 2025-12-19
amends Induction Period edit page

abdullaaden77:
### Context
Adds more context and fields to edit Induction Page


spec/requests/appropriate_bodies/induction_periods_spec.rb
```diff
@@ -4,14 +4,14 @@
   let(:appropriate_body) { FactoryBot.create(:appropriate_body) }
   let(:teacher) { FactoryBot.create(:teacher, trs_qts_awarded_on: 1.year.ago) }
   let!(:induction_period) do
-    FactoryBot.create(:induction_period, :ongoing, teacher:, started_on: 9.months.ago)
+    FactoryBot.create(:induction_period, :ongoing, teacher:, started_on: 9.months.ago, finished_on: 8.months.ago, number_of_terms: 2)
```
peterdavidhamilton: I think we want the `:ongoing` trait here. Only admins would edit a closed induction.

PR #1971, opened 2025-12-22
Add ECF1 API data extract scripts

tonyheadford:
### Context

I built these scripts when extracting data from the ECF1 API endpoints so that we could compare the data before and after changes made to ECF1 data.
These rely on `curl` and `jq` being installed locally.

You should have a local ECF1 instance running that is connected to a snapshot DB using `konduit` for example.

You also need the provider API tokens, you can recreate these in a snapshot DB with:
```ruby
 CpdLeadProvider.joins(:lead_provider).to_h do |cpd_lead_provider|
   token = LeadProviderApiToken.create_with_random_token!(
     cpd_lead_provider: cpd_lead_provider
   )

   [cpd_lead_provider.name, token]
 end
```

These tokens need to be added to the appropriate provider target in the `Makefile`. The `PROVIDER_KEY`s used in the `Makefile` are just values I've used locally and won't work anywhere else.

To capture all the data from the `participants`, `declarations`, `transfers` and `unfunded mentors` endpoints for all the providers just run:

```
make
```

Each page of results from an endpoint is stored in it's own file.  Once all the pages have been retrieved a combined file is generated from all the pages and the results are sorted by `id` in that file. This file is named `combined.json`
Having a sorted copy of all the data makes it easier to `diff` across 2 versions of the data:

```
diff -U 10 api_data/ambition/participants/combined.json api_data_previous/ambition/participants/combined.json
```

This removes issues where the same data might be on different pages of different the result sets for example.

You can also target individual providers and endpoints
eg
```
 make ambition participants
```

Each provider and endpoint is in its own directory under `api_data` (by default):
```
api_data
 |
 |-- ambition
 |       |
 |       |-- participants
 |       |-- declarations

etc
```
   
Best to look at the `Makefile` for an idea of the defaults and env vars used.

### DIFF
--- a/app/migration/ecf1_api_scripts/Makefile
+++ b/app/migration/ecf1_api_scripts/Makefile
@@ -1,0 +1,124 @@
+# extract data from ECF1 API into JSON files
+# You'll need to have a local ECF1 app running connected to a snapshot DB
+# and you'll need to have the provider keys or recreated new ones for this
+# purpose:
+#
+# CpdLeadProvider.joins(:lead_provider).to_h do |cpd_lead_provider|
+#   token = LeadProviderApiToken.create_with_random_token!(
+#     cpd_lead_provider: cpd_lead_provider
+#   )
+#
+#   [cpd_lead_provider.name, token]
+# end
+#
+# store the relevant token under each provider target at the bottom
+# of this file
+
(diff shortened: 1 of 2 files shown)


gen-sole:
@peteryates could you please have a look when you get a chance

PR #1975, opened 2025-12-23
[2670] transfer inductions to a new lead school

peterdavidhamilton:
### Context

An TSH AB has changed its lead school but continued to log in as the old school for a time registering inductions incorrectly with the old AB. 

We need to identify and transfer ownership of these inductions and their events, correcting any mention of the old AB's name in the event history. There are some inductions that span the cut-off date and these need to be curtailed and then reclaimed by the new AB. 

- A basic utility that can be reused should this ever occur again
- A one-off script to run the migration

image

### Guidance to review

1. Connect to the snapshot database
2. Run the service `#debug` method or
3. Enable rollback on `InductionPeriods::PartialTransferInductionPeriods.new(from:, to:, on:).call(rollback: true)`
4. Check `./tmp` for CSVs
5. Finally run the script (a retest needs to wait for a new snapshot)


peteryates (commented):
This looks great, easy to follow and code's all super clear.

I think introducing a new `RIAB` module feels a bit odd though, especially as the functionality within is dealing with `InductionPeriods` and we already have `services/induction_periods`.

Would it make more sense for these new classes to be in `app/services/induction_periods/transfers` or similar?

Also, not keen on bringing in an extra dependency just for displaying tabular output. If possible I think we're best dumping the CSV directly and keeping dependencies minimal.

We can always use `column`, [xan](https://github.com/medialab/xan) or similar to handle the formatting when debugging.

PR #1981, opened 2025-12-23
[2847] POST participant-declarations swagger docs

mooktakim:
### Context

Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/2847

* POST `/api/v3/participant-declarations`, following ECF api doc:
  * Pre 2025 (ECTs and mentors)
    * Started/retained/completed/extended
  * Post 2024 ECTs
    * Started/retained/completed/extended
  * Post 2024 Mentors
    * Started/completed

### DIFF
--- a/public/api/docs/v3/swagger.yaml
+++ b/public/api/docs/v3/swagger.yaml
@@ -1639,7 +1639,7 @@ components:
           items:
             "$ref": "#/components/schemas/Declaration"
     DeclarationCreateRequest:
-      description: An participant declaration request
+      description: A participant declaration request
       type: object
       required:
       - data
@@ -1655,9 +1655,10 @@ components:
           type: participant-declaration
           attributes:
             participant_id: 3452b1a6-cbaa-422f-9ca9-40afa28583a2
-            declaration_type: started
+            declaration_type: retained-1
             declaration_date: '2021-05-31T02:21:32.000Z'
             course_identifier: ecf-induction
+            evidence_held: training-event-attended
     DeclarationPre2025DataRequest:
       description: A participant declaration data request for participants in cohort
         2024 and previous years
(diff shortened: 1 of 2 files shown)


claire-hughez:
Is evidence_held missing from the request body? Or do we not include it because it's optional?

The grammar is a bit off in 'correct json data structure required. See API docs for reference' but if it's annoying to change we can leave it, should just be capitalised and have a full stop at the end.

I remember we spoke about idempotent before - do we specify this for the other idempotent endpoints out of interest?

I've left content suggestions in the code on removing references to ECF post 2024 which are in the success criteria of the ticket. I don't think we need to do this for pre-2025, just post-2024. If you could check I've corrected all of them that would be helpful, we shouldn't generally use 'ECF' for post-2024 _stuff_, but in some places we've had to like course-identifiers to avoid breaking changes.

claire-hughez:
This looks fine, but do you mind responding to my points here please - may its this over the API, just feels janky:

>The grammar is a bit off in 'correct json data structure required. See API docs for reference' but if it's annoying to change we can leave it, should just be capitalised and have a full stop at the end.

>I remember we spoke about idempotent before - do we specify this for the other idempotent endpoints out of interest?

mooktakim:
@claire-hughez 
> The grammar is a bit off in 'correct json data structure required. See API docs for reference' but if it's annoying to change we can leave it, should just be capitalised and have a full stop at the end.

I capitalised the text and added full stop:
https://github.com/DFE-Digital/register-early-career-teachers-public/pull/1981/files#diff-ec8f14e974b0dbd4558bef46a739db0b1c329aba6422b37003b817d626d87df6L16

> I remember we spoke about idempotent before - do we specify this for the other idempotent endpoints out of interest?

We don't specify any other endpoints to be idempotent, only declarations.

PR #1989, opened 2026-01-05
[3038] Update API seeds to also include dev seeds

leandroalemao:
### Context

Ticket: [3038](https://github.com.mcas.ms/DFE-Digital/register-ects-project-board/issues/3038)

Now that we no longer run dev seeds on API PRs, however, we are missing the users the dev seeds create so that reviewers can sign in to the backend.

- Run dev seeds after API seeds in CI
- Do not run API seeds for non-API PRs

### Guidance to review

[Review app](https://cpd-ec2-review-1989-web.test.teacherservices.cloud/)

### DIFF
--- a/.github/actions/deploy-environment-to-aks/action.yml
+++ b/.github/actions/deploy-environment-to-aks/action.yml
@@ -103,16 +103,6 @@ runs:
 
         echo "api_author=$api_author" >> $GITHUB_OUTPUT
 
-    - name: Run default seed task
-      if: steps.check_label.outputs.has_label == 'false' && steps.check_author.outputs.api_author == 'false' && inputs.pull-request-number != '' && inputs.environment != 'production'
-      shell: bash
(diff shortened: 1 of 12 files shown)


.github/actions/deploy-environment-to-aks/action.yml
```diff
@@ -132,3 +122,13 @@ runs:
             '
       env:
         PULL_REQUEST_NUMBER: ${{ inputs.pull-request-number }}
+
+    - name: Run default seed task
+      if: steps.check_label.outputs.has_label == 'false' && inputs.pull-request-number != '' && inputs.environment != 'production'
```
ethax-ross: I think we want to run this always now, so we should drop the `steps.check_label.outputs.has_label == 'false'` check? The label lets none-API team members raise a PR with the API seeds if they wanted

PR #1991, opened 2026-01-06
[1990] Allow changing training programme without mentor

avinhurry:
### Context

Changing a training programme when the ECT does not have a mentor assigned currently fails.

This PR makes the mentor optional in `SwitchTraining` and skips creating a mentor training period when absent.

### Guidance to review

1. Use an ECT without a mentor assigned
2. Go to `/school/ects/:id`
3. Start “Change training programme”
4. Continue to check-answers
5. Submit

It should now work.


ericaporter (approved):
Looks good!!

PR #2000, opened 2026-01-08
Rename super admin to user manager

ericaporter:
[Ticket](https://github.com/DFE-Digital/register-ects-project-board/issues/2618) 

_N.B. I kept this minimally backwards compatible and would need to raise another PR after deploying this to remove the remaining methods_

### Context

Previously access to the users section of the admin console was not restricted consistently by role, meaning any admin user could view or access user management pages directly.

This PR aligns the Users section with the existing Finance access controls, ensuring that only users explicitly responsible for managing access can view or interact with it.

As part of this work, the legacy super_admin role has been renamed to user_manager to better reflect its responsibilities.

- Renamed the super_admin role to user_manager (with backwards compatibility preserved during transition).
- Restricted access to the Admin → Users section so that:
- Only users with the user_manager role can see the Users navigation item
- Only users with the user_manager role can access Users pages directly
- Non-authorised users are redirected to the admin landing page with a clear error message.
- Updated the primary admin navigation to:
- Show Users only to user managers
- Continue to restrict Finance visibility based on finance access
- Updated request specs, component specs, and role delegation to reflect the new role name and access rules
- Ensured persona users and session users delegate user_manager? correctly
- Maintained compatibility for any existing references to super_admin? while the rename is rolled out

### Guidance to review
Requirements were updated so that Finance users should also have access to manage users.
The implementation and tests have been updated accordingly

### DIFF
--- a/spec/views/admin/users/show.html.erb_spec.rb
+++ b/spec/views/admin/users/show.html.erb_spec.rb
@@ -38,7 +38,7 @@ describe "admin/users/show.html.erb" do
   context "when the user is a user manager user" do
     let(:role) { :user_manager }
 
-    it "displays the finance role" do
+    it "displays the user manager role" do
       expect(rendered).to have_css("dt", text: "Role")
       expect(rendered).to have_css("dd", text: "User manager")
     end


db/migrate/20260107162705_rename_super_admin_to_user_manager_in_dfe_role_type.rb
```diff
@@ -0,0 +1,15 @@
...
+      RENAME VALUE 'super_admin' TO 'user_manager';
+    SQL
+  end
+
+  def down
+    execute <<~SQL
+      ALTER TYPE dfe_role_type
+      RENAME VALUE 'user_manager' TO 'super_admin';
+    SQL
+  end
```
peteryates: Rails now [supports dealing with enumerated types directly](https://guides.rubyonrails.org/active_record_postgresql.html#enumerated-types) so we should be able to do something like:

```suggestion
  def change
    rename_enum_value :dfe_role_type, from: "super_admin", to: "user_manager"
  end
```

avinhurry (approved):
I think `User#can_manage_users?`  is unused? If so we should remove it.

Also, the enum rename in the same PR could cause a brief blip if app code deploys before the migration runs (code expects `user_manager` but db still has `super_admin`) probably acceptable given it's admin only and not worth splitting out.

Otherwise, looks great.

ericaporter:
@avinhurry thanks for reviewing this - so I've left in a method aliasing super_admin and also left super_admin? as delegate methods hoping to avoid any blips....and to remove later. I'll consider though whether to leave out the enum renaming too!

spec/views/admin/users/show.html.erb_spec.rb
```diff
@@ -35,12 +35,12 @@
     expect(rendered).to have_css("dd", text: "Admin")
   end
 
-  context "when the user is a super admin user" do
-    let(:role) { :super_admin }
+  context "when the user is a user manager user" do
+    let(:role) { :user_manager }
 
     it "displays the finance role" do
```
avinhurry:
```suggestion
    it "displays the User manager role" do
```

Since we've updated the assertion we should update the description too

PR #2007, opened 2026-01-09
[3062] seeded teacher TRS statuses

peterdavidhamilton:
### Context

We have two issues with seeded ECTs because we use a mix of seeded teachers using TRNs found in pre-prod TRS and fake TRNs that were not.

1. If a TRN is not found in TRS then the teacher attributes cannot be synchronised. As a result the status indicator is not updated following a pass or fail outcome for example. This means testing using pre-populated ECTs does not produce the same result as claimed ECTs.
2. If a TRN is found in TRS then upon synchronisation their attributes (name, QTS, ITT) are also refreshed. As a result a seeded teacher's name is replaced. This means testing using these records results in unexpected behaviour and is again inconsistent.

- Update seeded teacher TRNs and make it clearer why we want our fake TRNs to not match any records in TRS. 
- Mimic updating the TRS induction attributes for ~seeded~ any teacher records not found in TRS so they still display correct statuses.
- Flag teachers who receive a `Not Found` from the API but only when using the pre-prod version of TRS and test data. Currently on production it would mistakenly flag deactived TRNs because the `Gone` response has not be released in the API yet. Once it has we can begin distinguishing between the two in all deployments.
- Add scopes to query this new flag
- Exclude flagged teachers from the recurring resync requiring manual intervention to reset them

### Guidance to review

cpd-ec2-review-2007-web test teacherservices cloud_appropriate-body_teachers_closed

- Pass or fail a pre-seeded ECT with a TRN starting with `000000`.
- The status should be displayed correctly. `Passed` or `Failed` respectively.
- The teacher's timeline will include an event that they were missing from a TRS lookup

cpd-ec2-review-2007-web test teacherservices cloud_admin_teachers_371_induction
cpd-ec2-review-2007-web test teacherservices cloud_admin_teachers_371_timeline

### DIFF
--- a/app/jobs/teachers/schedule_trs_sync_job.rb
+++ b/app/jobs/teachers/schedule_trs_sync_job.rb
@@ -5,7 +5,7 @@ module Teachers
     BATCH_SIZE = 1200
 
     def perform
-      teachers = Teacher.active_in_trs.ordered_by_trs_data_last_refreshed_at_nulls_first.limit(BATCH_SIZE)
+      teachers = Teacher.found_in_trs.active_in_trs.ordered_by_trs_data_last_refreshed_at_nulls_first.limit(BATCH_SIZE)
 
       teachers.each_with_index do |teacher, i|
         Teachers::SyncTeacherWithTRSJob.set(wait: (i * 3).seconds).perform_later(teacher:)
--- a/spec/jobs/teachers/schedule_trs_sync_job_spec.rb
+++ b/spec/jobs/teachers/schedule_trs_sync_job_spec.rb
@@ -7,6 +7,7 @@ RSpec.describe Teachers::ScheduleTRSSyncJob, type: :job do
     let!(:teacher4) { FactoryBot.create(:teacher, trs_data_last_refreshed_at: 2.days.ago) }
     let!(:teacher5) { FactoryBot.create(:teacher, trs_data_last_refreshed_at: 1.day.ago) }
     let!(:teacher6) { FactoryBot.create(:teacher, :deactivated_in_trs, trs_data_last_refreshed_at: 0.days.ago) }
+    let!(:teacher7) { FactoryBot.create(:teacher, :not_found_in_trs, trs_data_last_refreshed_at: 0.days.ago) }
 
     it "schedules sync jobs for teachers ordered by trs_data_last_refreshed_at" do
       [


peteryates:
This is what the [fake TRS API client](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/635) was intended for, really.

Perhaps that should be enabled for review apps and we use different seeds for staging which is configured to talk to TRS pre-prod.

app/services/teachers/manage.rb
```diff
@@ -85,14 +81,29 @@ def update_trs_attributes!(trs_qts_status_description:, trs_qts_awarded_on:, trs
...
 
     Teacher.transaction do
       teacher.update!(trs_deactivated: true, trs_data_last_refreshed_at:)
       record_teacher_deactivated_event
     end
   end
 
+  # FIXME: TRS induction values are populated using in-service data to preserve status indicators but we flag the teacher for investigation
+  def mark_teacher_as_not_found!(trs_data_last_refreshed_at:)
+    fail(AlreadyFlagged) if teacher.trs_not_found?
```
avinhurry: Will this raise an error on every subsequent `404`?

peterdavidhamilton: This'll only run where we use seed data. In those deployments we aren't currently running the TRS sync job. That should mean this'll run once for each seeded teacher when we pass or fail them. 

My thinking is once production's TRS API is the same as staging and can identify deactivated accounts, we can let this run in production. When we do I'm not yet sure if we'll encounter lost teachers in the wild still.
We can answer that before we proceeed.

app/services/teachers/manage.rb
```diff
@@ -85,14 +81,29 @@ def update_trs_attributes!(trs_qts_status_description:, trs_qts_awarded_on:, trs
...
+
+    Teacher.transaction do
+      teacher.update!(trs_not_found: true, trs_data_last_refreshed_at:)
+      record_teacher_not_found_event
+
+      trs_induction_status = INDUCTION_OUTCOMES.fetch(teacher.finished_induction_period&.outcome&.to_sym, "InProgress")
```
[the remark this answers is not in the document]
peterdavidhamilton: We should. The `teacher_histories` is only factorying inductions for `InProgress` statuses but that might evolve.

spec/models/teacher_spec.rb
```diff
@@ -415,6 +415,22 @@
         expect(Teacher.active_in_trs.to_sql).to end_with(expected_clause)
       end
     end
+
+    describe ".not_found_in_trs" do
+      it "only includes records where trs_not_found = TRUE" do
+        expected_clause = %("teachers"."trs_not_found" = TRUE)
+
+        expect(Teacher.not_found_in_trs.to_sql).to end_with(expected_clause)
```
craigmdavidson: This is interesting. 

I would normally expect this to be tested by:
- having a fixture of a teacher not marked `:trs_not_found`, and a teacher marked `:trs_not_found`
- executing the query, and ensuring that only the teacher marked `:trs_not_found` is returned.

I wouldn't have thought we would want to know the internals of the scope, only it's resultant behaviour?

peteryates (approved):
Looks good. Thanks for addressing my concerns 🙂

PR #2008, opened 2026-01-09
[3043] Data migration updates for API teacher names

mooktakim:
### Context

Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/3043

* Data migration sets `corrected_name` on Teacher with full name
* TRS sync clears `corrected_name` if it matches TRS name

### DIFF
--- a/spec/migration/teacher_history_converter_spec.rb
+++ b/spec/migration/teacher_history_converter_spec.rb
@@ -2,11 +2,21 @@ describe TeacherHistoryConverter do
   subject { TeacherHistoryConverter.new(ecf1_teacher_history:).convert_to_ecf2! }
 
   describe "Teacher" do
+    describe "setting trs names" do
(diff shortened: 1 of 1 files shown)


spec/migration/teacher_history_converter_spec.rb
```diff
@@ -78,5 +78,16 @@
         end
       end
     end
+
+    describe "setting corrected_name" do
+      let(:ecf1_teacher_history_user) { FactoryBot.build(:ecf1_teacher_history_user, full_name: "Test User") }
+      let(:ecf1_teacher_history) { FactoryBot.build(:ecf1_teacher_history, user: ecf1_teacher_history_user) }
+
+      it "selects the user updated at timestamp" do
```
ethax-ross: > it "sets the `corrected_name` to the full name from ECF"

Is it worth improving the specs around `trs_first_name` and `trs_last_name`, I know its not something you added but it doesn't seem to be covered atm so may be worth some housekeeping while we have a bit of time?

PR #2009, opened 2026-01-09
Disable Knapsack time offset warning locally

peteryates:
When enabled this prints a message/ad (below) when we run tests locally. We only really care about this warning in CI so it's set to be enabled there.

```
========= Knapsack Time Offset Warning ==========
Time offset: 30s
Max allowed node time execution: 07m 20s
Exceeded time: -07m 20s

Global time execution for this CI node is fine.
Happy testing!

Need explanation? See FAQ:
https://docs.knapsackpro.com/ruby/knapsack#faq
=================================================
Read up on the benefits of a dynamic test split with Knapsack Pro Queue Mode:
https://docs.knapsackpro.com/2020/how-to-speed-up-ruby-and-javascript-tests-with-ci-parallelisation

Sign up for Knapsack Pro here:
https://knapsackpro.com
=================================================
```

If you want to run it locally set `ENABLE_KNAPSACK_TIME_OFFSET_WARNING: true` in your .env (see example in .env.example)

We have to run this in `spec_helper.rb`, doing it in `rails_helper.rb` doesn't have any effect. This means we can't rely on having `ActiveModel::Type::Boolean,` so for now omit the line to disable. I'm going to implement an environment variable loader to make this a bit nicer and less repetitive.


ethax-ross (approved):
Thanks Pete, works for me 👍

avinhurry (approved):

PR #2010, opened 2026-01-12
Add release note for declaration endpoints in RECT API

tashonrp:
Added new release note for declaration endpoints in the RECT API, detailing available endpoints and changes to declaration behavior.

### DIFF
--- a/app/views/api/release_notes/release_notes.yml
+++ b/app/views/api/release_notes/release_notes.yml
@@ -22,7 +22,7 @@
 #     **bold** and _italic_ text
 
 - title: Declaration endpoints added, completing the RECT API in sandbox
-  date: 2025-01-16
+  date: 2025-01-19
   tags:
     - sandbox-release
   body: |


app/views/api/release_notes/release_notes.yml
```diff
@@ -21,6 +21,39 @@
 #
 #     **bold** and _italic_ text
 
+- title: Declaration endpoints added, completing the RECT API in the sandbox
```
claire-hughez: sorry if this is dumb but is it 'the' sandbox or just sandbox?

app/views/api/release_notes/release_notes.yml
```diff
@@ -21,6 +21,39 @@
...
+    * `POST /api/v3/participant-declarations`
+    * `GET /api/v3/participant-declarations/{id}`
+    * `PUT /api/v3/participant-declarations/{id}/void`
+
+    ## Changes to declaration behaviour
+
+    We have made the following changes to declaration validation:
+
+    * removed the `has_passed` field, as it was only relevant to NPQ declarations
+    * for participants in the 2025 cohort and later, declarations must be submitted in date order based on `declaration_date`
```
claire-hughez: I think this will confuse lead providers and we need to go in more detail. They don't need to be submitted in date order, the declaration_dates need to be in order based on the declaration type, e.g. a started declaration_date must be before a completed declaration_date.

app/views/api/release_notes/release_notes.yml
```diff
@@ -21,6 +21,41 @@
 #
 #     **bold** and _italic_ text
 
+- title: Declaration endpoints added, completing the RECT API in the sandbox
+  date: 2025-01-16
+  tags:
+    - sandbox-release
+    - api-update
```
ethax-ross: There is no tag for `api-updated`, these are the existing tags - we can add more if you'd like though:

```
"breaking-change" => "red",
"new-feature" => "green",
"bug-fix" => "yellow",
"new-field" => "turquoise",
"data-update" => "blue",
"contract-closure" => "orange",
"removed-field" => "pink",
"production-release" => "purple",
"sandbox-release" => "grey",
```

PR #2011, opened 2026-01-12
[2858] Show API get participant response on admin console

thenapking:
### Context
As an internal DfE user
I need to see how a teacher's record is shown over the API
So that I can understand the information a Lead Provider sees and resolve queries easily

API response

Update TrainingSummaryComponent to add API response

### Guidance to review
I've struggled to get the element with the API data to fill all the way to the right of the container, and after quite some time gave in.  If this is an issue for you when dev reviewing the PR perhaps you can suggest how I might get this to work?

### DIFF
--- a/app/controllers/admin/teachers/trainings_controller.rb
+++ b/app/controllers/admin/teachers/trainings_controller.rb
@@ -7,7 +7,8 @@ module Admin
         @teacher = TeacherPresenter.new(Teacher.find(params[:teacher_id]))
         @navigation_items = helpers.admin_teacher_navigation_items(@teacher, :training)
         @breadcrumbs = teacher_breadcrumbs
-        @ect_training_periods = @teacher.ect_training_periods.includes(ect_at_school_period: :teacher).order(started_on: :desc)
+        @ect_at_school_periods = @teacher.ect_at_school_periods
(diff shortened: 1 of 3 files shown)


app/components/admin/teachers/training_summary_component.rb
```diff
@@ -127,6 +128,41 @@ def training_programme_row
...
+
+      def lead_provider
+        return training_period.lead_provider if confirmed_partnership?
+
+        training_period.expression_of_interest_lead_provider
+      end
+
+      def serialized_teacher
+        API::TeacherSerializer.render(teacher, root: "data", **{ lead_provider_id: lead_provider.id })
+      end
```
avinhurry: Do we need to guard this for EOI only periods? Lead provider metadata might be missing there?

`API::TeacherSerializer` uses `.sole` which can raise.

app/components/admin/teachers/training_summary_component.rb
```diff
@@ -127,6 +128,41 @@ def training_programme_row
...
+
+      def formatted_teacher
+        parsed = JSON.parse(serialized_teacher)
```
avinhurry: `serialized_teacher` is called twice (guard + formatting).

Could we memoize it to avoid duplicate serializer/db work?

PR #2019, opened 2026-01-13
sort economy induction records

ltello:
### Context

[Issue](https://github.com/DFE-Digital/register-ects-project-board/issues/3079)

Migration economy mode: 
  - split a teacher induction record chain based on School, LP, Cohort. 
  - sort each group by end_date present?, created at
  - select the last IR on each group
  - return a list of selected IRs ordered by end_date present?, started_at, created_at

This resulting list is processed in reverse order (given so priority to the induction records starting more recently) to extract at_school_periods, training_periods and eventually mentorship_periods.

### DIFF
--- a/spec/migration/teacher_history_converter/ect/latest_induction_records_spec.rb
+++ b/spec/migration/teacher_history_converter/ect/latest_induction_records_spec.rb
@@ -83,28 +83,28 @@ describe "Latest induction records mode conversion" do
       #                                             ┌────────────────────┐
       #                                             │        A           │
       #                                             └────────────────────┘
-      #                            ┌────────────────────┐
-      #                            │         B          │
-      #                            └────────────────────┘
+      #                            ┌──────────────────────────┐
+      #                            │            B             │
+      #                            └──────────────────────────┘
       # ------------------------------------------------------------------------------------------
       #  To these ECT at school periods:
-      #                       ┌───┐
-      #                       │ A │
-      #                       └───┘
-      #                            ┌────────────────────┐
-      #                            │         B          │
-      #                            └────────────────────┘
+      #                            ┌───────────────┐┌─────────┐
+      #                            │       B       ││   ✂️    │
+      #                            └───────────────┘└─────────┘
+      #                                             ┌────────────────────┐
+      #                                             │         A          │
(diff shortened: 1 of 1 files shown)


spec/migration/teacher_history_converter/calculated_attributes_spec.rb
```diff
@@ -5,6 +5,101 @@ def migration_mode = :latest_induction_records
...
+      FactoryBot.build(:ecf1_teacher_history_induction_record_row,
+                       school: school_20,
+                       training_provider_info: provider_2,
+                       cohort_year: 2022,
+                       start_date: Date.new(2023, 3, 1),
+                       end_date:   Date.new(2023, 1, 1),
+                       created_at: Time.new(2023, 3, 1))
+    end
+
+    let(:ir_32) do
```
peteryates: Would it be clearer if the name described the record? i.e. `ecf1_induction_record_ongoing`?

Then in the assertion below we could see the expected order.

app/migration/teacher_history_converter/ect/latest_induction_records.rb
```diff
@@ -4,12 +4,12 @@ class TeacherHistoryConverter::ECT::LatestInductionRecords
...
 
   def initialize(induction_records)
-    @induction_records = induction_records
+    @induction_records = latest_induction_records(induction_records:)
     @ect_at_school_periods = [] # ECF2TeacherHistory::ECTAtSchoolPeriodRow[]
   end
 
   def ect_at_school_periods
-    induction_records.each_with_index do |induction_record, _i|
+    induction_records.reverse.each_with_index do |induction_record, _i|
```
peteryates: I wonder if we should do the whole economy sorting stuff here rather than before passing it in. Then we remove the need for our test data to be in the 'right' order.

ltello: @peteryates having 
`spec/migration/teacher_history_converter/latest_induction_records_spec.rb`
that tests all the potential cases, do we still need
 `spec/migration/teacher_history_converter/latest_induction_records/*_spec.rb` 
(the 5 initial scenarios to be tested)?

PR #2020, opened 2026-01-13
[2654] Enable schools to contact support

rwrrll:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/2654

- Add a support form for school users
- Automatically send submitted support queries to Zendesk

### DIFF
--- a/spec/features/schools/support_queries/submitting_a_support_query_spec.rb
+++ b/spec/features/schools/support_queries/submitting_a_support_query_spec.rb
@@ -1,0 +1,22 @@
+RSpec.describe "Submitting a support query", :enable_schools_interface do
+  let(:school) { FactoryBot.create(:school) }
+
+  scenario "happy path" do
+    sign_in_as_school_user(school:, first_name: "Jane", last_name: "Smith", email: "jane.smith@example.com")
+
+    page.get_by_role("link", name: "Contact support").click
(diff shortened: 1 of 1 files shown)


app/models/support_query.rb
```diff
@@ -0,0 +1,73 @@
+class SupportQuery < ApplicationRecord
+  validates :name, presence: true
```
thenapking: Shouldn't we link this model to the school and user that raised the query, rather than simply hardcoding in name, email and school name?  Even if we aren't going to display this data in the app, having foreign keys would be useful for SQL queries used in reporting for instance.

spec/requests/schools/support_queries_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+
+      it "renders the success message" do
+        request
+
+        expect(response).to be_successful
+        expect(response.body).to include("Your request for help has been submitted")
+      end
+    end
+
+    context "without a message" do
```
thenapking: There are validations on the SupportQuery model, which means there's a possibility the object could be invalid, and the columns they apply to have not null conditions.  We should test here that any validation failure leads to the new page being re-rendered with errors.  If in fact those items can't be invalid - because the data is all looked up from mandatory fields on other objects, then instead of adding validations, we should link up the object to the data sources.  Then there can be no possibility the code could try to save an invalid object.

PR #2021, opened 2026-01-14
[3083] Prevent duplicate declarations from being created

leandroalemao:
### Context

Ticket: [3083](https://github.com.mcas.ms/DFE-Digital/register-ects-project-board/issues/3083)

In RECT we want to solve duplicated declarations issue using database indexes to prevent these duplicates from being created. When we detect a duplicate would be created we instead want to return the existing declaration.

- Should we drop the duplicate model validation `declaration_does_not_already_exist`? @ethax-ross 
- Add unique index to declarations (billable declarations of the same type);
- Instead of creating a duplicate we return the existing declaration again;

### Guidance to review

Review app, but it requires multiple quick api calls to POST /declarations using same params to bypass the model validation.

### DIFF
--- a/app/services/declarations/create.rb
+++ b/app/services/declarations/create.rb
@@ -36,20 +36,36 @@ module Declarations
     end
 
     def create
-      ActiveRecord::Base.transaction do
-        declaration = create_declaration
+      # LPs making multiple same POST /declarations requests in a very quick succession may sometimes bypass the model validations
+      # So we wrap creation in an advisory lock to dedupe concurrent requests for the same training period and declaration type
+      # And return the existing declaration if found
+      Declaration.with_advisory_lock("lock_#{training_period.id}_#{declaration_type}") do
+        return existing_declaration if existing_declaration.present?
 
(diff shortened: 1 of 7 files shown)


app/services/declarations/create.rb
```diff
@@ -47,6 +47,16 @@ def create
 
         declaration
       end
+    rescue ActiveRecord::RecordNotUnique
```
ethax-ross: I wonder if we need to be more specific around which unique index we're referencing here? I'm thinking if declaration has another unique index that ends up firing when the duplicate declarations one doesn't it would result in the same behaviour 🤔 I'm thinking something like this (not tested, not sure if its even possible!):

```
def create  Declaration.create!(attrs)
  # create...
rescue ActiveRecord::RecordNotUnique => e
  raise unless exception.cause.constraint == "idx_unique_declarations"

  # return existing
end
```

Maybe that's overthinking things though. It's probably fine as-is.

cpjmcquillan (approved):

PR #2025, opened 2026-01-14
[2975] Send ECT registration confirmation emails via ActionMailer

avinhurry:
## Context

We are sending the early career teacher a confirmation email when they are registered. The email has a slight variation depending on whether it's provider-led or school-led.

## Changes proposed

- send ECT registration confirmation emails using ActionMailer + Notify generic template with ERB body templates for school‑led and provider‑led training
- add mailer specs that assert subject/to/body content
- enqueue the email after a successful registration

### provider-led

image

### school-led

image

## Guidance to review

Login and register an ECT using an email address that you have access to. Try for both provider-led and school-led. If you can't be bothered, i've done that for you and posted the emails above.

### DIFF
--- a/spec/services/schools/register_ect_spec.rb
+++ b/spec/services/schools/register_ect_spec.rb
@@ -83,6 +83,14 @@ RSpec.describe Schools::RegisterECT do
           it "raises an exception" do
             expect { service.register! }.to raise_error(ActiveRecord::RecordInvalid)
           end
+
+          it "does not enqueue a confirmation email" do
+            ActiveJob::Base.queue_adapter.enqueued_jobs.clear
+
+            expect { service.register! }.to raise_error(ActiveRecord::RecordInvalid)
+
+            expect(ActiveJob::Base.queue_adapter.enqueued_jobs).to be_empty
+          end
         end
 
         context "when a Teacher record with the same TRN exists and has multiple ect records at different schools" do


peteryates:
Sorry Avin, missed this. We're using the [Register early career teachers](https://www.notifications.service.gov.uk/services/186f4715-1d6b-4350-a137-736321375312) project on Notify - I just invited you.

Instead of the Notify templates we're using ActionMailer and a conventional Rails workflow. This means we can deploy changes to emails without having to manually adjust the templates on Notify in conjunction.

See the OTPMailer for inspiration:

* [mailer](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/mailers/otp_mailer.rb)
* [template](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/views/otp_mailer/otp_code_email.text.erb)

We can embed Notify styles (`^ inset text`) etc in our template but use Erb, and more importantly, write tests.

avinhurry:
> This looks fine. I realise that there will be a follow up ticket for bounce-backs, but I was wondering about other sources of failure. I can't see anything in the code which specifies that this job should be retried, and I can't see any generic configuration for ActiveJob or ActionMailer that would handle retries. I know we aren't recording events, but is it worth adding some sort of timestamp on the period, like `confirmation_sent_at` which the job updates when the email has been successfully sent? With this field if there is any failure which causes a batch of emails not to go out, even if the jobs don't exist it would be possible to requeue the emails.

Thanks @thenapking. As per [Notifys own docs](https://www.notifications.service.gov.uk/using-notify/message-status/email):

> Notify has sent the message to the provider. The provider will try to deliver the message to the recipient for up to 72 hours. Notify is waiting for delivery information.

So Notify will retry delivery once it's accepted. We're not doing explicit app side retries of the API call, so happy to create a card for retries or `confirmation_sent_at`/re-queue mechanism if we want that extra safety?

It's worth mentioning that this `ActionMailer` + `Notify` + `deliver_later` with queue defaults is a standard and proven pattern in DfE rails apps at scale which send easily 1k emails per day.

thenapking:
I haven't had any experience with Notify so wasn't aware of this.  If we've got previous experience with this, and it is reliable, then it sounds like we'd only want to do app retries if it becomes an issue.  Happy to approve this on that basis.

PR #2033, opened 2026-01-16
[3072] Bug: mismatch between mentorship assignments shown on ect page vs mentor page

thenapking:
### Context
If a mentee starting in the future was assigned to a mentor they did not show in the list of assigned mentees.  This bug became apparent in testing transfers between schools - not because the transfer caused it - but because transfers were logged in the future and suddenly it looked like the mentee had disappeared.  The logic to display the current mentor on the ECT details is completely different.

Update the `:currently_assigned_ects` scope on `mentor_at_school_period` to show the correct mentees - this should be the current mentees, the upcoming mentees and ones that are finishing, but not yet finished.

### Guidance to review
I noticed that in the component tests, this code was not functioning as expected:
```
within(".govuk-summary-list__value") do
        expect(page).to have_link(
          "Konohamaru Sarutobi",
          href: schools_ect_path(ect_period_1, back_to_mentor: true, mentor_id: mentor.id)
        )
end
```
This test will pass irrespective of whether we assert presence or absence of the link.  I am unsure exactly why - it must be some brittle aspect of the selector logic.  Therefore I have removed the `within(".govuk-summary-list__value")` selector.

### DIFF
--- a/app/models/mentor_at_school_period.rb
+++ b/app/models/mentor_at_school_period.rb
@@ -9,7 +9,7 @@ class MentorAtSchoolPeriod < ApplicationRecord
   has_many :training_periods, inverse_of: :mentor_at_school_period, dependent: :destroy
   has_many :declarations, through: :training_periods
   has_many :events
-  has_many :currently_assigned_ects,
+  has_many :current_or_future_ects,
            -> { current_or_future.includes(:teacher) },
            through: :mentorship_periods,
            source: :mentee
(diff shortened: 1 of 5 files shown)


app/models/mentor_at_school_period.rb
```diff
@@ -10,7 +10,7 @@ class MentorAtSchoolPeriod < ApplicationRecord
   has_many :declarations, through: :training_periods
   has_many :events
   has_many :currently_assigned_ects,
```
ericaporter: I'm wondering do we need to rename this association now since it contains future mentees?

PR #2034, opened 2026-01-16
Rename "trainee" to "at_school_period" on TrainingPeriod

craigmdavidson:
resolves https://github.com.mcas.ms/DFE-Digital/register-early-career-teachers-public/issues/1762

### Context

The `TrainingPeriond#trainee` method is not ideally named as the word "trainee" is normally used to refer to a teacher, however the `trainee` method returns either an `ect_at_school_period` or a `mentor_at_school_period`. 

This PR renames `trainee` to `at_school_period`.

Additionally, we since most of the references to `trainee` within the codebase are only calling `trainee` to get to `teacher` or `school` we provide additional delegate convenience methods. While [Demeter](https://en.wikipedia.org/wiki/Law_of_Demeter) is out of fashion, the cost/benefit is strongly in favour of encapsulation in this case. 

### Guidance to review

The most interesting stuff here is in the TrainingPeriod class.
Most of the rest of changes here are replacing things like `training_period.trainee.teacher` with `training_period.teacher`.

### DIFF
--- a/app/models/training_period.rb
+++ b/app/models/training_period.rb
@@ -45,7 +45,7 @@ class TrainingPeriod < ApplicationRecord
   has_many :events
 
   touch -> { self }, when_changing: %i[started_on finished_on], timestamp_attribute: :api_transfer_updated_at
-  touch -> { trainee&.teacher },
+  touch -> { teacher },
         on_event: %i[create destroy update],
         timestamp_attribute: :api_updated_at,
         when_changing: %i[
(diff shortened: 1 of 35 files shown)


app/models/training_period.rb
```diff
@@ -105,6 +105,8 @@ class TrainingPeriod < ApplicationRecord
   # Delegations
   delegate :name, to: :delivery_partner, prefix: true, allow_nil: true
   delegate :name, to: :lead_provider, prefix: true, allow_nil: true
+  delegate :teacher, :teacher_id, :school, :school_id, :mentorship_periods, :email, to: :at_school_period, allow_nil: true
+  delegate :started_on, :finished_on, to: :at_school_period, prefix: true, allow_nil: true
```
craigmdavidson: We need a prefix on these for disambiguation as both the `at_school_period` and self (TrainingPeriod) are both Intervals.

peteryates: Is the delegation worth adding? `at_school_period_started_on` vs  `at_school_period.started_on`.

craigmdavidson: I would say yes. 

It provides a cheap immediate convenience to TrainingPeriod, that someone previously felt warranted private methods, but it also provides an affordance to help preserve encapsulation external to **TrainingPeriod**, 

We notice that by providing a delegate `at_school_period_finished_on` we already used in multiple other places, to remove knowledge about the **TrainingPeriod**. 

Having `at_school_period_started_on` provides a symmetry to `at_school_period_finished_on`, and cheap affordance optionality to anything external to avoid extensive method chaining. 

We could make it private until there is at least one external caller, but as it is a simple delegate (rather than something that mutates) generally I wouldn't see benefit in that? 

What do you think?

PR #2057, opened 2026-01-21
[3094] Add seed data for participants in sandbox

leandroalemao:
### Context

Ticket: [3094](https://github.com.mcas.ms/DFE-Digital/register-ects-project-board/issues/3094)

We want to add seed data that we predict will be useful for lead providers to help them test in Sandbox.

Add for all lead providers in Sandbox:

-  participants in each contract period
- in 2023 and 2024, some participants that are new to the lead provider (so the school would be EOI = true) without a partnership that exists yet
- at least 5 ECTs with an induction_start_date
- at least 5 ECTs without an induction_start_date
- at least 2 ECTs and 2 mentors in each contract period from 2021-2025 with a range of billable declarations
- at least 5 participants with training status "leaving"

### Guidance to review

Review app.

And to be run in sandbox:
```
Rails.logger.silence do
    DeclarativeUpdates.skip do
      APISeedData::ParticipantScenarios.new.plant
    end
  end
end
```

### DIFF
--- a/spec/factories/ect_at_school_period_factory.rb
+++ b/spec/factories/ect_at_school_period_factory.rb
@@ -12,7 +12,7 @@ FactoryBot.define do
       end_date { (started_on || start_date) + rand(6.months..1.year) }
     end
 
-    teacher { association :teacher, api_ect_training_record_id: SecureRandom.uuid }
+    teacher { association :teacher, :with_realistic_name, api_ect_training_record_id: SecureRandom.uuid }
 
     after(:create) do |ect_at_school_period|
       teacher = ect_at_school_period.teacher
--- a/spec/factories/mentor_at_school_period_factory.rb
+++ b/spec/factories/mentor_at_school_period_factory.rb
@@ -13,7 +13,7 @@ FactoryBot.define do
     end
 
     association :school
-    teacher { association :teacher, api_mentor_training_record_id: SecureRandom.uuid }
+    teacher { association :teacher, :with_realistic_name, api_mentor_training_record_id: SecureRandom.uuid }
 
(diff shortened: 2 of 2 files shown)


app/services/api_seed_data/participant_scenarios.rb
```diff
@@ -0,0 +1,505 @@
...
+    def find_contract_period(year)
+      ContractPeriod.find_by(year:)
+    end
+
+    def find_or_create_active_lead_provider(lead_provider:, contract_period:)
+      ActiveLeadProvider.find_by(lead_provider:, contract_period:) ||
+        FactoryBot.create(:active_lead_provider, lead_provider:, contract_period:)
+    end
+
+    def random_appropriate_body
```
cpjmcquillan: Is it worth highlighting in the method name here that this can create an appropriate body too?

app/services/api_seed_data/participant_scenarios.rb
```diff
@@ -0,0 +1,505 @@
...
+        contract_period = find_contract_period(contract_period_year)
+
+        next unless contract_period
+
+        LeadProvider.find_each do |lead_provider|
+          active_lead_provider = find_or_create_active_lead_provider(lead_provider:, contract_period:)
+
+          existing_count = Teacher
+            .joins(ect_at_school_periods: { training_periods: [:schedule, { school_partnership: { lead_provider_delivery_partnership: :active_lead_provider } }] })
+            .left_joins(:induction_periods)
```
ethax-ross: We could just `join` here and avoid the `.where(induction_periods: { id: nil })`

PR #2086, opened 2026-01-27
Add glossary page to API guidance

ethax-ross:
Add glossary to API guidance section


app/views/api/guidance/guidance_for_lead_providers/glossary.md
```diff
@@ -0,0 +1,149 @@
...
+- not be exempt from induction
+- not have started induction before 1 September 2021
+- not have completed statutory induction before
+
+To be eligible for funding, a mentor must:
+
+- not be prohibited from teaching
+- not have completed early career teacher mentor training before
+- not have been marked as ineligible for future training by DfE, for example, because they started mentor training but took too long to complete it
+
```
claire-hughez:
```suggestion

## Expression of interest

This is a term used over the API when a school has indicated they want to train one or more of their ECTs and mentors with you. Lead providers can then submit a partnership, if one doesn't already exist,  to confirm they're training at least one participant at that school.

```

PR #2094, opened 2026-01-28
Add 'ECT Records you need to review' section to the Authorised Body landing page

craigmdavidson:
### Context

This PR implements the UI changes outlined in https://github.com.mcas.ms/DFE-Digital/register-ects-project-board/issues/3158 for the **AppropriateBody** landing page.

**This is the new landing page as specified:**
image

**This is the new landing page as built:**
as-built

1. The ECT record review panel should be restricted by a feature flag so we can withhold it until we are ready. This  flag is enabled in all environments except production.
2. At the top of the page, the claim ECT links are separated via sub-headings into individual management and bulk CSV actions. This incudes making the individual management a primary CTA button, and shortening the links of the CSV actions. The subordinate page titles should remain unchanged.
3. The _**"Check data from schools"**_ ECT record review panel uses a clickable "Action" card. This ActionCard will be used in multiple **AuthorisedBody** pages (e.g. https://github.com.mcas.ms/DFE-Digital/register-ects-project-board/issues/3160). This was adapted from the existing `app-card--{color}` structure, used in **admin** pages, but it is clickable, so we add a hover effect to make that clear. _Note:_ Due to the colour combination required (grey background, blue heading, black text), and the constancy of planned use, we do not add any future flexibility around colour. 

### Guidance to review

1. **Review in commit order** - the commits are relatively small and self contained.
2. The environment variables enable the feature everywhere except production.
3. The scope for the query to calculate the number of "ECT induction records to review" is non-trivial, as such the spec deviates from the current pattern of asserting the query structure to test the behaviour of the query is as expected.
4. This query uses outer joins rather than sub-selects, this allows us to express the query via a `has_many` relationship on **AuthorisedBody**
5. The execution plan for query to calculate the number of "ECT induction records to review" has been added as a PR comment. 
6. The **ActionCard** currently links to '#' - this will be amended in the PR for the subordinate page.
7. When there are zero "ECT induction records to review", this shows as just that (no adjustments), and continues to link to the subordinate page.
8. The **ActionCardComponent** and **CardContainerComponent** will be used in subordinate pages. The **CardContainerComponent** _could_ be retroactively introduced into existing app-cards. However, this is outside of the scope of this PR as this would also involve extracting an _**AppCardComponent**_.

### DIFF
--- a/app/components/teachers_index/review_section_component.html.erb
+++ b/app/components/teachers_index/review_section_component.html.erb
@@ -1,0 +1,28 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-full">
+    <h2 class="govuk-heading-m">
+      Check data from schools
+    </h2>
+    <p class="govuk-body">
+      When a school tells us you are their AB, 
+      the ECT induction record will show here. 
+      You can check these records and claim inductions, 
+      or take action if any information is not correct.
+    </p>
+    
+    <% if number_of_ect_records_to_review > 0 %>
+      <%= govuk_button_link_to "Review #{number_of_ect_records_to_review} ECT induction #{"record".pluralize(number_of_ect_records_to_review)}", "#" %>  
+    <% else %>
+      <p class="govuk-body govuk-inset-text">
+        You have no ECT induction records to review. 
+      </p>
+    <% end %>
+  </div>
+</div>
+
+
+<div class="govuk-grid-row govuk-!-margin-bottom-4">
+  <div class="govuk-grid-column-full">
+    <hr />
+  </div>
+</div>
--- a/app/models/ect_at_school_period.rb
+++ b/app/models/ect_at_school_period.rb
@@ -65,6 +65,17 @@ class ECTAtSchoolPeriod < ApplicationRecord
     with_expressions_of_interest_for_contract_period(year)
     .where(expression_of_interest: { lead_provider_id: })
   }
+  scope :unclaimed_by_school_reported_appropriate_body, -> {
+    current_or_future
+      .joins(:teacher)
+      .joins(<<~SQL)
+        LEFT OUTER JOIN induction_periods
+          ON induction_periods.teacher_id = ect_at_school_periods.teacher_id
+          AND induction_periods.finished_on IS NULL
+          AND induction_periods.appropriate_body_id = ect_at_school_periods.school_reported_appropriate_body_id
+      SQL
+      .where(induction_periods: { id: nil })
+  }
 
   def reported_leaving_by?(school)
     reported_leaving_by_school_id.present? && reported_leaving_by_school_id == school&.id
--- a/app/components/teachers_index/bulk_upload_links_component.html.erb
+++ b/app/components/teachers_index/bulk_upload_links_component.html.erb
@@ -1,13 +1,19 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-full">
+    <h2 class="govuk-heading-s">
+      Manage ECTs
+    </h2>    
     <p>
-      <%= govuk_link_to("Find and claim a new ECT", new_ab_claim_an_ect_find_path, no_visited_state: true) %>
+    <%= govuk_link_to "Find and claim a new ECT", new_ab_claim_an_ect_find_path, no_visited_state: true %>
+    </p>
+    <h2 class="govuk-heading-s">
+      Upload a CSV
+    </h2>
+    <p>
+      <%= govuk_link_to("Claim multiple ECTs", batch_claim_path, no_visited_state: true) %>
     </p>
     <p>
-      <%= govuk_link_to("Upload a CSV to record outcomes for multiple ECTs", batch_action_path, no_visited_state: true) %>
-    </p>
-    <p>
-      <%= govuk_link_to("Upload a CSV to claim multiple new ECTs", batch_claim_path, no_visited_state: true) %>
+      <%= govuk_link_to("Record outcomes for multiple ECTs", batch_action_path, no_visited_state: true) %>
     </p>
     <hr class="govuk-section-break govuk-section-break--l govuk-section-break--visible">
   </div>
--- a/app/components/teachers_index/review_section_component.rb
+++ b/app/components/teachers_index/review_section_component.rb
@@ -1,0 +1,21 @@
+module TeachersIndex
+  class ReviewSectionComponent < ApplicationComponent
+    include Rails.application.routes.url_helpers
+
+    def initialize(appropriate_body:)
+      @appropriate_body = appropriate_body
+    end
+
+    def render?
+      Rails.application.config.enable_appropriate_body_records_to_review
+    end
+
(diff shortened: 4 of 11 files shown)


app/models/teacher.rb
```diff
@@ -121,6 +121,16 @@ class Teacher < ApplicationRecord
...
+  scope :ects_with_school_reported_appropriate_body, ->(appropriate_body) {
+    joins(:ect_at_school_periods)
+      .merge(ECTAtSchoolPeriod.current_or_future)
+      .where(ect_at_school_periods: { school_reported_appropriate_body: appropriate_body })
+  }
+  scope :ects_with_school_reported_appropriate_body_but_unclaimed_by_that_appropriate_body, ->(appropriate_body) {
+    ects_with_school_reported_appropriate_body(appropriate_body)
+      .where.not(id: InductionPeriod.ongoing.where(appropriate_body:).select(:teacher_id))
+      .where.not(trn: PendingInductionSubmission.claimed.where(appropriate_body:).select(:trn))
+  }
```
craigmdavidson: Reviewing execution plan, the query first filters ect_at_school_periods by school_reported_appropriate_body_id using an index, then joins with teachers. Two subqueries exclude teachers: 
- one checks induction_periods for ongoing records with the same appropriate body (using an index on appropriate_body_id), and   
- another checks pending_induction_submissions for claimed records (also using an index on appropriate_body_id). 

**All three tables leverage indexes, making this query efficient even with large datasets.**


```
Aggregate  (cost=31.58..31.59 rows=1 width=8) (actual time=0.010..0.011 rows=1 loops=1)                                                          
    ->  Nested Loop  (cost=23.61..31.57 rows=1 width=0) (actual time=0.009..0.009 rows=0 loops=1)                                                  
          Join Filter: (teachers.id = ect_at_school_periods.teacher_id)                                                                            
          ->  Bitmap Heap Scan on ect_at_school_periods  (cost=4.16..9.51 rows=1 width=8) (actual time=0.009..0.009 rows=0 loops=1)                
                Recheck Cond: (school_reported_appropriate_body_id = 1)                                                                            
                Filter: ((range @> '2026-01-29'::date) OR (started_on >= '2026-01-30'::date))                                                      
                ->  Bitmap Index Scan on idx_on_school_reported_appropriate_body_id_01f5ffc90a  (cost=0.00..4.16 rows=2 width=0) (actual           
  time=0.009..0.009 rows=0 loops=1)                                                                                                                
                      Index Cond: (school_reported_appropriate_body_id = 1)                                                                        
          ->  Seq Scan on teachers  (cost=19.45..21.96 rows=8 width=8) (never executed)                                                            
                Filter: ((NOT (hashed SubPlan 1)) AND (NOT (hashed SubPlan 2)))                                                                    
                SubPlan 1                                                                                                                          
                  ->  Bitmap Heap Scan on induction_periods  (cost=4.17..11.28 rows=1 width=8) (never executed)                                    
                        Recheck Cond: (appropriate_body_id = 1)                                                                                    
                        Filter: (finished_on IS NULL)                                                                                              
                        ->  Bitmap Index Scan on index_induction_periods_on_appropriate_body_id  (cost=0.00..4.17 rows=3 width=0) (never executed) 
                              Index Cond: (appropriate_body_id = 1)                                                                                
                SubPlan 2                                                                                                                          
                  ->  Index Scan using index_pending_induction_submissions_on_appropriate_body_id on pending_induction_submissions                 
  (cost=0.14..8.16 rows=1 width=32) (never executed)                                                                                               
                        Index Cond: (appropriate_body_id = 1)                                                                                      
                        Filter: ((finished_on IS NULL) AND ((started_on IS NOT NULL) OR (delete_at IS NOT NULL)) AND (error_messages =             
  '{}'::character varying[]))                                                                                                                      
  Planning Time: 3.057 ms                                                                                                                          
  Execution Time: 0.051 ms           
```

app/components/teachers_index/review_section_component.html.erb
```diff
@@ -0,0 +1,27 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-full">
+    <h2 class="govuk-heading-m">
+      Check data from schools
+    </h2>
+    <div class="govuk-body">
```
peterdavidhamilton: Can this be a `p`?

app/models/ect_at_school_period.rb
```diff
@@ -65,6 +65,27 @@ class ECTAtSchoolPeriod < ApplicationRecord
...
+          ON pending_induction_submissions.trn = teachers.trn
+          AND pending_induction_submissions.appropriate_body_id = ect_at_school_periods.school_reported_appropriate_body_id
+          AND pending_induction_submissions.error_messages = '{}'
+          AND pending_induction_submissions.finished_on IS NULL
+          AND pending_induction_submissions.started_on IS NOT NULL
+          AND pending_induction_submissions.delete_at IS NOT NULL
+      SQL
+      .where(induction_periods: { id: nil })
+      .where(pending_induction_submissions: { id: nil })
+  }
```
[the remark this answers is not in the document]
peterdavidhamilton: I'll pull this down locally and review later. I'm guessing Pete might like to introduce an object to encapsulate this.

app/components/teachers_index/review_section_component.html.erb
craigmdavidson: ✅ @peterdavidhamilton

peteryates:
Ah, yeah that's fair.

I think we'd want to put in something to rework both sets a little after.

Just looking at the original batch cards they likely won't work too well for screenreaders:

https://github.com/DFE-Digital/register-early-career-teachers-public/blob/949fb181ef30e324626d13e75ecdbc490d06c733/app/components/admin/appropriate_bodies/batches/batch_cards_component.html.erb#L3-L10

Here the header will be announced without context. My name's on it last too (but I was just removing some outer tags! :sweat_smile:) - should probably be something more like:


```erb
    <div class="app-card--red">
      <h3 class="govuk-heading-l">
        <%= batch.tally[:errored_count] %>

        <div class="app-card--caption">
          <%= pluralize(batch.tally[:errored_count], 'error').split.last %>
        </div>
      </h3>
    </div>
```

app/components/teachers_index/review_section_component.html.erb
```diff
@@ -0,0 +1,27 @@
...
+  <%= render AppropriateBodies::ActionCardComponent.new(
+    url: "#",
+    heading: number_of_ect_records_to_review,
+    body: "ECT induction #{"record".pluralize(number_of_ect_records_to_review)} to review"
+  ) %>
+<% end %>
+
+<div class="govuk-grid-row govuk-!-margin-bottom-4">
+  <div class="govuk-grid-column-full">
+    <hr />
```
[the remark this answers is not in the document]
peterdavidhamilton: https://github.com/x-govuk/govuk-components/blob/main/app/helpers/govuk_components_helper.rb#L17

I assume it still works, there was at least one instance of it in this repo last week. Someone might have removed a template file.

PR #2126, opened 2026-01-30
[3213] Refresh metadata when training programme changes to school-led

leandroalemao:
### Context

Ticket: [3213](https://github.com.mcas.ms/DFE-Digital/register-ects-project-board/issues/3213)

When an ECT's `training_period` is removed, the `expression_of_interest` for the school does not turn `FALSE` on GET schools.

- When the training programme changes to `school-led`, a training period is deleted.. so we need to refresh the school metadata;

### Guidance to review

Review app

### DIFF
--- a/app/models/training_period.rb
+++ b/app/models/training_period.rb
@@ -61,7 +61,7 @@ class TrainingPeriod < ApplicationRecord
           school_partnership_id
         ]
 
-  refresh_metadata -> { school_partnership&.school }, on_event: %i[create destroy update], when_changing: %i[school_partnership_id expression_of_interest_id]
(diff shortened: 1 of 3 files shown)


app/models/training_period.rb
```diff
@@ -61,8 +61,9 @@ class TrainingPeriod < ApplicationRecord
           school_partnership_id
         ]
 
-  refresh_metadata -> { school_partnership&.school }, on_event: %i[create destroy update], when_changing: %i[school_partnership_id expression_of_interest_id]
```
ethax-ross: Why are we removing the update to `school_partnership`?

PR #2129, opened 2026-02-02
[3168] create realistic training started on dates

mooktakim:
### Context

Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/3168
Rebased off: https://github.com/DFE-Digital/register-early-career-teachers-public/pull/2087

* Most of the training starts in September 1 - 30 on the contract period year
* 5% will start late, Oct - Jan
* Set finish date to 200-300 days after starting
* In the spec replaced `Faker::Boolean.boolean` with constants so we can easily test different scenarios


ethax-ross (approved):
Looks good, thanks Mook. Can we have a script to sort the training/school periods in sandbox as well please

PR #2132, opened 2026-02-02
[3134] Block mentor registrations in contract periods that are not open

thenapking:
### Context

When registering a mentor we currently have validation to block a start date more than 4 months in advance. This validation should stay in place.  However we need to block registrations in contract periods that have not opened yet.  Nothing different should happen if the contract period is open.

AC: If the start date is in the future and the contract period is not yet open, user will be shown the screen to say they can't be registered yet, with content as per the screenshot above.

Update the register mentor wizard to add a screen saying the contract period is not yet open.

### DIFF
--- a/spec/wizards/schools/register_mentor_wizard/started_on_step_spec.rb
+++ b/spec/wizards/schools/register_mentor_wizard/started_on_step_spec.rb
@@ -10,19 +10,22 @@ RSpec.describe Schools::RegisterMentorWizard::StartedOnStep do
   end
 
   let(:mentoring_only) { "no" }
-  let(:ineligible) { false }
-  let(:provider_led) { true }
   let(:started_on) { { "day" => "10", "month" => "9", "year" => "2025" } }
-  let(:previous_training_period) { FactoryBot.build(:training_period) }
   let(:contract_period) { FactoryBot.create(:contract_period, year: 2025, enabled: true) }
-  let(:contract_period_enabled?) { contract_period.enabled }
-  let(:currently_mentor_at_another_school) { false }
 
   it_behaves_like "a started on step", current_step: :started_on
 
   describe "#next_step" do
+    let(:contract_period_enabled?) { contract_period.enabled }
+    let(:ineligible) { false }
+    let(:provider_led) { true }
+    let(:previous_training_period) { FactoryBot.build(:training_period) }
+
     before do
-      allow(wizard.mentor).to receive_messages(provider_led_ect?: provider_led, became_ineligible_for_funding?: ineligible, previous_training_period:, contract_period_enabled?: contract_period_enabled?)
+      allow(wizard.mentor).to receive_messages(provider_led_ect?: provider_led,
+                                               became_ineligible_for_funding?: ineligible,
+                                               previous_training_period:,
+                                               contract_period_enabled?: contract_period_enabled?)
     end
 
     context "when mentor is ineligible for funding" do
--- a/spec/features/schools/mentors/register_mentor/edge_cases/changing_start_date_spec.rb
+++ b/spec/features/schools/mentors/register_mentor/edge_cases/changing_start_date_spec.rb
@@ -3,7 +3,7 @@ RSpec.describe "Registering a mentor", :enable_schools_interface, :js do
   include SchoolPartnershipHelpers
 
   around do |example|
-    travel_to(Time.zone.local(2026, 2, 3)) do
+    travel_to(Time.zone.local(2025, 5, 30)) do
       example.run
     end
   end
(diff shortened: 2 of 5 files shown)


app/views/schools/register_mentor_wizard/cannot_register_mentor_yet.html.erb
```diff
@@ -0,0 +1,10 @@
+<% page_data(
+  title: "You cannot register #{@mentor.full_name} yet", 
+  error: false,
+  backlink_href: @wizard.previous_step_path,
+) %>
+
+<p class="govuk-body">Registration for the <%= formatted_year_range_for_registration_date(@mentor.started_on.to_date) %> academic year is not open.</p>
+<p class="govuk-body">We'll email you to let you know when registration will open.  This is usually in June.</p>
```
avinhurry:
```suggestion
<p class="govuk-body">We’ll email you to let you know when registration will open. This is usually in June.</p>
```

app/wizards/schools/register_mentor_wizard/started_on_step.rb
```diff
@@ -63,6 +65,14 @@ def started_on_within_4_months
         end
       end
 
+      def contract_period_enabled?
+        contract_period = ContractPeriod.containing_date(started_on_as_date)
```
avinhurry: Should this gate be future only? What if the user enters a past date in a disabled/missing contract period? Then the page wouldn't make sense.

Probably worth adding some specs to cover that scenario too.

spec/wizards/schools/register_mentor_wizard/started_on_step_spec.rb
```diff
@@ -83,6 +84,18 @@
 
       it { expect(step.next_step).to eq(:programme_choices) }
     end
+
+    context "when contract period is not open" do
+      let!(:contract_period) { FactoryBot.create(:contract_period, year: 2025, enabled: false) }
```
avinhurry: This spec uses a past date. The gate should be future only (per my earlier comment). We can keep this test with updated expectations and add a future date case. Freezing time would make it clearer and prevent it from becoming outdated.

app/views/schools/register_mentor_wizard/cannot_register_mentor_yet.html.erb
```diff
@@ -0,0 +1,10 @@
+<% page_data(
+  title: "You cannot register #{@mentor.full_name} yet", 
+  error: false,
+  backlink_href: @wizard.previous_step_path,
```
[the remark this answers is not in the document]
thenapking: I will double check, but I thought it was in the designs.  It allows you to go back if you entered the wrong date by accident without then having to go through the whole wizard again, which would be pretty onerous.

app/wizards/schools/register_mentor_wizard/started_on_step.rb
thenapking: This is now updated so it only applies to the future,

app/wizards/schools/register_mentor_wizard/wizard.rb
```diff
@@ -86,6 +87,12 @@ def allowed_steps
             steps << :lead_provider unless mentor.use_same_programme_choices == "yes"
             steps << :review_mentor_eligibility if mentor.eligible_for_funding?
             steps << :eligibility_lead_provider if mentor.eligible_for_funding?
+
+            if !mentor.contract_period_enabled? || store.revised_start_date_in_closed_contract_period
```
avinhurry: We're gating on `contract_period_enabled?` here, but that method is just `queries.contract_period&.enabled` (not future only).

Should we align it with `StartedOnStep#contract_period_enabled?` or move the method to the model or a shared helper so the step and wizard can't drift?

app/wizards/schools/register_mentor_wizard/started_on_step.rb
```diff
@@ -74,6 +76,17 @@ def started_on_obj
       def started_on_as_date
         @started_on_as_date ||= started_on_obj.value_as_date
       end
+
+      def contract_period_enabled?
+        return false unless contract_period
```
avinhurry: Given the AC is future only, should we allow past/today even if there's no matching contract period? As written, this can still block past dates.

avinhurry: Agreed. If we're worried about "no contract period" breaking things, that should be a follow up concern rather than handled here. The blocking page is specifically for future dates, so using it for past/today would need different content.

Please do feel free to card it up if you are concerned.

app/wizards/schools/register_mentor_wizard/registration_store/status.rb
```diff
@@ -81,6 +81,12 @@ def mentoring_at_new_school_only?
           registration_store.store.fetch("mentoring_at_new_school_only", "yes") == "yes"
         end
 
+        def contract_period_enabled?
+          return true if registration_store.started_on&.to_date&.< Date.current
```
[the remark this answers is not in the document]
thenapking: OK great.  I've gone over all the tests and standardized them with the same dates and test cases.  I've ensure the methods are as close as possible and that should hopefully now have nailed it.

PR #2151, opened 2026-02-03
[3010] Block access for closed/unlinked/unapproved schools 

avinhurry:
### Context

We need to block access for ineligible schools signing in via DfE Sign in.

- Block schools with GIAS status `closed` (includes `proposed_to_open` per existing GIAS logic and confirmation from the product team).
- Block schools that have a **GIAS::School** record but no linked **School**.
- Block independent schools where **S41 is not approved** and there are **no ongoing ECT/mentor training periods**.
- Add an information screen shown for the above cases.

> [!NOTE]
> This only applies to DfE Sign in sessions (personas are unaffected).

### Information Screenshot

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,3 +41,4 @@ We still aim to work as openly as possible and will respond to any issues or pul
 This diagram is generated from the application's models and shows their relationships.
 
 👉🏽 [View Mermaid ERD](documentation/domain-model.md)
+


app/views/schools/access/show.html.erb
```diff
@@ -0,0 +1,33 @@
+<%
+  page_data(
+    title: "You cannot access the register early career teacher service for #{@school_name}",
```
ericaporter: Very small nit - should this be 'Register' ?

ericaporter (commented):
LGTM & covered well with specs! A qu on intended behaviour for GIAS-only flow making sure they reach the info page rather than failing but that is mostly for peace of mind, happy to approve otherwise!

PR #2159, opened 2026-02-04
[3075] show schools when a mentor has been deferred or withdrawn

thenapking:
### Context
Add new details to the mentor list page and details page when the mentor's latest training period is deferred or withdrawn.

image

image

image

### Guidance to review
Whilst this ticket has some conceptual overlap with https://github.com/DFE-Digital/register-early-career-teachers-public/pull/2164, mostly this is about logic for the display which is quite different.  In order to remain in sync with that PR, but without branching off it, I have copied across the method added to the TrainingPeriod model.  I have not added tests for it here, as I would expect that to be done in that PR.  I am also awaiting confirmation on the naming of the method.  I have called it `status` here.

For product review, I have set the two mentors in Ackley Bridge to have withdrawn and deferred statuses.

### DIFF
--- a/app/components/schools/mentors/ect_mentor_training_details_component.rb
+++ b/app/components/schools/mentors/ect_mentor_training_details_component.rb
@@ -47,9 +47,11 @@ module Schools
       end
 
       def finished?
+        return false if latest_training_period.status == :withdrawn
+        return false if latest_training_period.status == :deferred
         return false unless latest_training_period.finished_on
 
-        latest_training_period.finished_on < Date.current
+        latest_training_period.finished_on <= Date.current
       end
 
       def started_not_completed_reason?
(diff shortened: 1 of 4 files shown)


app/components/schools/mentors/summary_component.rb
```diff
@@ -30,6 +34,64 @@ def assigned_ects_row
...
+
+        [
+          { key: { text: "Lead provider" }, value: { text: provider_display_with_status } },
+          { key: { text: "Delivery partner" }, value: { text: delivery_partner_name } },
+        ]
+      end
+
+      def latest_training_period
+        @latest_training_period ||= mentor_period_for_school&.latest_training_period
+      end
```
ericaporter: This covers latest but what about if the mentor has a future period starting soon or a current one that's deferred/ withdrawn. We have `scope :latest_first, -> { order(started_on: "desc") }` and latest_training_period uses latest_first so a future TP starting soon will always beat a current one - this was why I opted for `current_or_next_training_period || latest_training_period` Good to get your thoughts on this one!

PR #2162, opened 2026-02-04
Introduce banded fee structure contract models

cpjmcquillan:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/3223

#### Introduce Contract::FeeStructure::Banded model
This introduces a model for "banded" fee structure contracts, which is analogous
to the `CallOffContract` model in ECF1 that `has_many :participant_bands`.

#### Introduce Contract::FeeStructure::Bandeds::Band model
This introduces the `Bandeds::Band` model so we can associate bands to the
relevant banded fee structure.

Most validations are pretty straightforward, except the validation that ensures
bands are sequential without gaps.

The existing implementation assumes bands are added to a banded fee structure in
order.

This felt like a reasonable assumption to make given:

- The first band in a banded fee structure contract must have a minimum of 1
- Bands must be sequential
- Bands must not overlap
- Bands must not have any gaps

By definition, a banded fee structure contract would be invalid if the first
band had a minimum of anything other than 1.

Since we already validate the maximum declarations in a band is greater than the
minimum declarations, we only need to validate the minimum declarations for a
given band is equal to (maximum declarations + 1) for the last band in the
banded fee structure.

### DIFF
--- a/documentation/domain-model.md
+++ b/documentation/domain-model.md
@@ -1,9 +1,30 @@
 ```mermaid
 erDiagram
-  Contract_FeeStructure_FlatRate {
+  Contract_BandedFeeStructure_Band {
+    integer id
+    integer banded_fee_structure_id
+    integer min_declarations
+    integer max_declarations
+    integer fee_per_declaration
+    decimal output_fee_ratio
+    decimal service_fee_ratio
+    datetime created_at
+    datetime updated_at
+  }
+  Contract_BandedFeeStructure_Band }o--|| Contract_BandedFeeStructure : belongs_to
+  Contract_FlatRateFeeStructure {
     integer id
     integer recruitment_target
     decimal fee_per_declaration
+    datetime created_at
+    datetime updated_at
+  }
+  Contract_BandedFeeStructure {
+    integer id
+    integer recruitment_target
+    decimal uplift_fee_per_declaration
+    decimal monthly_service_fee
+    decimal setup_fee
     datetime created_at
     datetime updated_at
   }


app/models/contract/fee_structure/bandeds/band.rb
```diff
@@ -0,0 +1,50 @@
...
+      ordered_bands = banded_fee_structure.bands.to_a
+      previous_band_max_declarations = if ordered_bands.empty?
+                                         0
+                                       else
+                                         ordered_bands.last.max_declarations
+                                       end
+
+      errors.add(:base, "Declaration boundaries must be sequential without gaps") unless
+        min_declarations == previous_band_max_declarations + 1
+    end
```
cpjmcquillan: I think this logic holds based on the assumptions outlined in the commit message, but hoping people can interrogate it rigorously to see if it holds (or not)!

app/models/contract/fee_structure/bandeds/band.rb
```diff
@@ -0,0 +1,50 @@
+module Contract::FeeStructure::Bandeds
+  class Band < ApplicationRecord
+    self.table_name = "contract_fee_structure_bandeds_bands"
+
+    belongs_to :banded_fee_structure,
+               class_name: "Contract::FeeStructure::Banded"
+
+    validates :min_declarations,
```
ethax-ross: Worth adding messages to each of these, I think that's the general convention in RECT given we will add a UI at some point, hopefully?

ethax-ross (approved):
Looks great thanks Connor

PR #2175, opened 2026-02-06
[3225] Add `Contract` model

leandroalemao:
### Context

Ticket: [3225](https://github.com/DFE-Digital/register-ects-project-board/issues/3225)

We want to add the Contract model and associated validations for the upcoming payment calculator work.

- Add contract model and validations;

### DIFF
--- a/app/models/contract.rb
+++ b/app/models/contract.rb
@@ -14,6 +14,10 @@ class Contract < ApplicationRecord
   validates :contract_type,
             presence: { message: "Enter a contract type" },
             inclusion: { in: Contract.contract_types.keys, message: "Choose a valid contract type" }
+  validates :flat_rate_fee_structure,
+            uniqueness: { case_sensitive: false, message: "Contract with the same flat rate fee structure already exists" }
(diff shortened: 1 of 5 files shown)


app/models/contract.rb
```diff
@@ -0,0 +1,41 @@
...
+       validate: { message: "Contract type must be either `ecf` or `ittecf_ectp`" },
+       suffix: true
+
+  # Associations
+  belongs_to :contract_flat_rate_fee_structure, class_name: "Contract::FlatRateFeeStructure", optional: true
+  belongs_to :contract_banded_fee_structure, class_name: "Contract::BandedFeeStructure", optional: true
+  has_many :statements, inverse_of: :contract
+
+  # Validations
+  validates :contract_type,
```
ethax-ross: I wonder if this would be more readable if we used `with_options` to group the validation rules by contract type? 🤔

ethax-ross (commented):
Thanks Leandro just a couple of other things I noticed

cpjmcquillan (approved):

PR #2176, opened 2026-02-06
Update `/participants` error message for invalid course identifiers

cpjmcquillan:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/3232

Before, we returned a slightly misleading/confusing error message on the various PUT `/participants` endpoints.

In #2039 we updated the error message for the POST `/declarations` endpoint, and this makes the same change for the `/participants` endpoints.

Now, the error matches what lead providers would see in ECF1.


ethax-ross (commented):
Looks good just a question around the declarations spec change

PR #2178, opened 2026-02-06
[3116] View declarations for a participant in the admin console

rwrrll:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/3116

Add the "Declarations" tab described in the ticket

### Guidance to review

n.b. the state history table is intended to show all state transition events, but currently only three such events exist, because the features to trigger transition to the others don't yet exist.

### DIFF
--- a/spec/views/admin/teachers/declarations/index.html.erb_spec.rb
+++ b/spec/views/admin/teachers/declarations/index.html.erb_spec.rb
@@ -58,7 +58,7 @@ RSpec.describe "admin/teachers/declarations/index.html.erb" do
     end
   end
 
-  context "when the teacher has declarations" do
+  context "when the teacher has ECT declarations but no mentor declarations" do
     let(:ect_declarations) { [declaration] }
     let(:mentor_declarations) { [] }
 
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/admin.scss
```diff
@@ -187,3 +187,10 @@ $_current-link-border-width: govuk-spacing(1);
     }
   }
 }
+
+.app-mono {
+  padding: 2px 5px;
+  font-family: monospace;
+  background-color: $govuk-template-background-colour;
+  white-space: pre-wrap;
+}
```
rwrrll: Aware there's a similar class in `admin/teachers.scss`, but that one reduces `font-size`. Added this one to conform to prototype (which IMO does work better inline with non-monospaced elements)

app/models/declaration.rb
```diff
@@ -17,6 +17,7 @@ class Declaration < ApplicationRecord
   has_one :ect_teacher, through: :ect_at_school_period, source: :teacher
   has_one :mentor_at_school_period, through: :training_period
   has_one :mentor_teacher, through: :mentor_at_school_period, source: :teacher
+  has_many :events
```
thenapking: We should update the model spec to include this association:
```
it { is_expected.to have_many(:events) }
```

spec/views/admin/teachers/declarations/index.html.erb_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+    let(:ect_declarations) { [declaration] }
+    let(:mentor_declarations) { [] }
+
+    it "does not display the ECT empty state message" do
+      expect(rendered).not_to include("There are no ECT training declarations for")
+    end
+
+    it "renders the declaration partial" do
+      expect(rendered).to include(declaration.api_id)
+    end
```
thenapking: This should also assert that it does display the mentor empty state message.  For completeness we also need the same test to show it works when there are mentor training declarations, and no ECT training declarations.

PR #2180, opened 2026-02-06
RIAB 4 - data model revision

peterdavidhamilton:
### Context

The consensus was that a single table for AB data was sufficient and more maintainable. Therefore the TSH and NB models are combined. Compaired to `main` this means:

- `AppropriateBody` --> `AppropriateBodyPeriod`
- `AppropriateBody` (new)

image

- Address second part of renaming the original to `_period` tackling the factories.
- Replace models and tables
- Remove lead school from ABP for time being, might be replaced by a `lead_school_period`

### Guidance to review

`/admin/organisations/teaching-school-hubs`

cpd-ec2-review-2180-web test teacherservices cloud_admin_organisations_teaching-school-hubs

cpd-ec2-review-2180-web test teacherservices cloud_admin_organisations_teaching-school-hubs_24

cpd-ec2-review-2180-web test teacherservices cloud_admin_schools_141666_overview

[RIAB data model seeded TSHs.webm](https://github.com/user-attachments/assets/32882074-e5c0-4dce-8c0f-fa679feb79a8)

[RIAB data model DSI hub migration.webm](https://github.com/user-attachments/assets/f7b84f58-6276-4046-9fe7-f1b19e46a51e)

### DIFF
--- a/app/models/appropriate_body.rb
+++ b/app/models/appropriate_body.rb
@@ -1,6 +1,8 @@
 #
 # NB: This record is not production ready. Do not use until RIAB has removed this notice.
 #
+# An AppropriateBody is either a national body (like ISTIP) or a regional body (Teaching School Hub).
+# @see LegacyAppropriateBody for details of Local authorities, who used to perform this function.
 class AppropriateBody < ApplicationRecord
   # Independent schools in England and overseas
   ISTIP = "Independent Schools Teacher Induction Panel (ISTIP)"
--- a/app/models/dfe_sign_in_organisation.rb
+++ b/app/models/dfe_sign_in_organisation.rb
@@ -23,8 +23,8 @@ class DfESignInOrganisation < ApplicationRecord
             uniqueness: true
 
   # Scopes
-  scope :national_bodies, -> { joins(:appropriate_body).where.not(urn: nil) }
-  scope :schools, -> { includes(:school).where.not(urn: nil) }
+  scope :national_bodies, -> { where(urn: nil) }
+  scope :schools, -> { where.not(urn: nil) }
 
   def last_authenticated_at=(datetime)
     self.first_authenticated_at ||= datetime
--- a/db/scripts/migrate_legacy_appropriate_bodies.rb
+++ b/db/scripts/migrate_legacy_appropriate_bodies.rb
@@ -6,15 +6,12 @@
 # NB: Migration for active ABs is facilitated by AppropriateBodyMigrator during login
 #
 ActiveRecord::Base.transaction do
-  attributes =
-    AppropriateBodyPeriod.legacy.map do |appropriate_body_period|
-      {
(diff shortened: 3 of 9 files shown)


app/models/appropriate_body.rb
```diff
@@ -0,0 +1,37 @@
+#
+# NB: This record is not production ready. Do not use until RIAB has removed this notice.
+#
+class AppropriateBody < ApplicationRecord
```
craigmdavidson: Does this needs an additional class level comment, in that it represents organisations acting as AppropriateBodies including National Bodies and Teaching School hub? 

Alternatively maybe an alternate name might be better? Since we are creating applying new semantics to an existing concept?

app/models/dfe_sign_in_organisation.rb
```diff
@@ -23,7 +23,7 @@ class DfESignInOrganisation < ApplicationRecord
             uniqueness: true
 
   # Scopes
-  scope :national_bodies, -> { joins(:national_body) }
+  scope :national_bodies, -> { joins(:appropriate_body).where.not(urn: nil) }
```
craigmdavidson: Is this not inverted? I am write in thinking it is schools and not National Bodies that have urns?

Can we use 
```
scope :national_bodies, -> { joins(:appropriate_body).merge(AppropriateBody.national) } 
```

Since we have already defined above what it means for an **AppropriateBody** to be a _National Body_?

db/scripts/migrate_legacy_appropriate_bodies.rb
```diff
@@ -1,12 +1,20 @@
...
+    AppropriateBodyPeriod.legacy.map do |appropriate_body_period|
+      {
+        dqt_id: appropriate_body_period.dqt_id,
+        name: appropriate_body_period.name,
+        body_type: appropriate_body_period.body_type,
+        appropriate_body_period_id: appropriate_body_period.id,
+      }
+    end
+
+  LegacyAppropriateBody.insert_all(attributes)
```
craigmdavidson: This will be faster but will bypass any protections in ActiveRecord. Between the original approach and this, are we trading idempotency for speed here? Is that a good trade given there are relatively few appropriate bodies?

app/models/dfe_sign_in_organisation.rb
```diff
@@ -23,8 +23,8 @@ class DfESignInOrganisation < ApplicationRecord
             uniqueness: true
 
   # Scopes
-  scope :national_bodies, -> { joins(:national_body) }
-  scope :schools, -> { includes(:school).where.not(urn: nil) }
+  scope :national_bodies, -> { where(urn: nil) }
```
avinhurry: Are we safe to assume that all national bodies won't have a URN? Would it be safer to scope it via the `appropriate_body` association?

PR #2190, opened 2026-02-09
[3236] Add PaymentCalculator::FlatRate::DeclarationTypeOutput

mooktakim:
### Context

Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/3236

* New `refundable` scope on `Declaration`
* New service `PaymentCalculator::FlatRate::DeclarationTypeOutput`


app/models/payment_calculator/flat_rate/declaration_type_output.rb
```diff
@@ -0,0 +1,38 @@
+module PaymentCalculator
```
ethax-ross: Would this be better under `app/services` to keep it all together with the rest of the payment calculator service code? 🤔

PR #2191, opened 2026-02-09
Withdrawn/Deferred training periods cannot start in the future

cpjmcquillan:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/3274

Before, it was possible for a training period to be withdrawn or deferred even when it hadn't started yet.

This shape of data doesn't make much sense though, and we prevent lead providers from withdrawing or deferring participants via the API when their training hasn't started yet.

This reworks the `TeacherWithHistories` seed data so participants are only withdrawn/deferred if their training has started. This also adds a validation to the `TrainingPeriod` model to ensure withdrawal/deferral attributes can only be set if the `started_on` date is *not* in the future.

We will run a script on sandbox to update any impacted training periods.

### DIFF
--- a/app/models/training_period.rb
+++ b/app/models/training_period.rb
@@ -84,6 +84,12 @@ class TrainingPeriod < ApplicationRecord
   validate :contract_period_consistent_across_associations, if: :provider_led_training_programme?
   validate :schedule_absent_for_school_led, if: :school_led_training_programme?
   validate :schedule_applicable_for_trainee
+  with_options if: -> { started_on&.future? } do
+    validates :withdrawn_at, absence: true
+    validates :withdrawal_reason, absence: true
+    validates :deferred_at, absence: true
+    validates :deferral_reason, absence: true
+  end
 
   # Scopes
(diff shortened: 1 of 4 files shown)


app/models/training_period.rb
```diff
@@ -78,12 +78,16 @@ class TrainingPeriod < ApplicationRecord
...
   validates :withdrawn_at, presence: true, if: -> { withdrawal_reason.present? }
   validates :withdrawal_reason, presence: true, if: -> { withdrawn_at.present? }
-  validates :deferred_at, presence: true, if: -> { deferral_reason.present? }
-  validates :deferral_reason, presence: true, if: -> { deferred_at.present? }
   validates :schedule, presence: { message: "Schedule is required for provider-led training periods" }, if: :provider_led_training_programme?
   validate :contract_period_consistent_across_associations, if: :provider_led_training_programme?
   validate :schedule_absent_for_school_led, if: :school_led_training_programme?
   validate :schedule_applicable_for_trainee
+  with_options if: -> { started_on&.future? } do
+    validates :withdrawn_at, absence: true, if: -> { started_on&.future? }
```
peteryates: Is the second `if:` necessary here?

PR #2197, opened 2026-02-10
Overwrite first `*AtSchoolPeriod#created_at`

peteryates:
In ECF1 the API returns the `participant_profile`'s `created_at` but we don't have a natural counterpart in ECF2 so instead we use the first `at_school_period`'s `created_at` instead.


ethax-ross (approved):
Thanks Pete!

PR #2198, opened 2026-02-10
Record event when GIAS school eligibility changes

avinhurry:
### Context

We will shortly be enabling a recurring overnight sync with the GIAS API. This PR records an event when a school's eligibility changes so that we have a record of when things changed and why.

- Record an event when a GIAS school's eligibility changes during sync.
- Include the full set of changed fields using `Events::DescribeModifications`.
- Add a new `school_eligibility_changed` event type.

### Example event (DB only)

- Heading: `The Starship Children's Centre became eligible`
- Modifications: `["Eligible set to 'true'"]`


peteryates (approved):
Looks great to me 👌

PR #2200, opened 2026-02-10
Remove IRs linked to BSO schools in pre-wash

tonyheadford:
### Context

We will not have `British schools overseas` establishment type schools in RECT so we need to remove any induction records associated with those schools during migration.

Add a "pre-wash" cleaner for British schools overseas
Changes `Types::SchoolData` to a `Struct` to enable non-mandatory attribute `school_type_name`


peteryates (approved):
Nice!

PR #2201, opened 2026-02-10
[3102] Show mentorship periods in the admin console

rwrrll:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/3102

### DIFF
--- a/app/components/admin/teachers/school_summary_component.rb
+++ b/app/components/admin/teachers/school_summary_component.rb
@@ -81,11 +81,11 @@ module Admin
       end
 
       def mentorship_periods_for_ect
-        school_period.mentorship_periods.includes(mentor: :teacher).order(started_on: :desc)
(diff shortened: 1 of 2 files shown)


app/components/admin/teachers/school_summary_component.rb
```diff
@@ -88,32 +88,41 @@ def mentorship_periods_for_mentor
...
 
-        safe_join(mentor_links, tag.br)
+        mentorship_table(periods, :mentor)
       end
 
-      def mentees_text
-        mentee_links = mentorship_periods_for_mentor.map { |period| mentee_link(period) }.compact
-        return "None assigned" if mentee_links.empty?
+      def mentees_content
+        periods = mentorship_periods_for_mentor
```
thenapking: On the designs, this is sorted so that the latest period is at the bottom of the table.  But this is sorted by default so they are at the top.  Personally I think this is easier to read, but we should check that this is correct.  Same applies to the mentors table.

PR #2203, opened 2026-02-10
[3201] Enable GIAS import job in prod

avinhurry:
### Context

We're ready to enable the nightly GIAS sync. We have successfully run the job manually in prod already.

- Add `GIASImportJob` to run the importer via Active Job.
- Schedule the job in `config/recurring.yml` to run daily at 03:00 on the default queue.
- Update import log message to "may take several minutes".

### DIFF
--- a/lib/tasks/gias.rake
+++ b/lib/tasks/gias.rake
@@ -2,7 +2,7 @@ namespace :gias do
   desc "Import schools data from Get Information About Schools"
   task import: :environment do
     logger = Logger.new($stdout)
-    logger.info "Importing GIAS schools data, this may take a couple minutes..."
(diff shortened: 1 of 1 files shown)


lib/tasks/gias.rake
```diff
@@ -2,15 +2,15 @@ namespace :gias do
   desc "Import schools data from Get Information About Schools"
   task import: :environment do
     logger = Logger.new($stdout)
-    logger.info "Importing GIAS schools data, this may take a couple minutes..."
+    logger.info "Importing GIAS schools data, this may take several minutes..."
```
peteryates: Definitely more accurate 😅

PR #2206, opened 2026-02-11
RIAB 5 - appropriate body period / lead school period

peterdavidhamilton:
### Context

Exploring the option of having `lead_school_period` which overlaps with `appropriate_body_period` depends on first finishing the implemention of `Interval` for ABPs.

- make ABP behave like a period
- add LSP
- spike approach to have TSH's LS change without requiring ABP to end and IPs to be curtailed which was the intentional design of the first approach

### To Do

- [ ] what should the start dates be?
- [ ] what are the challenges of doing it this way?

### Guidance to review

Manual deployment due to conflicts... joy

https://cpd-ec2-review-2206-web.test.teacherservices.cloud/

```sh
$ kubectl exec -n cpd-development deployment/cpd-ec2-review-2206-web -- sh -c "DISABLE_DATABASE_ENVIRONMENT_CHECK=1 SEED_NEW_APPROPRIATE_BODY_MODELS=y rails db:seed:replant"
```


app/models/appropriate_body_period.rb
```diff
@@ -40,7 +39,17 @@ class AppropriateBodyPeriod < ApplicationRecord
   # Normalizations
   normalizes :name, with: -> { it.squish }
 
-  # TODO: consider removing once view components accept the new AB object not the ABP
-  # @return [School]
-  delegate :lead_school, to: :appropriate_body, allow_nil: true
+  # @return [LeadSchoolPeriod::ActiveRecord_AssociationRelation]
+  def lead_school_periods
+    return unless appropriate_body&.teaching_school_hub?
```
craigmdavidson: Does this guard actually help things or make it more brittle? 
If we remove it, we will just get an empty result-set, which will make call chaining safer.

app/models/lead_school_period.rb
```diff
@@ -0,0 +1,7 @@
+class LeadSchoolPeriod < ApplicationRecord
+  include Interval
+
+  # Associations
```
craigmdavidson: Do we need this kind of comment?

craigmdavidson (approved):
LGTM. Small handful of nitpicks ignore what you don't think is useful.

PR #2210, opened 2026-02-11
[3263] - extend the reporting for migration to cover mentorship periods

ltello:
### Context

[Issue](https://github.com/DFE-Digital/register-ects-project-board/issues/3263)

- Add new fields on DataMigrationTeacherCombination table to store ecf1 and ecf2 mentorships processed and created.
- Display totals in the main view of the migration
- Add exporters to download csv files with the mentorships accounted above for analysis.
- Fix bug that displays fewer migrated combinations than actually happened.

### DIFF
--- a/app/controllers/migration/detected_ecf1_mentorships_export_controller.rb
+++ b/app/controllers/migration/detected_ecf1_mentorships_export_controller.rb
@@ -1,0 +1,3 @@
+class Migration::DetectedECF1MentorshipsExportController < Migration::BaseExportController
+  def exporter_class = Migration::DetectedECF1MentorshipsExporter
+end
--- a/app/controllers/migration/ecf2_migrated_mentorships_export_controller.rb
+++ b/app/controllers/migration/ecf2_migrated_mentorships_export_controller.rb
@@ -1,0 +1,3 @@
+class Migration::ECF2MigratedMentorshipsExportController < Migration::BaseExportController
+  def exporter_class = Migration::ECF2MigratedMentorshipsExporter
+end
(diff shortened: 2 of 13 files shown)


app/migration/ecf1_teacher_history.rb
```diff
@@ -14,8 +14,8 @@ def initialize(user:, ect: nil, mentor: nil, participant_identity_updated_ats: [
 
   def self.build(teacher_profile:)
     user_record = teacher_profile.user
-    ect_profile = teacher_profile.participant_profiles.ect.first
-    mentor_profile = teacher_profile.participant_profiles.mentor.first
+    ect_profile = teacher_profile.participant_profiles.detect(&:ect?)
```
tonyheadford: Why change this to `detect`?  It is not a less efficient way of doing the same thing?

app/models/migration/participant_profile.rb
```diff
@@ -29,7 +29,7 @@ def previous_payments_frozen_cohort_start_year
       return nil unless cohort_changed_after_payments_frozen?
 
       induction_records
-        .find { |ir| ir.cohort.payments_frozen? && ir.cohort.id != schedule.cohort_id }
+        .detect { |ir| ir.cohort.payments_frozen? && ir.cohort.id != schedule.cohort_id }
```
tonyheadford: `detect` is an alias for `find` isn't it - not sure why you've changed this?

ltello: See my comment above. I was trying to speed the migration a bit

PR #2212, opened 2026-02-11
Introduce `PaymentCalculator::FlatRate` service

cpjmcquillan:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/3239

This introduces the payment calculator for flat rate fee structure contracts.

The calculator takes a statement, flat rate fee structure and a declaration
"selector" (a lambda function to be called on a collection on declarations) to
calculate the total net amount with (or without) VAT.

It also exposes outputs to be used in the UI

### DIFF
--- a/spec/services/payment_calculator/flat_rate_spec.rb
+++ b/spec/services/payment_calculator/flat_rate_spec.rb
@@ -1,0 +1,107 @@
+RSpec.describe PaymentCalculator::FlatRate do
+  subject(:flat_rate) do
+    described_class.new(statement:, flat_rate_fee_structure:, declaration_selector:)
+  end
+
+  let(:school_partnership) do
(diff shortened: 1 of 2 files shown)


spec/services/payment_calculator/flat_rate_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+  end
+  let(:flat_rate_fee_structure) do
+    FactoryBot.create(
+      :contract_flat_rate_fee_structure,
+      contract:,
+      fee_per_declaration: 100.00
+    )
+  end
+  let(:declaration_selector) do
+    ->(declarations) { declarations.select(&:for_ect?) }
```
ethax-ross: It's not a big deal, but since the flat rate calculator is only applied to mentors at the moment would it be less confusing to use `for_mentor?` here?

PR #2215, opened 2026-02-11
[3271] Replace Time.zone.local with Date ensure that impossible dates are invalid

thenapking:
### Context
`Time.zone.local` will silently transpose impossible dates, like 30th February 2026, into valid dates, eg 2nd March 2026.  This silent correction is confusing for users and we should instead raise a validation error. 
Using `Date.new` will raise an error on invalid entries for impossible dates, which can be used to raise the validation error.

### Guidance to review
Note I have another PR that significantly refactors the tests on start dates, adds missing tests for the step to change a start date, and pulls out common code to a shared example.  The updates here for start date will be moved into that shared example when that PR is merged.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -570,7 +570,7 @@ GEM
       nio4r (~> 2.0)
     raabro (1.4.0)
     racc (1.8.1)
-    rack (3.2.4)
+    rack (3.2.5)
     rack-attack (6.8.0)
       rack (>= 1.0, < 4)
     rack-oauth2 (2.2.1)


avinhurry (approved):
I think we want "Start date must be a real date" for this scenario.

Other than that looks good 👍🏽

thenapking:
> I think we want the validation error message "Start date must be a real date" for this scenario.
> 
> Other than that looks good 👍🏽

Unfortunately the Date library raises the same error when you enter an impossible date and when you enter a blank or partial date.  So we can't really differentiate the situation, and since there's a literally infinite number of invalid dates I've left this as is.  I could check with UX whether we could improve the generic error message?

PR #2233, opened 2026-02-12
Invalid delivery partner IDs return no declarations or school partnerships

cpjmcquillan:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/3272

Before, we returned all declarations when lead providers passed invalid delivery partner IDs to the API.

This was confusing and [inconsistent with ECF1][ecf].

We also made a similar change for filtering by teacher ID in #2032.

This makes the same change.

Now, filtering by an invlaid delivery partner ID returns no declarations.

[ecf]: https://github.com/DFE-Digital/early-careers-framework/blob/5652f140b212b9d323d55a749cee4ed913e47351/spec/services/api/v3/participant_declarations_query_spec.rb#L235-L241

### DIFF
--- a/app/services/api/school_partnerships/query.rb
+++ b/app/services/api/school_partnerships/query.rb
@@ -68,7 +68,7 @@ module API::SchoolPartnerships
     end
 
     def where_delivery_partner_is(delivery_partner_api_ids)
-      return if ignore?(filter: delivery_partner_api_ids)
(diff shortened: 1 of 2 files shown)


ethax-ross (approved):
Thanks @cpjmcquillan - could you give a quick check of the other queries to see if we have missed any others? I think we probably need to update this query as well:

https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/services/api/school_partnerships/query.rb#L71

PR #2244, opened 2026-02-16
Add contracts to seed data

ethax-ross:
### Context

We want to ensure all statements created in the seed data have a suitable contract associated with them (this is currently optional, but will be required in the future).

- Add contracts to seed data

Update dev and API seeds to create contracts and assign them to statements. If there are multiple contracts for an `ActiveLeadProvider` then the statements will be distributed across the contracts evenly.

Update contract factories with more realistic/random values.

### Guidance to review

Screenshot 2026-02-16 at 09 49 28

Script to run on sandbox:

```
APISeedData::Contracts.new.plant

ActiveLeadProvider.find_each do |active_lead_provider|
  statements = active_lead_provider.statements.order(:deadline_date)

  statements.each_with_index do |statement, index|
    contract_index = (index * active_lead_provider.contracts.size) / statements.size
    contract = active_lead_provider.contracts[contract_index]
    statement.update!(contract:)
  end
end
```

### DIFF
--- a/app/models/contract.rb
+++ b/app/models/contract.rb
@@ -9,6 +9,7 @@ class Contract < ApplicationRecord
 
   # Associations
   belongs_to :active_lead_provider
+  has_one :contract_period, through: :active_lead_provider
   belongs_to :flat_rate_fee_structure, class_name: "Contract::FlatRateFeeStructure", optional: true
   belongs_to :banded_fee_structure, class_name: "Contract::BandedFeeStructure", optional: true
   has_many :statements, inverse_of: :contract
--- a/app/services/api_seed_data/contracts.rb
+++ b/app/services/api_seed_data/contracts.rb
@@ -1,0 +1,46 @@
+module APISeedData
+  class Contracts < Base
+    def plant
+      return unless plantable?
+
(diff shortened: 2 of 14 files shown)


leandroalemao (approved):
Tks @ethax-ross looks good. Just one thing.. maybe it's ok to leave as it's.. but should we also maybe a tweak to the [lead providers seed](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/db/seeds/lead_providers.rb#L11) and add more contract periods to some of the lead providers? ie: why only Ambition is working with DfE in 2026 contract period?

ethax-ross:
> Tks @ethax-ross looks good. Just one thing.. maybe it's ok to leave as it's.. but should we also maybe a tweak to the [lead providers seed](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/db/seeds/lead_providers.rb#L11) and add more contract periods to some of the lead providers? ie: why only Ambition is working with DfE in 2026 contract period?

I'm not sure why there's only one LP in 2026 to be honest 🤔 one for another ticket I think though

PR #2245, opened 2026-02-16
[3327] Add fee proportions to flat rate calculators

leandroalemao:
### Context

Ticket: [3327](https://github.com/DFE-Digital/register-ects-project-board/issues/3327)

After some investigation into checking the consistency of the two calculators (ECF1 & RECT).. it's been spotted a few things we need to fix.

- Scope declarations in the flat rate calculator to started/completed with proportionate fees
- Ensure the declaration_type is supported as string or symbol

### Guidance to review

Review app

### DIFF
--- a/app/services/payment_calculator/flat_rate/declaration_type_output.rb
+++ b/app/services/payment_calculator/flat_rate/declaration_type_output.rb
@@ -1,5 +1,8 @@
 module PaymentCalculator
   class FlatRate::DeclarationTypeOutput
+    class DeclarationTypeNotSupportedError < StandardError; end
+    class FeeProportionsInconsistencyError < StandardError; end
+
     include ActiveModel::Model
     include ActiveModel::Attributes
 
@@ -7,6 +10,12 @@ module PaymentCalculator
     attribute :declaration_type
     attribute :fee_per_declaration
     attribute :fee_proportions
+
+    def fee_proportions=(value)
+      raise FeeProportionsInconsistencyError, "Fee proportions are inconsistent. Sum of proportions must be 1." if value.values.sum != 1
+
+      super(value)
+    end
 
     def billable_count
       @billable_count ||= declarations_of_matching_type.billable.count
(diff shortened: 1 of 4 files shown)


app/services/payment_calculator/flat_rate/declaration_type_output.rb
```diff
@@ -32,5 +33,10 @@ def total_net_amount
     def declarations_of_matching_type
       declarations.where(declaration_type:)
     end
+
+    def total_fee_per_declaration
+      output_ratio = fee_proportions[declaration_type.to_sym] || 0
```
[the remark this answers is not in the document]
ethax-ross: Yeah I think operating over the types in the declarations makes sense, we could maybe use `declarations.pluck("DISTINCT declaration_type")` to be efficient?

app/services/payment_calculator/flat_rate/declaration_type_output.rb
```diff
@@ -1,11 +1,14 @@
...
   class FlatRate::DeclarationTypeOutput
+    class DeclarationTypeNotSupportedError < StandardError; end
+
     include ActiveModel::Model
     include ActiveModel::Attributes
 
     attribute :declarations
     attribute :declaration_type
     attribute :fee_per_declaration
+    attribute :fee_proportions
```
ethax-ross: Is it worth verifying these fee proportions add up to 1? the other option is to have them against the flat rate fee structure in the database and validated/saved that way, but I think how it is just now is fine, but maybe just needs a check along the lines of:

```
def fee_proportions=(value)
  raise SomeError if fee_proportions.values.sum != 1
 super(value)
end
```

app/services/payment_calculator/flat_rate/outputs.rb
```diff
@@ -5,10 +5,11 @@ class FlatRate::Outputs
 
     attribute :declarations
     attribute :fee_per_declaration
+    attribute :fee_proportions
 
     def declaration_type_outputs
-      @declaration_type_outputs ||= Declaration.declaration_types.keys.map do |declaration_type|
-        FlatRate::DeclarationTypeOutput.new(declarations:, declaration_type:, fee_per_declaration:)
+      @declaration_type_outputs ||= declarations.pluck("DISTINCT declaration_type").map do |declaration_type|
```
ethax-ross: I think this should reference `declaration_types`?

I do wonder if using `fee_proportions.keys` would make more sense so we get an output for each type we expect/proportion a fee to 🤔 happy to leave it as is for now though and we can always revisit as part of the UI.

PR #2249, opened 2026-02-16
[3241] Remove induction tutor flag

thenapking:
### Context
Now we have tested the mechanism for setting an induction tutor when schools login, we can remove the flag.  One school, Greyfriars, is left without an induction tutor set, for testing purposes.

It turned out cleanest to update the factory to ensure that a confirmed induction tutor is assigned to each school.  Since this requires creating a contract period, some tests, which use contract periods with non-standard dates fail.  I have therefore added an override to the factory.

### DIFF
--- a/spec/services/schools/induction_tutor_details_spec.rb
+++ b/spec/services/schools/induction_tutor_details_spec.rb
@@ -82,7 +82,7 @@ RSpec.describe Schools::InductionTutorDetails do
 
   describe "#wizard_path" do
     context "when the school has an unconfirmed induction tutor" do
-      let(:school) { FactoryBot.create(:school, :with_induction_tutor) }
+      let(:school) { FactoryBot.create(:school, :with_unconfirmed_induction_tutor) }
 
       it "returns the confirm existing induction tutor wizard path" do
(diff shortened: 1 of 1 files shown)


spec/services/schools/induction_tutor_details_spec.rb
avinhurry: The context says "unconfirmed induction tutor" but the setup uses `:with_induction_tutor`, which is a confirmed state now.

Should we switch to `:with_unconfirmed_induction_tutor` (or rename the context) so the test matches the intent?

db/seeds/schools.rb
```diff
@@ -66,7 +66,10 @@ def describe_school(school)
   {
     urn: 6_384_201,
     name: "Greyfriars School",
-    type: :state_school_type
+    type: :state_school_type,
+    induction_tutor_email: nil,
+    induction_tutor_name: nil,
+    induction_tutor_last_nominated_in: nil,
```
avinhurry: Are we ok to 'block' this persona until we enter the induction details? I am not sure if it's used for any sort of testing atm, might be worth double checking or it could become a bit inconvenient.

PR #2259, opened 2026-02-17
Re-introduce foreign keys removed by mistake

ethax-ross:
These were removed accidentally in an early commit: https://github.com/DFE-Digital/register-early-career-teachers-public/commit/2d523aa6c276f408e17a9a57ff0b6121142d29f4


ethax-ross:
N/m they've already been fixed 🙂

PR #2260, opened 2026-02-17
Fix broken migration

cpjmcquillan:
I incorrectly assumed `update_column` would bypass any application logic, but we ran into [errors] running the migration.

I must have run the migration locally before adding the `attr_readonly` attribute.

This refactors the migration to execute the SQL directly.

Now, the migration runs successfully.

[errors]: https://github.com/DFE-Digital/register-early-career-teachers-public/actions/runs/22078959821/job/63805392864


cpjmcquillan:
> Looks good, I would have expected `update_column` to bypass `attr_readonly` as well tbf

https://github.com/rails/rails/blob/main/activerecord/lib/active_record/persistence.rb#L624

PR #2266, opened 2026-02-17
Fix typos in resources for release of Register ECTs

claire-hughez:
@StephenC-V1 I've suggested changing the H1 to match the name of the link to it from the main service manual page. I also think just 'documentation' might be too ambiguous?

Also removed the extra 'the'

### DIFF
--- a/documentation/resources-for-release/index.md
+++ b/documentation/resources-for-release/index.md
@@ -1,8 +1,8 @@
 ---
-title: Register early career teachers service documentation
+title: Resources for release of Register ECTs
 ---
 
-Resources to help the support schools using the register early career teachers service.
+Resources to help support schools using the register early career teachers service.
 
 These resources:
 


documentation/resources-for-release/index.md
```diff
@@ -1,8 +1,8 @@
 ---
-title: Register early career teachers service documentation
+title: Resources for release of Register ECTs
 ---
 
-Resources to help the support schools using the register early career teachers service.
+Resources to help support schools using the register early career teachers service.
```
avinhurry: Should the service name be capitalised

claire-hughez: online services
If an online service has a name that starts with a verb, you do not need to capitalise its name when linking to it. Write the link as part of a sentence so users know what action they can take.

For example:

‘You can [visit someone in prison](https://www.gov.uk/prison-visits) by booking online.’
‘You can [apply for Marriage Allowance online](https://www.gov.uk/apply-marriage-allowance).’

https://www.gov.uk/guidance/style-guide/a-to-z

I kinda hate it tho

PR #2271, opened 2026-02-17
[3282] Add banded payment calculator outputs service

leandroalemao:
### Context

Ticket: [3282](https://github.com.mcas.ms/DFE-Digital/register-ects-project-board/issues/3282)

The `Outputs` service is going to replicate the ECF1 `OutputCalculator` but with a more structured interface (rather than just exposing a method to a hash of banding values).

- Add `PaymentCalculator::Banded::Outputs` service

### Guidance to review

Review app

### DIFF
--- a/app/services/payment_calculator/banded/outputs.rb
+++ b/app/services/payment_calculator/banded/outputs.rb
@@ -8,7 +8,9 @@ module PaymentCalculator
     attribute :banded_fee_structure
 
     def declaration_type_outputs
-      @declaration_type_outputs ||= band_allocations.map { |band_allocation| Banded::DeclarationTypeOutput.new(band_allocation:) }
+      @declaration_type_outputs ||= band_allocations_by_declaration_type.map do |band_allocation|
+        Banded::DeclarationTypeOutput.new(band_allocation:)
+      end
     end
 
     def total_billable_amount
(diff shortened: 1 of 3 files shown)


app/services/payment_calculator/banded/outputs.rb
```diff
@@ -0,0 +1,38 @@
...
+
+    def band_allocator
+      Banded::BandAllocator.new(bands:, previous_declarations:, declarations:)
+    end
+
+    def band_allocations
+      @band_allocations ||= band_allocator.band_allocations.values.flatten
+    end
+
+    delegate :bands, to: :banded_fee_structure
```
ethax-ross: Shall we make this `private: true`?

spec/services/payment_calculator/banded/outputs_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+    end
+  end
+
+  describe "#declaration_type_outputs" do
+    it "returns outputs for each valid declaration type" do
+      expect(outputs.declaration_type_outputs.map(&:declaration_type).uniq).to match_array(declarations.map(&:declaration_type).uniq)
+    end
+  end
+
+  describe "#total_billable_amount" do
```
ethax-ross: It's maybe just me, but I'm finding these pretty hard to follow 😅 would it make things easier if we only had the minimum number of bands (2) with small capacities (maybe 2-3 declarations per band)?

app/services/payment_calculator/banded/outputs.rb
```diff
@@ -0,0 +1,38 @@
...
+    end
+
+  private
+
+    def band_allocator
+      Banded::BandAllocator.new(bands:, previous_declarations:, declarations:)
+    end
+
+    def band_allocations
+      @band_allocations ||= band_allocator.band_allocations.values.flatten
```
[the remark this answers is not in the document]
mooktakim: We'll need to band allocations by declaration type for the views as the data is shown per declaration type

PR #2280, opened 2026-02-18
[3326] Bug: Unable to register ECTs and mentors with a start date of 31/05/2026

thenapking:
### Context
Currently you cannot register an ECT or Mentor to start on the last day of a contract period, because the range query used to find the contract period excludes the last day of the period and thus returns `nil` giving the user an error.

image

To find the contract period containing the start date we rely on this method:
```
  def self.containing_date(date)
    find_by(*date_in_range(date))
  end
```

This uses range the attribute for very efficient quering, and which by default in Postgres, is start inclusive, end exclusive.  You can verify this by running the following SQL query:

```
SELECT range FROM contract_periods;
          range          
-------------------------
 [2021-06-01,2022-05-31)
 [2022-06-01,2023-05-31)
 [2023-06-01,2024-05-31)
 [2024-06-01,2025-05-31)
 [2025-06-01,2026-05-31)
 [2026-06-01,2027-05-31)
```

I can think of three different ways to address this, and this PR presents a conservative fix.  As discussed the most thorough fix for this is to run a migration which fixes the range attribute so it is end inclusive.  However if we did this on just the contract period table, this would leave it confusingly at variance to the other tables.  Fixing all of these will be quite a bit of work.  Therefore this change adds methods to `Queries::RangeQueries` for end exclusive/inclusive, which makes this explicit and set exclusive as default so no behaviour changes.  
I have then gone through all the classes which use ContractPeriod and one by one switched them to end inclusive methods, whilst adding test coverage.  This means that if in the future we update the underlying SQL range attribute we will be able to swap the queries in `Queries::RangeQueries` and see the new tests continue to pass

### Guidance to review
Along with the scenario described by this bug we need to test the following scenarios:

1) Registering an ECT starting to start in the future, on last day of the contract period
2) Registering an ECT starting to start in the future, on last day of the contract period
3) Registering a Mentor on last day of the contract period
4) Changing to an existing mentor on the last day of the contract period
5) Assigning an existing mentor from another school, starting at the new school on the last day of the contract period
6) Assigning an existing mentor from another school, on the last day of the contract period, and selecting a different lead provider
7) Changing lead provider for an ECT on the last day of the contract period
8) Changing the lead provider for a Mentor  on the last day of the contract period
9) Registering an ECT and using a reusable partnership on the last day of the contract period
10) Changing training programme for an ECT on the last day of the contract period
11) Changing induction tutor on the last day of the contract period
12) Logging into the service on the last day of the contract period, to confirming existing induction tutor 
13) Logging into the service on the last day of the contract period, when the induction tutor is confirmed
14) Logging into the service on the last day of the contract period, when the induction tutor is not set

Most of these need to be performed on the last day of the contract period. So for testing purposes we will update the seeds to set the last day on the day that the product review will happen.

### DIFF
--- a/spec/wizards/schools/ects/change_lead_provider_wizard/edit_step_spec.rb
+++ b/spec/wizards/schools/ects/change_lead_provider_wizard/edit_step_spec.rb
@@ -97,6 +97,7 @@ describe Schools::ECTs::ChangeLeadProviderWizard::EditStep do
 
   describe "#lead_providers_for_select" do
     let(:ect_at_school_period) { FactoryBot.create(:ect_at_school_period, school:, started_on:) }
+    let!(:training_period) { FactoryBot.create(:training_period, :for_ect, ect_at_school_period:, started_on:) }
     let!(:active_lead_provider) { FactoryBot.create(:active_lead_provider, :for_year, year: 2025) }
     let!(:other_lead_provider) { FactoryBot.create(:active_lead_provider, :for_year, year: 2025) }
     let!(:future_lead_provider) { FactoryBot.create(:active_lead_provider, :for_year, year: 2026) }


cpjmcquillan (commented):
Playing devil's advocate, could we create the next `ContractPeriod` for now?

And then do https://github.com/DFE-Digital/register-ects-project-board/issues/3413?

thenapking:
> Playing devil's advocate, could we create the next `ContractPeriod` for now?
> 
> And then do [DFE-Digital/register-ects-project-board#3413](https://github.com/DFE-Digital/register-ects-project-board/issues/3413)?

Well, to do 3413 we'd need to add the vast majority of tests in this PR, because they cover the scenarios where end-inclusive ranges are relevant to the code.  I deliberately wrote this PR so that once we update the SQL we just remove the couple of methods in ContractPeriod / RangeQueries that names ranges explictly as end-inclusive/exclusive, and the tests should pass again.  And isn't the next ContractPeriod already created, it's just not enabled yet?

PR #2282, opened 2026-02-18
Handle migrating ERO mentor training

tonyheadford:
### Context

There are a few ERO mentors in ECF1 that have had declarations that have been recorded in statements i.e. `paid` and `clawed_back`. In those cases we should add `training_periods` as normal
**In all other cases** we should not create training period records for ERO mentors.

- Adds service to check mentor's ERO details
- Adds ERO flags into ECF1TeacherHistory::Mentor to aid with decision making
- Prevents adding training periods for ERO mentors that do not have relevant declarations

### DIFF
--- a/app/migration/teacher_history_converter/mentor/latest_induction_records.rb
+++ b/app/migration/teacher_history_converter/mentor/latest_induction_records.rb
@@ -5,13 +5,14 @@
 class TeacherHistoryConverter::Mentor::LatestInductionRecords
   include TeacherHistoryConverter::CalculatedAttributes
 
-  attr_reader :trn, :profile_id, :induction_records, :states
+  attr_reader :trn, :profile_id, :induction_records, :states, :exclude_training_periods
 
-  def initialize(trn:, profile_id:, induction_records:, states:)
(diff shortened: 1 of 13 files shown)


app/migration/teacher_history_converter/mentor/latest_induction_records.rb
```diff
@@ -5,13 +5,14 @@
 class TeacherHistoryConverter::Mentor::LatestInductionRecords
   include TeacherHistoryConverter::CalculatedAttributes
 
-  attr_reader :trn, :profile_id, :induction_records, :states
+  attr_reader :trn, :profile_id, :induction_records, :states, :exclude_training_periods
 
-  def initialize(trn:, profile_id:, induction_records:, states:)
+  def initialize(trn:, profile_id:, induction_records:, states:, exclude_training_periods: false)
```
peteryates: I wonder if the double negative here might make it a bit more confusing than if it was reworded?

If we passed in `early_roll_out: true` perhaps  this would be more concise?

```
training_period = build_training_period(induction_record, { started_on:, finished_on: }) unless early_roll_out
```

PR #2283, opened 2026-02-18
Fix uplift validation and business rules

ethax-ross:
> ⚠️ Uplift data needs to be fixed in hosted environments for the new validation rules before this is merged down.

```
declarations_to_fix = Declaration
  .joins(:contract_period)
  .where(
    "(contract_periods.year >= 2025 OR declarations.declaration_type != 'started') AND 
    (declarations.pupil_premium_uplift = true OR declarations.sparsity_uplift = true)"
  )

declarations_to_fix.update_all(pupil_premium_uplift: false, sparsity_uplift: false)
```

> ⚠️ ~~Data migration will need to be updated to clear uplift flags on declarations that are no longer valid in RECT (so not-started or not in a contract period that supports uplifts)~~ I've let Lorenzo know

### Context

The uplift logic has been implemented incorrectly in RECT; it _mostly_ matches ECF, however we made the choice to not apply uplifts to any mentors in RECT when in fact mentors can have uplifts in ECF.

We want to only allow uplifts on ECT or mentors pre-2025 and uplifts should only be `true` on `started` declaration types.

- Add `uplift_fees_enabled` to `ContractPeriod`

It turns out that the uplift fee logic is contract period dependent. The contract periods <= 2024 allow uplift fees, however contract periods >= 2025 should not accept uplift fees.

Add `uplift_fees_enabled` to `ContractPeriod` so that we can apply relevant validation and business logic.

- Change validation/`uplift_paid?` behaviour on `Declaration`

Initially we thought uplifts were only paid for ECTs, however it turns out we also pay uplifts for mentors.

Change `uplift_paid?` to return `true` if the declaration has been paid and there are uplifts present.

Change validation to only allow uplifts to be specified as `true` if the declaration type is `started` and the declaration is associated to a contract period that allows uplift fees.

Remove scope to `started` declaration types in `Uplifts` service as its now impossible for other declaration types to have uplift flags set to `true`.

We will need to fix data in the various environments before we roll this change out.

- Only set uplifts on applicable declarations

Previously we only set uplift flags on ECT declarations if the teacher was eligible for uplifts.

It turns out that mentor declarations can have uplift flags as well, however we only want to set uplifts for `started` declaration types and if the declaration is in a contract period where uplift fees are enabled (>= 2025).

- Remove `ect_` prefix from `Teacher` uplifts

When we originally added these attributes we thought uplifts could only ever apply to ECTs, however this is not the case - mentors can also have uplift payments.

- Take into account contract period for teacher uplifts

When returning the uplift status of a `Teacher` enrolment we should take into account if the contract period is `uplift_fees_enabled`.

Even if the `Teacher` has uplifts specified, if they are currently in this contract period they will not be eligible for uplifts.

### Guidance to review

Best reviewed by-commit.

There is a loose dependency between the contract period and contracts with respect to when uplifts are paid. Uplifts can only be paid if:

- The declaration is in a pre-2025 'uplift fee enabled' cohort
- The contract takes into account uplift fees

We may be able to tighten this up in the future, but for now I've added comments where appropriate to hopefully highlight this in future.

We are going to take another look at how uplifts are applied to teachers in a separate PR/ticket as we think it could make more sense to derive them directly via the training period -> school.

### DIFF
--- a/app/serializers/api/teacher_serializer.rb
+++ b/app/serializers/api/teacher_serializer.rb
@@ -45,11 +45,19 @@ class API::TeacherSerializer < Blueprinter::Base
           teacher.mentor_first_became_eligible_for_training_at.present?
         end
       end
-      field(:pupil_premium_uplift) do |(training_period, teacher, _)|
-        training_period.for_ect? && teacher.pupil_premium_uplift
+      field(:pupil_premium_uplift) do |(training_period, teacher, metadata)|
+        if training_period.for_ect?
+          metadata.latest_ect_contract_period.uplift_fees_enabled? && teacher.pupil_premium_uplift
+        else
+          metadata.latest_mentor_contract_period.uplift_fees_enabled? && teacher.pupil_premium_uplift
+        end
       end
-      field(:sparsity_uplift) do |(training_period, teacher, _)|
-        training_period.for_ect? && teacher.sparsity_uplift
+      field(:sparsity_uplift) do |(training_period, teacher, metadata)|
+        if training_period.for_ect?
+          metadata.latest_ect_contract_period.uplift_fees_enabled? && teacher.sparsity_uplift
+        else
+          metadata.latest_mentor_contract_period.uplift_fees_enabled? && teacher.sparsity_uplift
+        end
       end
       field(:schedule_identifier) do |(training_period, _, _)|
         training_period.schedule.identifier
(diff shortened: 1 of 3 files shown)


leandroalemao (commented):
Hi @ethax-ross only a few minor questions really tkss

PR #2292, opened 2026-02-19
[2238] Introduce feature flagged OTP sign-in for school users

avinhurry:
## Summary

We need to run realistic user research sessions with school users in the migration environment which is refreshed nightly from production-like data.

Participants must be able to authenticate in a way that:

- uses realistic data
- avoids granting access to the admin interface
- does not create long term account management overhead
- is reversible and low risk

This PR introduces a **feature flagged OTP sign-in path** for explicitly marked school users.

---

### 1. Use DSI accounts

This was considered but not chosen because:

- Participants would need pre-prod DSI accounts, which do not reflect their real login experience.
- There is a risk of confusion later when users are using "real" production DSI accounts.
- Managing temporary DSI access introduces operational overhead.
- Migration contains production like data, misconfigured roles could accidentally grant admin access.

### 2. Grant temporary admin access

This was considered but not chosen because:

- Migration data is production-like.
- Admin access exposes functionality beyond the school journey.
- High risk of accidental data exposure or unintended actions.

### 3. Feature flagged OTP for marked school users (Chosen Approach)

Use a one time password email sign-in, gated by:

1. Feature flag: `ENABLE_OTP_SCHOOL_SIGN_IN`
2. Explicit marker: `users.otp_school_urn`

This approach:

- Requires no DSI setup.
- Is opt-in and environment controlled.
- Allows marked school users (when flag enabled) and always allows internal DfE domains.
- Is reversible by disabling the flag.
- Leaves no persistent state (migration DB refreshes nightly).

---

## Behaviour Matrix

- **Internal admin**: email domain is exactly `education.gov.uk` or `digital.education.gov.uk`
- **Flag**: `ENABLE_OTP_SCHOOL_SIGN_IN`

| User type | Flag false | Flag true |
|---|---|---|
| Internal admin | ✅ allowed | ✅ allowed |
| External user with `otp_school_urn` | ❌ blocked | ✅ allowed |
| All other users | ❌ blocked | ❌ blocked |

## Manual verification

| # | Scenario | Expected | Worked? |
|---|---|---|---|
| 1 | Flag `ON`, email not in `users` table | Blocked (cannot complete sign-in) | ✅ |
| 2 | Flag `ON`, internal DfE user (`@education.gov.uk`) | Allowed as DfE user (admin journey) | ✅ |
| 3 | Flag `ON`, external user with `otp_school_urn` and matching school | Allowed as OTP school user (school journey) | ✅ |
| 4 | Flag `ON`, external user without `otp_school_urn` | Blocked with `This account is not enabled for one time password sign in` | ✅ |
| 5 | Flag `OFF`, internal DfE user (`@education.gov.uk`) | Allowed as DfE user (admin journey) | ✅ |
| 6 | Flag `OFF`, external user with `otp_school_urn` and matching school | Blocked with `This account is not enabled for one time password sign in` | ✅ |
| 7 | Flag `OFF`, external user without `otp_school_urn` | Blocked with `This account is not enabled for one time password sign in` | ✅ |
| 8 | Flag `OFF`, internal DfE user with `otp_school_urn` | Allowed as DfE user | ✅ |

---

## Fail safe behaviour

- A non internal user with `otp_school_urn` is blocked when the flag is disabled. Internal DfE domains are always allowed and sign in as DfE users.
- Unmarked external users are always blocked.

---

## Screenshot of blocked access

image

## How To Use In Migration (or any env)

Because the migration DB refreshes nightly, perform setup on the morning of the UR session:

1. Set `ENABLE_OTP_SCHOOL_SIGN_IN=true` in migration  
   (`config/terraform/application/config/migration.yml`)
2. Create a `User` for the participant's email (must receive OTP emails)
3. Add the URN to the user
4. Verify the URN exists
5. Participant signs in via `/otp-sign-in`
6. Get the OTP from the notify API (real email delivery intentionally not enabled in migration to avoid outbound test emails)

Rollback:

- Set `ENABLE_OTP_SCHOOL_SIGN_IN` to `false`
- No cleanup required as the migration DB refresh removes data nightly

---

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -10,11 +10,11 @@ describe User do
 
     it { is_expected.to allow_value(nil).for(:otp_school_urn) }
 
-    %w[12345 20001 123456 212345 412345].each do |urn|
+    [12_345, 20_001, 123_456, 212_345, 412_345].each do |urn|
       it { is_expected.to allow_value(urn).for(:otp_school_urn) }
     end
 
-    %w[1234 1234567 abc123 -12345].each do |urn|
+    [1_234, 1_234_567, "abc123", -12_345].each do |urn|
       it { is_expected.not_to allow_value(urn).for(:otp_school_urn).with_message("URN must be 5 or 6 numbers") }
     end
   end
--- a/spec/requests/otp_sessions_spec.rb
+++ b/spec/requests/otp_sessions_spec.rb
@@ -40,7 +40,7 @@ RSpec.describe "OTP sessions", type: :request do
 
     context "when the user has a school urn that exists in GIAS" do
       let(:email) { "user@external.com" }
-      let(:otp_school_urn) { "123456" }
+      let(:otp_school_urn) { 123_456 }
 
       before do
         FactoryBot.create(:gias_school, :open, :state_school_type, urn: otp_school_urn)
@@ -72,7 +72,7 @@ RSpec.describe "OTP sessions", type: :request do
 
     context "when the user has a school urn that exists in GIAS and a non DfE email" do
       let(:email) { "user@external.com" }
-      let(:otp_school_urn) { "123456" }
+      let(:otp_school_urn) { 123_456 }
 
       before do
         FactoryBot.create(:gias_school, :open, :state_school_type, urn: otp_school_urn)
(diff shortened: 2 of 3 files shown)


spec/models/user_spec.rb
```diff
@@ -7,6 +7,16 @@
...
+
+    it { is_expected.to allow_value(nil).for(:otp_school_urn) }
+
+    [12_345, 20_001, 123_456, 212_345, 412_345].each do |urn|
+      it { is_expected.to allow_value(urn).for(:otp_school_urn) }
+    end
+
+    [1_234, 1_234_567, "abc123", -12_345].each do |urn|
+      it { is_expected.not_to allow_value(urn).for(:otp_school_urn).with_message("URN must be 5 or 6 numbers") }
+    end
```
peteryates: Are we sure about this? I thought URNs were six digits.

peteryates (approved):
Looks great!

PR #2294, opened 2026-02-20
Add wizard session data to Sentry context

cpjmcquillan:
Recently there have been some [bugs reported] that have so far proved difficult to reproduce.

This provides Sentry with session store data as extra context, which should help us find out more about these errors in non-Production environments if/when they happen.

This is disabled in Production to avoid any PII exposure.

For now this has only been added to the "Register ECT" and "Change" wizards, since this is where we have seen errors.

[bugs reported]: https://github.com/DFE-Digital/register-ects-project-board/issues/3370

### DIFF
--- a/app/controllers/concerns/schools/wizardable.rb
+++ b/app/controllers/concerns/schools/wizardable.rb
@@ -28,6 +28,10 @@ module Schools
                     unless: -> { wizard_class.step?(@previous_step) },
                     only: :new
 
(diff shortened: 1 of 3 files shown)


cpjmcquillan:
Once #2598 is merged, let's add the redis gid for the session to the Sentry context instead. That way we should be able to retrieve the session and dig around if we want (for a while at least)

PR #2303, opened 2026-02-23
Handle setting `api_transfer_updated_at` correctly

tonyheadford:
### Context

In ECF1 the `transfers` API endpoint uses a dynamically calculated `updated_at` value based upon a participants induction records if any school changes are present.  This is done in the [`TransferSerializer`](https://github.com/DFE-Digital/early-careers-framework/blob/main/app/serializers/api/v3/ecf/transfer_serializer.rb#L44-L51) in ECF

- Calculate the appropriate transfer updated_at timestamp per-provider
- make the list of transfers available to the migrator logic
- store the correct value for the lead provider in the `TrainingPeriod.api_transfer_updated_at`

### DIFF
--- a/app/migration/ecf1_teacher_history.rb
+++ b/app/migration/ecf1_teacher_history.rb
@@ -45,6 +45,9 @@ class ECF1TeacherHistory
   def self.build_ect_data(participant_profile:)
     induction_records = build_induction_records(participant_profile:)
 
+    user_updated_at = participant_profile.teacher_profile.user.updated_at
+    transfers = BuildParticipantTransfers.new(induction_records:, user_updated_at:).transfers
+
     ECT.new(
       participant_profile_id: participant_profile.id,
       created_at: participant_profile.created_at,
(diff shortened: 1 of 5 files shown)


app/migration/ecf1_teacher_history.rb
```diff
@@ -56,7 +56,8 @@ def self.build_ect_data(participant_profile:)
       payments_frozen_cohort_start_year: participant_profile.previous_payments_frozen_cohort_start_year,
       states: build_profile_states(participant_profile:),
       induction_records:,
-      mentor_at_school_periods: SchoolMentorsForECT.new(induction_records:).mentor_at_school_periods
+      mentor_at_school_periods: SchoolMentorsForECT.new(induction_records:).mentor_at_school_periods,
+      transfers: BuildParticipantTransfers.new(participant_profile:).transfers
```
ltello: Can we build these transfers from the ECF1TeacherHistory::InductionRecords created on line 46 instead of the actual ones from the database? I think that would save a lot of db queries and not slow too much the processing time of a migration run

PR #2305, opened 2026-02-23
Create new release note for fixing uplift fees

claire-hughez:
I think it makes sense for @ethax-ross to review this first when he has time for techy accuracy, and then I'll ask Tashon to review for content.

### DIFF
--- a/app/views/api/release_notes/release_notes.yml
+++ b/app/views/api/release_notes/release_notes.yml
@@ -35,7 +35,7 @@
     - `sparsity_uplift: true`
     - `pupil_premium_uplift: true`
   
-    Previously, we had said we would always show mentors as false, but this was incorrect.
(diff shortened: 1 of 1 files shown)


app/views/api/release_notes/release_notes.yml
```diff
@@ -20,6 +20,44 @@
...
+    We've now fixed how uplift fees are shown for early career teachers and mentors.
+    
+    ## What's changed
+    
+    ECTs and mentors in the 2024 or earlier cohorts, who are at schools eligible for uplifts, will show the following on `GET participants`:
+    
+    - `sparsity_uplift: true`
+    - `pupil_premium_uplift: true`
+  
+    Previously, we had said we would always show mentors as false, but this was incorrect.
```
claire-hughez:
```suggestion
    Previously, we had said we would always show mentors as false in the Register ECTs API, but this was incorrect.
```

PR #2308, opened 2026-02-23
Add counts for teachers with their `migration_mode`

peteryates:
This shows the total and percentage for economy, premium and not migrated (existing in RIAB already) teachers

image

### DIFF
--- a/app/controllers/migration/migrations_controller.rb
+++ b/app/controllers/migration/migrations_controller.rb
@@ -14,6 +14,11 @@ class Migration::MigrationsController < ::AdminController
       "SUM(ecf1_mentorships_count) AS total_ecf1_mentorships,
        SUM(ecf2_mentorships_count) AS total_ecf2_mentorships"
     ).take
+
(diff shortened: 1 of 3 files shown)


app/controllers/migration/migrations_controller.rb
```diff
@@ -14,6 +14,11 @@ def index
       "SUM(ecf1_mentorships_count) AS total_ecf1_mentorships,
        SUM(ecf2_mentorships_count) AS total_ecf2_mentorships"
     ).take
+
+    @premium_teachers = Teacher.where(migration_mode: "latest_induction_records").count
+    @economy_teachers = Teacher.where(migration_mode: "all_induction_records").count
```
avinhurry: Are these inverted?

https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/migration/ecf2_teacher_history.rb#L81-L87

avinhurry (approved):

PR #2342, opened 2026-02-26
[3285] Fix payment calculator ECF vs RECT discrepencies

mooktakim:
### Context

Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/3285

* RECT does not have `StatementLineItem`, when we do count for billable and refundable, we need to be more careful.
* `Declaration.billable` scope no longer requires clawback_status: :no_clawback.
* Declaration can be both billable and refundable at the same time (paid in one month, clawed back in another)
* Separate billable/refundable params for current and previous declarations for `BandAllocator`, `Banded::Outputs` and `Banded::Uplifts`
* `setup_fee` removed from subtotal in `Banded`
* `ServiceFees` deducts `setup_fee` proportionally from Band A before dividing by 29 months (to match ECF)

### DIFF
--- a/app/services/payment_calculator/banded/band_allocation.rb
+++ b/app/services/payment_calculator/banded/band_allocation.rb
@@ -19,7 +19,7 @@ module PaymentCalculator
       (previous_billable_count + billable_count) - (previous_refundable_count + refundable_count)
     end
 
-    def current_net_count
+    def total_net_count
       billable_count - refundable_count
     end
 
--- a/spec/services/payment_calculator/banded/band_allocation_spec.rb
+++ b/spec/services/payment_calculator/banded/band_allocation_spec.rb
@@ -43,25 +43,18 @@ RSpec.describe PaymentCalculator::Banded::BandAllocation do
     end
   end
 
-  describe "#current_net_count" do
+  describe "#total_net_count" do
     it "returns zero when all counts are zero" do
-      expect(allocation.current_net_count).to eq(0)
+      expect(allocation.total_net_count).to eq(0)
     end
 
-    it "returns current billable minus refundable only" do
+    it "returns current billable minus current refundable" do
       allocation.add_previous_billable(40)
       allocation.add_billable(20)
       allocation.remove_previous_refundable(5)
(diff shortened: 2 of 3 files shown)


app/models/declaration.rb
```diff
@@ -89,12 +89,14 @@ class Declaration < ApplicationRecord
...
     billable_or_changeable.where(declaration_type:)
   }
-  scope :billable, -> { where(payment_status: BILLABLE_PAYMENT_STATUSES, clawback_status: :no_clawback) }
-  scope :refundable, -> { where(clawback_status: REFUNDABLE_PAYMENT_STATUSES) }
   scope :ects, -> { joins(:training_period).where.not(training_periods: { ect_at_school_period_id: nil }) }
   scope :mentors, -> { joins(:training_period).where.not(training_periods: { mentor_at_school_period_id: nil }) }
   scope :with_declaration_types, ->(types) { where(declaration_type: types) }
 
+  # Declaration can be both billable and refundable, paid in one month and clawed_back in another
+  scope :billable, -> { where(payment_status: BILLABLE_PAYMENT_STATUSES) }
```
leandroalemao: Hi @mooktakim as we're changing the concept of a billable declaration, should we make any change to the index `idx_unique_declarations` that still includes `clawback_status: :no_clawback`?

ethax-ross: I don't think we want to update the index; its specifically protecting against two billable declarations that have not been clawed back, so two `paid` for example. If we remove the `no_clawback` check it would prevent a `paid` and `clawed_back`, which is a valid shape

app/services/payment_calculator/banded/uplifts.rb
```diff
@@ -3,21 +3,40 @@ class Banded::Uplifts
     include ActiveModel::Model
     include ActiveModel::Attributes
 
-    attribute :declarations
+    attribute :billable_declarations
```
ethax-ross: Do we need these to be split in this service? or is it just a byproduct of splitting them in the to-level `Banded` service?

ethax-ross (approved):

PR #2346, opened 2026-02-27
Give the migration workers more ram

peteryates:
We're hitting the limit when running migrations, as we're going live in <2 months this won't be a long standing setup.

Just testing it out for now, manually deployed to the `migration` env.


tonyheadford (approved):
1 or 2 webapps?

peteryates:
I think 1 will be fine, we can bump if needed.

PR #2348, opened 2026-02-27
Add retry logic for terraform apply in deploy action

leoapost:
### Context
Deployments are intermittently failing during Terraform apply with Azure API `context deadline exceeded` errors when reading existing PostgreSQL snapshot resources. These failures are transient and unrelated to application code changes, but currently fail the whole deployment on first occurrence.

- Add retry logic around `make ci ${{ inputs.environment }} terraform-apply` in the deploy GitHub Action
- Retry up to 3 attempts total
- Add incremental backoff between attempts (30s, then 60s)
- Preserve existing behavior after successful apply
- Exit with failure if all retry attempts fail

### Guidance to review
- Confirm retry loop behavior and failure handling in `.github/actions/deploy-environment-to-aks/action.yml`
- Confirm no changes to Terraform configuration or infrastructure resources themselves, only deployment robustness

### DIFF
--- a/.github/actions/deploy-environment-to-aks/action.yml
+++ b/.github/actions/deploy-environment-to-aks/action.yml
@@ -56,7 +56,26 @@ runs:
       id: apply-terraform
       shell: bash
       run: |
-        make ci ${{ inputs.environment }} terraform-apply
+        max_attempts=3
+        attempt=1
+        while [ "$attempt" -le "$max_attempts" ]; do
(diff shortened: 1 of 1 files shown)


avinhurry (commented):
Thank you for this Leo. I spoke to our Devops lead and they weren't aware of these failures. They would prefer to investigate and fix the root cause rather than add retries.

I've seen these failures before, but I couldn't find an example while looking through the logs. Do you have one to hand?

PR #2350, opened 2026-02-27
Add AppropriateBody name to events

craigmdavidson:
### Context

Resolves https://github.com.mcas.ms/DFE-Digital/register-ects-project-board/issues/3366

Adds AppropriateBody name to events

### Guidance to review

* The appropriate_body_period is passed in as a parameter, but may be nil when done by an admin e.g. https://github.com.mcas.ms/DFE-Digital/register-early-career-teachers-public/blob/main/app/controllers/admin/teachers/extensions_controller.rb#L49-L53 in these cases we describe the actioner as "admin".


app/services/events/record.rb
```diff
@@ -256,23 +256,26 @@ def self.record_teacher_trs_induction_end_date_updated_event!(author:, teacher:,
     def self.record_induction_extension_created_event!(author:, appropriate_body_period:, teacher:, induction_extension:, modifications:, happened_at: Time.zone.now)
       event_type = :induction_extension_created
       teacher_name = Teachers::Name.new(teacher).full_name
-      heading = "#{teacher_name}’s induction extended by #{induction_extension.number_of_terms} terms"
+      actioner_name = appropriate_body_period&.name || "admin"
+      heading = "#{teacher_name}’s induction extended by #{induction_extension.number_of_terms} terms by #{actioner_name}"
```
peteryates: Rather than defaulting to `admin` (which we'd be able to see from the `author`, could we just miss the "by" off the heading if for some reason the actioner's not present? Defaulting to admin doesn't feel quite right.

PR #2357, opened 2026-02-28
Make the `ECTAtSchoolPeriod` range have an inclusive end date

peteryates:
### Context

We're switching over the ranges so they include the end date. The non-inclusive end dates allows one range to finish on the day its sucecssor starts without overlapping. While this is fine for periods of training it makes less sense for Contract Periods, where people expect the final date to be included.

This change makes the switch for `ECTAtSchoolPeriod`, it updates the range itself and adjusts the tests accordingly.

To make fixing the tests easier I added a `transfer` param to `#create_school_period` and `#add_training_period`, which nudges the start date forward by one day. This allows us to still pass in dates a long time in the past like `3.years.ago` for the previous period's finish and the new period's start, without having to adjust every time or do calculations.

- **Replace ECT at school period [) range with []**
- **Fix specs broken by change to ECTAtSchoolPeriod range**
- **Failed attempt to fix edge case spec**

### See also

Fixes DFE-Digital/register-ects-project-board#3415

### To finish off

* [x] There's one spec I can't work out, in the final commit. Could do with a second pair of eyes cc @thenapking, @cpjmcquillan

### DIFF
--- a/spec/features/schools/ects/register/edge_cases/moving_school_backdated_start_date_spec.rb
+++ b/spec/features/schools/ects/register/edge_cases/moving_school_backdated_start_date_spec.rb
@@ -119,7 +119,7 @@ RSpec.describe "Moving School - backdated start spec", :enable_schools_interface
 
   def and_the_two_schools_want_to_register_the_same_teacher_on_different_dates_of_which_the_earliest_is_backdated
     @school_one_start_date = Date.new(2025, 4, 1)
-    @school_two_start_date = Date.new(2025, 6, 2)
+    @school_two_start_date = Date.new(2025, 6, 3)
   end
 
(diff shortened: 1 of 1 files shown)


peteryates:
> I think that its probably a good idea to revisit this feature test after the migration is done and check the scenario is working in the way we expect. This is to do with changes of school, when one or more date is backdated, and when these dates fall in different contract periods.

Sounds like a good idea to me.

I wonder if testing these tricky scenarios might be easier to do in `spec/services` - the wizard should only really be concerned with calling `Schools::RegistereECT` and Schools::RegisterMentor` with the provided arguments - so could this be moved there and we just use the services directly and inspect the output?

PR #2358, opened 2026-02-28
Split test assist components to their own files

peteryates:
Now they have templates associated with them it's cleaner to put them in their own files, otherwise it's hard to see which templates belong to which components.


avinhurry (approved):
🧹

PR #2362, opened 2026-03-02
Fix failing test

cpjmcquillan:
Now we're more than a year since the date specified in this test, it fails every time!

This reworks the test setup so the test always passes, irrespective of what today's date is.


avinhurry (approved):
Thanks @cpjmcquillan! 🙇🏽

PR #2379, opened 2026-03-04
Implement automated email in induction Fail Journey

craigmdavidson:
### Context

Resolves https://github.com.mcas.ms/DFE-Digital/register-ects-project-board/issues/3293

As part of the improvements we have made to the induction fail journey we wanted to automate an email so that whenever an AB records a fail on RIAB and gives us the relevant dates it sends an email to the TRA team. This is currently done manually by the engagement team (Komal) so that is rationale for streamlining this is that it removes the need for them to monitor this closely.

Whenever someone clicks the below button on appropriate-body/teachers/xxxx/record-failed-outcome/new we want an email to be sent to [TRA.SE@EDUCATION.GOV.UK](mailto:TRA.SE@EDUCATION.GOV.UK)

* Modifies **ApplicationMailer** to bypass Gov UK Notifier, to allow standard _ActionMailer_ behaviour for previews, etc.
* Sends a `InductionMailerFailure#tra_notification` email when **RecordFail** is triggered by an **AppropriateBody** or an _Admin_.
* Avoids inclusion of the ECT teachers Date of Birth in the email, as we don't have it in our database, and we don't really need to provide it in the email.
* **TODO: CC's the DfE group who would have normally sent the email**

### Guidance to review

When we have no `GOVUK_NOTIFY_API_KEY` we modify the configuration to bypass `:notify`, but because our **ApplicationMailer** inherits from a non-standard superclass (**Mail::Notify::Mailer**) we also add a bypass in the **ApplicationMailer**.

Other than that this is a small PR.

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -41,14 +41,18 @@ Rails.application.configure do
   config.active_storage.service = :local
 
   # GOVUK Notify
-  config.action_mailer.delivery_method = :notify
-
   if ENV.key?("GOVUK_NOTIFY_API_KEY")
+    config.action_mailer.delivery_method = :notify
     config.action_mailer.notify_settings = {
       api_key: ENV.fetch("GOVUK_NOTIFY_API_KEY"),
     }
   else
-    Logger.new($stdout).warn("GOVUK_NOTIFY_API_KEY is not set")
+    config.action_mailer.delivery_method = :test
+    Logger.new($stdout).warn("GOVUK_NOTIFY_API_KEY is not set, using test delivery method")
+
+    config.after_initialize do
+      ActionMailer::Base.preview_interceptors.delete(MailNotifyPreviewInterceptor)
+    end
   end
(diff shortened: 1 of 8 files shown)


config/environments/development.rb
```diff
@@ -41,14 +41,18 @@
...
       api_key: ENV.fetch("GOVUK_NOTIFY_API_KEY"),
     }
   else
-    Logger.new($stdout).warn("GOVUK_NOTIFY_API_KEY is not set")
+    config.action_mailer.delivery_method = :test
+    Logger.new($stdout).warn("GOVUK_NOTIFY_API_KEY is not set, using test delivery method")
+
+    config.after_initialize do
+      ActionMailer::Base.preview_interceptors.delete(MailNotifyPreviewInterceptor)
+    end
```
craigmdavidson: We need to remove the preview_interceptor after the fact, rather than clear it here, to take effect.

PR #2381, opened 2026-03-04
[3184] Run api seeds in background job

leandroalemao:
### Context

Ticket: [3184](https://github.com.mcas.ms/DFE-Digital/register-ects-project-board/issues/3184)

The API seeds take quite a while to run in CI now. 

- Instead of reducing the amount of data we could fire off the seed in a background job so it runs after deploy;
- Deploy is taking around ~3min now (from ~11min);
- The api seeds running in a background job is taking ~21min to complete;
- Tried to increase the workers memory `"worker_memory_max": "4Gi"` but it didn't make any difference at all;
- Haven't tried to increase the DB tier (maybe we could upgrade it from `B_Standard_B1ms` to `B_Standard_B2s`?)
- Sending this for a dev review, others could come up with other ideas;

### DIFF
--- a/app/jobs/run_api_seeds_job.rb
+++ b/app/jobs/run_api_seeds_job.rb
@@ -2,6 +2,8 @@ class RunAPISeedsJob < ApplicationJob
   queue_as :api_seeds
 
   def perform
+    return if Rails.env.production?
+
     Rails.application.load_tasks
 
     Rake::Task["api_seed_data:generate"].invoke
--- a/lib/tasks/api_seed_data.rake
+++ b/lib/tasks/api_seed_data.rake
@@ -1,6 +1,8 @@
 namespace :api_seed_data do
   desc "Generate seed data for the sandbox environment"
   task generate: :environment do
+    abort("Only available for non-production environments") if Rails.env.production?
+
     seeds = [
       APISeedData::ContractPeriods,
       APISeedData::LeadProviders,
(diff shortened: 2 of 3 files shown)


cpjmcquillan:
Are any groups of the API seeds completely independent of each other?

If there are, perhaps we can use a matrix in the GitHub action to parallelise things in the deploy itself?

app/services/api_seed_data/base.rb
```diff
@@ -28,5 +28,16 @@ def logger
...
+
+    def active_lead_providers
+      ActiveLeadProvider
+        .includes(:lead_provider, :contract_period)
+        .joins(:lead_provider, :contract_period)
+        .merge(lead_providers)
+    end
+
+    def lead_providers
+      LeadProvider.where.not("name ILIKE ?", "%Reuse%")
```
ethax-ross: Where are these LPs coming from? can we distinguish them in a more reliable way, like by not giving them an `api_id`?

ethax-ross (approved):
Looks good, just a couple of minor things I think would be worth adding it but otherwise good to go 👍

app/services/api_seed_data/base.rb
leandroalemao: Hi @ethax-ross It's coming from [here](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/db/seeds/reuse_choices_scenarios.rb). We're now running the normal seeds before the api seeds.. so the api seeds are finding these extra lead providers created by the normal seeds. All names include "Reuse" so we should be fine.. also these `reuse_choices_scenarios.rb` is not using FactoryBot when creating most of the records.

PR #2386, opened 2026-03-04
Update analytics allowlist and hidden PII

avinhurry:
### Context

Update DfE analytics to include additional columns and block others requested by D&I.

- Added requested tables to `config/analytics.yml`
- Removed the matching tables from `config/analytics_blocklist.yml` so those events can flow
- Updated `config/analytics_hidden_pii.yml` for sensitive fields

### Notes

- `analytics.yml` is shared, so this applies to all analytics enabled envs (`sandbox`, `staging`, `production`)

### Guidance to review

- Is the YAML valid?
- Any concerns with sending these events to BQ?

### DIFF
--- a/config/analytics.yml
+++ b/config/analytics.yml
@@ -42,6 +42,12 @@ shared:
     - trs_not_found
     - trs_induction_start_date
     - trs_induction_completed_date
+  :users:
+    - id
+    - name
+    - email
+    - created_at
+    - updated_at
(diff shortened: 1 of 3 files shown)


ericaporter (approved):
This all looks good! The syntax is fine - I believe it's how symbol keys are passed in a yaml but it also works with regular yaml syntax

PR #2401, opened 2026-03-06
Migrate special declarations from a csv file

ltello:
### Context

[Issue](https://github.com/DFE-Digital/register-ects-project-board/issues/3534)

The current declarations migrator works for most cases, but we have some old (slightly broken) data that cannot be imported to ECF2.

These records are old, mostly from 2021-2022.

In order to add these declarations we'll do the analysis up front and build a CSV which will we can simply 'copy' into the declarations table.

The CSV will have the following columns:

`participant_declaration_id,urn,delivery_partner_id`

Extend the current functionality of the declarations migrator to:

- build a `training_period` for those declarations that can't be associated to an existing teacher `training_period` and are included in the csv file in `app/migration/migrators/special_declarations.csv`
- build a `school_partnership` for that training period if none can be found for the given csv row's school and delivery partner and the declaration's contract period and lead provider.
- build an `at_school_period` for the training period created if none can be found for the teacher and school provided.

### DIFF
--- a/spec/migration/migrators/declaration_spec.rb
+++ b/spec/migration/migrators/declaration_spec.rb
@@ -59,6 +59,8 @@ describe Migrators::Declaration do
     let(:instance) { described_class.new(worker: 0) }
     let!(:data_migration) { FactoryBot.create(:data_migration, model: :declaration, worker: 0) }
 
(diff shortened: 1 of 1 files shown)


tonyheadford (approved):
LGTM

PR #2424, opened 2026-03-10
Wrap ECF2 teacher history-writing in a transaction

peteryates:
### Context

This change wraps the writing of complete ECF2 histories into a transaction, which means if any one of the records fails nothing persists in the database.

It's now necessary becuase we're retrying a teacher in the economy migrator if a premium migration fails (#2410).


peteryates:
This brings the numbers back up to around 99% (running against a slightly out of data ECF1 dataset). I'm re-running with a more recent one now to double check.

PR #2432, opened 2026-03-11
[3398] Create jobs to transition declaration states from 'eligible' to 'payable'

mooktakim:
### Context

Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/3398

* New service `Statements::MarkAsPayable`
* New job `Statements::MarkAsPayableJob`
* Scheduled to run 2 am every day

### DIFF
--- a/app/helpers/declaration_helper.rb
+++ b/app/helpers/declaration_helper.rb
@@ -23,6 +23,7 @@ module DeclarationHelper
     "teacher_declaration_created" => "Submitted",
     "teacher_declaration_voided" => "Voided",
     "teacher_declaration_awaiting_clawback" => "Awaiting clawback",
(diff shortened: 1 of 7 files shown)


config/recurring.yml
```diff
@@ -23,6 +23,11 @@ production:
     queue: metadata
     schedule: every day at 1am
 
+  mark_statements_as_payable:
+    class: Statements::MarkAsPayableJob
+    queue: default
+    schedule: every day at 2am
```
ethax-ross: Can we run it at midnight to match the ECF job?

https://github.com/DFE-Digital/early-careers-framework/blob/main/config/sidekiq_cron_schedule.yml#L25

It probably doesn't matter, but just in case anything we don't know about is setup when statements go to payable.

PR #2433, opened 2026-03-11
[3408] update payment overview

thenapking:
### Context

Update the layout as per the prototype, with some minor layout changes.

image

### Guidance to review

In order to review this please look at the statements for Teach First for September 2024 and 2025.

In each you should see 1x Clawed Back declaration (which has been clawed back from the August Statement). You should see 2x non-clawed back declarations. There will be uplifts and adjustments. Please note that the total includes the setup fee which isn't a listed row. The outputs have already be adjusted for the clawback so when checking the arithmetic do not double count this.

The total for 2025 should equal:
ECT/Mentor Outputs + Service Fees + Setup Fees + Adjustments + VAT

For 2024 it should equal:
Outputs + Service Fees + Uplift Fees + Setup Fees + Adjustments + VAT

### DIFF
--- a/spec/components/admin/statements/payment_overview/ecf_component_spec.rb
+++ b/spec/components/admin/statements/payment_overview/ecf_component_spec.rb
@@ -13,11 +13,9 @@ RSpec.describe Admin::Statements::PaymentOverview::ECFComponent, type: :componen
   let(:banded_fee_structure) do
     FactoryBot.create(
       :contract_banded_fee_structure,
-      :with_bands,
       monthly_service_fee:,
       uplift_fee_per_declaration: 50,
-      recruitment_target: 100,
-      declaration_boundaries: [{ min: 1, max: 200 }]
+      recruitment_target: 100
     )
   end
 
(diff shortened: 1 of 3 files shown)


app/components/admin/statements/payment_overview_component.rb
```diff
@@ -0,0 +1,73 @@
...
+        number_to_pounds(total_amount)
+      end
+
+      def formatted_deadline_date
+        statement.deadline_date.to_fs(:govuk)
+      end
+
+      def formatted_payment_date
+        statement.payment_date.to_fs(:govuk)
+      end
```
cpjmcquillan: It looks like we're using these via the Statement presenter, so can we lose these methods? In the tests we might need to pass in the delegated statement presenter rather than the record though 👌🏼

db/seeds/teacher_histories.rb
```diff
@@ -260,6 +268,67 @@ def find_or_create_school_partnership!(school:, delivery_partner:, lead_provider
   contract_period: cp_2022
 )
 
+FactoryBot.create(:contract_banded_fee_structure, :with_bands)
+
+teach_first_contract_2024 = Contract.where(active_lead_provider: teach_first_grain_2024, contract_type: :ecf).first
+teach_first_contract_2025 = Contract.where(active_lead_provider: teach_first_grain_2025, contract_type: :ittecf_ectp).first
```
cpjmcquillan: Should we call `.first!` so an error is raised explicitly if the contract doesn't exist?

PR #2435, opened 2026-03-11
Improve pre-migration Gantt chart

peteryates:
### Context

It's useful when viewing the Gantt chart to be able to see:

* who made which declaration (displayed in parens after the type)
* when the participant completed their induction/mentor training (shown as a ◇ instead of a ♦)

This PR makes those improvements.

### Standalone usage

Now you can generate a Gantt chart from the command line.

```bash
$ ruby script/build_ecf1_gantt.rb aaaaaaaa-bbbb-cccc-1111-222222222222
$ open /tmp/aaaaaaaa-bbbb-cccc-1111-222222222222.png
```

### Preview

gantt

- **Add induction and mentor completion to the Gantt**
- **Add script for generating Gantt charts**

### DIFF
--- a/script/build_ecf1_gantt.rb
+++ b/script/build_ecf1_gantt.rb
@@ -1,0 +1,10 @@
+require_relative "../config/environment"
+
+participant_profile_id = ARGV[0]
+participant_profile = Migration::ParticipantProfile.find(participant_profile_id)
+induction_records = Migration::InductionRecordExporter.new.where_participant_profile_id_is(participant_profile_id).rows
+declarations = participant_profile.participant_declarations
+
+gantt = Migration::PreMigrationGantt.new(induction_records, declarations, participant_profile)
+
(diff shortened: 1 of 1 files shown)


app/services/migration/pre_migration_gantt.rb
```diff
@@ -69,7 +71,25 @@ def withdrawn_note(identifier)
     end
 
     def declaration_descriptions
-      declarations.map { |d| %([#{d.declaration_type}] happens at #{d.declaration_date.to_date}) }
+      declarations.map { |d| %([#{d.declaration_type} (#{d.cpd_lead_provider.name})] happens at #{d.declaration_date.to_date}) }
```
avinhurry: I'm probably missing some domain context here, but whats the difference between `cpd_lead_provider` and `lead_provider` in the migration data? 

Noticed we use `lead_provider_name` for induction records and `cpd_lead_provider.name` for declarations here.

PR #2442, opened 2026-03-12
[3410] Update output payments section on financial statements view in admin console

rwrrll:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/3410


thenapking (commented):
I think this is OK as is, especially given we have some time pressure.  But there's a way of doing this inline with the declarations component which doesn't require having several presenters.

cpjmcquillan (approved):
🚀

I think the presenter pattern works well here, and there is probably some thinking we can use elsewhere.

Once we're live, I think it would be good to circle back round and align the style of these components —pulling our favourite parts out from each implementation.

PR #2453, opened 2026-03-13
[3393] Allow same-LP schedule changes on future training periods while blocking cross-LP changes during transfers

mooktakim:
### Context

Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/3393

* Replaces `lead_provider_is_currently_training_teacher` with `training_period_not_finished` — still blocks changes on finished TPs, but now also allows changes on future TPs (not just ongoing ones).
* Replaces `no_future_training_periods_exist` with two new validations:
  *  `future_training_period_not_blocked_by_another_lead_provider` - if the LP only has a future TP, it blocks the change when a different LP is currently training the participant (i.e. during a transfer, the new LP can't change the schedule while the old LP is still active).
  * `no_future_training_periods_with_different_lead_provider` - blocks the old LP from changing their schedule when a future TP exists with a different LP.
* Overrides `training_period` resolution - metadata picks the latest TP by `started_on`, which prefers a future TP over an ongoing one. The override now prefers the ongoing TP so the schedule change operates against the currently active record.

### DIFF
--- a/app/services/api/teachers/change_schedule.rb
+++ b/app/services/api/teachers/change_schedule.rb
@@ -34,15 +34,39 @@ module API::Teachers
       ).change_schedule
     end
 
-    # The metadata handler selects the training period with the latest started_on.
-    # When a future training period exists for the same lead provider, it gets selected
-    # instead of the ongoing one. We fall back to the ongoing training period so the
-    # schedule change can proceed against the correct record.
-    def training_period
-      tp = super
-      return tp unless tp&.started_on&.future?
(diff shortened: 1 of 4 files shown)


leandroalemao (approved):

ethax-ross (approved):
Thanks Mook, can we make sure this one goes to PO review please.

I feel like the `ChangeSchedule` service is getting overly complicated now 😞 maybe something we need to have a think about when we have some time

ethax-ross (approved):

PR #2479, opened 2026-03-18
Improve schools migrator

leandroalemao:
### Context

Seeing some schools being migrated but with different `api_id`.

ie:

This school is in ECF1 and it's been migrated to RECT but it has a different `api_id` for the same `URN` in RECT.

ECF1:
```
#<School:0x00007f9650be0288
 id: "4ebc3618-46c6-4313-b354-1b739cb87f01",
 created_at: Thu, 22 Apr 2021 12:35:24.639365000 UTC +00:00,
 updated_at: Mon, 02 Mar 2026 06:00:36.701988000 UTC +00:00,
 urn: "123BLA",
 name: "School Name",
...
```

RECT
```
#<School:0x00007f0920565010
 id: 11329,
 urn: 123BLA,
 name: "School Name",
 created_at: "2026-03-16 20:38:31.171168000 +0000",
 updated_at: "2026-03-16 20:38:31.171168000 +0000",
 api_id: "79d511f0-9b31-4f5a-8416-7e5f7bd7b663",
...
```

- Improve schools migrator in order to prevent the error from happening.

### DIFF
--- a/app/migration/migrators/school.rb
+++ b/app/migration/migrators/school.rb
@@ -41,10 +41,13 @@ module Migrators
 
       migrate(schools_with_associations) do |ecf_school|
         gias_school = find_gias_school_by_urn(ecf_school.urn.to_i) || migrate_school!(ecf_school)
+
         if check_gias_school(gias_school:, ecf_school:)
+          school = ::School.find_by!(urn: ecf_school.urn)
+
           [
             compare_fields(gias_school:, ecf_school:),
-            update_school!(school: gias_school.school, ecf_school:),
+            update_school!(school:, ecf_school:)
           ].all?
         end
       end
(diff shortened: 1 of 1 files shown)


avinhurry (approved):

tonyheadford (commented):
~~Looking at it a bit more I think the problem is because the `::School` records are created in the `GIASImportOnlyCreateSchools` migrator now which is not a dependency of the `School` migrator. I think making that change would fix the issue (i.e. adding `gias_import_only_create_schools` to the dependencies of the `School` migrator) - what do you think?~~. Nope I was wrong it's set OK as the new migrator uses `:gias_import` as it's model name as well - sorry

avinhurry (commented):
I had a deeper look at this and think there may be another failure mode here...

Previously the code assumed that if a `GIAS::School` existed, the paired `School` row existed too. That may no longer be true because we run the importer with `auto_create_school: false` every morning in production.
  
In the school migrator [we try to find a `GIAS::School`](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/migration/migrators/school.rb#L43), if it finds one, it skips the path that creates a new `School` + `GIAS::School` pair and then goes on to [updating the associated school](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/migration/migrators/school.rb#L47). That means the code is still assuming the counterpart `School` already exists in that branch. 

Apologies, I did not realise we had made that assumption in the migration code when I stopped auto creating schools.

Potentially then some `School` rows may be getting created elsewhere with a [default generated `api_id` from Postgres](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/db/schema.rb#L795) rather than the legacy ECF school id we copy into `api_id`.

There was also a `TRUNCATE gias_schools RESTART IDENTITY CASCADE` run in production last Monday. Because that was run with `CASCADE`, it would have truncated dependent tables including `schools` if rows were present at the time. `School.count` is 0 now in prod, and the subsequent daily import would have repopulated `GIAS::School` without recreating `School` rows. So my hunch is that this may have exposed or exacerbated the problem.

ethax-ross:
@leandroalemao I caught up with @avinhurry on this and given that the schools migrator is behaving as it currently is - whilst merging this down probably wouldn't hurt anything - I don't think its adding anything any more either? Unless you feel differently and want to go through rebasing/deploying and testing it on migration I think we should just close it off?

peteryates:
Closing this, don't think it's necessary now.

PR #2508, opened 2026-03-23
Extract closed provider-led period check

ericaporter:
For reference see [ticket](https://github.com/DFE-Digital/register-ects-project-board/issues/3330)
### Context
This logic was previously embedded within a larger registration flow change, which is currently blocking other work. To unblock those tickets, the predicate has been extracted into a standalone service to provide a single source of truth.

Following discussion, the implementation now uses payments_frozen? rather than hardcoded contract period years, to better align with the domain model.

- Extract ContractPeriods::MoveFromClosedProviderLedPeriod into its own service
- Encapsulates logic to determine whether:
the previous training period is provider-led
the associated contract period is payments frozen
- Replace hardcoded 2021/2022 checks with payments_frozen?
- Add unit tests for the service

### Guidance to review
- Check that the predicate correctly identifies provider-led training periods in payments-frozen contract periods
- Confirm this aligns with the intended domain behaviour discussed (subject to follow-up after further clarification)

### DIFF
--- a/spec/services/contract_periods/move_from_closed_provider_led_period_spec.rb
+++ b/spec/services/contract_periods/move_from_closed_provider_led_period_spec.rb
@@ -58,6 +58,25 @@ RSpec.describe ContractPeriods::MoveFromClosedProviderLedPeriod do
     end
   end
 
+  context "when previous training period uses expression_of_interest_contract_period that is not payments frozen" do
+    let(:contract_period) do
+      instance_double(ContractPeriod, payments_frozen?: false)
+    end
+
+    let(:previous_training_period) do
+      instance_double(
+        TrainingPeriod,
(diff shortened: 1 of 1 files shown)


spec/services/contract_periods/move_from_closed_provider_led_period_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+
+    it "returns false" do
+      expect(result).to be false
+    end
+  end
+
+  context "when previous training period uses expression_of_interest_contract_period" do
+    let(:contract_period) do
+      instance_double(ContractPeriod, payments_frozen?: true)
+    end
```
cpjmcquillan: For completeness should we add a test for `payments_frozen?: false` with the EOI contract period?

PR #2511, opened 2026-03-23
Declaration summary tweaks

cpjmcquillan:
Initially we thought we should change the "Clawed back" count so it only
included declarations that have been clawed back (i.e to be consistent with
ECF).

But after discussing with Product and Contract Managers, we decided we'd
actually prefer to include declarations that are awaiting clawback too.

Since we're already doing that, we can just change the label to "Clawbacks".

We have also tweaked the output for declaration types that don't apply to
mentors. Now, we display a "-" instead of "0" for retained and extended
declaration counts for mentors in statements for `ittecf_ectp` contracts.

### DIFF
--- a/app/components/admin/statements/declaration_component.html.erb
+++ b/app/components/admin/statements/declaration_component.html.erb
@@ -5,7 +5,7 @@
 
     table.with_head do |head|
       head.with_row do |row|
-        row.with_cell(header: true) { "" } 
+        row.with_cell(header: true) { "" }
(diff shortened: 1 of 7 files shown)


cpjmcquillan (commented):
@ethax-ross I _think_ this makes sense, but I want to check I'm not missing something. Things are slightly different here than in ECF-land since we model declarations slightly differently (i.e payment/clawback statements), but I think this covers that.

PR #2540, opened 2026-03-26
Mentors can be registered at a school any time after their old school start date

cpjmcquillan:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/3599

Before, users would run into a validation error if they tried to enter a start
date that was after the start date for a previous mentor at school period but
before the reported leaving date.

This wasn't intended, though.

Instead, we want the new school to be able to enter any start date after the
most recent previous school period start date. Any school, training, or
mentorship period end dates should be updated to use the new school's start date
to avoid overlaps.

This amends the `RegisterMentorWizard::StartedOnStep` to account for this.

Making this change (and adding the feature test) uncovered some other small bugs
too. These have been addressed as part of this change.

### Guidance to review

- [x] Register a mentor mentoring only at the new school, starting before the reported leaving date at the other school
- [x] Register a mentor mentoring only at the new school, starting after the reported leaving date at the other school
- [x] Register a mentor mentoring only at the new school, starting on the reported leaving date at the other school
- [x] Register a mentor mentoring not only at the new school

### DIFF
--- a/app/wizards/schools/register_mentor_wizard/started_on_step.rb
+++ b/app/wizards/schools/register_mentor_wizard/started_on_step.rb
@@ -4,15 +4,13 @@ module Schools
       attr_accessor :started_on
 
       validates :started_on, mentor_start_date: true
-      validate :started_on_cannot_be_before_previous_started_and_finished_dates, if: :started_on
+      validate :started_on_after_all_previous_period_started_on_dates
       validate :started_on_within_4_months, if: :currently_mentor_at_another_school?
 
-      def self.permitted_params
-        %i[started_on]
-      end
+      def self.permitted_params = %i[started_on]
 
       def next_step
-        if !contract_period_enabled?
+        if registrations_closed_for_contract_period?
           :cannot_register_mentor_yet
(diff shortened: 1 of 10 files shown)


app/wizards/schools/register_mentor_wizard/started_on_step.rb
```diff
@@ -43,21 +41,30 @@ def pre_populate_attributes
...
         return if errors.any?
 
-        date = mentor.previous_school_mentor_at_school_periods.pluck(:started_on, :finished_on).flatten.compact.max
-        return unless date
-
-        if started_on_as_date.before?(date.next_day)
-          errors.add(:started_on, "#{mentor.full_name} was registered as a mentor at their last school starting on the #{date.to_fs(:govuk)}. Enter a later date.")
+        latest_started_on_at_previous_school = mentor
+          .previous_school_mentor_at_school_periods
+          .maximum(:started_on)
```
cpjmcquillan: The key difference here is that now we only validate against the latest `started_on` date, rather than considering `finished_on` dates too

PR #2543, opened 2026-03-26
Comment on PR when `db/schema.rb` changes

cpjmcquillan:
Often, we make changes to `db/schema.rb` (e.g adding/removing fields) and forget to consider the impact on analytics.

This adds a comment that prompts the author to think about it, tagging the CPD Data & Insights team too.

Hopefully this light process helps us get ahead of things before they become issues and we block or disrupt the team.

### DIFF
--- a/.github/workflows/analytics-reminder.yml
+++ b/.github/workflows/analytics-reminder.yml
@@ -1,0 +1,47 @@
+name: Remind us to inform data & insights when we make changes that might impact analytics
+
+on:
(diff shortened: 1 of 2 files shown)


peteryates (approved):
Nice! I like the extra step in checking if they're already mentioned so we don't spam them 👌

avinhurry:
I'm a bit late here but this is cool! 

Much nicer than bloating every PR template with the guidance.

PR #2561, opened 2026-03-28
Pre wash future induction records that are withdrawn or deferred

peteryates:
We found in DFE-Digital/register-ects-project-board#3683 an edge case where the economy migrator wasn't dealing with future withdrawals/deferrals properly. They're prohibited by ECF2's validation.

This change adds a pre-wash step that removes induction records that:

* start in the future
* have a `training_status` of `withdrawn` or `deferred`

This is done as a pre-wash step rather than as part of the conversion because we don't want to risk throwing away a combination that might be needed to attach a declaration to, if the latest IR happens to be a future withdrawal/deferral.

### DIFF
--- a/app/migration/teacher_history_converter/cleaner.rb
+++ b/app/migration/teacher_history_converter/cleaner.rb
@@ -24,7 +24,8 @@ private
   end
 
   def economy_clean!
-    remove_british_schools_overseas(@raw_induction_records)
+    @raw_induction_records
+      .then { remove_british_schools_overseas(it) }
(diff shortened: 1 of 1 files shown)


peteryates:
Thinking about it, with ECF1's unconventional status behaviour, the `training_status` of `withdrawn` might not always mean that the participant is withdrawn (same for `deferred`) :thinking: 

If we move this back from pre-wash into the converter, that would work too but there's a chance that we'd lose a needed combo.

PR #2598, opened 2026-04-04
Add and configure redis-session-store

peteryates:
This gem moves our sessions from cookies (where there's a chance they'll overflow at 4k) to Redis.

We opted for Redis over PostgreSQL for this because we're able to take advantage of its inbuilt [TTL](https://redis.io/docs/latest/commands/ttl/) functionality to avoid having to write extra functionality to clean up old sessions.

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -32,6 +32,16 @@ Rails.application.configure do
     config.cache_store = :null_store
   end
 
+  config.session_store(
+    :redis_session_store,
+    serializer: :json,
+    redis: {
+      expire_after: ENV.fetch("MAX_SESSION_IDLE_TIME", 7200.seconds).to_i,
+      key_prefix: "_ecf2_session:",
+      url: ENV.fetch("REDIS_CACHE_URL")
+    }
+  )
+
   # Store uploaded files on the local file system (see config/storage.yml for options).
   config.active_storage.service = :local
 
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -26,14 +26,9 @@ Rails.application.configure do
   if Rails.root.join("tmp/caching-dev.txt").exist?
     config.action_controller.perform_caching = true
     config.action_controller.enable_fragment_cache_logging = true
-
-    config.cache_store = :memory_store
-    config.public_file_server.headers = {
-      "Cache-Control" => "public, max-age=#{2.days.to_i}"
-    }
   else
     config.action_controller.perform_caching = false
-
+    config.action_controller.enable_fragment_cache_logging = false
     config.cache_store = :null_store
   end
 
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -73,6 +73,15 @@ Rails.application.configure do
 
   # Use a different cache store in production.
   config.cache_store = :redis_cache_store, { url: ENV.fetch("REDIS_CACHE_URL") }
+  config.session_store(
+    :redis_session_store,
+    serializer: :json,
+    redis: {
+      expire_after: ENV.fetch("MAX_SESSION_IDLE_TIME", 7200.seconds).to_i,
+      key_prefix: "_ecf2_session:",
+      url: ENV.fetch("REDIS_CACHE_URL")
+    }
+  )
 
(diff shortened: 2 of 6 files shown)


cpjmcquillan (approved):

avinhurry (approved):
Since we are moving from cookie backed to redis backed sessions, will this log out existing AB users? If so are we happy with that?

Perhaps it could be worth deploying out of hours to minimise disruption?

peteryates:
Yeah good call @avinhurry - will do this out of hours.

peteryates:
Added the `DO NOT MERGE` label as a reminder that this should happen when nobody's logged in. I'll make a note to do this over a weekend.

cpjmcquillan (approved):

config/environments/production.rb
```diff
@@ -73,6 +73,15 @@
 
   # Use a different cache store in production.
   config.cache_store = :redis_cache_store, { url: ENV.fetch("REDIS_CACHE_URL") }
+  config.session_store(
+    :redis_session_store,
+    serializer: :json,
+    redis: {
+      expire_after: ENV.fetch("MAX_SESSION_IDLE_TIME", 7200.seconds).to_i,
+      key_prefix: "_ecf2_session:",
+      url: ENV.fetch("REDIS_CACHE_URL")
```
avinhurry, line 82: Do we expect sessions and cache to share the same Redis instance here?

I was just looking at the Redis config and noticed we're on the default eviction policy `allkeys-lru`. My understanding is that if we store sessions in the same instance, they'd also be eligible for eviction under memory pressure, which could mean unexpected logouts.

[The Redis docs](https://redis.io/docs/latest/develop/reference/eviction/) seem to suggest separate instances if possible when mixing cache data with less disposable data.

> The volatile-lru, volatile-lrm, and volatile-random policies are mainly useful when you want to use a single Redis instance for both caching and for a set of persistent keys. However, you should consider running two separate Redis instances in a case like this, if possible.

Curious about your view on that in case i'm missing something.

config/environments/development.rb
```diff
@@ -26,17 +26,22 @@
...
     config.cache_store = :null_store
   end
 
+  config.session_store(
+    :redis_session_store,
+    serializer: :json,
+    redis: {
+      expire_after: ENV.fetch("MAX_SESSION_IDLE_TIME", 7200.seconds).to_i,
+      key_prefix: "_ecf2_session:",
+      url: ENV.fetch("REDIS_CACHE_URL")
```
avinhurry, line 41: Do we need to add Redis to the devcontainer/codespaces setup here:

- https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/.devcontainer/docker-compose.yml

Since we require `REDIS_CACHE_URL` here now, I was wondering if codespaces will also need Redis config/service?

config/environments/development.rb
```diff
@@ -26,17 +26,22 @@
   if Rails.root.join("tmp/caching-dev.txt").exist?
     config.action_controller.perform_caching = true
     config.action_controller.enable_fragment_cache_logging = true
-
-    config.cache_store = :memory_store
-    config.public_file_server.headers = {
-      "Cache-Control" => "public, max-age=#{2.days.to_i}"
-    }
```
avinhurry, line 33: Did we drop this by accident? My understanding is that in development, if we remove this, cached data will fall back to disk (`tmp/cache`) rather than `:memory_store`.

Not a big deal, just wanted to check whether it was intentional?

peteryates: Yeah, thought it'd make sense to require Redis for developers as we'll be using it for more than caching in prod.

config/environments/production.rb
peteryates: This is a really good point. We don't really cache anything at the moment in the app so I don't think this will be an immediate problem, but it's definitely something we should fix sooner rather than later.

I wonder if changing from `allkeys-lru` to `volatile-ttl` might make more sense. We have quite a long timeout (2 hours) so we'd have to be in a really unhealthy state for people to be being logged out while they're using the app.

I guess we could use the database for sessions, it's a bit safer - but I like that Redis has a TTL setting and we don't need to worry about cleanup jobs etc.

PR #2611, opened 2026-04-08
[3514] Add API::Teachers::EligibilityForFunding service

mooktakim:
### Context

Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/3514

* New service `API::Teachers::EligibilityForFunding` to determine eligibility for funding
* Updated teacher serializer to use the service.

### DIFF
--- a/app/serializers/api/teacher_serializer.rb
+++ b/app/serializers/api/teacher_serializer.rb
@@ -39,11 +39,8 @@ class API::TeacherSerializer < Blueprinter::Base
       field(:training_status) { |(training_period, _, _)| API::TrainingPeriods::TrainingStatus.new(training_period:).status }
       field(:participant_status) { |(training_period, teacher, _)| API::TrainingPeriods::TeacherStatus.new(latest_training_period: training_period, teacher:).status }
       field(:eligible_for_funding) do |(training_period, teacher, _)|
-        if training_period.for_ect?
-          teacher.ect_first_became_eligible_for_training_at.present?
-        else
-          teacher.mentor_first_became_eligible_for_training_at.present?
-        end
+        teacher_type = training_period.for_ect? ? :ect : :mentor
+        API::Teachers::EligibilityForFunding.new(teacher:, teacher_type:).eligible?
       end
       field(:pupil_premium_uplift) do |(training_period, _, metadata)|
         uplift_value(training_period, metadata, :pupil_premium_uplift)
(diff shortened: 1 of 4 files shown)


app/serializers/api/teacher_serializer.rb
```diff
@@ -39,11 +39,8 @@ class TrainingPeriodSerializer < Blueprinter::Base
       field(:training_status) { |(training_period, _, _)| API::TrainingPeriods::TrainingStatus.new(training_period:).status }
       field(:participant_status) { |(training_period, teacher, _)| API::TrainingPeriods::TeacherStatus.new(latest_training_period: training_period, teacher:).status }
       field(:eligible_for_funding) do |(training_period, teacher, _)|
-        if training_period.for_ect?
-          teacher.ect_first_became_eligible_for_training_at.present?
-        else
-          teacher.mentor_first_became_eligible_for_training_at.present?
-        end
+        teacher_type = training_period.for_ect? ? :ect : :mentor
+        API::Teachers::EligibilityForFunding.new(teacher:, teacher_type:).eligible?
```
leandroalemao, line 43: Hi @mooktakim How about sending only the training period to the service? So we can call `training_period.for_ect?` in the service only once instead of `teacher_type == :ect` twice

app/services/api/teachers/eligibility_for_funding.rb
```diff
@@ -0,0 +1,33 @@
+class API::Teachers::EligibilityForFunding
+  attr_reader :teacher, :teacher_type
+
+  def initialize(teacher:, teacher_type:)
+    @teacher = teacher
+    @teacher_type = teacher_type
+  end
+
+  def eligible?
```
ethax-ross, line 9: Take it or leave it:

```
earliest = [eligible_on, ineligible_on].compact.min
earliest && earliest == eligible_on
```

spec/serializers/api/teacher_serializer_spec.rb
```diff
@@ -209,19 +209,57 @@
             end
           end
 
-          context "when `eligible_for_funding` is true" do
-            let(:teacher) { FactoryBot.create(:teacher, ect_first_became_eligible_for_training_at: Time.zone.now) }
+          describe "`eligible_for_funding`" do
```
ethax-ross, line 212: Is it worth just stubbing the service in the serializer to avoid repeating all the tests?

ethax-ross (approved):

app/serializers/api/teacher_serializer.rb
mooktakim: I think the service makes more sense passing in `teacher_type` as we dont use training period to return the value. The calculation is done with the columns on `Teacher` model.

leandroalemao (approved):

PR #2616, opened 2026-04-09
Prevent multiple ongoing training periods

cpjmcquillan:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/3740

As part of testing before launch, we uncovered a small [bug].

When an ECT is reported as leaving by a school (or reported as joining by another school), their current school are still able to make changes to their training (e.g via training programme or lead provider) until they have left.

This means it is possible to create training histories that don't make sense semantically. Specifically, you can end up with a Teacher with multiple ongoing training periods.

This isn't supposed to be possible!

We have the `trainee_distinct_period` validation, which is supposed to ensure training periods don't overlap (including preventing multiple ongoing periods), but this wasn't working as expected.

Before, the `#siblings` method only accounted for training periods belonging to the associated `*_at_school_period`.

But in this context, we actually care about all the training periods belonging to `*_at_school_periods` for the same Teacher.

This reworks that method so it returns those records. We have also extended the overlap validation to check whether the `finished_on` is blank (which was previously unaccounted for).

As part of this change, we have added some unit tests to the relevant models that explicitly cover concurrent ongoing periods. This should give us more confidence this isn't possible.

[bug]: https://github.com/DFE-Digital/register-ects-project-board/issues/3651

### DIFF
--- a/app/models/concerns/interval.rb
+++ b/app/models/concerns/interval.rb
@@ -60,7 +60,7 @@ module Interval
 
     if siblings.any? { |s| s.range.include?(started_on) }
       errors.add(:started_on, "Start date cannot overlap another #{name} period")
-    elsif siblings.any? { |s| s.range.include?(finished_on) }
+    elsif siblings.any? { |s| s.range.include?(finished_on) } || finished_on.nil?
       errors.add(:finished_on, "End date cannot overlap another #{name} period")
     end
   end
--- a/app/models/training_period.rb
+++ b/app/models/training_period.rb
@@ -138,7 +138,11 @@ class TrainingPeriod < ApplicationRecord
   def siblings
     return TrainingPeriod.none unless at_school_period
 
-    at_school_period.training_periods.excluding(self)
+    training_periods_for_teacher = TrainingPeriod
+      .includes(at_school_period.model_name.element.to_sym)
+      .where(at_school_period.model_name.element => { teacher: })
+
+    training_periods_for_teacher.excluding(self)
   end
 
   def only_expression_of_interest?
--- a/spec/models/ect_at_school_period_spec.rb
+++ b/spec/models/ect_at_school_period_spec.rb
@@ -344,6 +344,62 @@ describe ECTAtSchoolPeriod do
         end
       end
     end
+
+    describe "concurrent ongoing periods" do
+      subject(:save_concurrent_ongoing_period!) do
+        concurrent_ongoing_period.save!
+      end
+
+      before { freeze_time }
+
+      let(:teacher) { FactoryBot.create(:teacher) }
+      let(:school) { FactoryBot.create(:school) }
+
+      let!(:ongoing_ect_at_school_period) do
+        FactoryBot.create(
+          :ect_at_school_period,
+          :ongoing,
+          teacher:,
+          school:,
+          started_on: 1.year.ago
+        )
+      end
+
+      context "at different schools" do
+        let(:other_school) { FactoryBot.create(:school) }
+        let(:concurrent_ongoing_period) do
+          FactoryBot.build(
+            :ect_at_school_period,
+            :ongoing,
+            teacher:,
+            school: other_school,
+            started_on: 1.week.from_now
+          )
+        end
+
+        it do
+          expect { save_concurrent_ongoing_period! }
+            .to raise_error(ActiveRecord::RecordInvalid)
+        end
+      end
+
+      context "at the same school" do
+        let(:concurrent_ongoing_period) do
+          FactoryBot.build(
+            :ect_at_school_period,
+            :ongoing,
+            teacher:,
+            school:,
+            started_on: 1.week.from_now
+          )
+        end
+
+        it do
+          expect { save_concurrent_ongoing_period! }
+            .to raise_error(ActiveRecord::RecordInvalid)
+        end
+      end
+    end
   end
 
   describe "check constraints" do
--- a/spec/models/mentorship_period_spec.rb
+++ b/spec/models/mentorship_period_spec.rb
@@ -93,6 +93,167 @@ describe MentorshipPeriod do
       end
     end
 
+    describe "concurrent ongoing periods" do
+      subject(:save_concurrent_ongoing_period!) do
+        concurrent_ongoing_period.save!
+      end
+
+      before { freeze_time }
+
+      context "for mentees" do
+        let(:mentee_teacher) { FactoryBot.create(:teacher) }
+
+        # Mentor mentoring mentee
+        let(:ect_at_school_period) do
+          FactoryBot.create(
+            :ect_at_school_period,
+            teacher: mentee_teacher,
+            started_on: 1.year.ago,
+            finished_on: 2.weeks.from_now
+          )
+        end
+        let(:mentor_at_school_period) do
+          FactoryBot.create(
+            :mentor_at_school_period,
+            :ongoing,
+            school: ect_at_school_period.school,
+            started_on: 1.year.ago
+          )
+        end
+        let!(:mentorship_period) do
+          FactoryBot.create(
+            :mentorship_period,
+            mentee: ect_at_school_period,
+            mentor: mentor_at_school_period,
+            started_on: 1.year.ago,
+            finished_on: 2.weeks.ago
+          )
+        end
+
+        # Mentee transferring to a new school and being mentored by a different teacher
+        let(:next_ect_at_school_period) do
+          FactoryBot.create(
+            :ect_at_school_period,
+            :ongoing,
+            teacher: mentee_teacher,
+            started_on: ect_at_school_period.finished_on.advance(days: 1)
+          )
+        end
+        let(:next_mentor_at_school_period) do
+          FactoryBot.create(
+            :mentor_at_school_period,
+            :ongoing,
+            school: next_ect_at_school_period.school,
+            started_on: 1.year.ago
+          )
+        end
+        let!(:ongoing_mentorship_period) do
+          FactoryBot.create(
+            :mentorship_period,
+            :ongoing,
+            mentee: next_ect_at_school_period,
+            mentor: next_mentor_at_school_period,
+            started_on: next_ect_at_school_period.started_on
+          )
+        end
+
+        # Mentee being mentored by the original mentor at the old school
+        let(:concurrent_ongoing_period) do
+          FactoryBot.build(
+            :mentorship_period,
+            :ongoing,
+            mentee: ect_at_school_period,
+            mentor: mentor_at_school_period,
+            started_on: mentorship_period.finished_on
+          )
+        end
+
+        it do
+          expect { save_concurrent_ongoing_period! }
+            .to raise_error(ActiveRecord::RecordInvalid)
+        end
+      end
+
+      context "for mentors" do
+        let(:mentor_teacher) { FactoryBot.create(:teacher) }
+
+        # Mentor mentoring mentee at one school
+        let(:ect_at_school_period) do
+          FactoryBot.create(
+            :ect_at_school_period,
+            :ongoing,
+            started_on: 1.year.ago
+          )
+        end
+        let(:mentor_at_school_period) do
+          FactoryBot.create(
+            :mentor_at_school_period,
+            school: ect_at_school_period.school,
+            teacher: mentor_teacher,
+            started_on: 1.year.ago,
+            finished_on: 2.weeks.from_now
+          )
+        end
+        let!(:mentorship_period) do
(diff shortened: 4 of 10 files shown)


app/models/training_period.rb
```diff
@@ -138,7 +138,11 @@ def at_school_period
   def siblings
     return TrainingPeriod.none unless at_school_period
 
-    at_school_period.training_periods.excluding(self)
+    training_periods_for_teacher = TrainingPeriod
+      .includes(at_school_period.model_name.element.to_sym)
+      .where(at_school_period.model_name.element => { teacher: })
```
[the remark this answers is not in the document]
thenapking: I think it is fine as it is.  I think it is easier to read.  `at_school_period.model_name` is obviously just for switching between the two period types.

app/models/concerns/interval.rb
```diff
@@ -62,6 +62,8 @@ def overlap_validation(name:)
       errors.add(:started_on, "Start date cannot overlap another #{name} period")
     elsif siblings.any? { |s| s.range.include?(finished_on) }
       errors.add(:finished_on, "End date cannot overlap another #{name} period")
+    elsif finished_on.nil?
```
thenapking: My devil's advocate question here is: is it sufficient to add this at the validation level, rather than having a database constraint?  Is it possible that race conditions, data migrations or other bits of raw SQL might bypass this?  For example I am thinking of a uniqueness constraint, eg on `ect_at_school_period` where there can only be one period with `NULL` `finished_at` scoped to `teacher_id`.  Not sure how practical that is for each different period.

thenapking:
> Worth noting that all of these tests passed without any code changes **except** the training period tests.

That's right, but it was for registering ECTs.  Is it worth adding a specific test to cover the scenario that discovered this, around an ECT leaving school?

app/models/concerns/interval.rb
cpjmcquillan: I think that makes sense.

There is a uniqueness constraint at the minute, but I'm not sure what its intention is 🤔

https://github.com/DFE-Digital/register-early-career-teachers-public/blob/5064b1054a1e18d4ed857388d4429c79da75d017/db/schema.rb#L1056

If we wanted to add a constraint for training periods, I think we'd want something like this:

```ruby
add_index(
  :training_periods, 
  "(SELECT teacher_id FROM ect_at_school_periods WHERE id = ect_at_school_period_id), ect_at_school_period_id",
  name: "idx_one_active_ect_training_period_per_teacher",
  unique: true,
  where: "ect_at_school_period_id IS NOT NULL AND finished_on IS NULL"
)
```

for ECTs, plus the equivalent for mentors.

It feels a little complex (since `teacher_id` isn't directly on the `training_periods` table), but perhaps that isn't an issue.

Let's see what the rest of the team think!

thenapking: I'm guessing that unique constraint is to ensure that an ECT can only have one ongoing period with each mentor.  The question is how slow the index will be with the SELECT query.  As you say, we don't really want to denormalise `training_periods` to add the `teacher_id`, which would definitely make the index easier to create.

spec/services/schools/register_ect_spec.rb
```diff
@@ -103,36 +170,34 @@
           end
         end
 
-        context "when a Teacher record with the same TRN exists and has multiple ect records at different schools" do
+        context "when a Teacher has multiple ECT periods at different schools" do
           let(:school_one) { FactoryBot.create(:school) }
           let(:school_two) { FactoryBot.create(:school) }
-          let(:started_on) { Date.current + 1.day }
-          let!(:future_contract_period) { FactoryBot.create(:contract_period, :with_schedules, year: started_on.year) }
-          let!(:future_active_lead_provider) { FactoryBot.create(:active_lead_provider, lead_provider:, contract_period: future_contract_period) }
```
cpjmcquillan, line 111: These didn't seem to be doing anything. 

I don't want to be the guy that just deletes stuff he doesn't understand, _but_ removing this has no impact on the outcome of the test.

app/models/concerns/interval.rb
```diff
@@ -60,7 +60,7 @@ def overlap_validation(name:)
 
     if siblings.any? { |s| s.range.include?(started_on) }
       errors.add(:started_on, "Start date cannot overlap another #{name} period")
-    elsif siblings.any? { |s| s.range.include?(finished_on) }
+    elsif siblings.any? { |s| s.range.include?(finished_on) } || finished_on.nil?
```
cpjmcquillan, line 63: It's possible for there to be siblings, but for none of the sibling ranges to include the `started_on` or `finished_on`.

This meant errors could have slipped through the cracks silently!

I've added explicit tests for each period, and made this change (combined with [reworking the `TrainingPeriod#siblings` ⤵️](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/6801345d54cd09c33d17adc70428d71cd83b64fc/app/models/training_period.rb#L138-L146)) so we should avoid any more weird scenarios (like this [bug](https://github.com/DFE-Digital/register-ects-project-board/issues/3651)).

**But perhaps we want to be more rigorous with these overlap validations?**

We could add an `else` clause to this conditional that raises an error to catch this at the application layer, or we could add some database constraints for some of these more complex scenarios, or both.

What do you think? @peteryates

cpjmcquillan:
> > Worth noting that all of these tests passed without any code changes **except** the training period tests.
> 
> That's right, but it was for registering ECTs. Is it worth adding a specific test to cover the scenario that discovered this, around an ECT leaving school?

I'll cover that as part of https://github.com/DFE-Digital/register-ects-project-board/issues/3651 👌🏼

app/models/concerns/interval.rb
peteryates: Originally the constraints were added before we thought we'd be storing future (potential) periods along side actual periods.

A teacher should be allowed one ongoing:
* ECT at school period
* training period as an ECT
* mentor at school at school period **per school**
* training period as a mentor

Now we're able to add training periods to both current and future at school periods, we shouldn't be thinking in terms of siblings but cousins.

I wonder if we can make overlap validation take an extra argument that supplies the things it might overlap with, which will give us finer control over how it works. Or, just remove the method and introduce a little duplication across the `Periods` classes, but the contents will be simpler as we're just dealing with one case at a time.

Happy to have a chat about options here!

I think what's here in this PR is a good first step regardless though.

spec/services/ect_at_school_periods/switch_mentor_spec.rb
```diff
@@ -245,7 +245,7 @@ module ECTAtSchoolPeriods
         end
 
         context "on the last day of the contract period" do
-          let(:travel_date) { Date.new(2026, 5, 31) }
```
thenapking, line 248: Just to note that May 31st is hardcoded into the contract period factory as the end date of the contract period, so there should be no problem with this being hardcoded here.

app/models/concerns/interval.rb
peteryates: Sorry, missed this. I think my other comment overlapped it though in that we need to think about cousins rather than siblings here so maybe this method needs a slight rejig.

spec/services/schools/register_ect_spec.rb
thenapking: I've double checked this, and these are in fact identical to the current contract period, because they use the started_on year.  They aren't as I had initially assumed the next contract period.

spec/services/schools/register_ect_spec.rb
```diff
@@ -155,67 +220,14 @@
...
+            let(:started_on) { contract_period.finished_on }
 
-            it "registers the teach with the correct contrat period" do
+            it "allows registration at the current school" do
               expect { service.register! }.to change(ECTAtSchoolPeriod, :count).by(1)
             end
           end
         end
 
-        context "when a Teacher record with the same TRN has a future period at different school" do
```
thenapking, line 167: Although something's probably not right with the set up of this test, with the future periods, I think we might still need to test this scenario.  Shouldn't we be asserting here that with a future date, the service closes any other periods to ensure there are not multiple ongoing periods?

spec/models/ect_at_school_period_spec.rb
```diff
@@ -344,6 +344,62 @@
...
+      end
+
+      context "at the same school" do
+        let(:concurrent_ongoing_period) do
+          FactoryBot.build(
+            :ect_at_school_period,
+            :ongoing,
+            teacher:,
+            school:,
+            started_on: 1.week.from_now
```
thenapking, line 393: I think we might need more tests around the start date.  So a new ongoing period which starts before, and on the start date should also raise the error.

spec/models/mentorship_period_spec.rb
```diff
@@ -93,6 +93,162 @@
...
+        end
+        let(:mentor_at_school_period) do
+          FactoryBot.create(
+            :mentor_at_school_period,
+            :ongoing,
+            school: ect_at_school_period.school,
+            started_on: 1.year.ago
+          )
+        end
+        let!(:mentorship_period) do
```
thenapking, line 123: It was not clear to me at first why these tests are different from the ect and mentor period tests.  Perhaps you could rename some of the test objects to make it clear we are testing logic around sibling periods with overlapping, but not identical start/end dates.

app/models/concerns/interval.rb
thenapking: I'm just asking this because it took some time for this problem to surface on the database layer.  Are we absolutely sure that `range` here is end-inclusive?  I *think* it is but I just want to double check.

thenapking (approved):
Looks good.  There's a couple of small things perhaps to address and a couple of queries I wanted to raise.

spec/services/ect_at_school_periods/switch_mentor_spec.rb
cpjmcquillan: It's not that 31st May isn't significant today, and that was the significant date for the bug this addressed, but rather the only reason 31st May is significant is because it's the **last date of the current contract period**.

This felt clearer, and will still give the same value once we have a 2027 contract period.

app/models/concerns/interval.rb
cpjmcquillan: I'm going to add an issue to the backlog to revisit these overlap validations. I think a little duplication and making things really explicit would be helpful and remove some ambiguity.

For launch, I'll move forward with this to mitigate any "bad" data being created.

app/models/concerns/interval.rb
cpjmcquillan: I think wires have got a little crossed here.

Currently there is a mismatch between `has_overlap_with_siblings?` and this conditional that adds the error message.

There can be overlaps where the start or end are not in the range —specifically when there are multiple ongoing periods (the `finished_on` is `nil`).

A range with nil will resolve to something like `14 Jan 2026..Infinity`, and `nil` isn't in that range.

This fills in the gap in that logic, so now we add an error in every scenario we expect the validation to fail.

Definitely agree the `siblings` terminology is confusing in this instance, though. I'm hoping to circle back and make things clearer ASAP!

spec/models/ect_at_school_period_spec.rb
cpjmcquillan: I think we're fine for now. The start date is arbitrary, we just care about them both being "ongoing".

I chose the dates so the tests told a story that had some basis in reality (i.e original ongoing period started in the past, with a "concurrent" ongoing period starting in the future). 

Also, to avoid the error being raised because of overlapping start dates.

We already have a bunch of "sibling" specs that test a variety of dates. These are intended to act as regression tests that protect us specifically against multiple "ongoing" periods.

These tests already passed for all periods, except `TrainingPeriod`.

spec/services/schools/register_ect_spec.rb
cpjmcquillan: I think we need to decide what to do in this scenario. 

This corresponds to multiple schools trying to register an ECT.

For example,

- School A registers ECT on 10th September 2025
- School A reports ECT leaving on 31st December 2025
- School B registers ECT as starting on 10th January 2026
- School C registers ECT as starting on 5th January 2026

How do we know which is correct?

I think it's reasonable for us to raise an error in this scenario. That prevents the third school from registering the ECT, and should prompt them to contact support if it is something that needs resolving.

I've added a test to assert that explicitly as part of https://github.com/DFE-Digital/register-ects-project-board/issues/3740

spec/models/mentorship_period_spec.rb
cpjmcquillan: I've added some comments that should clear things up a little bit here 👌🏼

PR #2617, opened 2026-04-09
[2786] Unhide sections of the teacher page in the admin console

thenapking:
### Context

For launch we want the admin view of a teacher to have the secondary nav visible, with Overview as the first item, selected by default.  Currently the nav is turned off in production by a flag, and we link to the Induction panel by default.  [This PR](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/1735) updated the links.  I have reverted all of these from `admin_teacher_induction_path` to `admin_teacher_path` except where the admin user should actually be directed back to the `admin_teacher_induction_path`, eg when they've created a new induction period.  I *think* I have selected the correct ones, but would appreciate a second pair of eyes.

As per @peteryates comment this should not be merged now, but should be ready to merge into a branch we can use for launching the service.


spec/views/admin/teachers/timeline/show.html.erb_spec.rb
```diff
@@ -0,0 +1,13 @@
+RSpec.describe "admin/teachers/timeline/show.html.erb" do
+  let(:teacher) { FactoryBot.create(:teacher) }
+
+  before do
+    assign(:teacher, teacher)
+    assign(:events, [])
+    render
+  end
+
+  it "includes a back link to the teacher induction page" do
```
cpjmcquillan:
```suggestion
  it "includes a back link to the teacher page" do
```

cpjmcquillan (approved):

peteryates (approved):

PR #2618, opened 2026-04-09
[3642] Allow future event dates

avinhurry:
### Context

Remove the `Event` validation that rejected future `happened_at` values.

This fixes [failures](https://dfe-teacher-services.sentry.io/issues/6871425829/?environment=migration&project=4508369974788096&query=is%3Aunresolved&referrer=issue-stream) when future dated leaving events are recorded asynchronously by `RecordEventJob`.

### Changes proposed

Remove the custom past only validation from `Event`

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -108,7 +108,6 @@ class Event < ApplicationRecord
   validates :event_type, presence: true, inclusion: { in: EVENT_TYPES }
 
   validate :check_author_present
-  validate :event_happened_in_the_past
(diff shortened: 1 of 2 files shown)


peteryates (approved):
Maybe we should rename the column to `possibly_happend_at` too?

avinhurry:
> Maybe we should rename the column to `possibly_happend_at` too?

@peteryates 😆 continuing the theme of making things less sensible...

PR #2619, opened 2026-04-09
[3646] Fix disparity on `updated_at` on GET schools between ECF1 and RECT, so `updated_at` only shows cohort specific `updated_at` changes in RECT

leandroalemao:
### Context

Ticket: [3646](https://github.com/DFE-Digital/register-ects-project-board/issues/3646)

The school `updated_at` should reflect the filtered cohort in the schools API response.
In ECF it does this by looking at the most recent `updated_at` from the school or the school cohort.

- De-couple the metadata so that it doesn't touch the school's `api_updated_at`;
- Add a new `api_updated_at` to the metadata;
- Touch the metadata `api_updated_at` when relevant/serialized school attribute change or when the `in_partnership` and `induction_programme_choice` changes;
- Surface the metadata `api_updated_at` in the serializer and use it in the query;
- Remove the school `api_updated_at`;
- Create the metadata for each contract period and populate the `api_updated_at` in the migrator instead of the school's `api_updated_at` (using the max of the ECF school `updated_at` and ECF school cohort `updated_at` as in the ECF serializer);

### Guidance to review

Review app

### DIFF
--- a/app/migration/migrators/school.rb
+++ b/app/migration/migrators/school.rb
@@ -63,8 +63,7 @@ module Migrators
 
         [
           compare_fields(gias_school:, ecf_school:),
-          update_school!(school: gias_school.school, ecf_school:),
-          create_or_update_contract_period_metadata!(school: gias_school.school, ecf_school:),
+          update_school!(school: gias_school.school, ecf_school:)
         ].all?
       end
     end
@@ -126,16 +125,6 @@ module Migrators
       }
 
       school.update_columns(attrs)
-    end
-
-    def create_or_update_contract_period_metadata!(school:, ecf_school:)
-      Metadata::Manager.new.refresh_metadata!(school)
-
-      ecf_school.school_cohorts.map { |school_cohort|
-        contract_period_year = school_cohort.cohort.start_year
-        metadata = school.contract_period_metadata.find { it.contract_period_year == contract_period_year }
-        metadata.update!(api_updated_at: [school.updated_at, school_cohort.updated_at].max)
-      }.all?
     end
   end
 end
--- a/app/migration/migrators/school.rb
+++ b/app/migration/migrators/school.rb
@@ -13,7 +13,7 @@ module Migrators
 
     MISMATCH_FIELD_MESSAGE = ->(school, field, gias_value, ecf_value) { ":#{field} - School #{school.urn} (#{school.name}) mismatch value on field named '#{field}': '#{ecf_value}' on ECF whilst '#{gias_value}' expected on RECT!" }
 
-    def self.dependencies = %i[gias_import gias_childrens_centres]
+    def self.dependencies = %i[contract_period gias_import gias_childrens_centres]
 
     def self.model = :school
 
@@ -64,6 +64,7 @@ module Migrators
         [
           compare_fields(gias_school:, ecf_school:),
           update_school!(school: gias_school.school, ecf_school:),
+          create_or_update_contract_period_metadata!(school: gias_school.school, ecf_school:),
         ].all?
       end
     end
@@ -121,10 +122,20 @@ module Migrators
         api_id: ecf_school.id,
         induction_tutor_name: induction_coordinator&.full_name,
         induction_tutor_email: induction_coordinator&.email,
-        created_at: ecf_school.created_at,
-        api_updated_at: ecf_school.updated_at
+        created_at: ecf_school.created_at
       }
+
       school.update_columns(attrs)
+    end
+
+    def create_or_update_contract_period_metadata!(school:, ecf_school:)
+      Metadata::Manager.new.refresh_metadata!(school)
+
+      ecf_school.school_cohorts.map { |school_cohort|
+        contract_period_year = school_cohort.cohort.start_year
+        metadata = school.contract_period_metadata.find { it.contract_period_year == contract_period_year }
+        metadata.update!(api_updated_at: [school.updated_at, school_cohort.updated_at].max)
+      }.all?
     end
   end
 end
--- a/app/models/metadata/school_lead_provider_contract_period.rb
+++ b/app/models/metadata/school_lead_provider_contract_period.rb
@@ -7,6 +7,7 @@ module Metadata
     belongs_to :school
     belongs_to :lead_provider
     belongs_to :contract_period, foreign_key: :contract_period_year
+    has_many :contract_period_metadata, class_name: "Metadata::SchoolContractPeriod", through: :school
 
     validates :school, presence: true
     validates :lead_provider, presence: true
@@ -14,6 +15,6 @@ module Metadata
     validates :expression_of_interest_or_school_partnership, inclusion: { in: [true, false] }
     validates :school_id, uniqueness: { scope: %i[lead_provider_id contract_period_year] }
 
-    touch -> { school }, on_event: :update, when_changing: %i[expression_of_interest_or_school_partnership], timestamp_attribute: :api_updated_at
+    touch -> { contract_period_metadata }, on_event: :update, when_changing: %i[expression_of_interest_or_school_partnership], timestamp_attribute: :api_updated_at
   end
 end
--- a/app/services/metadata/manager.rb
+++ b/app/services/metadata/manager.rb
@@ -9,8 +9,9 @@ module Metadata
     end
 
     class << self
-      def refresh_all_metadata!(async: false, track_changes: false)
-        Resolver.all_handlers.each { it.refresh_all_metadata!(async:, track_changes:) }
+      def refresh_all_metadata!(async: false, track_changes: false, excluding_handlers: [])
+        handlers = Resolver.all_handlers.reject { |handler| handler.in?(excluding_handlers) }
+        handlers.each { it.refresh_all_metadata!(async:, track_changes:) }
       end
 
       def destroy_all_metadata!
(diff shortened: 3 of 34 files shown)


db/migrate/20260408171244_switch_not_null_from_metadata_schools_contract_periods.rb
```diff
@@ -0,0 +1,9 @@
+class SwitchNotNullFromMetadataSchoolsContractPeriods < ActiveRecord::Migration[8.0]
+  def change
+    change_table :metadata_schools_contract_periods, bulk: true do |t|
+      t.change_null :contract_period_year, false
+      t.change_null :in_partnership, true
```
ethax-ross: Why do these need to be nullable? I don't think we'd ever want to create the metadata without these values?

app/models/metadata/school_lead_provider_contract_period.rb
```diff
@@ -7,14 +7,15 @@ class SchoolLeadProviderContractPeriod < Metadata::Base
...
 
     validates :school, presence: true
     validates :lead_provider, presence: true
     validates :contract_period, presence: true
     validates :expression_of_interest_or_school_partnership, inclusion: { in: [true, false] }
     validates :school_id, uniqueness: { scope: %i[lead_provider_id contract_period_year] }
 
-    touch -> { school }, on_event: %i[create destroy], timestamp_attribute: :api_updated_at
-    touch -> { school }, on_event: :update, when_changing: %i[expression_of_interest_or_school_partnership], timestamp_attribute: :api_updated_at
+    touch -> { contract_period_metadata }, on_event: %i[create destroy], timestamp_attribute: :api_updated_at
```
ethax-ross: I think since the metadata would always be created/destroyed with the school we can probably drop this?

app/migration/migrators/school.rb
```diff
@@ -121,10 +122,22 @@ def update_school!(school:, ecf_school:)
...
-        created_at: ecf_school.created_at,
-        api_updated_at: ecf_school.updated_at
+        created_at: ecf_school.created_at
       }
+
       school.update_columns(attrs)
     end
+
+    def create_or_update_contract_period_metadata!(school:, ecf_school:)
+      ecf_school.school_cohorts.map { |school_cohort|
```
ethax-ross: I think I see why you made the fields nullable now 🙂 it looks like we're replicating some of what the handler already does; instead could we just run it through the metadata manager to create it? so something like:

```
Metadata::Manager.refresh_metadata!(school)
```

spec/migration/migrators/school_spec.rb
```diff
@@ -29,10 +30,16 @@ def create_gias_school(ecf_school)
                       ukprn: ecf_school.ukprn)
   end
 
-  it_behaves_like "a migrator", :school, %i[gias_import gias_childrens_centres] do
+  it_behaves_like "a migrator", :school, %i[contract_period gias_import gias_childrens_centres], creates_metadata: true do
     def create_migration_resource = create_ecf_school
 
-    def create_resource(migration_resource) = create_gias_school(migration_resource)
+    def create_resource(migration_resource)
+      migration_resource.school_cohorts.each do |school_cohort|
```
ethax-ross, line 37: Instead of creating the metadata here can we have a test that ensures the migration creates metadata for the school? I think we're maybe not exercising the new code by doing this?

spec/requests/api/v3/schools_spec.rb
```diff
@@ -26,9 +26,17 @@ def apply_expected_order(resources)
     it_behaves_like "a token authenticated endpoint", :get
     it_behaves_like "an index endpoint"
     it_behaves_like "a paginated endpoint"
-    it_behaves_like "a sortable endpoint"
+    it_behaves_like "a sortable endpoint", updated_at_column: "contract_period_metadata.api_updated_at" do
+      def set_updated_at(resource:, updated_at_column:, value:)
```
ethax-ross: If we're overriding this could we not just pass `api_updated_at` into the `updated_at_column`? so it then becomes something like:

```
it_behaves_like "a sortable endpoint", updated_at_column: :api_updated_at do
  def set_updated_at(resource:, updated_at_column:, value:)
    resource.contract_period_metadata.update_all(updated_at_column => value)
  end
end
```

spec/support/shared_contexts/api/sortable_endpoint.rb
```diff
@@ -1,13 +1,23 @@
...
+      travel_to(5.days.ago) { create_resource(active_lead_provider:) }.tap { set_updated_at(resource: it, updated_at_column:, value: 1.day.ago) },
     ]
   end
   let(:updated_at_sort_attribute) { updated_at_column.to_s }
 
+  def set_updated_at(resource:, updated_at_column:, value:)
+    resource.update_columns("#{updated_at_column}": value)
+  end
+
+  def resolve_sort_value(resource, sort_attribute)
```
ethax-ross: Instead of having this method which is pretty hard to follow, would it be cleaner to extract the sorting into a method that can be overridden in the cases where the sorting is obscure/not necessarily on the main model, so something like:

```
def sort_resources(resources, sort_attribute)
  resources.sort_by!(&:"#{sort_attribute}")
end

sort_resources(resources).tap { |l| l.reverse! if sort[0] == "-" }
```

spec/support/shared_examples/migrator_support.rb
```diff
@@ -159,21 +159,23 @@
...
     end
 
-    it "does not create any metadata" do
-      dump_metadata_database_state = -> {
-        ActiveRecord::Base.connection.tables.select { it.start_with?("metadata_") }.to_h do |table|
-          rows = ActiveRecord::Base.connection.execute("SELECT * FROM #{table}").to_a
-          [table, rows]
-        end
-      }
+    unless creates_metadata
```
ethax-ross: Worth adding a opposing test that covers when this is `true` just to test that some/any metadata is created? (with the specific migrator specs checking the right metadata is created/values)

ethax-ross (commented):
Looks good overall thanks Leandro, just a few comments/suggestions.

Can we also deploy this to the parity check environment if you haven't already and see how long a full async metadata refresh takes (maybe after addressing any comments as some might effect the runtime)? As we're touching from the metadata now I want to check its not going to balloon the runtime.

app/migration/migrators/school.rb
leandroalemao: Hi @ethax-ross we can't call the Metadata handler here as we don't have the instance of `SchoolCohort` from ECF1 in the Metadata handler. The only place/time we have it is in the migrator in order to get `school_cohort.updated_at`.

db/migrate/20260408171244_switch_not_null_from_metadata_schools_contract_periods.rb
leandroalemao: Hi @ethax-ross for the same reason as https://github.com/DFE-Digital/register-early-career-teachers-public/pull/2619#discussion_r3082776407

app/migration/migrators/school.rb
ethax-ross: Ah I see. I'd still rather avoid making those fields nullable if possible 🤔 can we get the metadata handler to create the metadata and then just update the `api_updated_at` in the migrator? something like:

```
Metadata::Manager.refresh_metadata!(school)

ecf_school.school_cohorts.each do |school_cohort| 
  contract_period_year = school_cohort.cohort.start_year
  metadata = school.contract_period_metadata.find { it.contract_period_year == contract_period_year }
  metadata.update!(api_updated_at: [school.updated_at, school_cohort.updated_at].max }
end
```

spec/requests/api/v3/schools_spec.rb
```diff
@@ -26,9 +26,23 @@ def apply_expected_order(resources)
     it_behaves_like "a token authenticated endpoint", :get
     it_behaves_like "an index endpoint"
     it_behaves_like "a paginated endpoint"
-    it_behaves_like "a sortable endpoint"
+    it_behaves_like "a sortable endpoint", updated_at_column: :api_updated_at do
```
ethax-ross: I think we can drop passing in the `updated_at_column` and go back to hard coding it in the shared context? if a particular request spec like this one uses a different column it would override `set_updated_at` anyway and it would make it easier to follow if its explicit in there I think?

spec/support/shared_contexts/api/filterable_endpoint.rb
```diff
@@ -276,11 +276,15 @@ def set_updated_at(resource:, updated_at_column:, value:)
   end
 end
 
-RSpec.shared_examples "a does not filter by updated_since endpoint" do
+RSpec.shared_examples "a does not filter by updated_since endpoint" do |updated_at_column: :api_updated_at|
```
ethax-ross: Ditto for not passing in `updated_at_column`

ethax-ross (commented):
Just some minor points then I think its good to go providing we have tested on the parity check env, thanks Leandro

ethax-ross (approved):

mooktakim (approved):

app/services/metadata/manager.rb
```diff
@@ -9,13 +9,20 @@ def refresh_metadata!(objects, track_changes: false)
...
+        handlers = Resolver.all_handlers.reject { |handler| excluding_handlers.include?(demodulize_handler_name(handler)) }
+        handlers.each { it.refresh_all_metadata!(async:, track_changes:) }
       end
 
       def destroy_all_metadata!
         Resolver.all_handlers.each(&:destroy_all_metadata!)
       end
+
+    private
+
```
ethax-ross: Might be cleaner having this method on the `Base` handler, so it becomes:

```
handler.demodularized_name.in?(excluding_handlers)
```

Or we could pass the full handler class names into `excluding_handlers` and use:

```
handler.in?(excluding_handlers)
```

I don't feel strongly about it though its fine as is

ethax-ross (approved):

ethax-ross (approved):

PR #2626, opened 2026-04-10
Fix missing delivery partner name on Check Your Answers page

ericaporter:
### Context

When a school registers a new ECT and chooses to reuse their previous programme choices, the Check Your Answers page was showing a blank Delivery partner row despite a confirmed partnership existing for the current contract period. 

When a school reuses previous programme choices, the lead provider step is skipped so lead_provider_id is never stored in the session. The delivery partner therefore has to be looked up from the confirmed current year school partnership, but delivery_partner_name was never implemented on RegistrationStore, causing it to silently return nil.

- Added confirmed_delivery_partner_for_contract_period(school:) to Queries
- Added delivery_partner_name(school:) to RegistrationStore
- Updated the check answers view to call @ect.delivery_partner_name(school: @school) when rendering the delivery partner row
- Added a unit test to queries_spec.rb for confirmed_delivery_partner_for_contract_period
- Updated the feature spec to reflect a realistic data setup
- Updated ReusablePartnershipHelpers to create a current year lead_provider_delivery_partnership and school_partnership alongside the existing previous year ones

### DIFF
--- a/app/wizards/schools/register_ect_wizard/registration_store/queries.rb
+++ b/app/wizards/schools/register_ect_wizard/registration_store/queries.rb
@@ -96,6 +96,12 @@ module Schools
           previous_training_period&.lead_provider
         end
 
+        def confirmed_delivery_partner_for_contract_period(school:)
+          lead_provider_partnerships_for_contract_period(school:)
+          .first
+          &.delivery_partner
+        end
+
       private
 
         attr_reader :registration_store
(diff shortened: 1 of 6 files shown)


app/wizards/schools/register_ect_wizard/registration_store/queries.rb
```diff
@@ -96,6 +96,12 @@ def previous_lead_provider
           previous_training_period&.lead_provider
         end
 
+        def confirmed_delivery_partner_for_contract_period(school:)
+          lead_provider_partnerships_for_contract_period(school:)
+          .first
+          &.delivery_partner
+        end
```
cpjmcquillan, line 103: Surprised this isn't picked up by a linter, but how do you feel about indenting this so it's a little easier to read? 💭

```suggestion
        def confirmed_delivery_partner_for_contract_period(school:)
          lead_provider_partnerships_for_contract_period(school:)
            .first
            &.delivery_partner
        end
```

cpjmcquillan (approved):

thenapking (approved):
LGTM

PR #2628, opened 2026-04-10
[2317] UUID presentation  

### LINKED ISSUE #2317: Format UUIDs in `<code>` tags
UUIDs are long strings of characters and numbers. They're easier to read in a monospaced font, we should probably add a `format_uuid` helper that wraps it in a `tag.code(uuid)` so we can show them consistently throughout the app.

| API participant ID | Declaration ID |
| ------ | -----------|
|  Image | Image |

I spotted this in the admin / teachers / declarations section - we might have a few others floating around too.

peterdavidhamilton:
### Context

This change adds a `format_uuid` helper which

* wraps the UUID in a code tag
* adds a shortened version of the UUID for screenreader users (see [this comment](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/2628#issuecomment-4305685105))

### Guidance to review

Is the change actually better for visually impaired users?

Are there any sneaky UUIDs missing?

Note, it also fixed this weird link underline glitch where the line is broken.

| Before | After | 
| ------ | ------- |
| Screenshot From 2026-05-07 15-53-50|Screenshot From 2026-05-07 15-54-13 |

### DIFF
--- a/app/components/admin/teachers/overview_summary_component.rb
+++ b/app/components/admin/teachers/overview_summary_component.rb
@@ -71,7 +71,7 @@ module Admin
       end
 
       def api_participant_id_row
-        { key: "API participant ID", value: content_tag(:code, teacher.api_participant_id, class: "app-code") }
+        { key: "API participant ID", value: helpers.format_uuid(teacher.api_participant_id) }
       end
 
       def current_school_value
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -92,4 +92,11 @@ module ApplicationHelper
   def govuk_html_element(&block)
     tag.html(lang: "en", class: %w[govuk-template govuk-template--rebranded], &block)
   end
+
+  def format_uuid(uuid)
+    visually_hidden_span = govuk_visually_hidden("Identifier starting with #{uuid.first(6)}")
+    aria_hidden_span = tag.span(uuid, aria: { hidden: true })
+
+    tag.code(safe_join([visually_hidden_span, aria_hidden_span]))
+  end
 end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -94,9 +94,8 @@ module ApplicationHelper
   end
 
   def format_uuid(uuid)
-    visually_hidden_span = govuk_visually_hidden("Identifier starting with #{uuid.first(6)}")
-    aria_hidden_span = tag.span(uuid, aria: { hidden: true })
+    return if uuid.blank?
 
-    tag.code(safe_join([visually_hidden_span, aria_hidden_span]))
+    tag.code(uuid)
   end
 end
--- a/app/views/admin/appropriate_bodies/show.html.erb
+++ b/app/views/admin/appropriate_bodies/show.html.erb
@@ -6,7 +6,3 @@
 %>
 
 <%= render Admin::AppropriateBodies::DetailsComponent.new(appropriate_body_period: @appropriate_body) %>
-
-<p class='govuk-body'>
-  <%# govuk_link_to('Timeline of events', admin_appropriate_body_timeline_path(@appropriate_body)) %>
-</p>
--- a/app/views/admin/dfe_sign_in_organisations/show.html.erb
+++ b/app/views/admin/dfe_sign_in_organisations/show.html.erb
@@ -1,7 +1,7 @@
 <%
   page_data(
     title: @organisation.name,
-    caption: "DFE Sign-In Organisation ID: #{@organisation.uuid}"
+    caption: "DFE Sign-In Organisation ID: #{format_uuid(@organisation.uuid)}"
   )
 %>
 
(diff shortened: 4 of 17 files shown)


app/helpers/application_helper.rb
```diff
@@ -92,4 +92,9 @@ def trs_alerts_text(alerts_present)
   def govuk_html_element(&block)
     tag.html(lang: "en", class: %w[govuk-template govuk-template--rebranded], &block)
   end
+
+  def format_uuid(uuid)
+    # tag.code(uuid, class: "app-code")
+    content_tag(:code, uuid, class: "app-code") }
```
peteryates: If `:code` isn't dynamic here `tag.code` makes more sense I think.

peteryates:
Thanks for picking this up! :)

peterdavidhamilton:
@peteryates we have `app-mono` and `app-code` styles. Virtually identical but increases the scope of work if you want them consolidating. Unless there is a reason we have both I think it is convergent evolution.

peteryates:
I think we should just have one. I think using `<code>` is fine, it's intended for computer code and UUIDs aren't _code_, but tbh they're not really intended for human consumption.

peteryates:
Had a chat about this in the RECT channel and from an accessibility point of view think this might be the best approach:


```
<code>
  <span class="govuk-visually-hidden">Identifier starting with 757b3a</span>
  <span aria-hidden="true">757b3a93-9e7d-426c-81d6-891846e8db95</span>
</code>
```

Without it, screen readers just read the whole UUID out which will be super confusing. If screenreader users _do_ actually want the full UUID, they can still expand the accordion and get at it.

https://github.com/user-attachments/assets/6403b5a1-45ab-449d-8d90-cc69486f8767

app/helpers/application_helper.rb
```diff
@@ -92,4 +92,11 @@ def trs_alerts_text(alerts_present)
   def govuk_html_element(&block)
     tag.html(lang: "en", class: %w[govuk-template govuk-template--rebranded], &block)
   end
+
+  def format_uuid(uuid)
+    visually_hidden_span = govuk_visually_hidden("Identifier starting with #{uuid.first(6)}")
```
peterdavidhamilton: 🙌

app/views/admin/appropriate_bodies/show.html.erb
```diff
@@ -6,7 +6,3 @@
 %>
 
 <%= render Admin::AppropriateBodies::DetailsComponent.new(appropriate_body_period: @appropriate_body) %>
-
-<p class='govuk-body'>
```
peterdavidhamilton, line 10: 👍🏼

peterdavidhamilton (commented):
So I assume we then don't need either `.app-code` or `.app-mono`?

app/components/admin/teachers/overview_summary_component.rb
```diff
@@ -71,7 +71,7 @@ def induction_status_row
       end
 
       def api_participant_id_row
-        { key: "API participant ID", value: content_tag(:code, teacher.api_participant_id, class: "app-code") }
+        { key: "API participant ID", value: helpers.format_uuid(teacher.api_participant_id) }
```
avinhurry, line 74: `api_participant_id` falls back to "Not available", that means we end up formatting a non UUID string as a UUID.

Image

app/views/admin/dfe_sign_in_organisations/show.html.erb
```diff
@@ -1,7 +1,7 @@
 <%
   page_data(
     title: @organisation.name,
-    caption: "DFE Sign-In Organisation ID: #{@organisation.uuid}"
+    caption: "DFE Sign-In Organisation ID: #{format_uuid(@organisation.uuid)}"
```
avinhurry: The helper can't safely be interpolated into caption, `page_data` wraps captions in `tag.span(...)`, so any `format_uuid(...)` HTML passed through a caption string gets escaped and rendered as literal markup.

Image

avinhurry (commented):
This makes me slightly uncomfortable because it gives screen reader users a less complete experience. The current experience is noisy, but it does at least expose the full identifier if needed.

[MDN guidance](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-hidden) says:

> Only when identical or equivalent meaning and functionality is exposed to assistive technologies can removing visible content from the accessibility API be considered.

Not blocking, just food for thought 💭

app/components/admin/teachers/overview_summary_component.rb
peteryates: Ah well caught. It's a bit odd that a method called `_id` can return something other than the ID :thinking: 

Perhaps we should move the `format_uuid` to the presenter.

peteryates: Or just leave it blank? It's missing rather than not available. Not available suggests it soon might be!

peterdavidhamilton: Ternary formatted UUID or some text? But I agree, the wording makes it sound like a temporary omission.

PR #2635, opened 2026-04-12
Security strengthening

peteryates:
Some security tweaks based on the ITHC findings. Nothing major here, just some tweaking of headers etc.

- **Remove https from content security policy**
- **Set up hsts header**
- **Enable the permissions_policy**

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -44,7 +44,19 @@ Rails.application.configure do
 
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
   config.force_ssl = true
-  config.ssl_options = { redirect: { exclude: ->(request) { request.path.include?("/healthcheck") } } }
+  config.ssl_options = {
+    hsts: {
+      preload: false,
+      expires: 3.hours
+      # FIXME: when we're happy everything's working correctly, replace the setings above
+      #        with the ones below
+      # preload: true,
+      # expires: 1.year
+    },
+    redirect: {
+      exclude: ->(request) { request.path.start_with?("/healthcheck") },
+    },
+  }
 
   # Log to STDOUT by default
   config.logger = ActiveSupport::Logger.new($stdout)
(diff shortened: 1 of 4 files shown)


cpjmcquillan (approved):

config/environments/production.rb
```diff
@@ -44,7 +44,15 @@
 
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
   config.force_ssl = true
-  config.ssl_options = { redirect: { exclude: ->(request) { request.path.include?("/healthcheck") } } }
+  config.ssl_options = {
+    hsts: {
+      preload: true,
+      expires: 3.hours # FIXME: increase this when we're happy it's working correctly
```
avinhurry: I'm not sure what "working correctly" means here. What are the criteria for increasing this and what would we increase it to?

avinhurry (approved):
Looks good overall. I just had a question around the rollout logistics.

config/environments/production.rb
peteryates: So I was doing a bit of research while implementing it and on [this post](https://medium.com/@2nick2patel2/fastapi-hsts-hpkp-csp-playbook-ship-secure-by-default-apis-without-breaking-browsers-b8170811c1ff) there was this excerpt which frightened me a little :sweat_smile: 

> ```
> Strict-Transport-Security: max-age=300
> ```
> That’s 5 minutes. Why so small? Because HSTS is sticky. If you accidentally ship it on the wrong domain (or break TLS), you can lock users into failure until max-age expires.

So I thought let's go with a few hours and then maybe bump it to a month (or a year) if there are 0 complaints.

avinhurry: Makes sense! From that article: 

> Preload has strict requirements: max-age must be at least 31536000 (1 year),

Would it make sense to keep `preload: false` until we increase to a year or maybe two years (since thats the[ rails recommended and default](https://api.rubyonrails.org/v8.0.0.1/classes/ActionDispatch/SSL.html)) otherwise it feels a bit misleading.

peteryates: Done.

avinhurry: Nice, informative commit message. Thanks!

PR #2639, opened 2026-04-13
Disable schools interface in production

cpjmcquillan:
This was enabled in #2617, but I don't think it should have been!

We probably only want to turn this on once we're about to go live.


avinhurry (approved):

peteryates (approved):

peteryates:
Let's go with the reversion one instead DFE-Digital/register-early-career-teachers-public#2640 as some bits of the admin UI will be affected too.

PR #2640, opened 2026-04-13
Revert "[2786] Unhide sections of the teacher page in the admin console"

thenapking:
Reverts DFE-Digital/register-early-career-teachers-public#2617


peteryates (approved):

config/terraform/application/config/production.yml
```diff
@@ -8,7 +8,6 @@ ENABLE_BLAZER: true
 ENABLE_API: false
 ENABLE_MIGRATION_TESTING: false
 ENABLE_MIGRATION_INTERFACE: false
-ENABLE_SCHOOLS_INTERFACE: true
```
avinhurry, line 11: Don't we want to keep this in but set it to false?

avinhurry: Ignore me, [it's false by default](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/config/application.rb#L55).

avinhurry (approved):

config/terraform/application/config/production.yml
peteryates: I closed that PR to avoid the risk :D I know it's poor manners but didn't want to cause more delay.

avinhurry: Thanks Pete!

PR #2643, opened 2026-04-13
[3718] Bug: Changing LP for deferred or withdrawn ECTs and Mentors changes the contract period

thenapking:
### Context

When a school changes LP for an ECT or mentor in the 2023 or 2024 contract period, a new training period is created with 2025 as the contract period.  This was because the schedule reuse logic was wrong.  We've already applied a similar change to the training programme wizard to fix a similar [bug](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/2595).  Both the change LP and change training period wizard must override the schedule that `Schedule::Find` class would assign because of the potential for unconfirmed training periods to be deleted during the wizard.  As this logic has become more complex,  the service classes for these wizards have grown bloated.  Therefore I have introduced a new concern `Schedules::Reuse`.  This could perhaps be named better.

### Guidance to review

I found using George Cole and Imogen Stubbs in Malory Towers would work.  You will need to ensure they are deferred or withdrawn.

### DIFF
--- a/app/services/schedules/find.rb
+++ b/app/services/schedules/find.rb
@@ -2,6 +2,8 @@ module Schedules
   class Find
     attr_accessor :period, :training_programme, :started_on, :period_type_key, :mentee
 
+    include Schedules::Reuse
+
     def initialize(period:, training_programme:, started_on:, period_type_key:, mentee:)
       @period = period
       @training_programme = training_programme
@@ -39,17 +41,11 @@ module Schedules
       teacher.mentor_training_periods if teacher.mentor_at_school_periods.exists?
     end
 
-    def most_recent_provider_led_period
-      training_periods&.provider_led_training_programme&.latest_first&.first
-    end
-
     def most_recent_schedule
       @most_recent_schedule ||= most_recent_provider_led_period&.schedule
     end
 
-    def latest_start_date
-      [started_on, Time.zone.today].max
-    end
+    alias_method :latest_start_date, :date_of_transition
 
     def schedule_month
       return "september" if extended_schedule?
@@ -106,20 +102,10 @@ module Schedules
       previous_mentor_started_training?
     end
 
-    def extended_schedule?
-      return false unless period_type_key == :ect_at_school_period
-
-      contract_period_reassignment.required?
-    end
-
-    def contract_period_reassignment
-      @contract_period_reassignment ||= ContractPeriods::Reassignment.new(training_period: most_recent_provider_led_period)
-    end
+    alias_method :extended_schedule?, :contract_period_reassignment_required?
 
     def extended_schedule
       successor_contract_period.schedules.find_by!(identifier:)
     end
-
-    delegate :successor_contract_period, to: :contract_period_reassignment
   end
 end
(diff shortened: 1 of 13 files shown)


app/services/schedules/find.rb
```diff
@@ -39,17 +41,11 @@ def training_periods
...
-    end
-
     def most_recent_schedule
       @most_recent_schedule ||= most_recent_provider_led_period&.schedule
     end
 
-    def latest_start_date
-      [started_on, Time.zone.today].max
-    end
+    alias_method :latest_start_date, :date_of_transition
```
avinhurry: We no longer seem to use the explicit `started_on` passed into this class here. Is that intentional? If so, do we still need `started_on` as an initializer arg and are we happy this doesn't change the behaviour for callers that pass a different training period start date?

app/services/schedules/reuse.rb
```diff
@@ -0,0 +1,61 @@
+module Schedules
+  module Reuse
+    def contract_period
+      if confirmed_training_reassignment_required?
+        successor_contract_period
+      elsif reuse_existing_schedule?
+        existing_schedule.contract_period
```
avinhurry, line 7: Contract period only re-assigns when `current_or_next_training_period_confirmed?` but `reuse_existing_schedule?` still returns `false` whenever `contract_period_reassignment_required?` is `true`. That seems to push the case back into `contract_period_at_transition` rather than preserving the old contract period.

Could be worth adding a spec?

avinhurry (commented):
I like the extraction into `Schedules::Reuse` overall. The intent around reusing old schedules, using the contract period for the transition date and reassigning is much easier to follow than when the logic was duplicated across the services.

app/services/schedules/find.rb
thenapking: Will double check whether that can be removed.

thenapking: Looking at this again, I think replacing this with an alias was wrong.  `date_of_transition` looks at the started_on attribute of the `ect_at_school_period`, whereas in `Schedules::Find` `started_on` refers to the beginning of the `training_period` that will be created. So I have reverted this change

app/services/schedules/reuse.rb
thenapking: That's correct.  When a training period from 2021/2 was never confirmed, it's going to get deleted during a change of LP or training.  In that case we can't reuse the existing schedule, because we deleted it. And we assign the  `contract_period_at_transition`, which in this case will always be on today's date.  

For changing an LP this scenario is already covered by [this](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/8668892bafdc9e2a1792882875fed6daf5031245/spec/services/ect_at_school_periods/change_lead_provider_spec.rb#L299) test, which was added as part of the successor contract period feature.  In the case of switching training, it's a bit more complex.  There must be at least two previous training periods, the first of which is provider-led, the second school-led.  We're looking at the provider-led period to decide whether reassignment is necessary.  However that period cannot be unconfirmed, because if it were, it would have been deleted when the ECT was switched to school-led.  

I think that's correct, but it is quite tricky this stuff.  If there's another test I could add which makes this scenario more explicit I am happy to.

avinhurry: That makes sense, thanks. I'm happy the spec you referenced above covers this sufficiently.

avinhurry (approved):
Nice work!

PR #2644, opened 2026-04-13
[2786] Unhide sections of the teacher page in the admin console

thenapking:
### Context

For launch we want the admin view of a teacher to have the secondary nav visible, with Overview as the first item, selected by default. Currently the nav is turned off in production by a flag, and we link to the Induction panel by default. https://github.com/DFE-Digital/register-early-career-teachers-public/pull/1735 updated the links. I have reverted all of these from admin_teacher_induction_path to admin_teacher_path except where the admin user should actually be directed back to the admin_teacher_induction_path, eg when they've created a new induction period. I think I have selected the correct ones, but would appreciate a second pair of eyes.

As per @peteryates comment this should not be merged now, but should be ready to merge into a branch we can use for launching the service.

I have also ensured the Timelines page is laid out like the others, adding in the missing nav, h2 and breadcrumbs
image


avinhurry (approved):
Looks good overall. 

One issue I spotted was when you click through to timeline from the secondary nav, the page shows a back link rather than breadcrumbs. Not sure if we've already carded that, so up to you whether to fix it here or separately.

PR #2645, opened 2026-04-13
3527 - updated email addresses on 500 page

JSJohal:

Update to email addresses shown on the internal server error page to include email addresses for schools and ABs.


avinhurry (approved):
Looks great, nice work!

spec/views/errors/internal_server_error.html.erb_spec.rb
```diff
@@ -3,5 +3,6 @@
 
   it "display support email" do
```
avinhurry, line 4:
```suggestion
  it "shows both support email links" do
```

PR #2647, opened 2026-04-13
Include voided declarations in CSVs

cpjmcquillan:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/3772

In #2406, we updated the declaration summary section of the financial statements to ensure voided declarations were displayed.

In #2416, we introduced the declarations CSV. But we didn't update the `Statements::DeclarationSelection` to include voided declarations too.

This separation should be addressed, otherwise we will end up with similar discrepancies again.

For now, this addresses the issue though.

Now, voided declarations are included in CSVs of the relevant statement.

### DIFF
--- a/app/services/statements/declaration_selection.rb
+++ b/app/services/statements/declaration_selection.rb
@@ -39,13 +39,17 @@ module Statements
     end
 
     def selected_ids_for_flat_rate(calculator)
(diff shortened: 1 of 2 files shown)


avinhurry (approved):
Looks great, thanks!

PR #2649, opened 2026-04-13
ParityCheck::Response save calculated 'rect_performance_gain_ratio' in column

mooktakim:
### Context

`ParityCheck::Response.rect_performance_gain_ratio` is a calculated method, is timing out for declarations endpoint.

* Add new column `ParityCheck::Response.rect_performance_gain_ratio`, save calculated value on before_validation

### DIFF
--- a/app/models/parity_check/request.rb
+++ b/app/models/parity_check/request.rb
@@ -68,7 +68,7 @@ module ParityCheck
     end
 
     def rect_performance_gain_ratio
-      ratios = responses.map(&:rect_performance_gain_ratio).compact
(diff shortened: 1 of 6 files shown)


ethax-ross (approved):
Looks good, can we avoid changing the blocklist though I expect if we commit this it will revert when someone else regenerates it via the CLI? has your IDE applied some formatting?

mooktakim:
> Looks good, can we avoid changing the blocklist though I expect if we commit this it will revert when someone else regenerates it via the CLI? has your IDE applied some formatting?

ah man, got some YAML linting going on recent nvim update

PR #2655, opened 2026-04-14
[3648] Add metadata refreshable endpoint shared context to api requests specs

leandroalemao:
### Context

Ticket: [3648](https://github.com/DFE-Digital/register-ects-project-board/issues/3648)

We're seeing some alerts on Sentry about metadata refresh.

- Add metadata refreshable endpoint shared context to api requests specs so we make sure that the alerts are not being created when an API endpoint is hit;
- Do we need to `track_changes` set to true on the daily job `RefreshAllMetadataJob`?

On the Sentry alerts:
- All events on Sentry are happening after 12am (utc):
  - Screenshot 2026-04-14 at 09 25 30
- `RefreshAllMetadataJob` is running every day at 12am (utc).
  - `Metadata::Manager.refresh_all_metadata!(async: true, track_changes: true)` is run with `track_changes` flag `true`;
  - `Metadata::Handlers` spawns multiple `RefreshMetadataJob` with `track_changes` flag `true`;
  - `Metadata::Manager.new.refresh_metadata!(objects, track_changes:)` is run with `track_changes` flag `true`;
  - Metadata is refreshed and alerts are created on Sentry as `alert_on_changes` is flag is set to `true`;
    ```
      Array.wrap(objects).each do
        handler = resolve_handler(it)
        handler.track_changes! if track_changes
        handler.refresh_metadata!
      end

    def track_changes!
      @alert_on_changes = true
    end

    def upsert_all(model:, changes_to_upsert:, unique_by:)
      return if changes_to_upsert.empty?

      model.upsert_all(changes_to_upsert.values, unique_by:)
      alert_on_changes(changes: changes_to_upsert)
    end
    ```

### DIFF
--- a/spec/support/shared_contexts/api/metadata_refreshable_endpoint.rb
+++ b/spec/support/shared_contexts/api/metadata_refreshable_endpoint.rb
@@ -1,0 +1,14 @@
+shared_examples "an endpoint that refreshes metadata" do |request_method|
+  it "refreshes all required metadata" do
+    send(request_method, path, headers: api_headers, params:)
+
+    dump_metadata_database_state = -> {
+      ActiveRecord::Base.connection.tables.select { it.start_with?("metadata_") }.to_h do |table|
+        rows = ActiveRecord::Base.connection.execute("SELECT * FROM #{table}").to_a
+        [table, rows]
+      end
+    }.call
+
+    expect { Metadata::Manager.refresh_all_metadata! }.not_to(change { dump_metadata_database_state })
+  end
+end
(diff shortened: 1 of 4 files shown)


spec/support/shared_contexts/api/metadata_refreshable_endpoint.rb
```diff
@@ -0,0 +1,14 @@
+shared_examples "an endpoint that refreshes metadata" do |request_method|
+  it "does not touch any existing metadata" do
```
ethax-ross: `it "refreshes all required metadata"`?

ethax-ross (approved):
Thanks @leandroalemao looks good.

We want `track_changes: true` on the nightly refresh job as that's what is driving the errors reporting in Sentry.

Given these tests are passing it isn't anything API-related that is resulting in out of date metadata; do you have any other ideas on what's causing it? I would maybe look at the individual records reported in Sentry more closely, I have a suspicion its maybe when new instances of objects with metadata are being created somewhere, maybe bypassing our callbacks (as the tracked changes seem to include all attributes getting refreshed which would happen on creation)

PR #2658, opened 2026-04-14
Improve metadata change tracking

ethax-ross:
Previously we were passing all attributes to Sentry when any of them are updated on some existing metadata. This is misleading because it looks like all the attributes have changed. Instead, we want to just surface the attributes that actually had their value changed due to refreshing the metadata.

It's awkward to extend the existing tests to cover this change, but they should already do a good enough job of testing the bits we care about.

### DIFF
--- a/app/services/metadata/handlers/base.rb
+++ b/app/services/metadata/handlers/base.rb
@@ -47,10 +47,17 @@ module Metadata::Handlers
       return unless @alert_on_changes
 
       changes.each do |metadata, changed_attributes|
+        changed_attributes_with_history = changed_attributes.transform_values.with_index do |new_value, key|
+          {
+            old: metadata.attributes[key],
+            new: new_value
+          }
+        end
+
         attrs = {
           class: metadata.class.name,
           id: metadata.id,
-          changed_attributes:,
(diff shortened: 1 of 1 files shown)


app/services/metadata/handlers/base.rb
```diff
@@ -47,10 +47,17 @@ def alert_on_changes(changes:)
...
+            new: new_value
+          }
+        end
+
         attrs = {
           class: metadata.class.name,
           id: metadata.id,
-          changed_attributes:,
+          changed_attributes: changed_attributes_with_history,
         }
```
mooktakim, line 61: Can we also send the caller like this:

```ruby
attrs = {
  class: metadata.class.name,
  id: metadata.id,
  changed_attributes: changed_attributes_with_history,
  backtrace: caller(1, 15) # skip self, keep top 15 frames
}
```

That way it'll make it easier to see where the problem is.

mooktakim (approved):

app/services/metadata/handlers/base.rb
mooktakim: ignore, we see the stacktrace on sentry

PR #2660, opened 2026-04-14
Fix Homepage Sign-In Journey

avinhurry:
### Context

The homepage was sending users to the sign-in method chooser instead of initiating the intended journey.

This updates the school facing homepage sign-in CTA so production end users go straight to DfE Sign-in, while non-production environments still use the existing sign-in chooser.

It also updates the appropriate body link on the school homepage to go directly to the appropriate body landing page, because once `ENABLE_SCHOOLS_INTERFACE` is enabled the service root will no longer redirect to `/appropriate-body`.

### Changes proposed

- Production school CTA -> direct to DfE Sign-in
- Non-production school CTA -> unchanged, still goes via `/sign-in`


peterdavidhamilton (approved):

spec/helpers/pages_helper_spec.rb
```diff
@@ -1,3 +1,66 @@
...
+        )
+      end
+
+      it "renders a POST button to DfE Sign-in" do
+        expect(rendered_cta).to have_css("form[action='/auth/dfe_sign_in'][method='post']")
+        expect(rendered_cta).to have_button(button_text)
+      end
+    end
+
+    context "when in production but personas are enabled" do
```
cpjmcquillan: Would we ever want personas enabled in Production? 🤔

cpjmcquillan (approved):

spec/helpers/pages_helper_spec.rb
avinhurry: Just to check, are you suggesting we simplify the test only, or remove the personas check from the logic as well?

I kept the personas check because the AB landing page already uses the same rule to decide whether to bypass the chooser. 

Happy to remove it if we're comfortable with that wider behaviour change?

app/helpers/pages_helper.rb
```diff
@@ -1,2 +1,15 @@
...
+  def school_sign_in_cta(text)
+    return govuk_button_to(text, "/auth/dfe_sign_in") if direct_school_dfe_sign_in?
+
+    govuk_button_link_to(text, sign_in_path)
+  end
+
+private
+
+  def direct_school_dfe_sign_in?
+    Rails.env.production? &&
```
peteryates: I think if possible here we shouldn't be mentioning the env and instead wholly rely on the app config, which we have full control over via env vars and can adjust them by env if we need to.

spec/helpers/pages_helper_spec.rb
peteryates: Like my comment above, I think the logic here should rely on the config rather than the env.

In reality, we wouldn't ever want personas to be available in prod, but we might want to allow `#direct_school_dfe_sign_in?`  in a non-prod env.

avinhurry: Thank you, I updated this now so the school CTA no longer relies on `Rails.env` and is controlled by config via `ENABLE_DIRECT_SCHOOL_DFE_SIGN_IN`.

peteryates (approved):
Looks great, thanks!

PR #2662, opened 2026-04-14
[3744] Add validation for leaving date and new school start date when teacher is moving schools

thenapking:
### Context
NB please note that the error message has changed slightly from the card with agreement from @mason-emily.   
ACs
1. Validation on new school start date added to block a start date earlier than the latest training period start date at the previous school.

2. Error message for this scenario will say: Our records show that {{teacher name}} started their latest training at {{school}} on {{latest training period start date}}. Enter a start date after {{earliest end date of latest training period}}.

3. The existing school start date error message will take priority over this new one if there are no training periods with a later start date.

4. Validation on school date leaving date to block a date that is earlier than the latest training period start date at the current school.

Error message for leaving school scenario should say: Our records show that {{teacher name}} started their latest training on {{latest training period start date}}. Enter a date after {{earliest end date of latest training period}}.

When we were discussing this we thought there might be hidden complexity. The first issue we discovered was to do with backdated training periods.  The second issue was to do with the interplay of `ect_at_school_periods` and `training_periods`.

image

Here we are trying to create a new ECT period at a separate school.  The existing ECT and training period may not start on the same day.  There may be several training periods.  We need to ensure that the latest start date of either period is used for validation.  In addition if the entered date is before the start of both the ECT and training period we need to prioritise errors from the ECT period.

Since any open periods need to be finished, and can't be zero length we need to take this into account when giving the earliest date the ECT period can be created at a new school.  For instance, consider an ECT with a training period that starts on 1st Jan.  This cannot be zero length so the earliest it can finish is 2nd Jan.  Therefore the earliest valid start date at a new school is 3rd Jan.

Additionally there may be several training periods, of which we only consider the latest.  If that training period starts in the future it will be deleted by the ECT registration process and we therefore don't let it affect validation.  However if the ECT was registered today, and their first training period started today, then that ECT period would block registration at another school today.  The earliest registration date would be the day after tomorrow.  Similarly we now ensure that if a training period has started today, the earliest registration date at another school is the day after tomorrow.  

Due to the variety of combinations present to make this validation work, I have introduced a custom validation class, which exhaustively tests these combinations.  The wizard steps then handle error messages given the period which overlaps and the earliest valid date.

In the tests for the service class there are shared examples which include the logic above that states an entered date is invalid yesterday, today, and tomorrow, but valid the day after tomorrow, if it clashes with a period that starts today.  I've tried to make that as readable as possible.  The easiest way I found to think of it was this four day period around the entered date, with the last day being a valid date.

We have recently updated the logic around backdated ECT periods, in this [PR](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/2688).  This means that when registering an ECT in the past their first training period starts today.  Previously we allowed the niche case where two schools do that on the same day.  This validation blocks that, and insists that the second school cannot back date their registration.

### Guidance to review

I found that for some reason many of the listed teachers could not be registered, and therefore you will have to use these very specific ones listed below.

| TRN     | Name           | Type                     | Period Start | Training Period 1 | Training Period 2 | Training Period 3 |
|---------|----------------|--------------------------|--------------|-------------------|-------------------|-------------------|
| 3002577 | Jonas Bloggs   | ect_at_school_period     | 2025-09-01   | provider_led (2025-09-01 → ongoing) |  |  |
| 3002578 | Cynthia Parks  | ect_at_school_period     | 2025-09-01   | provider_led (2025-10-01 → ongoing) |  |  |
| 3002579 | Taylor Hawkins | ect_at_school_period     | 2025-09-01   | provider_led (2025-10-01 → 2025-10-31) | provider_led / ambition (2025-11-01 → ongoing) |  |
| 3002580 | Muhammed Ali   | ect_at_school_period     | 2025-09-01   | school_led (2025-10-01 → 2025-10-31) | provider_led / ambition (2025-11-01 → ongoing) |  |
| 3002582 | Robson Scottie | ect_at_school_period     | 2025-09-01   | provider_led (2025-10-01 → 2025-10-31) | school_led (2025-11-01 → 2025-11-30) | provider_led (2025-12-01 → ongoing) |
| 3012858 | Dave Teacher   | mentor_at_school_period  | 2025-09-01   | provider_led (2025-09-01 → ongoing) |  |  |
| 3003943 | Donna Msa      | mentor_at_school_period  | 2025-09-01   | provider_led (2025-10-01 → ongoing) |  |  |
| 3012235 | Claire Cool    | mentor_at_school_period  | 2025-09-01   | provider_led (2025-10-01 → 2025-10-31) | provider_led / ambition (2025-11-01 → ongoing) |  |

### DIFF
--- a/app/wizards/schools/ects/teacher_leaving_wizard/edit_step.rb
+++ b/app/wizards/schools/ects/teacher_leaving_wizard/edit_step.rb
@@ -6,6 +6,7 @@ module Schools
 
         validate :leaving_on_valid
         validate :leaving_after_start_date
+        validate :leaving_after_training_started
 
         def self.permitted_params
           %i[
@@ -40,15 +41,43 @@ module Schools
           errors.add(:leaving_on, leaving_on_input.error_message)
         end
 
+        def skip_leaving_on_validation?
+          errors[:leaving_on].any?
+        end
+
         def leaving_after_start_date
-          return unless leaving_on_input.valid?
-          return if leaving_on_input.value_as_date > ect_at_school_period.started_on
+          return if skip_leaving_on_validation?
+          return unless period_start_date_valid?(ect_at_school_period)
+          return if after_started_on(ect_at_school_period)
 
           errors.add(
             :leaving_on,
-            "Our records show that #{name_for(ect_at_school_period.teacher)} started teaching at your school on
-            #{ect_at_school_period.started_on.to_formatted_s(:govuk)}. Enter a later date."
+            "Our records show that #{name_for(ect_at_school_period.teacher)} started teaching at your school on #{ect_at_school_period.started_on.to_formatted_s(:govuk)}. Enter a later date."
           )
+        end
+
+        def leaving_after_training_started
+          return if skip_leaving_on_validation?
+          return unless period_start_date_valid?(latest_started_training_period)
+          return if after_started_on(latest_started_training_period)
+
+          errors.add(
+            :leaving_on,
+            "Our records show that #{name_for(ect_at_school_period.teacher)} started their latest training at your school on #{latest_started_training_period.started_on.to_formatted_s(:govuk)}. Enter a later date."
+          )
+        end
+
+        def after_started_on(period)
+          leaving_on_input.value_as_date > period.started_on
+        end
+
+        def period_start_date_valid?(period)
+          period&.started_on.present?
+        end
+
+        # Unstarted training periods (ie starting today or in the future) will be deleted and should not prevent the leaving date from being valid
+        def latest_started_training_period
+          ect_at_school_period&.training_periods&.started_before(Time.zone.today)&.latest_first&.first
         end
 
         def leaving_on_input
--- a/app/wizards/schools/ects/teacher_leaving_wizard/edit_step.rb
+++ b/app/wizards/schools/ects/teacher_leaving_wizard/edit_step.rb
@@ -5,8 +5,7 @@ module Schools
         attr_accessor :leaving_on
 
         validate :leaving_on_valid
-        validate :leaving_after_start_date
-        validate :leaving_after_training_started
+        validate :leaving_on_after_previous_school_or_training_period_start
 
         def self.permitted_params
           %i[
@@ -45,44 +44,36 @@ module Schools
           errors[:leaving_on].any?
         end
 
-        def leaving_after_start_date
+        def leaving_on_after_previous_school_or_training_period_start
           return if skip_leaving_on_validation?
-          return unless period_start_date_valid?(ect_at_school_period)
-          return if after_started_on(ect_at_school_period)
+          return if leaving_on_boundary_validator.valid?
 
-          errors.add(
-            :leaving_on,
-            "Our records show that #{name_for(ect_at_school_period.teacher)} started teaching at your school on #{ect_at_school_period.started_on.to_formatted_s(:govuk)}. Enter a later date."
-          )
+          errors.add(:leaving_on, invalid_period_error_message)
         end
 
-        def leaving_after_training_started
-          return if skip_leaving_on_validation?
-          return unless period_start_date_valid?(latest_started_training_period)
-          return if after_started_on(latest_started_training_period)
-
-          errors.add(
-            :leaving_on,
-            "Our records show that #{name_for(ect_at_school_period.teacher)} started their latest training at your school on #{latest_started_training_period.started_on.to_formatted_s(:govuk)}. Enter a later date."
-          )
-        end
-
-        def after_started_on(period)
-          leaving_on_input.value_as_date > period.started_on
-        end
-
-        def period_start_date_valid?(period)
-          period&.started_on.present?
-        end
-
-        # Unstarted training periods (ie starting today or in the future) will be deleted and should not prevent the leaving date from being valid
-        def latest_started_training_period
-          ect_at_school_period&.training_periods&.started_before(Time.zone.today)&.latest_first&.first
+        def invalid_period_error_message
+          "Our records show that #{name_for(ect_at_school_period.teacher)} started " \
+          "#{invalid_period_type} at your school on " \
+          "#{invalid_period_started_on_formatted}." \
+          " Enter a date after #{earliest_valid_input_date_formatted}."
         end
 
         def leaving_on_input
           @leaving_on_input ||= Schools::Validation::LeavingDate.new(date_as_hash: leaving_on)
         end
+
+        def leaving_on_boundary_validator
+          @leaving_on_boundary_validator ||= Schools::Validation::PeriodBoundary.new(
+            ect_at_school_period:,
+            input_date: leaving_on_input.value_as_date
+          )
+        end
+
+        delegate  :type,
+                  :started_on_formatted,
+                  to: :leaving_on_boundary_validator,
+                  prefix: :invalid_period
+        delegate :earliest_valid_input_date_formatted, to: :leaving_on_boundary_validator
       end
     end
   end
--- a/app/wizards/schools/ects/teacher_leaving_wizard/edit_step.rb
+++ b/app/wizards/schools/ects/teacher_leaving_wizard/edit_step.rb
@@ -64,7 +64,7 @@ module Schools
 
         def leaving_on_boundary_validator
           @leaving_on_boundary_validator ||= Schools::Validation::PeriodBoundary.new(
-            ect_at_school_period:,
+            input_period: ect_at_school_period,
             input_date: leaving_on_input.value_as_date
           )
         end
--- a/app/wizards/schools/register_ect_wizard/start_date_step.rb
+++ b/app/wizards/schools/register_ect_wizard/start_date_step.rb
@@ -4,8 +4,7 @@ module Schools
       attr_accessor :start_date
 
       validates :start_date, ect_start_date: true
-      validate :start_date_not_before_previous_school
-      validate :start_date_not_before_last_training_period
+      validate :start_date_after_previous_school_or_training_period_start
       validate :start_date_within_4_months, if: :currently_ect_at_another_school?
 
       def self.permitted_params
@@ -26,33 +25,6 @@ module Schools
 
     private
 
-      def start_date_not_before_previous_school
-        return if skip_start_date_validation?
-
-        return unless period_start_date_valid?(previous_period)
-        return if start_date_after_period_starts?(previous_period)
-
-        errors.add(
-          :start_date,
-          "Our records show that #{wizard.ect.full_name} started teaching at #{previous_period.school&.name} on #{previous_period.started_on.to_formatted_s(:govuk)}. Enter a later start date."
-        )
-      end
-
-      def start_date_not_before_last_training_period
-        return if skip_start_date_validation?
-        return unless period_start_date_valid?(latest_started_training_period)
-        return if start_date_after_period_starts?(latest_started_training_period)
-
-        errors.add(
-          :start_date,
-          "Our records show that #{wizard.ect.full_name} started their latest training at #{latest_started_training_period.school&.name} on #{latest_started_training_period.started_on.to_formatted_s(:govuk)}. Enter a later start date."
-        )
-      end
-
-      def start_date_after_period_starts?(period)
-        start_date_as_date > period.started_on
-      end
-
       def currently_ect_at_another_school?
         ect.previously_registered? && previous_period.ongoing?
       end
@@ -61,9 +33,18 @@ module Schools
         ect.previous_ect_at_school_period
       end
 
-      # Unstarted training periods (ie starting today or in the future) will be deleted and should not prevent a start date at a new school from being valid
-      def latest_started_training_period
-        previous_period&.training_periods&.started_before(Time.zone.today)&.latest_first&.first
+      def start_date_after_previous_school_or_training_period_start
+        return if skip_start_date_validation?
+        return if start_date_boundary_validator.valid?
+
+        errors.add(:start_date, invalid_period_error_message)
+      end
+
+      def invalid_period_error_message
+        "Our records show that #{wizard.ect.full_name} started " \
+        "#{invalid_period_type} at #{previous_period&.school&.name} on " \
+        "#{invalid_period_started_on_formatted}." \
+        " Enter a start date after #{earliest_valid_input_date_formatted}."
       end
 
       def start_date_within_4_months
@@ -82,10 +63,6 @@ module Schools
 
       def registrations_closed_for_contract_period?
         start_date_as_date.future? && !start_date_contract_period&.enabled?
-      end
-
-      def period_start_date_valid?(period)
-        period&.started_on.present?
       end
 
       def skip_start_date_validation?
@@ -121,6 +98,19 @@ module Schools
       def start_date_obj
         @start_date_obj ||= Schools::Validation::ECTStartDate.new(date_as_hash: start_date)
       end
+
+      def start_date_boundary_validator
+        @start_date_boundary_validator ||= Schools::Validation::PeriodBoundary.new(
+          ect_at_school_period: previous_period,
+          input_date: start_date_as_date
+        )
+      end
+
+      delegate :type,
+               :started_on_formatted,
+               to: :start_date_boundary_validator,
+               prefix: :invalid_period
+      delegate :earliest_valid_input_date_formatted, to: :start_date_boundary_validator
     end
   end
 end
--- a/app/wizards/schools/register_ect_wizard/start_date_step.rb
+++ b/app/wizards/schools/register_ect_wizard/start_date_step.rb
@@ -101,7 +101,7 @@ module Schools
 
       def start_date_boundary_validator
         @start_date_boundary_validator ||= Schools::Validation::PeriodBoundary.new(
-          ect_at_school_period: previous_period,
+          input_period: previous_period,
           input_date: start_date_as_date
         )
       end
--- a/app/wizards/schools/register_ect_wizard/start_date_step.rb
+++ b/app/wizards/schools/register_ect_wizard/start_date_step.rb
@@ -5,6 +5,7 @@ module Schools
 
       validates :start_date, ect_start_date: true
       validate :start_date_not_before_previous_school
+      validate :start_date_not_before_last_training_period
       validate :start_date_within_4_months, if: :currently_ect_at_another_school?
 
       def self.permitted_params
@@ -28,16 +29,41 @@ module Schools
       def start_date_not_before_previous_school
         return if skip_start_date_validation?
 
-        previous_period = ect.previous_ect_at_school_period
-        return unless previous_start_date_invalid?(previous_period)
+        return unless period_start_date_valid?(previous_period)
+        return if start_date_after_period_starts?(previous_period)
 
-        if start_date_as_date <= previous_period.started_on
-          add_start_date_too_early_error(previous_period)
-        end
+        errors.add(
+          :start_date,
+          "Our records show that #{wizard.ect.full_name} started teaching at #{previous_period.school&.name} on #{previous_period.started_on.to_formatted_s(:govuk)}. Enter a later start date."
+        )
+      end
+
+      def start_date_not_before_last_training_period
+        return if skip_start_date_validation?
+        return unless period_start_date_valid?(latest_started_training_period)
+        return if start_date_after_period_starts?(latest_started_training_period)
+
+        errors.add(
+          :start_date,
+          "Our records show that #{wizard.ect.full_name} started their latest training at #{latest_started_training_period.school&.name} on #{latest_started_training_period.started_on.to_formatted_s(:govuk)}. Enter a later start date."
+        )
+      end
+
+      def start_date_after_period_starts?(period)
+        start_date_as_date > period.started_on
       end
 
       def currently_ect_at_another_school?
-        ect.previously_registered? && ect.previous_ect_at_school_period.ongoing?
+        ect.previously_registered? && previous_period.ongoing?
+      end
+
+      def previous_period
+        ect.previous_ect_at_school_period
+      end
+
+      # Unstarted training periods (ie starting today or in the future) will be deleted and should not prevent a start date at a new school from being valid
+      def latest_started_training_period
+        previous_period&.training_periods&.started_before(Time.zone.today)&.latest_first&.first
       end
 
       def start_date_within_4_months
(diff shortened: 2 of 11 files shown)


app/wizards/schools/ects/teacher_leaving_wizard/edit_step.rb
```diff
@@ -40,17 +41,39 @@ def leaving_on_valid
...
+            "Our records show that #{name_for(ect_at_school_period.teacher)} started #{description} at your school on #{period.started_on.to_formatted_s(:govuk)}. Enter a later date."
           )
         end
 
+        def previous_start_date_invalid?(period)
+          period&.started_on.present?
+        end
+
+        def latest_started_training_period
+          ect_at_school_period&.training_periods&.started_before(Time.zone.today)&.latest_first&.first
```
avinhurry: Is the `started_before(Time.zone.today)` check here because backdated registrations can create training periods with a `started_on` of today at sign up so we don't want to validate against those? 🤔

app/wizards/schools/ects/teacher_leaving_wizard/edit_step.rb
```diff
@@ -40,17 +41,39 @@ def leaving_on_valid
           errors.add(:leaving_on, leaving_on_input.error_message)
         end
 
+        def skip_leaving_on_validation?
+          errors[:leaving_on].any?
+        end
```
avinhurry, line 45: This feels a little brittle because the order is being driven by validation order and existing errors. 

Would it be possible to make the error priority explicit in a single validation method instead?

app/wizards/schools/ects/teacher_leaving_wizard/edit_step.rb
```diff
@@ -40,17 +41,39 @@ def leaving_on_valid
...
-          return unless leaving_on_input.valid?
-          return if leaving_on_input.value_as_date > ect_at_school_period.started_on
+          leaving_on_before(ect_at_school_period, "teaching")
+        end
+
+        def leaving_after_training_started
+          leaving_on_before(latest_started_training_period, "their latest training")
+        end
+
+        def leaving_on_before(period, description)
```
avinhurry: The abstraction feels a little awkward to me, especially passing `description` in as a string. `period` makes sense, but `description` makes the call sites fairly opaque and couples the validation to the message construction.

Would it be clearer to keep these as explicit validation methods and only extract the shared comparison if needed?

avinhurry (approved):
I had one question on the date logic and a few cosmetic comments, but otherwise this looks good to me.

app/wizards/schools/ects/teacher_leaving_wizard/edit_step.rb
thenapking: That's precisely why.  I could add a comment in the code - I feel like that is something that we'll forget.

app/wizards/schools/ects/teacher_leaving_wizard/edit_step.rb
thenapking: I agree about the naming of `description`.  I'll see if there's a way I can refactor this so we aren't verbosely repeating logic.  It was a little unclear I felt before, but I am sure I can balance it.

app/wizards/schools/ects/teacher_leaving_wizard/edit_step.rb
cpjmcquillan, line 1: Agree with Avin here, I think pretty much all of this would be more readable if it was inline in the validation method. I don't think the de-duplication is helping us here, and I think it's ok to be verbose!

app/wizards/schools/ects/teacher_leaving_wizard/edit_step.rb
thenapking: That turned out somewhat less clear to read than what we have already.  This pattern has been used elsewhere and there are tests which assert the priority of the validation error messages so I think we should be ok.

app/wizards/schools/ects/teacher_leaving_wizard/edit_step.rb
```diff
@@ -40,17 +41,45 @@ def leaving_on_valid
...
         end
 
+        def skip_leaving_on_validation?
+          errors[:leaving_on].any?
+        end
+
         def leaving_after_start_date
-          return unless leaving_on_input.valid?
-          return if leaving_on_input.value_as_date > ect_at_school_period.started_on
+          return if skip_leaving_on_validation?
```
cpjmcquillan, line 48: I don't think the extra method here conveys any extra meaning. The guard clause communicates that the validation is being skipped already IMO.

```suggestion
          return if errors[:leaving_on].any?
```

app/wizards/schools/ects/teacher_leaving_wizard/edit_step.rb
```diff
@@ -40,17 +41,45 @@ def leaving_on_valid
...
+          return unless period_start_date_valid?(ect_at_school_period)
```
cpjmcquillan: Again, I don't think the extra method definition makes things clearer.


```suggestion
          return unless ect_at_school_period.started_on
```

app/wizards/schools/ects/teacher_leaving_wizard/edit_step.rb
```diff
@@ -40,17 +41,45 @@ def leaving_on_valid
...
+          return if after_started_on(ect_at_school_period)
```
cpjmcquillan: [`ActiveSupport`](https://api.rubyonrails.org/classes/DateAndTime/Calculations.html#method-i-after-3F) already gives us a bunch of methods that help us compare dates. 

So I think we'd be better off just using that if we want something more readable than `>` 👌🏼

```suggestion
          return if leaving_on_input.value_as_date.after?(ect_at_school_period.started_on)
```

app/wizards/schools/ects/teacher_leaving_wizard/edit_step.rb
```diff
@@ -40,17 +41,45 @@ def leaving_on_valid
...
+          errors.add(
+            :leaving_on,
+            "Our records show that #{name_for(ect_at_school_period.teacher)} started teaching at your school on #{ect_at_school_period.started_on.to_formatted_s(:govuk)}. Enter a later date."
+          )
+        end
+
+        def leaving_after_training_started
+          return if skip_leaving_on_validation?
+          return unless period_start_date_valid?(latest_started_training_period)
+          return if after_started_on(latest_started_training_period)
```
cpjmcquillan: Same comments apply

```suggestion
          return if errors[:leaving_on].any?
          return unless latest_started_training_period&.started_on
          return if leaving_on_input.value_as_date.after?(latest_started_training_period.started_on)
```

app/wizards/schools/ects/teacher_leaving_wizard/edit_step.rb
```diff
@@ -40,17 +41,45 @@ def leaving_on_valid
...
+        end
+
+        def period_start_date_valid?(period)
+          period&.started_on.present?
+        end
+
+        # Unstarted training periods (ie starting today or in the future) will be deleted and should not prevent the leaving date from being valid
+        def latest_started_training_period
+          ect_at_school_period&.training_periods&.started_before(Time.zone.today)&.latest_first&.first
+        end
```
cpjmcquillan: Do we need to be this defensive?

If the user is reporting an ECT as leaving, the `ect_at_school_period` should always be present. Or perhaps I'm missing something 🤔

Provided the `ect_at_school_period` is there, then we can drop all the `&.`


```suggestion
        def latest_started_training_period
          ect_at_school_period.training_periods.started_before(Time.zone.today).latest_first.first
        end
```

app/wizards/schools/register_ect_wizard/start_date_step.rb
cpjmcquillan, line 1: Same comments apply here

cpjmcquillan (approved):
Tests look great! I just added some minor comments about the validation methods. I think things would be easier to understand without the extra method definitions.

Let's add a issue to the backlog to add a custom validator, though, if you think that would be worthwhile 💚

app/wizards/schools/ects/teacher_leaving_wizard/edit_step.rb
rwrrll: I agree with the above, it doesn't look like `skip_leaving_on_validation?` is used anywhere else, so another vote from me to inline as per the suggestion (unless there was a reason to keep this that I'm not privvy to, in which case all good!)

app/wizards/schools/ects/teacher_leaving_wizard/edit_step.rb
```diff
@@ -40,20 +40,53 @@ def leaving_on_valid
...
+
+          errors.add(:leaving_on, invalid_period_error_message)
+        end
+
+        def invalid_period_error_message
+          "Our records show that #{name_for(ect_at_school_period.teacher)} started " \
+          "#{invalid_period_type} at your school on " \
+          "#{invalid_period_formatted_date}." \
+          " Enter a date after #{invalid_period_earliest_end_date_formatted}."
+        end
```
rwrrll: The recurring `invalid_period_` prefix on 5 helper methods is making my POODR senses tingle! It's a hint that a lot of stuff specific to that validation is spilling over the boundary between the domain model and the validator class.

The code is also almost identical to the version in the other step, but with small differences buried in them that are effortful to parse out.

Since we already instantiate a boundary validator anyway, can we just pass in the small number of remaining differences as additional parameters, and have all these `invalid_period_` methods live privately in that class (then we can remove the method name prefixes too)?

I haven't done the actual work, but I think it'd look something like:

```ruby
def leaving_on_after_previous_school_or_training_period_start
  return if errors[:leaving_on].any?
  return if boundary.valid?

  errors.add(:leaving_on, boundary.error_message)
end

def boundary
  @boundary ||= Schools::Validation::PeriodBoundary.new(
    ect_at_school_period:,
    date: leaving_on_input.value_as_date,
    teacher_name: name_for(ect_at_school_period.teacher),
    school_name: "your school",
    date_noun: "date",
  )
end
```

spec/lib/schools/validation/period_boundary_spec.rb
```diff
@@ -0,0 +1,408 @@
+RSpec.shared_examples "it is only valid two days after" do |base_date_proc|
+  let(:base_date) { instance_exec(&base_date_proc) }
```
rwrrll: Clever use of `instance_exec` but hard to follow. IMO it'd be simpler and more idiomatic to lean on RSpec example group hierarchy/scoping, e.g.:

```ruby
  RSpec.shared_examples "it is only valid two days after" do
    context "when the date is before the period started" do
      let(:date) { base_date - 1.day }

      it { is_expected.not_to be_valid }
    end

    context "when the date is the same as the period started" do
      let(:date) { base_date }

      it { is_expected.not_to be_valid }
    end

    context "when the date is one day after the period started" do
      let(:date) { base_date + 1.day }

      it { is_expected.not_to be_valid }
    end

    context "when the date is more than one day after the period started" do
      let(:date) { base_date + 2.days }

      it { is_expected.to be_valid }
    end
  end

  RSpec.describe Schools::Validation::PeriodBoundary do
    describe "#valid?" do
      context "when there is an ect_at_school_period" do
        context "when there are no training periods" do
          let(:base_date) { ect_at_school_period.started_on }

          it_behaves_like "it is only valid two days after"
        end
      end
    end
  end
```

spec/features/schools/ects/register/edge_cases/moving_school_backdated_start_date_spec.rb
```diff
@@ -263,9 +269,19 @@ def when_i_choose_to_reuse_previous_choices
   end
 
   def then_i_should_i_should_be_taken_to_the_working_pattern_page
+    page.screenshot(path: "tmp/error-working-pattern.png")
```
rwrrll: Stray debug screenshot

spec/features/schools/ects/register/edge_cases/moving_school_backdated_start_date_spec.rb
```diff
@@ -263,9 +269,19 @@ def when_i_choose_to_reuse_previous_choices
...
     expect(page).to have_path("/school/register-ect/working-pattern")
   end
 
+  def then_the_second_school_cannot_register_the_same_teacher
+    expect(page).to have_path("/school/register-ect/start-date")
+    expect(page.get_by_text("There is a problem")).to be_visible
+  end
+
+  def and_the_second_school_should_see_an_error_message_about_the_teachers_latest_training_period
```
rwrrll: Is this called anywhere? Should it be?

spec/wizards/schools/register_ect_wizard/start_date_step_spec.rb
```diff
@@ -6,7 +6,16 @@
...
+
+  let(:wizard) do
+    FactoryBot.build(:register_ect_wizard, current_step: :start_date, store:, school:, step_params:).tap do |instance|
+      allow(instance).to receive(:ect).and_return(
+        Schools::RegisterECTWizard::RegistrationStore.new(store).tap do |registration_store|
+          allow(registration_store).to receive_messages(trs_first_name: "Johnnie", trs_last_name: "Walker")
+        end
+      )
+    end
+  end
```
rwrrll: Quite hefty tapping and nesting, might be nice to pull the store out to a separate let

lib/tasks/product_review/2662.rake
```diff
@@ -0,0 +1,175 @@
...
+        teacher.mentor_at_school_periods.each do |mentor|
+          mentor.training_periods.destroy_all
+          mentor.destroy!
+        end
+      end
+    end
+  end
+end
+
+CANDIDATE_TEACHERS = [
```
rwrrll, line 144: It's not really the end of the world since this rake task should get deleted soon, but this and the other top-level constant will leak into `Object`, as a general rule these and the method could do with being wrapped in a Module

spec/lib/schools/validation/period_boundary_spec.rb
```diff
@@ -0,0 +1,408 @@
...
+
+  context "when the date is before the period started" do
+    let(:date) { base_date - 1.day }
+
+    it { is_expected.not_to be_valid }
+  end
+
+  context "when the date is the same as the period started" do
+    let(:date) { base_date  }
```
rwrrll: Stray double space here (mega tiny nitpick!)

app/wizards/schools/ects/teacher_leaving_wizard/edit_step.rb
```diff
@@ -40,20 +40,53 @@ def leaving_on_valid
...
+
+        def invalid_period_type
+          case invalid_period
+          when ECTAtSchoolPeriod then "teaching"
+          when TrainingPeriod    then "their latest training"
+          end
```
rwrrll: Using `case` will return `nil` on an unexpected future class. Could add an `else raise`, or my personal preference is `Hash#fetch`:


```suggestion
          {
            ECTAtSchoolPeriod => "teaching",
            TrainingPeriod => "their latest training"
          }.fetch(invalid_period.class)
```

which will raise `KeyError` if the key doesn't exist

lib/schools/validation/period_boundary.rb
```diff
@@ -0,0 +1,52 @@
...
+
+      def validate_periods
+        return if date.blank?
+        return unless ect_at_school_period
+
+        period = ect_at_school_period
+        return period if date_before_period_starts?(period)
+
+        period = latest_started_training_period
+        period if date_before_period_starts?(period)
```
rwrrll: Assigning to `period` caught me out here, I didn't expect it to be re-assigned and got confused until I realised it can be two different things depending on how far in the method we get. Also the early return is a result rather than a guard.

For me, a boring old conditional represents what's happening more clearly:

```suggestion
        if date_before_period_starts?(ect_at_school_period)
          ect_at_school_period
        elsif date_before_period_starts?(latest_started_training_period)
          latest_started_training_period
        end
```

(n.b. above probably wants `latest_started_training_period` memoized)

lib/schools/validation/period_boundary.rb
```diff
@@ -0,0 +1,52 @@
...
+
+      def date_before_period_starts?(period)
+        return false if period&.started_on.blank?
+
+        date <= earliest_possible_end_date(period)
+      end
+
+      def earliest_possible_end_date(period)
+        period.started_on.next_day
+      end
```
rwrrll: Not critical, but not a huge fan of passing the `period` argument forward again, although I do see the upside of having a separate method name. With my pedant hat on I'd also say `date_before_period_starts?` is a misnomer because that's not exactly what it checks.

Serving suggestion:

```suggestion
      def date_permitted_by_period?(period)
        return false if period&.started_on.blank?

        # a period must last at least 1 day
        date <= period.started_on.next_day
      end
```

I think even that name is a bit vague, but I can't say I 100% thoroughly understand what's going on here (e.g. what precisely is `date`?), maybe there's an even more idiot-proof name! Anyway, just a suggestion 🙃

rwrrll (commented):
Looks pretty good to me, big fan of the validator extraction and how that's covered by a dedicated spec. I reckon there's a little more juice to squeeze!

app/wizards/schools/ects/teacher_leaving_wizard/edit_step.rb
thenapking: I wanted it to return nil, when not matched, eg when there is no invalid period.  I could make that explicit.

app/wizards/schools/ects/teacher_leaving_wizard/edit_step.rb
thenapking: Originally, I had done this as you expected, but feedback I received at that point was that it was spilling over from the wizard too much.  I agree the code is almost identical apart from the messaging, and that messaging could stay in the wizard.  I'll have a look at it again with a fresh brain, and see where I can get.

lib/schools/validation/period_boundary.rb
thenapking: I am happy with conditional returns, but I am used to working places where people hate them, and it's conditioned me to always return early.  Now altered.

spec/wizards/schools/register_ect_wizard/start_date_step_spec.rb
thenapking: Yes, none of that was really necessary it turns out!

lib/schools/validation/period_boundary.rb
thenapking: I've tried to make this a little clearer, although it's more verbose.  `date` is the input that this validator is validating.  I've changed that to `input_date` to make that context a little clearer.  Or I could change the name of this validator to `DateBeforePeriodValidator` or something, and leave it as `date`?  I've changed the name of this method then to `input_date_present_and_before_period_starts?`.  It's pretty verbose, but I think it explains exactly what the validator does, whereas `permitted` is too generic.

spec/lib/schools/validation/period_boundary_spec.rb
thenapking: Now altered

spec/features/schools/ects/register/edge_cases/moving_school_backdated_start_date_spec.rb
thenapking: Thanks yes, that was a bit more of a thorough check than the test above I replaced it with. I missed that I'd left it in.  I've combined the two.

lib/tasks/product_review/2662.rake
thenapking: Hmm. Yes. I am going to delete this before merge.  I've found writing rake tasks for the seeds not to be that convenient so I might look for another way.  But I will bear this in mind if I write another.

lib/schools/validation/period_boundary.rb
```diff
@@ -0,0 +1,71 @@
...
+      def invalid_period
+        @invalid_period ||= validate_periods
+      end
+
+      def type
+        case invalid_period
+        when ECTAtSchoolPeriod then "teaching"
+        when TrainingPeriod    then "their latest training"
+        end
+      end
```
rwrrll, line 25: I know you said returning nil on unhandled case was your intention, but just re-iterating that if it does then the error message will be malformed

lib/schools/validation/period_boundary.rb
```diff
@@ -0,0 +1,71 @@
+module Schools
+  module Validation
+    class PeriodBoundary
+      attr_reader :ect_at_school_period, :input_date, :error_message
```
rwrrll: `error_message` attr_reader appears to be unused anywhere

spec/features/schools/ects/register/edge_cases/moving_school_backdated_start_date_spec.rb
```diff
@@ -49,29 +49,31 @@
...
-      scenario "the training period at the first school which has not started is deleted" do
+      scenario "the schools are prevented from registering the ECTs on the same day" do
         given_there_are_two_schools
         and_there_are_partnerships_at_the_schools
         and_the_two_schools_want_to_register_the_same_teacher_on_different_dates_of_which_are_both_backdated
 
         given_a_school_has_registered_a_teacher
-        and_another_school_has_registered_the_same_teacher_at_a_later_date
-
-        then_the_training_period_at_the_first_school_which_has_not_started_yet_should_be_deleted
```
rwrrll, line 60: I think we can delete the method definition too

spec/wizards/schools/register_ect_wizard/start_date_step_spec.rb
```diff
@@ -126,41 +128,99 @@
...
-          allow(instance).to receive(:ect).and_return(
-            Schools::RegisterECTWizard::RegistrationStore.new(store).tap do |registration_store|
-              allow(registration_store).to receive_messages(trs_first_name: "Johnnie", trs_last_name: "Walker")
-            end
-          )
+      context "when the start_date is blank" do
+        let(:start_date) { nil }
+
+        it { is_expected.to have_error(:start_date, "Enter the date the ECT started or will start teaching at your school") }
+      end
```
rwrrll: Looks like this duplicates `context "when the start_date is not present"` around line 36

app/wizards/schools/ects/teacher_leaving_wizard/edit_step.rb
```diff
@@ -40,20 +40,38 @@ def leaving_on_valid
...
+
+        def leaving_on_boundary_validator
+          @leaving_on_boundary_validator ||= Schools::Validation::PeriodBoundary.new(
+            ect_at_school_period:,
+            input_date: leaving_on_input.value_as_date
+          )
+        end
+
+        delegate :type, to: :leaving_on_boundary_validator, prefix: :invalid_period
+        delegate :started_on_formatted, to: :leaving_on_boundary_validator, prefix: :invalid_period
```
rwrrll:
```suggestion
        delegate :type,
                 :started_on_formatted, 
                 to: :leaving_on_boundary_validator, 
                 prefix: :invalid_period
```

Little bit quicker to read by grouping them together, bonus if the other method could follow the same naming convention and it all rolls into one delegate (also similar in the other step)

rwrrll (commented):
Tried to just focus on the changes here, assuming other bits that didn't change have been worked through... n.b. PR is already has approval so just offering extra thoughts here

spec/wizards/schools/register_ect_wizard/start_date_step_spec.rb
thenapking: Neither this nor the example it clashes with were my code, but are being highlighted because of sections I've removed. I've cleaned it up though as you are right - and I will check for more examples that can be removed.

PR #2670, opened 2026-04-15
Fix mentor search to preserve search box when no results found

ericaporter:
This ticket

### Context
- The mentor list controller was setting @number_of_mentors from the filtered search result rather than the total count of mentors at the school so when a search returned no results, the view couldn't distinguish between 'no mentors registered' and 'no mentors matching the search' therefore incorrectly rendered the no-mentors state, hiding the search box entirely

- Controller now makes two separate `Teachers::Search` calls — one without a query to get the true total, and one with the query string to get the filtered results
- This mirrors the existing pattern in `ECTsController`

### Guidance to review
1. Go to mentors list page
2. Search for a name/trn that doesn't exist and confirm the search box stays visible and "There are no mentors that match your search" appears 
3. Clear the search — confirm the mentor list returns to normal
4. Search for a name that does exist — confirm only the matching mentor appears
5.  Find a school with no mentors and go to the mentor list — confirm "Your school currently has no registered mentors" appears with no search box

### DIFF
--- a/app/controllers/schools/mentors_controller.rb
+++ b/app/controllers/schools/mentors_controller.rb
@@ -11,10 +11,9 @@ module Schools
     before_action :set_ects, only: :show
 
     def index
-      search = Teachers::Search.new(mentor_at_school: school, query_string: params[:q])
-
+      @number_of_mentors = Teachers::Search.new(mentor_at_school: school).count
+      search                   = Teachers::Search.new(mentor_at_school: school, query_string: params[:q])
       @pagy, @filtered_mentors = pagy_array(search.search, limit: 20)
-      @number_of_mentors       = search.count
     end
 
     def show
--- a/spec/features/schools/mentors/search_spec.rb
+++ b/spec/features/schools/mentors/search_spec.rb
@@ -21,6 +21,21 @@ RSpec.describe "Searching for a mentor", :enable_schools_interface do
     and_i_should_not_see_the_non_matching_teacher
   end
 
+  scenario "when no results are found, keeps the search box visible and shows a no-results message" do
+    when_i_search_for_a_teacher_that_does_not_exist
+
+    then_i_should_see_the_search_box
+    and_i_should_see_the_no_results_message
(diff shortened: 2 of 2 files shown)


avinhurry (approved):
Looks great and now consistent with the ECTs.

app/controllers/schools/mentors_controller.rb
```diff
@@ -11,10 +11,9 @@ class MentorsController < SchoolsController
     before_action :set_ects, only: :show
 
     def index
-      search = Teachers::Search.new(mentor_at_school: school, query_string: params[:q])
-
+      @number_of_mentors = Teachers::Search.new(mentor_at_school: school).count
+      search                   = Teachers::Search.new(mentor_at_school: school, query_string: params[:q])
```
cpjmcquillan, line 15: If there aren't any mentors, there doesn't feel like much point in searching for any.

Do you think it's worth passing `Teacher.none` to `pagy` in that case and saving ourselves a query?

cpjmcquillan (approved):

app/controllers/schools/mentors_controller.rb
ericaporter: @cpjmcquillan We could do something like this (below) but it won't mirror the ECT search and is a little less readable. I’d probably lean towards keeping the current approach for consistency since its only a single query but also happy to update it to the below

```
def index
  @number_of_mentors = Teachers::Search.new(mentor_at_school: school).count

  if @number_of_mentors.zero?
    @pagy, @filtered_mentors = pagy_array([], limit: 20)
    return
  end

  search = Teachers::Search.new(mentor_at_school: school, query_string: params[:q])
  @pagy, @filtered_mentors = pagy_array(search.search, limit: 20)
end
```

cpjmcquillan: That's fine if we want to keep things consistent. If we run into performance issues we can revisit later 👍🏼

PR #2672, opened 2026-04-15
[2672] Speed up participant declarations endpoint

### LINKED ISSUE #2549: Declarations benchmarking/optimisation
> ⚠️ Timebox to 1 day; we will spin up separate tickets if the performance improvements are going to be complicated.

## Description

We need to ensure the API endpoints in RECT perform at least as well as those in ECF, and ideally they should be more performant.

## Acceptance criteria

- [ ] Benchmark/optimise the GET /api/v3/participant-declarations response
- [ ] Benchmark/optimise the GET /api/v3/participant-declarations/:id response
- [ ] Benchmark/optimise the POST /api/v3/participant-declarations response
- [ ] Benchmark/optimise the PUT /api/v3/participant-declarations/:id/void response

⚠️ Make sure we test all filtering/sorting combinations, as that can have a big impact on performance of an endpoint.

mooktakim:
### Context

Ticket: https://github.com/DFE-Digital/register-early-career-teachers-public/issues/2549

- Add index on `declarations.updated_at`
- Filter delivery partners by foreign key instead of joining on api_id.
- Rewrite teacher filter to use a single predicate over ECT/mentor teacher_id, replacing the OR-of-scopes pattern.

### DIFF
--- a/app/services/api/declarations/query.rb
+++ b/app/services/api/declarations/query.rb
@@ -58,7 +58,14 @@ module API::Declarations
     def where_lead_provider_is(lead_provider_id)
       return if ignore?(filter: lead_provider_id)
 
-      @scope = scope
+      @scope = scope.where(id: declarations_matching_lead_provider(lead_provider_id).select(:id))
+    end
+
+    # Keeps the lead-provider join chain in a sub-relation so the outer scope
+    # is filtered via WHERE id IN (subquery). This makes pagination COUNT(*)
+    # cheaper and lets other filters compose against a clean outer scope.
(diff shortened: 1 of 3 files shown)


leandroalemao (approved):

ethax-ross (approved):
What kind of impact did this have on the endpoint performance in the end @mooktakim?

mooktakim:
> What kind of impact did this have on the endpoint performance in the end @mooktakim?

From the requests that did complete, there was improvement from 4.1x slower to about 3x slower.

PR #2674, opened 2026-04-15
Update the post completion date combo check lists

ltello:
## Context
Records added for ECTs after they have completed induction are not valid. However in some instances there are declarations recorded that mean we need to keep some of these periods.

 update the post completion combo check lists of induction records we need to keep


tonyheadford (approved):
Nice!

peteryates (approved):
Nice

PR #2676, opened 2026-04-15
[3639] funding eligibility should be set for all ect_at_school_periods include future ones

mooktakim:
### Context

Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/3639

* Updated `Teachers::SetFundingEligibility.eligible_for_ect_training?` to check for all `teacher.ect_at_school_periods`, including future

### DIFF
--- a/spec/services/teachers/set_funding_eligibility_spec.rb
+++ b/spec/services/teachers/set_funding_eligibility_spec.rb
@@ -65,6 +65,22 @@ RSpec.describe Teachers::SetFundingEligibility do
       end
     end
 
+    context "when teacher has an ongoing induction but only a finished ECT at school period" do
+      before do
+        FactoryBot.create(:induction_period, :ongoing, teacher:)
+        FactoryBot.create(:ect_at_school_period,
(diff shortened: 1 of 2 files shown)


spec/services/teachers/set_funding_eligibility_spec.rb
```diff
@@ -49,6 +49,32 @@
       end
     end
 
+    context "when teacher has an ongoing induction but only a future-dated ECT at school period" do
```
ethax-ross, line 52: Do we want to cover when they have only a past/finished school period as well?

ethax-ross (approved):

spec/services/teachers/set_funding_eligibility_spec.rb
claire-hughez: I think so - otherwise we run the risk of historical declarations when they were at school not being paid out, which LPs are due. @ethax-ross  @mooktakim

leandroalemao (approved):

PR #2684, opened 2026-04-16
Upgrade to Rails 8.1.3

peteryates:
### Context

Upgrade to Rails 8.1.3

One benefit of this upgrade is that we should be sending structured logs to Logit and Rails 8.1 includes [structured event reporting](https://guides.rubyonrails.org/8_1_release_notes.html#structured-event-reporting), so we can hopefully get this without needing additional dependencies.

- **Upgrade Rails from 8.0.4 to 8.1.3**
- **Upgrade internal version references to 8.1**
- **Refresh the schema so it's alphabetical**
- **Change Normalization include to new location**
- **Fix failing array assertion**
- **Add comment on where tests are failing**
- **Ignore `Metadata::Base` in namespace restriction**


config/application.rb
```diff
@@ -14,7 +14,7 @@
 
 module RegisterEarlyCareerTeachers
   class Application < Rails::Application
-    config.load_defaults 8.0
+    config.load_defaults 8.1
```
cpjmcquillan, line 17: Do we need to check these new defaults? 🤔

In the past I've brought along the `new_framework_defaults` config and shipped each new default to production incrementally.

That might be overkill, though.

spec/services/teachers/manage_spec.rb
```diff
@@ -140,12 +140,12 @@
               trs_qts_awarded_on: [nil, 3.years.ago.to_date],
               trs_qts_status_description: [nil, "QTS status description"]
             }.with_indifferent_access,
-            modifications: [
+            modifications: array_including([
```
cpjmcquillan: One slight nuance here is that now the test would pass if there were any extra modifications.

Could we use `contain_exactly` instead?

cpjmcquillan (approved):

config/application.rb
peteryates: I went through the diff when running `app:update` and it looks fairly innocuous.

spec/services/teachers/manage_spec.rb
peteryates: Yep, good call. Will amend.

PR #2685, opened 2026-04-16
Ensure schools can change an ECTs training before reported leaving date

cpjmcquillan:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/3651

Before, a school could report that an ECT was leaving (or another school could register the ECT as joining) *and* make changes to their training up until the date they left.

Until #2616, this resulted in multiple ongoing training periods —which didn't make sense!

After #2616, this was no longer possible and the validation would prevent the change from happening.

We do want schools to be able to make these changes, though.

This adds some feature tests to handle these edge cases, as well as the changes to ensure training periods finish on the relevant date (i.e the `ECTAtSchoolPeriod#finished_on` if it's present).

Now, we pass the `finished_on` date from the `ect_at_school_period` for the school making the change. If the school period is due to finish in the future (i.e they have reported the teacher is leaving), then the new training period created by the change will be curtailed at that date. If the school period is ongoing, then the new training period will be ongoing too.

### DIFF
--- a/app/services/ect_at_school_periods/switch_training.rb
+++ b/app/services/ect_at_school_periods/switch_training.rb
@@ -106,7 +106,8 @@ module ECTAtSchoolPeriods
     def create_school_led_training_period!
       TrainingPeriods::Create.school_led(
         period: @ect_at_school_period,
-        started_on: date_of_transition
+        started_on: date_of_transition,
+        finished_on: @ect_at_school_period.finished_on
       ).call
     end
 
@@ -114,6 +115,7 @@ module ECTAtSchoolPeriods
       TrainingPeriods::Create.provider_led(
         period: @ect_at_school_period,
         started_on: date_of_transition,
+        finished_on: @ect_at_school_period.finished_on,
         school_partnership: earliest_matching_school_partnership,
         expression_of_interest:,
         author: @author,
@@ -129,6 +131,7 @@ module ECTAtSchoolPeriods
       TrainingPeriods::Create.provider_led(
         period: @mentor_at_school_period,
         started_on: date_of_transition,
+        finished_on: @mentor_at_school_period.finished_on,
         school_partnership: earliest_matching_school_partnership,
         expression_of_interest:,
         author: @author
--- a/app/services/schedules/reuse.rb
+++ b/app/services/schedules/reuse.rb
@@ -56,6 +56,5 @@ module Schedules
     delegate :successor_contract_period, to: :contract_period_reassignment
     delegate :assigned_contract_period, to: :contract_period_reassignment
     alias_method :previous_contract_period, :assigned_contract_period
-    alias_method :started_on, :date_of_transition
   end
 end
--- a/spec/services/schools/register_ect_spec.rb
+++ b/spec/services/schools/register_ect_spec.rb
@@ -228,6 +228,47 @@ RSpec.describe Schools::RegisterECT do
           end
         end
 
+        context "when a Teacher has an ECT period finishing in future at a different school (reported leaving)" do
+          let(:other_school) { FactoryBot.create(:school) }
+          let(:started_on) { Date.new(2025, 8, 1) + 1.year }
+
+          before do
+            FactoryBot.create(
+              :ect_at_school_period,
+              teacher:,
+              school: other_school,
+              started_on: 6.months.ago,
+              finished_on: 6.months.from_now
+            )
+          end
+
+          it "allows registration with non-overlapping future date" do
+            expect { service.register! }.to change(ECTAtSchoolPeriod, :count).by(1)
+          end
+        end
+
+        context "when a Teacher already has an ongoing ECT period starting in future at a different school (reported joining)" do
+          let(:other_school) { FactoryBot.create(:school) }
+          let(:started_on) { Date.new(2025, 8, 1) + 1.year }
+
+          before do
+            FactoryBot.create(
+              :ect_at_school_period,
+              :ongoing,
+              teacher:,
+              school: other_school,
+              started_on: started_on.advance(weeks: 2)
+            )
+          end
+
+          it "raises an error" do
+            expect { service.register! }.to raise_error(
+              ActiveRecord::RecordInvalid,
+              "Validation failed: Finished on End date cannot overlap another Teacher ECT period"
+            )
+          end
+        end
+
         context "when no SchoolPartnerships exist" do
           it "creates a TrainingPeriod linked to the ECTAtSchoolPeriod and with an expression of interest for the ActiveLeadProvider" do
             expect { service.register! }.to change(TrainingPeriod, :count).by(1)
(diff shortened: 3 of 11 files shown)


spec/features/schools/ects/change/edge_cases/changing_training_before_reported_leaving_date_spec.rb
```diff
@@ -0,0 +1,321 @@
...
+    leaving_date_fieldset.get_by_label("Day").fill(on.day.to_s)
+    leaving_date_fieldset.get_by_label("Month").fill(on.month.to_s)
+    leaving_date_fieldset.get_by_label("Year").fill(on.year.to_s)
+    and_i_continue
+    and_i_confirm_and_continue
+    confirmation_message = "Mr Teacher will be removed from your school’s ECT list"
+    then_panel_is_visible_with(message: confirmation_message)
+  end
+
+  def given_the_ect_has_been_registered_by_another_school(on:)
```
cpjmcquillan: This feels a bit cheeky for a feature test, but it seemed pragmatic. 

The alternative was logging in as the other school and stepping through the whole journey.

We can't just create another `ECTAtSchoolPeriod` and `TrainingPeriod`, because we depend on periods at the original school getting finished when the ECT is registered by the new school.

spec/services/schools/register_ect_spec.rb
```diff
@@ -228,6 +228,44 @@
...
+              finished_on: 6.months.from_now
+            )
+          end
+
+          it "allows registration with non-overlapping future date" do
+            expect { service.register! }.to change(ECTAtSchoolPeriod, :count).by(1)
+          end
+        end
+
+        context "when a Teacher already has an ongoing ECT period starting in future at a different school (reported joining)" do
```
cpjmcquillan, line 250: Are we comfortable with this? @mason-emily 

Essentially the question here to answer is:

**What should happen if multiple schools register an ECT?**

- School A registers ECT on 10th September 2025
- School A reports ECT leaving on 31st December 2025
- School B registers ECT as starting after 31st December 2025
- School C registers ECT as starting after 31st December 2025

Is this actually possible? Should we add a check to the `find-ect` step to prevent this?

Currently if this was possible and it did happen, the underlying service would raise an error because there can't be multiple ongoing periods, which I think is fine.

cpjmcquillan: We discussed and think this is fine 👍🏼

app/services/ect_at_school_periods/switch_training.rb
```diff
@@ -106,14 +106,16 @@ def finish_training_period!
...
+        finished_on: @ect_at_school_period.finished_on
       ).call
     end
 
     def create_provider_led_training_period_for_ect_at_school_period!
       TrainingPeriods::Create.provider_led(
         period: @ect_at_school_period,
         started_on: date_of_transition,
+        finished_on: @ect_at_school_period.finished_on,
         school_partnership: earliest_matching_school_partnership,
```
ericaporter, line 119: Do we need to do the same thing (i.e. pass finished_on) for create_provider_led_training_period_for_mentor_at_school_period! in this file too?

spec/features/schools/ects/change/edge_cases/changing_training_before_reported_leaving_date_spec.rb
```diff
@@ -0,0 +1,322 @@
...
+      )
+
+      when_i_visit_the_admin_teacher_training_page
+      then_i_see_the_correct_training_periods(
+        { type: "School-led", start: Date.current, end: @leaving_date },
+        { type: "Provider-led", start: @ect_at_school_period.started_on, end: Date.current }
+      )
+    end
+  end
+
```
ericaporter: Do you think it's worth adding in a switching to provider-led when a future end date exists example?

spec/features/schools/ects/change/edge_cases/changing_training_before_reported_leaving_date_spec.rb
ericaporter: Makes sense!

spec/features/schools/ects/change/edge_cases/changing_training_before_reported_leaving_date_spec.rb
```diff
@@ -0,0 +1,322 @@
...
+      summary_card = summary_cards[index]
+
+      urn_row = summary_card
+        .locator("dl dt", hasText: "School URN")
+        .locator("..") # Navigate "up" to the parent summary list row
+      expect(urn_row.locator("dd")).to have_text(period_hash[:urn].to_s)
+
+      start_date_row = summary_card
+        .locator("dl dt", hasText: "School start date")
+        .locator("..")
```
ericaporter: Not a blocker, but I think we often target summary cards by identifying text e.g. (page.find(".govuk-summary-card", text: ...), hasText: etc.) rather than relying on card order. Might be worth doing that here and below as well, so the assertions are a bit less brittle if the ordering changes?

spec/services/schools/register_ect_spec.rb
```diff
@@ -228,6 +228,44 @@
...
+              :ect_at_school_period,
+              :ongoing,
+              teacher:,
+              school: other_school,
+              started_on: started_on.advance(weeks: 2)
+            )
+          end
+
+          it "raises an error" do
+            expect { service.register! }.to raise_error ActiveRecord::RecordInvalid
```
ericaporter: Is there an error message we can add here to know why it fails? If not maybe we can just describe it in the expectation

ericaporter (commented):
This looks good 👍 just wondering if you think we need to cover `LeadProviderChanger` in this update too - it calls TrainingPeriods::Create.provider_led without passing finished_on, so changing lead provider while an ECT has a future school period would likely create a similarly open-ended training period

spec/features/schools/ects/change/edge_cases/changing_training_before_reported_leaving_date_spec.rb
cpjmcquillan: Yes, good idea 👌🏼. I think I need to add tests for switching from provider-led to school-led too, those would have flagged all the things I missed!

spec/features/schools/ects/change/edge_cases/changing_training_before_reported_leaving_date_spec.rb
cpjmcquillan: I'm happy to make that change for the part that asserts against the school periods, but I don't think it makes sense for the training periods. There isn't really a reliable way to retrieve a given training period just on some text.

Also, the order of the training periods is important —since the cards are supposed to be displayed in order.

ericaporter (approved):
Looks good!! 👍

app/services/schedules/reuse.rb
```diff
@@ -56,6 +56,5 @@ def existing_schedule
     delegate :successor_contract_period, to: :contract_period_reassignment
     delegate :assigned_contract_period, to: :contract_period_reassignment
     alias_method :previous_contract_period, :assigned_contract_period
-    alias_method :started_on, :date_of_transition
```
cpjmcquillan, line 59: While I was digging into https://github.com/DFE-Digital/register-ects-project-board/issues/3815, this confused me _alot_. I think it makes things less clear, so I removed it!

PR #2686, opened 2026-04-16
Move patches CSV to `DATA_PATCHES` global constant

peteryates:
This change will allow us to load the CSV once when the app boots rather than once per participant. It's intended to ease the migration and will be removed immediately once it's done.

The other options were:

* loading it into the database - which would be fine but also be more work as we'd need migrations/model and a way of syncing/overwriting data
* loading it into Redis when the app boots - also a good option but some more legwork would be required in setting it up and making sure it refreshes nicely each time the app boots (without leaving remnants of previous versions behind, etc)

This is the simplest approach and I don't think there's really much that can go wrong, providing the CSV doesn't grow too large. We can boost the RAM in prod for a fortnight if necessary, but I think it'll be fine as is.

### DIFF
--- a/app/migration/ecf1_teacher_history/data_patcher.rb
+++ b/app/migration/ecf1_teacher_history/data_patcher.rb
@@ -1,9 +1,7 @@
 class ECF1TeacherHistory::DataPatcher
-  PATCHES_PATH = Rails.root.join("app/migration/ecf1_teacher_history/data_patches.csv")
-
   attr_reader :data_patches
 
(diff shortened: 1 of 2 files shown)


tonyheadford (approved):

avinhurry (approved):
Probably worth adding a card to the backlog so we remember to remove it

peteryates:
I don't think we'll be able to forget because once the file's gone the app won't boot! :sweat_smile:

PR #2562, opened 2026-03-30
Support closed-contract ECT registration reassignment to 2024 for partnership and EOI reuse

ericaporter:
### Context
Ticket: here

This PR implements the reassignment logic for ECTs whose latest training period is in a closed 2021 or 2022 contract period.

When such ECTs are registered again, their training is reassigned to the 2024 contract period, ensuring they can continue training under the appropriate contract terms.

### Changes proposed in this pull request

- Introduced ContractPeriods::ForECTRegistration to resolve the correct contract period for registration
   - Repurposed the previously unused ContractPeriods::ForECT into ContractPeriods::ForECTRegistration to explicitly handle contract period resolution during ECT registration flows
- Using ContractPeriods::Reassignment to encapsulate logic for moving ECTs from closed 2021/2022 provider-led periods to 2024
- Updated registration queries to pass previous_training_period into the resolver
- Ensured provider-led ECTs from closed 2021/2022 periods are reassigned to 2024
- Reuse of previous training choices now evaluated against the resolved (2024) contract period
- New training periods are created as:
   - confirmed partnership if available in 2024
   - expression of interest (EOI) if no partnership exists

Specs:
- Added coverage for 2021/2022 → 2024 reassignment, reuse (partnership/EOI), school view, and API behaviour
- Added unit tests for reuse logic and contract period resolution
- Refactored specs to reduce duplication, simplify setup, and group scenarios by reuse type

### Guidance to review

New Commits: 
Several edge cases in the reuse screen logic were found during testing and fixed:
    - EOI lookups fixed to scope to the current school and last chosen lead provider
    - payments_frozen_reassignment? now correctly handles EOI-based previous training periods 
    - Reuse screen now correctly shows for frozen contract period ECTs regardless of whether their previous training period used a confirmed partnership or EOI

PR #2688, opened 2026-04-16
[3771] Update logic for training period start date

rwrrll:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/3771

Unless the ECT is having their contract period reassigned from a frozen one, the effective training start date should be the later of when they started at the school and the registration date.

### Guidance for review

There was concern about side-effects elsewhere in the application, I've tried to research these as much as I can, I'd welcome any knowledge of other places to check:

#### Schedule assignment no change

[`Schedules::Find`](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/72ccb48b7794edcd10638cd59ff7367ceda671f9/app/services/schedules/find.rb#L49) already uses `[started_on, Time.zone.today].max` internally to select the schedule. Since both the old and new logic feed into this same max, the schedule selected is identical in all cases.

#### Contract period reassignment (PR #2562 interaction)

`ContractPeriods::ForECTRegistration` can return a past contract period (2024) for ECTs reassigned from closed 2021/2022 contract periods. Confirmed that [`Schedules::Find` bypasses `started_on` for the extended-schedule path](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/72ccb48b7794edcd10638cd59ff7367ceda671f9/app/services/schedules/find.rb#L24), so the schedule is unaffected. In the changes to the registration service in this PR, I preserved the old logic for the reassignment case to keep `training_period.started_on` within its contract period.

#### Validations on `TrainingPeriod`

- `enveloped_by_trainee_at_school_period` only fires when `finished_on` is present; new training periods are open-ended, so no risk.
- `contract_period_consistent_across_associations` still passes because schedule contract period selection is unchanged.

#### Registration flows (partnership reuse, EOI reuse, no reuse)

`RegisterECT#training_period_started_on` is the single point where the date is determined for all paths. The wizard's reuse steps populate the store, then `RegisterECT#register!` creates the training period — no parallel creation logic exists.

#### Transfer / moving school behaviour change

When a backdated ECT transfers to another school, `ECTAtSchoolPeriods::Finish` closes their ECT period. Under the old logic the training period started on the contract period start (often before the transfer date), so it survived. Under the new logic it starts on `Date.current`, which may be after the transfer date, in which case the training period is treated as "not started yet" and deleted. This is consistent with the existing "both backdated" scenario behaviour.

#### API surface

`training_period.started_on` is exposed in the [school transfer serializer](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/serializers/api/teachers/school_transfer_serializer.rb). Lead providers will see the registration date instead of the contract period start for backdated registrations. Based on ticket comments from Aamina / Emily I think this is fine / expected.

#### Mentor registration

`Schools::RegisterMentor` uses `mentor_at_school_period.started_on` directly, no `max` logic. Unchanged and already inconsistent with ECT registration, not in scope for this ticket. (Maybe warrants looking at separately?)

#### Misc unaffected

  - Events/audit trail (records the training period as-is, no date-specific logic)
  - Declarations/statements (tied to contract period via schedule, which is unchanged)
  - Scopes filtering by started_on (queries in the registration wizard use <= against contract period finished_on, still satisfied)
  - `ContractPeriods::Reassignment` (never reads training_period.started_on)

### DIFF
--- a/spec/features/schools/ects/register/edge_cases/moving_school_backdated_start_date_spec.rb
+++ b/spec/features/schools/ects/register/edge_cases/moving_school_backdated_start_date_spec.rb
@@ -32,7 +32,7 @@ RSpec.describe "Moving School - backdated start spec", :enable_schools_interface
     end
 
     context "when the first period has a back-dated start date" do
-      scenario "the training period at the first school is aligned to the current contract period" do
+      scenario "the training period at the first school starts on the registration date" do
         given_there_are_two_schools
         and_there_are_partnerships_at_the_schools
         and_the_two_schools_want_to_register_the_same_teacher_on_different_dates_of_which_the_earliest_is_backdated
@@ -40,7 +40,7 @@ RSpec.describe "Moving School - backdated start spec", :enable_schools_interface
         given_a_school_has_registered_a_teacher
         and_another_school_has_registered_the_same_teacher_at_a_later_date
 
-        then_the_training_period_for_the_first_school_should_start_on_the_first_day_of_the_current_contract_period
+        then_the_training_period_for_the_first_school_should_start_on_the_registration_date
         and_the_training_period_for_the_second_school_should_start_on_the_date_entered_by_the_school
 
         and_the_training_period_for_the_first_school_should_finish_on_the_day_before_training_starts_at_the_second_school
@@ -113,13 +113,13 @@ RSpec.describe "Moving School - backdated start spec", :enable_schools_interface
   end
 
   def and_the_two_schools_want_to_register_the_same_teacher_on_different_dates_which_are_not_backdated
-    @school_one_start_date = Date.new(2025, 6, 1)
-    @school_two_start_date = Date.new(2025, 6, 3)
+    @school_one_start_date = Date.new(2025, 6, 16)
+    @school_two_start_date = Date.new(2025, 6, 18)
   end
 
   def and_the_two_schools_want_to_register_the_same_teacher_on_different_dates_of_which_the_earliest_is_backdated
     @school_one_start_date = Date.new(2025, 4, 1)
-    @school_two_start_date = Date.new(2025, 6, 3)
+    @school_two_start_date = Date.new(2025, 6, 18)
   end
 
   def and_the_two_schools_want_to_register_the_same_teacher_on_different_dates_of_which_are_both_backdated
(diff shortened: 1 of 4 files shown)


spec/services/schools/register_ect_spec.rb
```diff
@@ -12,24 +12,26 @@
...
-  include_context "safe_schedules"
-
+  let(:store) { nil }
   let(:author) { FactoryBot.create(:school_user, school_urn: school.urn) }
   let(:school_reported_appropriate_body) { FactoryBot.create(:appropriate_body_period) }
   let(:corrected_name) { "Randy Marsh" }
   let(:email) { "randy@tegridyfarms.com" }
   let(:school) { FactoryBot.create(:school) }
-  let(:started_on) { mid_year - 1.day }
+  let(:started_on) { mid_year }
```
rwrrll, line 25: The only thing the `- 1.day` was doing was making `started_on < Date.current`, which had no effect under the old logic (both values were well after `registration_contract_period.started_on`). No test was deliberately exercising "school start in the recent past" behaviour, it was incidental.

cpjmcquillan:
> enveloped_by_trainee_at_school_period only fires when finished_on is present; new training periods are open-ended, so no risk

This isn't strictly true, if an ECT or Mentor is leaving (i.e their `at_school_period` is not `nil`) then we will pass that through to the training period so it finishes on the same date. (at least we will once #2685 is merged)

cpjmcquillan (approved):
Looks good to me.

I think it's worth us adding a ticket to the backlog to investigate whether we should align the register mentor service, too.

I added a comment about newly-created training periods with a finished on in the future. Since those will always be in the future (and never beyond the school period), _I think_ dates should still always be enveloped and the training periods should be valid. It might be worth adding some tests to verify that explicitly, though.

Nice work!

spec/features/schools/ects/register/edge_cases/moving_school_backdated_start_date_spec.rb
```diff
@@ -32,15 +32,15 @@
     end
 
     context "when the first period has a back-dated start date" do
-      scenario "the training period at the first school is aligned to the current contract period" do
+      scenario "the training period at the first school starts on the registration date" do
```
thenapking, line 35: It's a small thing, but I find this new test description a little less easy to understand.  The "aligned to" was indicating that the service had made a change to the date, and it had changed it to the beginning of the current contract period.  In this context, registration date is ambiguous, so it doesn't seem like the service is doing anything.  Could we say something like "the training period at the first school is set to start on the current date"?

thenapking (approved):
LGTM!

PR #2689, opened 2026-04-16
[2250] Redesign Admin Teacher Index Page

avinhurry:
### Summary

A redesign of the admin teacher index page from cards to a table, including the supporting search/filter behaviour.

| Before | After |
|--------|-------|
|image|image|

> [!IMPORTANT]
> This index paginates teachers in postgres and expands the current page into ECT/mentor rows in Ruby.
>
> Filtering, ordering and pagination remain database backed across the full dataset, while avoiding the complexity of row level SQL pagination.
>
> A small amount of targeted SQL is used to ensure contract period filtering reflects the value shown for each row (latest at school period -> latest training period -> contract period).

### Notes

- Search now supports API participant ID as well as existing name/TRN
- Teachers with no ECT or mentor role history appear as `No role assigned`
- `school-led` ECT rows show `Not available` for contract period
- Invalid full text search input now returns no results instead of an internal server error


rwrrll (commented):
I see the performance concern and why we’ve gone down this route. Mega work rolling this query! 💪

The complexity makes me a bit uncomfortable. Pure row-level pagination is coming in at a heavy price: it looks like the RowQuery service, CTE + UNION ALL + LATERAL joins, derived-table wrapping, separate count override, and role_sort_order tiebreaker all exist to meet A/C 1e (dual-role teachers appear twice).

A small tradeoff that seems like it would eliminate most of the complexity would be to consider the pagination as “25 teachers per page”, rather than strictly “25 rows per page”.  If we keep all the pagination at the teacher level, but expand to rows in Ruby, each page would contain 25 teachers and render 25–50 rows (depending on how many are dual-role). That feels like a minor UX wrinkle, and in exchange we'd get:

- no RowQuery (~270 lines gone)
- Teachers::Search stays as-is
- controller is search -> filter -> paginate in a handful of lines
- role/contract-period filters become normal joins
- "current or next else latest" can live as a scope on the period models, reusing the current_or_next_* associations already on Teacher
- teachers with both roles never have their results split across a pagination boundary

Even at hundreds of thousands of teachers, this way should be fine in terms of scalability -- the extra work in Ruby has an upper bound of 50 rows per request, so shouldn’t meaningfully hurt scalability. It looks like all the search/filter stuff we want would work with this approach, too. What do you think?

avinhurry:
> I see the performance concern and why we’ve gone down this route. Mega work rolling this query! 💪
> 
> The complexity makes me a bit uncomfortable. Pure row-level pagination is coming in at a heavy price: it looks like the RowQuery service, CTE + UNION ALL + LATERAL joins, derived-table wrapping, separate count override, and role_sort_order tiebreaker all exist to meet A/C 1e (dual-role teachers appear twice).
> 
> A small tradeoff that seems like it would eliminate most of the complexity would be to consider the pagination as “25 teachers per page”, rather than strictly “25 rows per page”. If we keep all the pagination at the teacher level, but expand to rows in Ruby, each page would contain 25 teachers and render 25–50 rows (depending on how many are dual-role). That feels like a minor UX wrinkle, and in exchange we'd get:
> 
> * no RowQuery (~270 lines gone)
> * Teachers::Search stays as-is
> * controller is search -> filter -> paginate in a handful of lines
> * role/contract-period filters become normal joins
> * "current or next else latest" can live as a scope on the period models, reusing the current_or_next_* associations already on Teacher
> * teachers with both roles never have their results split across a pagination boundary
> 
> Even at hundreds of thousands of teachers, this way should be fine in terms of scalability -- the extra work in Ruby has an upper bound of 50 rows per request, so shouldn’t meaningfully hurt scalability. It looks like all the search/filter stuff we want would work with this approach, too. What do you think?

Thanks @rwrrll, this was really helpful, and I agree with the direction of simplifying.

I've reworked this to paginate at the teacher level and expand the current page to ECT/mentor rows in Ruby. That removes the CTE/UNION/lateral query and custom count logic, so the overall shape is much simpler.

One place this didn't quite work was contract period filtering becoming a normal join. The value we display is derived from the selected training period (latest at school period -> then latest training period), so the filter needs to answer:

> does the row we would display resolve to X?

rather than:

> has this teacher ever had X?

A simple join gives the latter, which can be incorrect (for example matching an older training period instead of the one shown).

So I've kept a small amount of targeted SQL in `Search` to filter on the contract period of the latest at school/training period.

Everything else now happens in Ruby on the paginated set.

All changes since your last review are in the [latest commit](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/2689/changes/102c3a8c82da492eb9c6fc3159c96c502a055012).

app/helpers/admin/teachers_helper.rb
```diff
@@ -0,0 +1,28 @@
+module Admin::TeachersHelper
+  FilterOption = Data.define(:value, :name)
+
+  def admin_teacher_index_params
+    {
+      page: params[:page],
+      q: params[:q],
+      role: params[:role],
+      contract_period: params[:contract_period]
+    }.compact
```
rwrrll, line 10: Optionally `params.slice(:page, :q, :role, :contract_period).compact` would work I think since keys are aligned

app/helpers/admin/teachers_helper.rb
```diff
@@ -0,0 +1,28 @@
+module Admin::TeachersHelper
+  FilterOption = Data.define(:value, :name)
```
rwrrll, line 2: `Data` 😍

app/services/admin/teachers/rows.rb
```diff
@@ -0,0 +1,124 @@
...
+        # Rows are role based, so a single teacher can produce multiple rows
+        # (e.g. one for ECT and one for mentor).
+
+        rows = []
+        rows << build_row(teacher:, role: "ect", role_period: role_period_for(teacher, "ect")) if ect?(teacher)
+        rows << build_row(teacher:, role: "mentor", role_period: role_period_for(teacher, "mentor")) if mentor?(teacher)
+
+        return rows if rows.any?
+
+        [build_row(teacher:, role: NO_ROLE_ASSIGNED)]
```
rwrrll:
```suggestion
        rows = []
        rows << build_row(teacher:, role: "ect", role_period: role_period_for(teacher, "ect")) if ect?(teacher)
        rows << build_row(teacher:, role: "mentor", role_period: role_period_for(teacher, "mentor")) if mentor?(teacher)
        rows << build_row(teacher:, role: NO_ROLE_ASSIGNED) if rows.none?

        rows
```

Mildly prefer this, pretty subjective though, so a serving suggestion only!

app/services/admin/teachers/rows.rb
```diff
@@ -0,0 +1,124 @@
...
+      end
+
+      def ect?(teacher)
+        teacher.latest_ect_at_school_period.present? ||
+          teacher.induction_periods.any?
+      end
```
rwrrll: I was gonna suggest maybe wrapping `teacher` with a `TeacherPresenter` to call common definition of `#full_name`, `#ect?`, `#mentor?` etc. on the object, but I noticed [the presenter defines this differently](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/951492ca5901f0e77b97919ba1b61995b7a3039c/app/presenters/admin/teacher_presenter.rb#L16), so checking that's intentional?

app/services/admin/teachers/rows.rb
```diff
@@ -0,0 +1,124 @@
...
+        rows.sort_by { |row| row_sort_key(row) }
+      end
+
+      def row_sort_key(row)
+        [
+          row.name.to_s.downcase,
+          ROLE_SORT_ORDER.fetch(row.role),
+          row.teacher.id
+        ]
+      end
```
rwrrll: This could live in Row as `Row#sort_key`, or even better the comparator method

app/models/teacher.rb
```diff
@@ -42,6 +42,7 @@ class Teacher < ApplicationRecord
   has_one :earliest_mentor_at_school_period, -> { earliest_first }, class_name: "MentorAtSchoolPeriod"
   has_one :current_appropriate_body_period, through: :ongoing_induction_period, source: :appropriate_body_period # NB or TSH through period
   has_one :current_or_next_ect_at_school_period, -> { current_or_future.earliest_first }, class_name: "ECTAtSchoolPeriod"
+  has_one :current_or_next_mentor_at_school_period, -> { current_or_future.earliest_first }, class_name: "MentorAtSchoolPeriod"
```
rwrrll: Not necessarily against adding this for consistency, but it's not called anywhere, so just checking it wasn't a leftover

spec/services/admin/teachers/rows_spec.rb
```diff
@@ -0,0 +1,286 @@
...
+          :training_period,
+          :for_mentor,
+          :ongoing,
+          mentor_at_school_period:,
+          school_partnership: mentor_school_partnership
+        )
+      end
+
+      it "returns one row for each role" do
+        matching_rows = rows_builder.rows(teachers).select { |row| row.teacher == teacher }
```
rwrrll, line 55: Very small thing but this occurs in a few places, maybe worth a `subject` (or maybe not!)

spec/services/admin/teachers/rows_spec.rb
```diff
@@ -0,0 +1,286 @@
...
+  let(:teacher_ids) { Teacher.select(:id) }
+  let(:training_period_preload) do
+    {
+      latest_training_period: :schedule
+    }
+  end
+
+  describe "#rows" do
+    context "when a teacher has both ECT and mentor roles" do
+      let!(:teacher) { FactoryBot.create(:teacher, trs_first_name: "Naruto", trs_last_name: "Uzumaki") }
```
rwrrll, line 16: This one comes up 4 times, maybe it could be hoisted and DRYed?

app/services/admin/teachers/rows.rb
rwrrll, line 75: I was gonna suggest delegating to a TeacherPresenter to call common definition of `#full_name`, `#ect?`, `#mentor?` etc. on the object, but I noticed [the presenter defines this differently](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/951492ca5901f0e77b97919ba1b61995b7a3039c/app/presenters/admin/teacher_presenter.rb#L16), so checking that's intentional?

rwrrll (approved):
LGTM, a few extra ideas but only if we have time and appetite!

app/services/admin/teachers/rows.rb
```diff
@@ -0,0 +1,124 @@
...
+        sort_rows(rows)
+      end
+
+    private
+
+      attr_reader :role, :contract_period
+
+      def rows_for_teacher(teacher)
+        # Rows are role based, so a single teacher can produce multiple rows
+        # (e.g. one for ECT and one for mentor).
```
peteryates, line 63: This feels a bit weird to me but I can see why we did it.

peteryates:
It all looks good to me, code's neat and easy to follow.

The querying seems really complex though, I'd have thought we'd be able to get at what we need with the existing relationships and scopes. Maybe we can walk through it a bit so I can understand the thinking behind it?

app/helpers/admin/teachers_helper.rb
avinhurry: We could but I'd rather keep the helper decoupled from the controller param behaviour.

app/services/admin/teachers/rows.rb
avinhurry: Agreed, thank you! Updated

app/models/teacher.rb
avinhurry: Good spot. I added and used this for my first iteration, I've now removed it.

app/services/admin/teachers/rows.rb
avinhurry: I've gone with your `Row#sort_key` suggestion, the comparator method felt a bit overkill here.

spec/services/admin/teachers/rows_spec.rb
avinhurry: I've left this as it is, `row_builder` is already the real subject. The repetition is small and I didn't think a helper buys enough to justify the indirection.

spec/services/admin/teachers/rows_spec.rb
avinhurry: I reckon we probably leave that one too, each context sets up a different history, so keeping the teacher local keeps the examples clearer.
 
I like a bit of intentional duplication in specs, I think it makes things easier to reason about.

app/services/admin/teachers/rows.rb
avinhurry: My understanding was that a teacher should be treated as an ECT here if they have any induction history, even if they don’t have a current `ect_at_school_period`.

I've kept that logic local for now to avoid widening the behaviour change, but happy to take a deeper look at this in a follow up to check if we can align them.

Great suggestion, cheers 👍🏽

avinhurry:
> It all looks good to me, code's neat and easy to follow.
> 
> The querying seems really complex though, I'd have thought we'd be able to get at what we need with the existing relationships and scopes. Maybe we can walk through it a bit so I can understand the thinking behind it?

Thanks @peteryates, happy to go through it 👍🏽 

The tricky bit is that `contract_period` is derived from the latest at school period -> latest training period, so we need to filter on that specific path across the full dataset before pagination, rather than any historical one. A normal join doesn't quite allow us to do that, which is why there’s still some SQL in there.

app/services/admin/teachers/rows.rb
avinhurry: Closing the loop here, it turns out this is more complicated. I've added a ticket to try and get to the bottom of it, then we can align the definitions in the codebase:

https://github.com/DFE-Digital/register-ects-project-board/issues/3846

PR #2698, opened 2026-04-17
update patch csv with 12k records

tonyheadford:
### Context

Update the patch CSV with 12k records
Fix real examples affected by patch


gen-sole:
There are some issues with the specs and Tony will fix them today

ltello (approved):

PR #2702, opened 2026-04-20
[3780] API transfers index endpoint performance

peterdavidhamilton:
### Context

https://cpd-ec2-paritycheck-web.teacherservices.cloud/migration/parity-checks/11

Whilst `GET /api/v3/participants/:id/transfers` was fine, the index endpoint and how it filters were performing poorly.

- `GET /api/v3/participants/transfers` 
- `GET /api/v3/participants/transfers?filter[updated_since]=xxx` 

I believe the performance issue is in `API::Teachers::SchoolTransfers::Query` and manifests in the `.for_lead_provider` scope and `#where_updated_since` filtering and can be mitigated by adding indices that will make querying large datasets more efficient.

Benchmarking each provider (20 iterations against seed data) suggests that the transfers index endpoint will be more consistent in the parity check with this in place.

### Guidance to review

chart_improvement_percentages


app/services/api/teachers/school_transfers/query.rb
```diff
@@ -71,29 +71,22 @@ def where_lead_provider_is(lead_provider_id)
...
-          .where(
-            "(training_periods.id IS NOT NULL AND training_periods.api_transfer_updated_at >= :updated_since AND active_lead_providers.lead_provider_id = :lead_provider_id) OR
-            (latest_training_periods_ect_at_school_periods.id IS NOT NULL AND latest_training_periods_ect_at_school_periods.api_transfer_updated_at >= :updated_since AND active_lead_providers_lead_provider_delivery_partnerships.lead_provider_id = :lead_provider_id) OR
-            (earliest_training_periods_mentor_at_school_periods.id IS NOT NULL AND earliest_training_periods_mentor_at_school_periods.api_transfer_updated_at >= :updated_since AND active_lead_providers_lead_provider_delivery_partnerships_2.lead_provider_id = :lead_provider_id) OR
-            (latest_training_periods_mentor_at_school_periods.id IS NOT NULL AND latest_training_periods_mentor_at_school_periods.api_transfer_updated_at >= :updated_since AND active_lead_providers_lead_provider_delivery_partnerships_3.lead_provider_id = :lead_provider_id)",
-            updated_since:, lead_provider_id:
-          )
+      training_periods =
```
ethax-ross: I think this is a behaviour change when compared to the original filtering implementation?

It looks like you're filtering to teachers that have a training period that has been updated since the provided date, which isn't quite what we want here. 

The original code is complex as what it's doing is filtering to teachers that have a training period that has been updated since the provided date, but **only** where that training period is either the **first** or **last** in its school period. 

It's effectively trying to get the training periods that make up a transfer (which is always the first/last in the school period). It doesn't _quite_ achieve this though, as it will also include the very first training period for a teacher and their very last training period, even if they have completed training (the transfer history service correctly excludes these).

It looks like we're missing a test for this though! it would be interesting to see what effect this has on the results we return to lead providers; if its minimal (so we return mostly the same teacher IDs before/after this change) then I think having it simplified like this would be worth it.

leandroalemao (approved):

app/services/api/teachers/school_transfers/query.rb
```diff
@@ -68,36 +75,63 @@ def where_lead_provider_is(lead_provider_id)
         )
     end
 
-    def where_updated_since(updated_since, lead_provider_id)
-      return if ignore?(filter: updated_since)
+    def participant_ids
```
ethax-ross: Can we avoid the `participant` terminology here and go for `teacher_ids`?

app/services/api/teachers/school_transfers/query.rb
```diff
@@ -68,36 +75,63 @@ def where_lead_provider_is(lead_provider_id)
...
-          .where(
-            "(training_periods.id IS NOT NULL AND training_periods.api_transfer_updated_at >= :updated_since AND active_lead_providers.lead_provider_id = :lead_provider_id) OR
-            (latest_training_periods_ect_at_school_periods.id IS NOT NULL AND latest_training_periods_ect_at_school_periods.api_transfer_updated_at >= :updated_since AND active_lead_providers_lead_provider_delivery_partnerships.lead_provider_id = :lead_provider_id) OR
-            (earliest_training_periods_mentor_at_school_periods.id IS NOT NULL AND earliest_training_periods_mentor_at_school_periods.api_transfer_updated_at >= :updated_since AND active_lead_providers_lead_provider_delivery_partnerships_2.lead_provider_id = :lead_provider_id) OR
-            (latest_training_periods_mentor_at_school_periods.id IS NOT NULL AND latest_training_periods_mentor_at_school_periods.api_transfer_updated_at >= :updated_since AND active_lead_providers_lead_provider_delivery_partnerships_3.lead_provider_id = :lead_provider_id)",
-            updated_since:, lead_provider_id:
-          )
+    def scoped_ect_at_school_periods
```
ethax-ross: `boundary_ect_at_school_periods` and `boundary_mentor_at_school_periods`?

app/services/api/teachers/school_transfers/query.rb
```diff
@@ -68,36 +75,63 @@ def where_lead_provider_is(lead_provider_id)
...
+    def boundary_training_period_ids
+      [*scoped_ect_at_school_periods, *scoped_mentor_at_school_periods].flat_map { |at_school_period|
+        [
+          at_school_period.earliest_training_period&.id,
+          at_school_period.latest_training_period&.id
+        ]
+      }.compact.uniq
+    end
+
+    def recent_boundary_periods
```
ethax-ross: `boundary_training_periods_updated_since`?

app/services/api/teachers/school_transfers/query.rb
```diff
@@ -68,36 +75,63 @@ def where_lead_provider_is(lead_provider_id)
...
+      @recent_boundary_periods ||=
+        TrainingPeriod
+          .joins(school_partnership: { lead_provider_delivery_partnership: :active_lead_provider })
+          .where(active_lead_providers: { lead_provider_id: })
+          .where(api_transfer_updated_at: updated_since..)
+          .where(id: boundary_training_period_ids)
+    end
+
+    def ect_ids
```
ethax-ross, line 112: Worth rolling the `ect_ids` and `mentor_ids` into a `teacher_ids_updated_since` or similar?

```
rect_boundary_periods
  .joins(:ect_at_school_period, :mentor_at_school_period)
  .pluck("mentor_at_school_periods.teacher_id", "ect_at_school_periods.teacher_id")
   .flatten
```

app/services/api/teachers/school_transfers/query.rb
```diff
@@ -68,36 +75,63 @@ def where_lead_provider_is(lead_provider_id)
...
-    def set_sort_by(sort)
-      @scope = scope.order(sort)
+    def scoped_mentor_at_school_periods
+      MentorAtSchoolPeriod
+        .where(teacher_id: participant_ids)
+        .includes(:earliest_training_period, :latest_training_period)
+    end
+
+    def boundary_training_period_ids
+      [*scoped_ect_at_school_periods, *scoped_mentor_at_school_periods].flat_map { |at_school_period|
```
ethax-ross: Rather than loading the whole objects into memory we could pluck here

```
ect_ids = ECTAtSchoolPeriod
  .where(teacher_id: participant_scope)
  .pluck(:earliest_training_period_id, :latest_training_period_id)

mentor_ids = MentorAtSchoolPeriod
  .where(teacher_id: participant_scope)
  .pluck(:earliest_training_period_id, :latest_training_period_id)

(ect_ids + mentor_ids).flatten.compact.uniq
```

ethax-ross (commented):
Much more readable 👏 just a few suggestions mainly around method names

app/services/api/teachers/school_transfers/query.rb
peterdavidhamilton: You know I had that originally and changed it...

ethax-ross: It's a confusing bit of baggage from ECF 😞 generally speaking we try and keep ECF terms (so `participant`, `cohort`, `course`, etc) from going any further down into the architecture than the controller layer

app/services/api/teachers/school_transfers/query.rb
peterdavidhamilton: I reasoned that because they come in two flavours, it might be prudent to keep them separate until the last, in case we want to query them individually. So I intentionally wrote it with the ect and mentor sides split apart.

ethax-ross: I figured rolling them into a single query saves a trip to the DB, but its not going to be a significant saving so I don't feel strongly about it 👍

PR #2707, opened 2026-04-20
Fix change name event name

ericaporter:
### Context
This [ticket](https://github.com/DFE-Digital/register-ects-project-board/issues/2742) 
When a school user manually corrects an ECT or mentor's name, the event recorded was teacher_name_updated_by_trs, implying the change originated from TRS. This was misleading and possibly added in error

- Adds `teacher_name_updated_by_user` to `Event::EVENT_TYPES`
- Adds `Events::Record.teacher_name_updated_by_user_event!` method to sit alongside the existing `teacher_name_changed_in_trs_event!`
- Updates both `Schools::ECTs::ChangeNameWizard::CheckAnswersStep` and `Schools::Mentors::ChangeNameWizard::CheckAnswersStep` to call the new method
- Tightens the existing specs to assert the correct event type, rather than just that a RecordEventJob was enqueued

### Guidance to review
To verify the fix, trigger a name change as a school user for both an ECT and a mentor, then check the event log to confirm the event type shows `teacher_name_updated_by_user` rather than `teacher_name_updated_by_trs`

### DIFF
--- a/app/services/events/record.rb
+++ b/app/services/events/record.rb
@@ -180,6 +180,12 @@ module Events
       new(event_type:, author:, appropriate_body_period:, teacher:, heading:, happened_at:).record_event!
     end
 
+    def self.teacher_name_updated_by_user_event!(old_name:, new_name:, author:, teacher:, happened_at: Time.zone.now)
+      event_type = :teacher_name_updated_by_user
+      heading = TransitionDescription.for("name", from: old_name, to: new_name)
+      new(event_type:, author:, appropriate_body_period: nil, teacher:, heading:, happened_at:).record_event!
+    end
+
     def self.teacher_induction_status_changed_in_trs_event!(old_induction_status:, new_induction_status:, author:, teacher:, appropriate_body_period: nil, happened_at: Time.zone.now)
       event_type = :teacher_trs_induction_status_updated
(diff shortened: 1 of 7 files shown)


app/services/events/record.rb
```diff
@@ -180,6 +180,12 @@ def self.teacher_name_changed_in_trs_event!(old_name:, new_name:, author:, teach
       new(event_type:, author:, appropriate_body_period:, teacher:, heading:, happened_at:).record_event!
     end
 
+    def self.teacher_name_updated_by_user_event!(old_name:, new_name:, author:, teacher:, happened_at: Time.zone.now)
+      event_type = :teacher_name_updated_by_user
+      heading = TransitionDescription.for("name", from: old_name, to: new_name)
+      new(event_type:, author:, appropriate_body_period: nil, teacher:, heading:, happened_at:).record_event!
+    end
```
cpjmcquillan, line 187: Is it worth adding a test here too? 🤔

The tests feel a little tautological, but I think we have them for the other event types

cpjmcquillan (approved):

avinhurry (approved):
Looks good 👍🏽

peteryates (approved):

app/services/events/record.rb
ericaporter: Yeah probably a good idea especially since it's a bug fix!

PR #2708, opened 2026-04-20
Fix: Exclude cross-type transferred teachers in seed scenarios

mooktakim:
`select_random_teacher` only excluded teachers already transferred for the same type (`@transferred_teachers[type]`). A teacher picked as an ECT in one scenario could be re-picked as a mentor in a later scenario.

`TrainingPeriod#siblings` treats all training periods for a teacher as siblings regardless of ECT vs mentor at_school_period, so reusing a teacher across types produced overlapping training periods and failed the `Trainee` overlap validation.

Exclude all already-transferred teachers across both types before sampling.

### DIFF
--- a/app/services/api_seed_data/teachers/school_transfers.rb
+++ b/app/services/api_seed_data/teachers/school_transfers.rb
@@ -11,7 +11,7 @@ module APISeedData
 
       log_plant_info("teacher school transfers")
 
-      @transferred_teachers = { ect: [], mentor: [] }
+      @transferred_teachers = []
 
       lead_providers.each do |from_lead_provider|
         log_seed_info("Processing Lead provider: #{from_lead_provider.name}")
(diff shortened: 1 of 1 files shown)


app/services/api_seed_data/teachers/school_transfers.rb
```diff
@@ -37,7 +37,8 @@ def select_different_lead_provider(excluding_lead_providers: [])
     end
 
     def select_random_teacher(excluding_teachers: [], type: :ect)
-      teacher = (available_teachers - excluding_teachers).sample.presence ||
+      already_transferred = @transferred_teachers.values.flatten
+      teacher = (available_teachers - already_transferred - excluding_teachers).sample.presence ||
```
ethax-ross: Can we not drop the `excluding_teachers` here as it will always be a subset of `already_transferred`?

ethax-ross (approved):
Makes sense, I think we could maybe clean it up a bit though - probably don't need to track excluded teachers by type at all if we want to exclude all of them anyway?

PR #2713, opened 2026-04-20
[3746] Stop erasing schedule changes on defer/withdraw

mooktakim:
### Context

Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/3746

- `Teachers::Defer` and `Teachers::Withdraw`: removed `rollback_redundant_training_period!` and the `previous_training_period_for_lead_provider` helper. Services no longer mutate training periods outside the one they were given.
- `API::Teachers::Defer` and `API::Teachers::Withdraw`: `training_for_at_least_one_day` now always errors when the training period started today (previously bypassed when predecessors existed for the LP). New error message per ticket:
  > You cannot defer or withdraw this participant today. You need to try again tomorrow as the training was recently changed for this participant.
- Service specs: removed the "destroys the redundant training period" contexts.
- API validator specs: flipped the "earlier training period exists" context to expect the new validation error.

### DIFF
--- a/app/services/api/teachers/defer.rb
+++ b/app/services/api/teachers/defer.rb
@@ -55,9 +55,8 @@ module API::Teachers
     def training_for_at_least_one_day
       return if errors[:teacher_api_id].any?
       return unless training_period&.started_on&.today?
-      return if training_period.predecessors.any? { it.lead_provider.id == lead_provider_id }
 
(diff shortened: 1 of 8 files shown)


ethax-ross (approved):
Looks good thanks 👍

PR #2719, opened 2026-04-21
[3717] Fix N+1 queries on api POST/PUT action responses

leandroalemao:
### Context

Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/3717

We have a number of N+1 queries in the serializers that were noticed for participant action endpoints

- Check/fix N+1 queries in POST partnerships
- Check/fix N+1 queries in PUT partnerships
- Check/fix N+1 queries in POST declarations
- Check/fix N+1 queries in PUT declarations/void
- Check/fix N+1 queries in PUT participant actions (resume, withdraw, defer, change schedule)

### Guidance to review

```
all request specs runs with no N+1 exceptions:

rspec ./spec/requests/api/v3/
```

- Remove [Prosopite.scan to API controller](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/2719/changes/19d4168f5cc8871e516b3ecc304b7e8856f055dc) before merging.

### DIFF
--- a/app/controllers/api_controller.rb
+++ b/app/controllers/api_controller.rb
@@ -9,6 +9,17 @@ class APIController < ActionController::API
   include API::ConditionExtractable
   include DfE::Analytics::Requests
 
+  unless Rails.env.production?
+    around_action :n_plus_one_detection
+
+    def n_plus_one_detection
+      Prosopite.scan
+      yield
+    ensure
+      Prosopite.finish
+    end
+  end
+
 private
 
   # `current_user` needed for DfE::Analytics
--- a/app/controllers/api_controller.rb
+++ b/app/controllers/api_controller.rb
@@ -9,17 +9,6 @@ class APIController < ActionController::API
   include API::ConditionExtractable
   include DfE::Analytics::Requests
 
-  unless Rails.env.production?
-    around_action :n_plus_one_detection
-
-    def n_plus_one_detection
-      Prosopite.scan
-      yield
-    ensure
-      Prosopite.finish
-    end
-  end
-
 private
 
   # `current_user` needed for DfE::Analytics
--- a/app/serializers/api/teacher_serializer.rb
+++ b/app/serializers/api/teacher_serializer.rb
@@ -121,8 +121,12 @@ class API::TeacherSerializer < Blueprinter::Base
     field(:api_updated_at, name: :updated_at)
 
     association :ecf_enrolments, blueprint: TrainingPeriodSerializer do |teacher, options|
-      metadata = ::API::TeacherSerializer.lead_provider_metadata(teacher:, options:)
-      [[metadata.latest_ect_training_period, teacher, metadata], [metadata.latest_mentor_training_period, teacher, metadata]].reject { |period, _| period.nil? }
+      metadata ||= {}
+      metadata[teacher.id] ||= ::API::TeacherSerializer.lead_provider_metadata(teacher:, options:)
+
+      [[metadata[teacher.id].latest_ect_training_period, teacher, metadata[teacher.id]],
+       [metadata[teacher.id].latest_mentor_training_period, teacher, metadata[teacher.id]]]
+        .reject { |period, _| period.nil? }
     end
 
     association :teacher_id_changes, blueprint: TeacherIdChangeSerializer, name: :participant_id_changes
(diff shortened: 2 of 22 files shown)


app/controllers/api_controller.rb
```diff
@@ -9,6 +9,17 @@ class APIController < ActionController::API
   include API::ConditionExtractable
   include DfE::Analytics::Requests
 
+  unless Rails.env.production?
```
ethax-ross: Since the gem is only in the test group would it be safer to run `scan/finish` in a request spec around the call to each endpoint via a shared context?

app/serializers/api/teacher_serializer.rb
```diff
@@ -121,8 +121,12 @@ def uplift_value(training_period, metadata, uplift_type)
     field(:api_updated_at, name: :updated_at)
 
     association :ecf_enrolments, blueprint: TrainingPeriodSerializer do |teacher, options|
-      metadata = ::API::TeacherSerializer.lead_provider_metadata(teacher:, options:)
-      [[metadata.latest_ect_training_period, teacher, metadata], [metadata.latest_mentor_training_period, teacher, metadata]].reject { |period, _| period.nil? }
+      metadata ||= {}
```
ethax-ross: Why the hash here? 🤔

ethax-ross (commented):
Hey Leandro, it looks like we're duplicating a lot of the preloading that goes on in the query services already, are we not able to utilise them instead?

I thought if we just re-query the object via the relevant query service before returning it from the API service it would then be preloaded ready for the serializer?

app/serializers/api/teacher_serializer.rb
leandroalemao: Hi @ethax-ross I'm memoising the metadata as the teacher is always the same we don't need to query metatada for every training period that will be displayed in `ecf_enrolments` field in the serializer.

ethax-ross: Not sure I'm following; `ecf_enrolments` is going to be called once for each teacher that is serialized? it's also not going to span `ecf_enrolments` as `metadata` is just a scoped instance variable? unless I'm missing something? 🤔

app/controllers/api_controller.rb
leandroalemao: Hi @ethax-ross the problem of doing that is that we'd be scanning for N+1 queries in the request specs factories creation as well.. that's the reason why adding to the controller instead. You can see here [this failed spec](https://github.com/DFE-Digital/register-early-career-teachers-public/actions/runs/24748993355/job/72407584697?pr=2719).. it's generating a N+1 query when creating factories for a payment statement, that creates a contract, that creates a band fee structures.

ethax-ross: We just need to call the factories before making the request to ensure any setup is already done, so something like:

```
  it "does not introduce N+1 queries" do
    # Preload the path to ensure N+1 queries are only detected during the API
    # request and not during the path resolution (which calls factories with N+1s).
    resolved_path = path

    expect {
      Prosopite.scan do
        send(request_method, resolved_path, headers: api_headers, params: parameters)
      end
    }.not_to raise_error
  end
```

leandroalemao: Hi @ethax-ross yeah I had added a workaround too

```
# Ensure that all factories data is created before the request is made
  # and Prosopite starts tracking queries, to avoid N+1 queries being raised
  # due to lazy loading of associations in the factories.
  before do
    resource if defined?(resource)
    resources if defined?(resources)
    service_args if defined?(service_args)
    serializer_options if defined?(serializer_options)
  end
```

but yours looks simpler.. will make the changes tks

leandroalemao: All done @ethax-ross tks

config/environments/test.rb
```diff
@@ -8,6 +8,15 @@
 Rails.application.configure do
   # Settings specified here will take precedence over those in config/application.rb.
 
+  config.after_initialize do
+    Prosopite.rails_logger = true
+    Prosopite.raise = true
+
+    # Allow N+1 from raised from Metadata refreshes
+    # as these are expected and not a problem in tests
+    Prosopite.allow_stack_paths = [/Metadata::Handlers/, /DeclarativeUpdates/]
```
ethax-ross: I think we only need `/DeclarativeUpdates/` in here, I don't really see where the N+1 is coming from still, but metadata should all be optimised/free of N+1s

ethax-ross (approved):

PR #2720, opened 2026-04-21
Fix failing specs on main

avinhurry:
### Summary

Replaces hard coded dates in the claim an ECT register spec with relative past dates so the examples stay valid regardless of the day the suite runs.


mooktakim (approved):

peteryates (approved):
Thanks @avinhurry

PR #2723, opened 2026-04-21
Fix `api_updated_at` migration from ECF

ethax-ross:
The previous query had an issue; if a `SchoolCohort` doesn't exist the `api_updated_at` would be set to the default/migration date.

Query on `School` rather than `SchoolCohort` to ensure all schools are covered and fill in dates for cohorts when there is no `SchoolCohort`.

Tested on parity check env.


leandroalemao (approved):

leandroalemao (approved):
Hi @ethax-ross any specs for this scenario? or is it already covered? tks

PR #2735, opened 2026-04-22
Refactor TeacherStatus

craigmdavidson:
### Context

This PR is to address issue: https://github.com/DFE-Digital/register-ects-project-board/issues/3752 

> We added an InductionStatus service for this a while ago:
>
> [https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/services/api/teachers/induction_status.rb#L10](https://github.com.mcas.ms/DFE-Digital/register-early-career-teachers-public/blob/main/app/services/api/teachers/induction_status.rb?McasCtx=4&McasTsid=11760#L10)
>
> But I noticed we still only look at the InductionPeriod when determining TeacherStatus:
>
> [https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/services/api/training_periods/teacher_status.rb#L18](https://github.com.mcas.ms/DFE-Digital/register-early-career-teachers-public/blob/main/app/services/api/training_periods/teacher_status.rb?McasCtx=4&McasTsid=11760#L18)

Originally the plan was to improve how we evaluate whether an induction is complete by using InductionStatus. But removing the logical anomaly hidden in the original structure makes that change redundant — it turns out the status is determined entirely by the training period's start and end dates.

If a teacher isn't joining, the only thing that matters is whether they have a finish date. If they do, they're either leaving (finish date in the future) or left (finish date in the past). If they don't, they're active. That's it. There's no situation where `mentor_became_ineligible_for_funding_on` or `finished_induction_period` could change that outcome — so any logic referencing them is redundant.

The only callers of `active?`, `joining?`, `leaving?`, and `left?` were the specs — and those specs were testing internal implementation rather than desired behaviour. When the methods are initially made private, then removed the whole thing collapses down to a few simple date checks.

### Guidance to review

* ~Review on a commit by commit basis.~ Only 1 commit ;-P
* The initial commit leaves the specs intact to demonstrate the behaviour preserving change. 
* I was originally going to then simplify the specs, but I think I will leave these in place to provide confidence initially. These can then be superseded come API v4.

### DIFF
--- a/app/serializers/api/teacher_serializer.rb
+++ b/app/serializers/api/teacher_serializer.rb
@@ -37,7 +37,7 @@ class API::TeacherSerializer < Blueprinter::Base
         end
       end
       field(:training_status) { |(training_period, _, _)| API::TrainingPeriods::TrainingStatus.new(training_period:).status }
-      field(:participant_status) { |(training_period, teacher, _)| API::TrainingPeriods::TeacherStatus.new(latest_training_period: training_period, teacher:).status }
+      field(:participant_status) { |(training_period, _, _)| API::TrainingPeriods::TeacherStatus.new(latest_training_period: training_period).status }
       field(:eligible_for_funding) do |(training_period, teacher, _)|
         teacher_type = training_period.for_ect? ? :ect : :mentor
         API::Teachers::EligibilityForFunding.new(teacher:, teacher_type:).eligible?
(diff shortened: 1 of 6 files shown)


peterdavidhamilton:
Good spot @craigmdavidson. With only a single public method it is clear what its purpose is.

peterdavidhamilton (approved):

ethax-ross (approved):
Looks great! can we make sure this goes through PO review before merging please

cpjmcquillan (approved):

PR #2739, opened 2026-04-22
Add validation on change schedule service so we return an error message when the schedule identifier is invalid

leandroalemao:
### Context

Sending an invalid schedule identifier in the params to change-schedule endpoint is raising an exception.

Screenshot 2026-04-22 at 12 01 32

- Add validation on change schedule service so we return an error message when the schedule identifier is invalid;

Screenshot 2026-04-22 at 12 02 46

### Guidance to review

Review app

### DIFF
--- a/spec/requests/api/v3/participants_spec.rb
+++ b/spec/requests/api/v3/participants_spec.rb
@@ -100,7 +100,21 @@ RSpec.describe "Participants API", :with_touches, type: :request do
     end
 
     it_behaves_like "a token authenticated endpoint", :put
-    it_behaves_like "an API update endpoint"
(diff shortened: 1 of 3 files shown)


mooktakim (approved):

spec/requests/api/v3/participants_spec.rb
```diff
@@ -100,7 +100,21 @@ def apply_expected_order(resources)
     end
 
     it_behaves_like "a token authenticated endpoint", :put
-    it_behaves_like "an API update endpoint"
```
ethax-ross, line 103: Can we cover this in the model rather than the request specs?

ethax-ross (approved):

PR #2740, opened 2026-04-22
Add missing full stops

avinhurry:
### Summary

Add missing full stops on support pages


StephenC-V1 (commented):
Looks good to me. Thanks Avin.

peteryates (approved):
Nice to see that apostrophes was too narrow a remit so you've progressed to other forms of punctuation.

PR #2741, opened 2026-04-22
[3824] Require mentor registration for mentor eligibility

mooktakim:
### Context

Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/3824

* Set `teacher.mentor_first_became_eligible_for_training_at` only if mentor has `mentor_at_school_periods.any?`

### Guidance to review

Script to update existing teachers:

```ruby
class ClearIncorrectMentorEligibility
  def initialize(dry_run: true)
    @dry_run = dry_run
  end

  def call
    scope = Teacher.where.not(mentor_first_became_eligible_for_training_at: nil)
                   .where.missing(:mentor_training_periods)

    teacher_count = scope.count
    event_scope   = Event.where(event_type: "teacher_funding_eligibility_set",
                                teacher_id: scope.select(:id))
    events_count  = event_scope.count

    puts "Teachers to clear:              #{teacher_count}"
    puts "Eligibility events to delete:   #{events_count}"

    if @dry_run
      puts "DRY RUN — nothing changed. Call with dry_run: false to apply."
      return
    end

      deleted_events = event_scope.delete_all
      # update_all bypasses the immutable_once_set validator — corrective data fix.
      updated_teachers = scope.in_batches.update_all(mentor_first_became_eligible_for_training_at: nil)
      puts "APPLIED — events deleted=#{deleted_events} teachers updated=#{updated_teachers}"
    end
  end
end
```


app/services/teachers/set_ect_funding_eligibility.rb
ethax-ross, line 42: Not related to this change, but can we pass if this was the ECT or mentor eligibility down into the event? I think that'll be useful information to have

spec/services/teachers/set_ect_funding_eligibility_spec.rb
```diff
@@ -0,0 +1,164 @@
...
+        ect_at_school_period = FactoryBot.create(:ect_at_school_period,
+                                                 teacher:,
+                                                 started_on: 2.months.ago,
+                                                 finished_on: nil)
+        training_period = FactoryBot.create(:training_period,
+                                            :for_ect,
+                                            :ongoing,
+                                            ect_at_school_period:)
+        FactoryBot.create(:declaration, training_period:)
+        FactoryBot.create(:statement, :open, :output_fee,
```
ethax-ross: If we don't `and_call_original` on the mark declaration service I think we could drop this setup?

spec/services/teachers/set_ect_funding_eligibility_spec.rb
```diff
@@ -0,0 +1,164 @@
...
+          teacher.update!(ect_first_became_eligible_for_training_at: 1.day.ago)
+        end
+
+        it "does not change `ect_first_became_eligible_for_training_at`" do
+          expect { service.set! }.not_to change(teacher, :ect_first_became_eligible_for_training_at)
+        end
+      end
+
+      it "calls `Declarations::Actions::MarkDeclarationsEligible` service" do
+        expect(Declarations::Actions::MarkDeclarationsEligible).to receive(:new).with(declarations: teacher.ect_declarations, author:).and_call_original
```
ethax-ross: Can we add an existing, eligible declaration to assert that it doesn't get passed in (and only the `no_payment` declarations do)?

spec/services/teachers/set_ect_funding_eligibility_spec.rb
```diff
@@ -0,0 +1,164 @@
...
+      before do
+        FactoryBot.create(:induction_period, :ongoing, teacher:)
+      end
+
+      it "does not set `ect_first_became_eligible_for_training_at`" do
+        expect { service.set! }.not_to change(teacher, :ect_first_became_eligible_for_training_at)
+      end
+    end
+
+    context "when teacher is not eligible for ECT training" do
```
ethax-ross: `when the teacher has no induction periods or ECT at school periods`?

app/services/teachers/set_ect_funding_eligibility.rb
ethax-ross, line 17:
```
return unless eligible_for_ect_training?
return if teacher.ect_first_became_eligible_for_training_at

teacher.touch(:ect_first_became_eligible_for_training_at)
```

app/services/teachers/set_ect_funding_eligibility.rb
```diff
@@ -45,10 +40,6 @@ def eligible_for_ect_training?
     teacher.induction_periods.ongoing_today.any? && teacher.ect_at_school_periods.any?
   end
 
-  def eligible_for_mentor_training?
-    teacher.mentor_became_ineligible_for_funding_on.blank? && teacher.mentor_became_ineligible_for_funding_reason.blank?
-  end
-
   def record_teacher_set_funding_eligibility_event!
     return unless teacher.saved_changes?
```
ethax-ross: Is it worth calling `record_teacher_set_funding_eligibility_event!` and `make_declarations_eligible!` from set_eligibility! to avoid this and the guard on line 28?

app/services/teachers/set_mentor_funding_eligibility.rb
```diff
@@ -0,0 +1,55 @@
+class Teachers::SetMentorFundingEligibility
+  attr_reader :teacher, :author
+
+  def initialize(teacher:, author:)
```
ethax-ross, line 4: Just the same thoughts as in the ECT eligibility service

spec/services/teachers/set_mentor_funding_eligibility_spec.rb
```diff
@@ -0,0 +1,155 @@
+RSpec.describe Teachers::SetMentorFundingEligibility do
+  let(:teacher) { FactoryBot.create(:teacher, api_ect_training_record_id: SecureRandom.uuid) }
+  let(:author) { Events::SystemAuthor.new }
+  let(:service) { described_class.new(teacher:, author:) }
+
+  describe "#set!" do
```
ethax-ross, line 6: Ditto for the ECT service comments

spec/features/appropriate_bodies/process_batch_combo_spec.rb
```diff
@@ -53,11 +53,9 @@
       "Induction status set to 'InProgress'",
       "TRS attributes updated",
       /was claimed by The Appropriate Body/,
-      /funding eligibility was set/,
```
ethax-ross, line 56: Not sure I follow the changes in this class? 🤔 would we not still get _some_ eligibility events?

ethax-ross (commented):
Thanks for doing this so quickly Mook! I've added some minor thoughts/suggestions

spec/features/appropriate_bodies/process_batch_combo_spec.rb
mooktakim: Before this change we would always run `record_teacher_set_funding_eligibility_event`, which would create event when `teacher.saved_changes?` is true, even though the eligibility was not set because `teacher.ect_at_school_periods.any?` was false (as the test doesn't create ect_at_school_periods).

Now we only create event if the eligibility is set.

ethax-ross: Ah nice, fixing a bug as well then! 🙂

ethax-ross (approved):
Looks great ✅ I think we may want to look at DRYing up these services at some point as they're so similar, but lets get this moved along as it's critical for going live. Thanks for putting this together on short notice 🙏

leandroalemao (approved):

peteryates (approved):
Nice - thanks for remembering the events!

PR #2743, opened 2026-04-22
[3827] Migrate remaining cohort flags to contract periods

peterdavidhamilton:
### Context

ECF1 has cohort attributes we are not yet migrating

**Migrate:**

- `detailed_evidence_types` => `detailed_evidence_types_enabled` ([required an attribute correction](https://github.com/DFE-Digital/register-early-career-teachers-public/commit/0ceb5f02cde8d6255c9944a099a6872f94f61cc3#diff-d0f6b1bce616292a512815fc441a4a034a48a793751fdd84ac8f16e81d6d099f)) 
- `payments_frozen_at`
- `mentor_funding` => `mentor_funding_enabled`

**Add:**

`uplift_fees_enabled`

### Guidance to review

A 2024 cohort will receive uplift but a 2025 will not.

### DIFF
--- a/app/migration/migrators/contract_period.rb
+++ b/app/migration/migrators/contract_period.rb
@@ -28,11 +28,17 @@ module Migrators
       contract_period = ::ContractPeriod.find_or_initialize_by(year: cohort.start_year)
 
       contract_period.update!(
+        # dates
         started_on: cohort.registration_start_date,
         finished_on: finished_on_for(cohort),
-        enabled: contract_period_enabled?(cohort),
+        payments_frozen_at: cohort.payments_frozen_at,
         created_at: cohort.created_at,
-        updated_at: cohort.updated_at
+        updated_at: cohort.updated_at,
+        # flags
+        detailed_evidence_types_enabled: cohort.detailed_evidence_types_enabled,
+        mentor_funding_enabled: cohort.mentor_funding,
+        uplift_fees_enabled: uplift_fees_enabled?(cohort),
+        enabled: contract_period_enabled?(cohort)
       )
 
       contract_period
@@ -43,6 +49,10 @@ module Migrators
     def contract_period_enabled?(cohort)
       cohort.start_year.to_s != "2020" &&
         !(cohort.payments_frozen_at.present? && Time.current >= cohort.payments_frozen_at)
+    end
+
+    def uplift_fees_enabled?(cohort)
+      cohort.start_year <= 2024
     end
 
     def finished_on_for(cohort)
--- a/app/migration/migrators/contract_period.rb
+++ b/app/migration/migrators/contract_period.rb
@@ -35,7 +35,7 @@ module Migrators
         created_at: cohort.created_at,
         updated_at: cohort.updated_at,
         # flags
-        detailed_evidence_types_enabled: cohort.detailed_evidence_types_enabled,
+        detailed_evidence_types_enabled: cohort.detailed_evidence_types,
         mentor_funding_enabled: cohort.mentor_funding,
         uplift_fees_enabled: uplift_fees_enabled?(cohort),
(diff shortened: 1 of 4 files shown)


app/migration/migrators/contract_period.rb
```diff
@@ -28,11 +28,17 @@ def migrate_one!(cohort)
...
+        # dates
         started_on: cohort.registration_start_date,
         finished_on: finished_on_for(cohort),
-        enabled: contract_period_enabled?(cohort),
+        payments_frozen_at: cohort.payments_frozen_at,
         created_at: cohort.created_at,
-        updated_at: cohort.updated_at
+        updated_at: cohort.updated_at,
+        # flags
+        detailed_evidence_types_enabled: cohort.detailed_evidence_types_enabled,
```
ethax-ross: This should be `detailed_evidence_types` in ECF - it looks like the ECF schema (in RECT) was mistakenly updated here 😞 

https://github.com/DFE-Digital/register-early-career-teachers-public/commit/0ceb5f02cde8d6255c9944a099a6872f94f61cc3#diff-b5529da91d157ce12f09ba0f2fe6e6c7524aa4d2142a00fc351c9502a2da53e2L265-R265

Here's the correct reference:

https://github.com/DFE-Digital/early-careers-framework/blob/main/db/schema.rb#L264

spec/migration/migrators/contract_period_spec.rb
```diff
@@ -14,15 +14,42 @@ def setup_failure_state
...
+                                                     "created_at",
+                                                     "updated_at",
+                                                     "payments_frozen_at",
+                                                     "detailed_evidence_types_enabled"
+                                                   ))
         expect(contract_period.id).to eq(migration_resource1.start_year)
         expect(contract_period.started_on.to_date).to eq(migration_resource1.registration_start_date.to_date)
         expect(contract_period.finished_on).to eq(contract_period.started_on.next_year.prev_day)
         expect(contract_period.enabled).to be(true)
+        expect(contract_period.mentor_funding_enabled).to eq(migration_resource1.mentor_funding)
```
ethax-ross, line 33: Can we add a test for `detailed_evidence_types_enabled` as well? I think if we add `detailed_evidence_types` to the ECF cohort factory that would have caused it to blow up due to the typo in the schema 🤔

ethax-ross (commented):
Looks good, just spotted a mistake in our ECF schema from an earlier change that we'll need to fix

app/migration/migrators/contract_period.rb
peterdavidhamilton: Will do

spec/migration/migrators/contract_period_spec.rb
```diff
@@ -14,15 +14,42 @@ def setup_failure_state
...
+      end
+
+      describe "uplift fees" do
+        def create_migration_resource
+          FactoryBot.create(:migration_cohort, start_year:)
+        end
+
+        context "when start year is 2024 or earlier" do
+          let(:start_year) { (2021..2024).to_a.sample }
```
craigmdavidson: I'm not familiar in this space but is it worth adding a start year of 2020 since for some reason that is specifically excluded from `contract_period_enabled?` - this means we can have a uplift fees enabled but the contract period not enabled. 

Is this okay?

craigmdavidson (approved):
LGTM but WDIK

app/migration/migrators/contract_period.rb
peterdavidhamilton: Done.

The thing that confused me was I started seeing variance in timestamps after I touched the schema and couldn't see how that was related.

Screenshot 2026-04-22 at 17 31 30

ethax-ross: That's strange 🤔 and it just resolved itself?

PR #2747, opened 2026-04-22
Always show api/guidance and swagger docs

craigmdavidson:
### Context

Resolves https://github.com/DFE-Digital/register-ects-project-board/issues/3690

> At the minute there is a single enable_api flag that will switch on all the API related endpoints:
>
> https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/config/routes/api.rb#L1
>
> We will need to add a new enable_api_guidance flag and pull out the relevant route(s) so that we can enable them separately.
>
> As it stands I think the API guidance links to the API swagger docs (and also includes release notes), so we probably want to include them all but confirm with Claire/Andy:
>
> https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/config/routes/api.rb#L3C5-L6C71

Since we always want to show the guidance, we do not need the `enable_api_guidance` flag.
So this PR just removes the `enable_api` restrictions on the guidance and swagger doc routes. 

Additionally API Guidance UI updates to de-prioritise Sandbox:
1. Remove Sandbox "card"
2. Remove Sandbox from API Guidance Navbar
3. Move 'Using the API Sandbox' guide to guidance-for-lead-providers (adds an entry in the sidebar).

### Guidance to review

Review in commit order.


app/components/navigation/primary_navigation_component.rb
```diff
@@ -69,7 +69,6 @@ def items_by_area
           { text: "Home", href: "/api/guidance" },
           { text: "Swagger API documentation", href: "/api/guidance/swagger-api-documentation" },
           { text: "Release notes", href: "/api/guidance/release-notes", active_when: "/api/guidance/release-notes" },
-          { text: "Sandbox", href: "/api/guidance/sandbox" },
```
ethax-ross, line 72: I wonder if we should add in a redirect in case LPs have bookmarked this page 🤔 one for @claire-hughez

ethax-ross (approved):
Looks great thanks Craig 👍  can you run it past @claire-hughez and then merge it into the bundled PR we have ready to merge later on please 🙏

app/components/navigation/primary_navigation_component.rb
craigmdavidson: > I wonder if we should add in a redirect in case LPs have bookmarked this page 🤔 one for @claire-hughez

@claire-hughez - FWIW: this is minimal additional work if you want to do this.

claire-hughez:
This looks great to me, thank you @craigmdavidson. It's highlighted a minor content change we need to fix in the Sandbox guidance page but I'll make a note to get back to it after this has merged. We were meant to fix it in another ticket but looks like with our old content designer leaving it got missed somehow.

app/components/navigation/primary_navigation_component.rb
claire-hughez: We originally had Sandbox and Production separate because we thought they may need to vary. I don't think they do strongly tbh, but they are separate environments and we may want to reconsider that so best to leave without the redirect I think for now. I'll just start sharing the production URL more with LPs I think.

PR #2748, opened 2026-04-22
[3695] Return only one prioritised error message from the API when multiple error messages

leandroalemao:
### Context

Ticket: [3695](https://github.com.mcas.ms/DFE-Digital/register-ects-project-board/issues/3695)

When a lead provider submits a POST or PUT declaration, if there are multiple errors, we show all of them.

- Most endpoints only show one in RECT, ECF1 also only shows one, and RECT should do the same.
- Checked the precedence/order of the validations/error messages, and all seems fine to me.

### Guidance to review

Review app

### DIFF
--- a/spec/support/matchers/have_one_error_only_matcher.rb
+++ b/spec/support/matchers/have_one_error_only_matcher.rb
@@ -1,0 +1,17 @@
+RSpec::Matchers.define :have_one_error_only do
+  match do |actual|
+    actual.invalid? && actual.errors.count == 1
+  end
+
+  failure_message do |_actual|
+    "expected #{actual.class} to have 1 error, but it had #{actual.errors.count} with messages: #{actual.errors.full_messages.join("\n")}"
+  end
+
+  failure_message_when_negated do |_actual|
+    "expected exactly one error message in total, but found more"
+  end
+
+  description do
+    "have exactly one error message in total across all attributes"
+  end
+end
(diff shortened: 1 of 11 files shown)


peterdavidhamilton (approved):

ethax-ross (commented):
I like the idea of this change, but as I put in the ticket 😅 :

> It may not be something we want to change so drastically this close to launch

If we're bringing in this type of change I would want to be very sure we have the validation rules in a sensible order. 

Can we put this one on the back burner for post-launch and raise a separate PR to address the specific issue in the ticket around the multiple error messages when submitting a declaration? and do a sweep of the services in general to see if we can identify others where we may be returning multiple errors?

When we come to do this change I think we should also look to unwind the logic we already have in the services that restricts the errors being returned -- the services could return all errors, in priority order, to later be truncated to the first error only using this approach.

mooktakim (approved):

mooktakim (approved):

ethax-ross (commented):
Thanks Leandro, are we able to capture this change with a test?

ethax-ross (approved):

spec/support/matchers/have_one_error_only_matcher.rb
```diff
@@ -0,0 +1,17 @@
+RSpec::Matchers.define :have_one_error_only do
```
peterdavidhamilton, line 1: 👍🏼

PR #2752, opened 2026-04-23
Launch

cpjmcquillan:
Launch the service!


app/views/admin/appropriate_bodies/current_ects/index.html.erb
```diff
@@ -33,7 +33,7 @@
     <% @teachers.each do |teacher| %>
       <%=
         govuk_summary_card(title: teacher_full_name(teacher)) do |card|
-          card.with_action { govuk_link_to("Show", admin_teacher_induction_path(teacher)) }
+          card.with_action { govuk_link_to("Show", admin_teacher_path(teacher)) }
```
cpjmcquillan, line 36: Do we still want to link to inductions here, given this is the AB list of current ECTs? 🤔

cc @mdoyle99 @Aaminap

thenapking: That sounds right, I don't think we should be linking to inductions

config/terraform/application/config/production.yml
```diff
@@ -9,6 +9,7 @@ ENABLE_BLAZER: true
 ENABLE_API: false
 ENABLE_MIGRATION_TESTING: false
 ENABLE_MIGRATION_INTERFACE: false
+ENABLE_SCHOOLS_INTERFACE: true
```
cpjmcquillan, line 12: This is the main thing ⏯️

config/terraform/application/config/production.yml
```diff
@@ -6,11 +6,12 @@ ENABLE_PERSONAS: false
 ENABLE_FAKE_TRS_API: false
 ENABLE_DIRECT_SCHOOL_DFE_SIGN_IN: true
 ENABLE_BLAZER: true
-ENABLE_API: false
+ENABLE_API: true
```
cpjmcquillan, line 9: This is another main thing!

config/terraform/application/config/production.yml
```diff
@@ -6,11 +6,12 @@ ENABLE_PERSONAS: false
...
 ENABLE_MIGRATION_TESTING: false
 ENABLE_MIGRATION_INTERFACE: false
+ENABLE_SCHOOLS_INTERFACE: true
 ENABLE_INCIDENT_BANNER: false
-ENABLE_MAINTENANCE_BANNER: true
+ENABLE_MAINTENANCE_BANNER: false
```
cpjmcquillan, line 14: This felt sensible too

config/terraform/application/config/production.yml
peteryates: Well caught 😂

app/controllers/admin/teachers/timeline_controller.rb
```diff
@@ -1,6 +1,25 @@
...
-  def show
-    @teacher = Teacher.find(params[:teacher_id])
-    @events = Events::List.new.for_teacher(@teacher)
+module Admin
+  module Teachers
+    class TimelineController < AdminController
+      layout "full"
+
+      def show
+        teacher = Teacher.find_by(id: params[:teacher_id])
```
avinhurry: Minor point but probably worth keeping this as `find`?

If the teacher id is missing `find` gives us a `404` whereas `find_by` can return `nil` and lead to a `500`.

cpjmcquillan: Yes definitely, good spot. Not sure we intended to change this!

cpjmcquillan: Fixed here @avinhurry 3edf1ce69efbc69b50db63404c7a2191c80e1ed5

avinhurry: Nice, thanks. Good idea with those tests too.

avinhurry (approved):

PR #2756, opened 2026-04-23
Fix `ECTDeclarationScenarios` api seeds

leandroalemao:
### Context

When we run normal seeds and api seeds after, the api seeds `ECTDeclarationScenarios` is broken.

- Fix it so we can run both seeds with no errors;

### Guidance to review

Run locally:

`rake db:reset`
and straight after
`rake api_seed_data:generate`

No errors should be displayed.


peterdavidhamilton (approved):
@leandroalemao I've been finding that seeds hang for me right at the end, but not always. Any idea why that might be?

leandroalemao:
> @leandroalemao I've been finding that seeds hang for me right at the end, but not always. Any idea why that might be?

Hi @peterdavidhamilton That might be the metadata handlers refreshing all records in the end.. takes awhile.. or are you seeing any exceptions being raised?

ethax-ross (approved):

PR #2761, opened 2026-04-23
[3592] Fix teacher metadata handler bug

leandroalemao:
### Context

Ticket: [3592](https://github.com.mcas.ms/DFE-Digital/register-ects-project-board/issues/3592)

When a teacher is moved to another school, we `finish!` the latest mentorship period.

- When the latest mentorship period is finished, we set `api_mentor_id` as nil in the metadata handler;
- It triggers the handler updates and the API is updated correctly;

### Guidance to review
Review app

### DIFF
--- a/app/services/metadata/handlers/teacher.rb
+++ b/app/services/metadata/handlers/teacher.rb
@@ -27,7 +27,9 @@ module Metadata::Handlers
         latest_mentor_training_period = latest_mentor_training_period_by_lead_provider(teacher:)[lead_provider_id]
         latest_ect_contract_period = latest_ect_training_period&.contract_period
         latest_mentor_contract_period = latest_mentor_training_period&.contract_period
-        api_mentor_id = latest_ect_training_period&.at_school_period&.latest_mentorship_period&.mentor&.teacher&.api_id
+        api_mentor_id = if latest_ect_training_period&.at_school_period&.latest_mentorship_period&.ongoing_today?
+                          latest_ect_training_period.at_school_period.latest_mentorship_period.mentor&.teacher&.api_id
+                        end
         involved_in_school_transfer = school_transfers_exist_for(teacher.ect_at_school_periods, lead_provider_id) ||
           school_transfers_exist_for(teacher.mentor_at_school_periods, lead_provider_id)
 
--- a/spec/services/metadata/handlers/teacher_spec.rb
+++ b/spec/services/metadata/handlers/teacher_spec.rb
@@ -330,6 +330,114 @@ RSpec.describe Metadata::Handlers::Teacher do
         end
       end
 
+      context "when the latest ECT training period has a finished mentorship period" do
+        let!(:ect_at_school_period) do
+          FactoryBot.create(
+            :ect_at_school_period,
+            school: school1,
+            teacher: teacher1,
+            started_on: 1.year.ago,
+            finished_on: nil
+          )
+        end
+        let!(:mentor_at_school_period) do
+          FactoryBot.create(
+            :mentor_at_school_period,
+            school: school1,
+            started_on: 1.year.ago,
+            finished_on: nil
+          )
+        end
+        let!(:ect_training_period) do
+          FactoryBot.create(
+            :training_period,
+            :for_ect,
+            started_on: ect_at_school_period.started_on + 1.month,
+            finished_on: nil,
+            ect_at_school_period:,
+            school_partnership: school_partnership1
+          )
+        end
+        let!(:latest_mentorship_period) do
+          FactoryBot.create(
+            :mentorship_period,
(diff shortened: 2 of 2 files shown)


ethax-ross (commented):
I'm not sure this is the right approach 🤔 won't this result in participants being returned for a lead provider before the school partnership is created? I'm not sure how this fixes the `mentor_id` not updating for the original lead provider when the mentorship periods are changed? Did you see Connor's comment on the ticket, that's looks more like what I would expect the issue to be as the `mentor_id` is backed by metadata?

https://github.com/DFE-Digital/register-ects-project-board/issues/3592#issuecomment-4204540702

leandroalemao:
> I'm not sure this is the right approach 🤔 won't this result in participants being returned for a lead provider before the school partnership is created? I'm not sure how this fixes the `mentor_id` not updating for the original lead provider when the mentorship periods are changed? Did you see Connor's comment on the ticket, that's looks more like what I would expect the issue to be as the `mentor_id` is backed by metadata?
> 
> [DFE-Digital/register-ects-project-board#3592 (comment)](https://github.com/DFE-Digital/register-ects-project-board/issues/3592#issuecomment-4204540702)

Hi @ethax-ross 

trying to explain the scenario.

before the fix:

- a participant is already being returned by the api for a specific lead provider;
- via admin UI, you assign a mentor to the participant;
- API response `mentor_id` and `updated_at` are updated correctly;
- via admin UI, you move the ect to a school with no partnerships and tick the same lead provider in the forms/wizards;
- a new training period is created with EOI only with the lead provider;
- teacher lead provider metadata handlers does not take in consideration EOIs in the query to fetch the latest training periods;
- API response does not change at all.. `mentor_id` and `updated_at` stays the same (with old `mentor_id` and `updated_at`)

after the fix:

- a participant is already being returned by the api for a specific lead provider;
- via admin UI, you assign a mentor to the participant;
- API response `mentor_id` and `updated_at` are updated correctly;
- via admin UI, you move the ect to a school with no partnerships and tick the same lead provider in the forms/wizards;
- a new training period is created with EOI only with the lead provider;
- teacher lead provider metadata handlers fetch latest training periods correctly also taking in consideration EOIs in the query;
- API response `mentor_id` and `updated_at` are updated correctly;

app/services/metadata/handlers/teacher.rb
```diff
@@ -27,7 +27,9 @@ def upsert_lead_provider_metadata!
         latest_mentor_training_period = latest_mentor_training_period_by_lead_provider(teacher:)[lead_provider_id]
         latest_ect_contract_period = latest_ect_training_period&.contract_period
         latest_mentor_contract_period = latest_mentor_training_period&.contract_period
-        api_mentor_id = latest_ect_training_period&.at_school_period&.latest_mentorship_period&.mentor&.teacher&.api_id
+        api_mentor_id = if latest_ect_training_period&.at_school_period&.latest_mentorship_period&.ongoing?
```
ethax-ross: Worth extracting `latest_mentorship_period` to private method to make this easier to read and we can memoize it as it'll be the same for all LPs?

spec/services/metadata/handlers/teacher_spec.rb
```diff
@@ -330,6 +330,60 @@ def perform_refresh_metadata
...
+          FactoryBot.create(
+            :training_period,
+            :for_ect,
+            started_on: ect_at_school_period.started_on + 1.month,
+            finished_on: nil,
+            ect_at_school_period:,
+            school_partnership: school_partnership1
+          )
+        end
+        let!(:latest_mentorship_period) do
```
ethax-ross, line 361: So much setup 😿 would it work just creating the mentorship period and referencing through that in the test? so `mentorship_period.mentee.teacher` instead of `teacher1` etc? maybe it doesn't scaffold everything we need though 🤔 just a thought

ethax-ross (approved):
Looks good some minor suggestions

mooktakim (approved):

spec/services/metadata/handlers/teacher_spec.rb
leandroalemao: Hi @ethax-ross I've decided to leave as it's.. as I tried that but the mentorship period factory does not create a at school periods alongside a training period with specific started and finished on dates.

ethax-ross: No worries its fine as-is. I just remembered we have this which might help, but take it or leave it 🙂  https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/db/seeds/support/mentorship_period_helpers.rb#L2

ethax-ross (approved):

ethax-ross (approved):

PR #2762, opened 2026-04-24
[3672] schedule change validation error for a participant who has left

peterdavidhamilton:
### Context

Once a participant's status means they have "left" an establishment changes are not permitted and the error message clearly communicates this.

A status of "leaving" still permits edits until the leaving date has passed.

Update "change schedule" validation and release notes.

~`API::Teachers::ChangeSchedule` now uses `API::TrainingPeriods;;TrainingStatus` as the source of truth. `#ongoing_today?` is the same as `active? || leaving?` and felt right to replace as we check `left?`.~

This meant restoring some predicates that were pruned.

Post-review changes: reordered validations so this new validation takes precedent over `withdrawn`

### DIFF
--- a/app/services/api/teachers/change_schedule.rb
+++ b/app/services/api/teachers/change_schedule.rb
@@ -99,8 +99,10 @@ module API::Teachers
     end
 
     def participant_status
-      @participant_status ||= API::TrainingPeriods::TeacherStatus
-        .new(latest_training_period: training_period)
+      @participant_status ||= API::TrainingPeriods::TeacherStatus.new(
+        latest_training_period: training_period,
+        teacher:
+      )
     end
 
     def lead_provider_is_currently_training_teacher
--- a/app/services/api/teachers/change_schedule.rb
+++ b/app/services/api/teachers/change_schedule.rb
@@ -57,6 +57,7 @@ module API::Teachers
 
     def training_period_not_withdrawn
       return if errors[:teacher_api_id].any?
+      return unless training_period
       return unless training_status&.withdrawn?
 
       errors.add(:teacher_api_id, "Cannot perform actions on a withdrawn participant")
@@ -97,11 +98,21 @@ module API::Teachers
         )
     end
 
+    def participant_status
+      @participant_status ||= API::TrainingPeriods::TeacherStatus.new(
+        latest_training_period: training_period,
+        teacher:
+      )
+    end
+
     def lead_provider_is_currently_training_teacher
       return if errors[:teacher_api_id].any?
       return unless training_period
+      return if participant_status.active? || participant_status.leaving?
 
-      errors.add(:teacher_api_id, "You cannot change this participant's schedule. Only the lead provider currently training this participant can update their schedule.") unless training_period.ongoing_today?
+      if participant_status.left?
+        errors.add(:teacher_api_id, "You cannot change this participant’s schedule. This is because the participant has a 'left' participant_status, so they are not training with you currently.")
+      end
     end
 
     def no_future_training_periods_exist
--- a/app/services/api/teachers/change_schedule.rb
+++ b/app/services/api/teachers/change_schedule.rb
@@ -99,10 +99,8 @@ module API::Teachers
     end
 
     def participant_status
-      @participant_status ||= API::TrainingPeriods::TeacherStatus.new(
-        latest_training_period: training_period,
-        teacher:
-      )
+      @participant_status ||= API::TrainingPeriods::TeacherStatus
+        .new(latest_training_period: training_period)
     end
 
     def lead_provider_is_currently_training_teacher
--- a/app/services/api/teachers/change_schedule.rb
+++ b/app/services/api/teachers/change_schedule.rb
@@ -106,11 +106,9 @@ module API::Teachers
     def lead_provider_is_currently_training_teacher
       return if errors[:teacher_api_id].any?
       return unless training_period
-      return if participant_status.active? || participant_status.leaving?
+      return unless participant_status.left?
 
-      if participant_status.left?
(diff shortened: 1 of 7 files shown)


app/services/api/training_periods/teacher_status.rb
```diff
@@ -16,5 +16,21 @@ def status
...
+      status == :active
+    end
+
+    def leaving?
+      status == :leaving
+    end
+
+    def left?
+      status == :left
+    end
```
craigmdavidson, line 39: 😭 😆

app/services/api/teachers/change_schedule.rb
```diff
@@ -97,11 +98,19 @@ def school_partnership
...
 
+    def participant_status
+      @participant_status ||= API::TrainingPeriods::TeacherStatus
+        .new(latest_training_period: training_period)
+    end
+
     def lead_provider_is_currently_training_teacher
       return if errors[:teacher_api_id].any?
       return unless training_period
+      return if participant_status.active? || participant_status.leaving?
```
craigmdavidson: Do we need this guard? Since we have the error.add wrapped in a conditional?

craigmdavidson (approved):
LGTM.

Nitpick around unnecessary guard clause?

app/services/api/teachers/change_schedule.rb
peterdavidhamilton: For the reason in the description it felt right to leave it, and I half expected someone to say actually:
> Let's also stop edits once they've committed to leaving too

ethax-ross: I think we may be confusing training status and participant status here? training status is a separate status and can also be `active` (or `withdrawn`/`deferred`):

https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/services/api/training_periods/training_status.rb#L10

This method was previously sound behaviour-wise (show an error unless the training period is ongoing today), but the error message itself was a bit vague from the LPs point of view. I think all we want to do is call out the scenario where the teacher has left training with that LP with its own error, so something like:

```
def lead_provider_is_currently_training_teacher
  return if errors[:teacher_api_id].any?
  return unless training_period
  return if training_period.ongoing_today?

  if participant_status.left?
    errors.add(:teacher_api_id, "You cannot change this participant’s schedule. This is because the participant has a 'left' participant_status, so they are not training with you currently.")
  else
    errors.add(:teacher_api_id, "You cannot change this participant's schedule. Only the lead provider currently training this participant can update their schedule.")
  end
end
```

It may be even better if we just pass the participant status into the error message directly (it will only ever be `left` or `joining` - `active` and `leaving` will both be `ongoing_today?`):

```
errors.add(:teacher_api_id, "You cannot change this participant’s schedule. This is because the participant has a '#{participant_status}' participant_status, so they are not training with you currently.")
```

The `joining` state is also covered by the `no_future_training_periods_exist` rule, so maybe we just change this rule to explicitly cover the `left` state?:

```
def lead_provider_is_currently_training_teacher
  return if errors[:teacher_api_id].any?
  return unless training_period
  return unless participant_status.left?

  errors.add(:teacher_api_id, "You cannot change this participant’s schedule. This is because the participant has a 'left' participant_status, so they are not training with you currently.")
end
```

peterdavidhamilton: Yes, there was a moment when I was thinking, hang on, have I been looking at the wrong status service. I think what is not immediately apparent is what the source of truth is for a given status. I'd argue it makes sense to always defer to the status service(s) even when there is predicate in the `Interval` module because our service is scoped to the API. The convention of having dedicated services vs. methods on models is a recurring one (`TrainingPeriod#status`)

leandroalemao (approved):

app/services/api/training_periods/teacher_status.rb
```diff
@@ -22,6 +22,22 @@ def status
       end
     end
 
+    def joining?
```
ethax-ross, line 25: Can we add tests for these

app/services/api/teachers/change_schedule.rb
```diff
@@ -57,6 +57,7 @@ def schedule_exists
 
     def training_period_not_withdrawn
       return if errors[:teacher_api_id].any?
+      return unless training_period
```
ethax-ross, line 60: This is covered by the optional chaining below, but happy if you want to leave it in for explicitness

ethax-ross (approved):
Looks good just a couple of minor points - can we make sure this one goes through PO review please 🙏

app/services/api/teachers/change_schedule.rb
```diff
@@ -97,11 +98,19 @@ def school_partnership
...
+      )
+    end
+
     def lead_provider_is_currently_training_teacher
       return if errors[:teacher_api_id].any?
       return unless training_period
+      return unless participant_status.left?
 
-      errors.add(:teacher_api_id, "You cannot change this participant's schedule. Only the lead provider currently training this participant can update their schedule.") unless training_period.ongoing_today?
+      errors.add(:teacher_api_id, "You cannot change this participant’s schedule. This is because the participant has a 'left' participant_status, so they are not training with you currently.")
```
mooktakim: minor issue, lets replace curvy `’` with straight `'`

mooktakim (approved):

app/services/api/teachers/change_schedule.rb
mooktakim: @avinhurry its for error response not view ok 🤪

avinhurry: @mooktakim contrary to what you may believe I actually think this one should be plain ASCII... 😱

Because its API JSON rather than content rendered on the UI but the surrounding code previously used smart quotes, now that it's all consistent, it makes me happy 😄

PR #2765, opened 2026-04-24
Return :active status for ECTs and mentors who have completed with finished training period

craigmdavidson:
### Context

Resolves https://github.com.mcas.ms/DFE-Digital/register-ects-project-board/issues/3838

Currently if a training period is finished the participant_status returns as `:left` (or `:leaving` if the date is in the future).

If the teacher has "completed" in their role in this training period (for an ECT this would be completed induction, for a mentor this would be that they became ineligible for funding), then their status should be :active rather than :left when the training_period is finished. 

### Guidance to review

Relatively small change. Some restructuring of test logic.

### DIFF
--- a/spec/factories/training_period_factory.rb
+++ b/spec/factories/training_period_factory.rb
@@ -38,8 +38,8 @@ FactoryBot.define do
     end
 
     trait :finished do
-      started_on  { ect_at_school_period&.started_on || 1.year.ago }
-      finished_on { ect_at_school_period&.started_on&.+(1.month) || 2.weeks.ago }
+      started_on  { ect_at_school_period&.started_on || mentor_at_school_period&.started_on || 1.year.ago }
+      finished_on { ect_at_school_period&.started_on&.+(1.month) || mentor_at_school_period&.started_on&.+(1.month) || 2.weeks.ago }
     end
 
     trait(:school_led) do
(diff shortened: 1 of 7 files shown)


spec/factories/training_period_factory.rb
```diff
@@ -38,8 +38,8 @@
     end
 
     trait :finished do
-      started_on  { ect_at_school_period&.started_on || 1.year.ago }
-      finished_on { ect_at_school_period&.started_on&.+(1.month) || 2.weeks.ago }
+      started_on  { ect_at_school_period&.started_on || mentor_at_school_period&.started_on || 1.year.ago }
```
ethax-ross, line 41: Good spot!

ethax-ross (approved):
Nicely done 👍

claire-hughez:
This works for mentors when they complete training, their participant_status stays 'active', but I can't test for ECTs without help with seed data.

However I think it's not working where the mentor is already 'left' and then completes training. It then turns the participant_status to 'active' when I think it should stay 'left'? Based on this ticket, which was identified in the parity check before: https://github.com/DFE-Digital/register-ects-project-board/issues/3663

craigmdavidson:
> However I think it's not working where the mentor is already 'left' and then completes training. It then turns the participant_status to 'active' when I think it should stay 'left'? Based on this ticket, which was identified in the parity check before: https://github.com/DFE-Digital/register-ects-project-board/issues/3663

@claire-hughez Updated with commit **[A teacher acting as a mentor should be deemed left if they become ineligible for funding after the given training_period has finished.](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/2765/changes/dd4ec72f289edb5a907f6158750d72dfa79a02ba)**

ethax-ross (approved):

PR #2772, opened 2026-04-24
Travel in time!

cpjmcquillan:
### Context

Driven by #2922 

Since https://github.com/DFE-Digital/register-early-career-teachers-public/commit/f8d8513bd8a50acb62953c0cb92fb0cdd937112f, it has been possible to travel in time when submitting API requests.

This has proved really useful for testing, especially when we want to setup data
or simulate scenarios happening on specific dates.

This was limited to the API though, and there was no capability for us to do the
same across the application.

This extends the functionality enabled by setting `ENABLE_TIME_TRAVEL` to
`true`.

We're storing the dates in the session and leaning on `Current` attributes to
set and access the dates on a per-request basis.

Users can change or reset the date via a simple form.

Now, (except in Production) users can "Choose a date" they want to travel to,
and be transported through space and time!

**Note:** this is definitely not thread-safe, and it might break if pushed too
hard. So bear that in mind! It definitely isn't intended to ever be run in
Production, and its usefulness might degrade the closer an environment gets to
being Production-like (e.g Staging). So also bear that in mind!

### Guidance to review

Play around with the functionality in the review app 📆

### DIFF
--- a/app/components/time_traveller_component.rb
+++ b/app/components/time_traveller_component.rb
@@ -1,0 +1,39 @@
+class TimeTravellerComponent < ApplicationComponent
+  erb_template <<~ERB
+    <% if Current.date_after_time_travel.present? %>
+      <%= govuk_phase_banner(tag: {colour: "purple", text: "Warning 📆🚨"}) do %>
+        <%= message %>
+        <%= govuk_link_to("Change it or reset it", new_time_traveller_path) %>
+      <% end %>
+    <% else %>
+      <%= govuk_phase_banner(tag: {colour: "purple", text: "Travel in time ⏪"}) do %>
+        <%= govuk_link_to("Choose a date", new_time_traveller_path) %>
+      <% end %>
+    <% end %>
+  ERB
+
+  def render? = Rails.application.config.enable_time_travel
+
+private
+
+  def message
+    <<~TXT.squish
+      Today is #{Current.date_after_time_travel.to_fs(:govuk)}, which is
+      #{relative_date_in_words}!
+    TXT
+  end
+
+  def relative_date_in_words
+    distance_in_words = helpers.distance_of_time_in_words(
+      Current.date_before_time_travel,
+      Current.date_after_time_travel
+    )
+    suffix = if Current.date_after_time_travel.after?(Current.date_before_time_travel)
+               "from now"
+             else
+               "ago"
+             end
+
+    "#{distance_in_words} #{suffix}"
+  end
+end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,6 +1,7 @@
 class ApplicationController < ActionController::Base
   include Authentication
   include DfE::Analytics::Requests
+  include TimeTravellable
 
   before_action :set_sentry_user
 
--- a/app/controllers/concerns/time_travellable.rb
+++ b/app/controllers/concerns/time_travellable.rb
@@ -1,0 +1,25 @@
+require "active_support/testing/time_helpers"
+
+module TimeTravellable
+  extend ActiveSupport::Concern
+  include ActiveSupport::Testing::TimeHelpers
+
+  included do
+    around_action :travel_in_time,
+                  if: -> { Rails.application.config.enable_time_travel }
+  end
+
+private
+
+  def travel_in_time
+    if session["date_after_time_travel"].present?
+      Current.date_before_time_travel = Date.current
+      Current.date_after_time_travel = Date.parse(session["date_after_time_travel"])
+      travel_to Current.date_after_time_travel
+    end
+
+    yield
+  ensure
+    travel_back
+  end
+end
--- a/app/services/time_traveller.rb
+++ b/app/services/time_traveller.rb
@@ -1,0 +1,21 @@
+class TimeTraveller
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+  include ActiveRecord::AttributeAssignment
+
+  attribute :travel_to, :date
+
+  validate :travel_to_valid
+
+  def travel_to_date
+    @travel_to_date ||= Schools::Validation::HashDate.new(travel_to)
+  end
+
+private
+
+  def travel_to_valid
+    return if travel_to_date.valid?
+
+    errors.add(:travel_to, travel_to_date.error_message)
+  end
+end
--- a/app/views/time_travellers/new.html.erb
+++ b/app/views/time_travellers/new.html.erb
@@ -1,0 +1,30 @@
+<% page_data(
+    title: "What date do you want to travel to?",
+    error: @time_traveller.errors.any?,
+    backlink_href: backlink_with_fallback(fallback: root_path)
+  ) %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_with(model: @time_traveller, url: time_traveller_path) do |form| %>
+    <%= content_for(:error_summary) { form.govuk_error_summary } %>
+
+    <%= form.govuk_date_field(
+          :travel_to,
+          width: "two-thirds",
+          legend: { text: "Travel to date", size: "m", tag: "h3", hidden: true },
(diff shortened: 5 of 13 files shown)


app/views/time_travellers/new.html.erb
```diff
@@ -0,0 +1,18 @@
...
+  <%= content_for(:error_summary) { form.govuk_error_summary } %>
+  <%= form.govuk_date_field(
+        :travel_to,
+        width: "two-thirds",
+        legend: { text: "Travel to date", size: "m", tag: "h3", hidden: true },
+        hint: { text: "For example, 17 9 #{Date.current.year.next}" }
+      ) %>
+  <%= form.govuk_submit "Back to the future!" do %>
+    <%= govuk_button_to "Back to the real world!", time_traveller_path, method: :delete, warning: true %>
+  <% end %>
```
peteryates: Image

cpjmcquillan: Grouping the two buttons like this means the "Submit" actually submits the "destroy" form too

app/services/time_traveller.rb
```diff
@@ -0,0 +1,21 @@
+class TimeTraveller
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+  include ActiveRecord::AttributeAssignment
+
+  attribute :travel_to, :date
+
+  validate :travel_to_valid
```
cpjmcquillan, line 8: Also need to validate the date is within a `ContractPeriod`, otherwise a bunch of bad stuff happens

cpjmcquillan: So a bunch of bad stuff isn't happening, as long as we don't try and check for an induction tutor. But that feels a bit strange in itself.

I wonder whether it makes sense to at least make sure the date travelling to is within the bounds of **all** `ContractPeriod` records 🤔

app/components/time_traveller_component.rb
```diff
@@ -0,0 +1,39 @@
+class TimeTravellerComponent < ApplicationComponent
+  erb_template <<~ERB
+    <% if Current.date_after_time_travel.present? %>
+      <%= govuk_phase_banner(tag: {colour: "purple", text: "Warning 📆🚨"}) do %>
+        <%= message %>
+        <%= govuk_link_to("Change it", new_time_traveller_path) %>,
```
thenapking: I think it would be super useful to also have a reset link here to just remove it.  It could just be an X icon.

app/services/time_traveller.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  def travel_to_date
+    @travel_to_date ||= Schools::Validation::HashDate.new(travel_to)
+  end
+
+private
+
+  def travel_to_valid
```
thenapking, line 16: It is still possible to enter invalid dates which pass validation but then cause a redirect because the error is unhandled.  For instance if you type the date in US format as 1/23/2024 you get redirected to "Sorry, there’s a problem with the service".  I'm not really sure if this is important though as this is really only for us to test.

app/controllers/application_controller.rb
```diff
@@ -1,6 +1,7 @@
 class ApplicationController < ActionController::Base
   include Authentication
   include DfE::Analytics::Requests
+  include TimeTravellable
```
thenapking, line 4: Are we really sure we want to add this at the top level?  Wouldn't it be better to excluded this from authentication routes for instance?  Also I'm not so familiar with the TimeTravelerMiddleware, but if that is intended just for use with the API then we wouldn't want this TimeTravelling to also apply to the routes for the API.

thenapking (approved):
This is a really great idea and I think it will be extremely useful for testing.  I think it might even be worth extracting the time-travel functionality in this project into a gem which could be made available to the department or gov-uk in general.

app/controllers/concerns/time_travellable.rb
```diff
@@ -0,0 +1,28 @@
...
+  def set_time_travelled_dates
+    if session["date_after_time_travel"].present?
+      Current.date_before_time_travel = Date.current
+      Current.date_after_time_travel = Date.parse(session["date_after_time_travel"])
+    end
+  end
+
+  def travel_in_time
+    if Current.date_after_time_travel.present?
+      travel_to Current.date_after_time_travel
```
avinhurry, line 18: `travel_to` comes from `ActiveSupport::Testing::TimeHelpers`, so I wanted to check the concurrency implications of using it in `ApplicationController`.

Puma is configured with 5 threads by default, and my understanding is that `travel_to` stubs core time methods process wide rather than per request/session.

Could a concurrent request handled by the same Puma process see the travelled to date, even though it belongs to a different user/session?

app/components/time_traveller_component.rb
cpjmcquillan: Yeah I agree, that's what I'd tried to have initially. 

Since we're not using `turbo`, and since `rails-ujs` isn't a thing anymore, we can't have links with `method: :delete`. And a button looks hellish in that position. 

I've changed the link so it says "Change it, or reset it" instead, which hopefully makes things a little clearer 🙌🏼

app/controllers/concerns/time_travellable.rb
cpjmcquillan: Yeah, I don't think `travel_to` is thread-safe.

But, I don't think we care too much. 

This is going to provide immediate testing value in review apps, plus it's something we're already doing with the API. And it's not something we will ever use in Production.

If we run into issues, we can rethink.

What do you think?

app/services/time_traveller.rb
cpjmcquillan: Now we're rescuing `ActiveRecord::MultiparameterAssignmentErrors` so we handle this more gracefully 👍🏼

app/controllers/application_controller.rb
cpjmcquillan: I don't think there's really an issue with defining it at the top-level.  It's something we want to do for _every_ request if it's enabled and being used.

The `API::TimeTravelerMiddleware` lets API users travel to a given time using headers. If somebody is trying to do both at the same time, then 🤷🏼‍♂️.

IMO, I don't think it's worth adding any complexity to the codebase to mitigate people doing something weird with a testing tool.

app/controllers/concerns/time_travellable.rb
avinhurry: It still makes me slightly uncomfortable because review apps are used by multiple people at the same time. If the travelled date leaks between concurrent requests, people could see false errors or odd state and spend time debugging something that is not real. The risk may be small but it's still a risk.

Also, the API implementation uses a block to `travel_to` [which restores the original time at the end of the block](https://github.com/rails/rails/blob/v8.1.3/activesupport/lib/active_support/testing/time_helpers.rb#L193-L204). Here we're using `before_action`/`after_action`, which does not give the same guarantee. For example, if a later `before_action` redirects, or an exception is raised after `travel_to`, the `after_action` [may not run](https://guides.rubyonrails.org/action_controller_overview.html#after-action-and-around-action).

Could we at least switch this to an `around_action`/block form `travel_to`, even if we accept the review app concurrency tradeoff?

cpjmcquillan: I've switched to an `around_action` and ensure we `travel_back` each time.

Some people might have issues sometimes, some people might get value out of it, and some people might not.

Ultimately it's a non-Production change that can be easily turned off and easily reverted, so it's inherently low-risk IMO.

If we don't like it, we can get rid of it.

I don't think it's worth agonising over much more, so unless there are strong objections I'll move it to product review tomorrow, then merge it so we can start using it.

thenapking: Good idea.  I think this is likely to be super useful.  I think we should just keep an eye on it, and we can perhaps make some improvements in later updates if we have time.

avinhurry (approved):
Thanks, the switch to `around_action`/`ensure` addresses my cleanup concern.

I'm happy to approve given this is constrained to non production environments and we understand the tradeoff that there could occasionally be date leakage between concurrent users. It may be worth making that limitation visible during product review/testing.

mason-emily (approved):

PR #2776, opened 2026-04-27
API bundled release branch

ethax-ross:
Contains changes that need to go live before we open the floodgates to lead providers:

- #2774
- #2765
- #2777
- #2747 
- #2773
- #2781


cpjmcquillan:
@ethax-ross does this mean we only want to set `ENABLE_API` to true as part of this PR, rather than in #2752? 🤔

ethax-ross:
> @ethax-ross does this mean we only want to set `ENABLE_API` to true as part of this PR, rather than in #2752? 🤔

@cpjmcquillan lets keep the flags in the other launch PR, this is just some odds and ends we didn't get done in time for the cutoff yesterday 🙂

mooktakim (approved):

peteryates (approved):

PR #2780, opened 2026-04-27
Optimise `Declarations::Query`

ethax-ross:
### Context

The declarations query is hard to understand and not very performant. We want to improve the readability and make it faster/closer to ECF performance-wise.

Remove re-query of declarations by id, instead we can chain onto the existing `@scope` which is more optimal now that `pagy_countless` doesn't issue a `COUNT(*)`.

Optimise join on `training_periods` for `latest_training_period` so that we only end up with one training period (instead of an ECT and mentor training period) so that the subsequent filtering is simpler. Tweak ordering of filtering on previous declarations to be more efficient.

Join on the school periods at the top-level to avoid duplicating as most paths need these tables.

### Guidance to review

I checked this in the parity check with ambition 2023 declarations and it returned the same results before/after the refactor. The performance improved from ~3s to ~1-2s (with ECF being around 1s).

### DIFF
--- a/app/services/api/declarations/query.rb
+++ b/app/services/api/declarations/query.rb
@@ -12,7 +12,7 @@ module API::Declarations
       updated_since: :ignore
     )
       @lead_provider_id = lead_provider_id
-      @scope = Declaration.order(:created_at)
+      @scope = join_school_period(Declaration.order(:created_at))
 
       where_lead_provider_is(lead_provider_id)
       where_contract_period_year_in(contract_period_years)
@@ -58,14 +58,11 @@ module API::Declarations
     def where_lead_provider_is(lead_provider_id)
       return if ignore?(filter: lead_provider_id)
 
-      @scope = scope.where(id: declarations_matching_lead_provider(lead_provider_id).select(:id))
+      declarations_matching_lead_provider(lead_provider_id)
     end
 
-    # Keeps the lead-provider join chain in a sub-relation so the outer scope
-    # is filtered via WHERE id IN (subquery). This makes pagination COUNT(*)
-    # cheaper and lets other filters compose against a clean outer scope.
     def declarations_matching_lead_provider(lead_provider_id)
-      Declaration
+      @scope = scope
         # Join the lead provider for the declaration.
         .joins(
           training_period: {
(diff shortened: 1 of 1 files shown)


leandroalemao (approved):

craigmdavidson (approved):
As a general note our exclusive arc approach to database table design for ECTs and Mentors really hurts us here in terms of query complexity. If we had used a more standard Rails approach the resultant queries would have been far, _far_ simpler.

It would be useful when reviewing an optimisation PR to have some sense of the number of records in the production tables? When I run an `explain` on the generated query locally, we see a lot of seq scans (including two full table scans on `training_periods`) but this is likely because the tables locally are very small.

ethax-ross:
> As a general note our exclusive arc approach to database table design for ECTs and Mentors really hurts us here in terms of query complexity. If we had used a more standard Rails approach the resultant queries would have been far, _far_ simpler.
> 
> It would be useful when reviewing an optimisation PR to have some sense of the number of records in the production tables? When I run an `explain` on the generated query locally, we see a lot of seq scans (including two full table scans on `training_periods`) but this is likely because the tables locally are very small.

Thanks @craigmdavidson I agree we are hamstrung by the data model to a degree. For the `explain` when you finally get Azure access it's possible to create a konduit to a snapshot/production-like environment that has a copy of production data in order to test the queries with production data. 

Here is the explain output of the query in this PR against production data - I managed to eliminate all but one sequence scan, which I think is just down to the really nested `OR` that filters previous declarations to billable/by declaration date that prevents index usage: https://gist.github.com/ethax-ross/025f6cbd945b58a72c064251e7dab0e1

I have a feeling if we split it into two distinct queries (one to get declarations direct on the LP and the other to get previous, billable/dated declarations) and then `UNION` the result it would be more efficient, but that's a larger refactor than I wanted to pursue.

mooktakim (approved):

craigmdavidson:
> I have a feeling if we split it into two distinct queries (one to get declarations direct on the LP and the other to get previous, billable/dated declarations) and then UNION the result it would be more efficient, but that's a larger refactor than I wanted to pursue.

💯agree splitting would help, but worth keeping up our sleeve though as the number of declarations increase. 

The `lead_provider_id` seems a long way from the **declarations**? Would denormalizing not help also?

ethax-ross:
> > I have a feeling if we split it into two distinct queries (one to get declarations direct on the LP and the other to get previous, billable/dated declarations) and then UNION the result it would be more efficient, but that's a larger refactor than I wanted to pursue.
> 
> 💯agree splitting would help, but worth keeping up our sleeve though as the number of declarations increase.
> 
> The `lead_provider_id` seems a long way from the **declarations**? Would denormalizing not help also?

It would, but there's been push back on doing so in the core data model when we've brought it up in the past; the metadata does do this for some things, though -- usually with some specificity, like 'the latest of x'.

PR #2782, opened 2026-04-27
Spike: add `rails_semantic_logger`

ethax-ross:
> ⚠️ WIP - up for discussion, do not merge!

### Context

We don't have rich logging in production as we're formatting logs as a single text line that logit can't parse (it expects JSON). We have tagging ability but we'd have to manually add a bunch of tags for it to be useful. `rails_semantic_logger` would give us all this for free.

- Our logger is not formatting in JSON or structuring the log in a way that is useful for logit to consume and parse, so we result in only a couple of fields we are able to filter or query on in Kibana.

Add `rails_semantic_logger`, which gives us a richer, pre-structured JSON log message.

Include the user type and id in the payload.

### Guidance to review

> ⚠️ Before moving this forward we would need to filter out any PII, probably with a custom appender.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -66,6 +66,10 @@ gem "diffy"
 
 gem "zendesk_api", "~> 3.1", ">= 3.1.1"
 
+group :development, :production do
+  gem "rails_semantic_logger", group: %w[development production]
+end
+
 group :development do
   gem "better_errors"
   gem "binding_of_caller"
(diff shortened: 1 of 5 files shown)


peteryates:
Before we go ahead with this can we have a quick look at achieving the same goal via Rails' structured event reporting which was introduced in Rails 8.1? #2684 

Maybe that doesn't cover all our needs, I thought it included the ability to generate regular log messages in JSON but I'm less sure now having read the release notes.

ethax-ross:
Closing - ticketed up for someone else to pick it up

PR #2787, opened 2026-04-28
Fix database bottlenecking on metadata jobs

ethax-ross:
### Context

The metadata refresh in production is causing the database to be CPU-bound for the duration of the refresh, which results in dropped connections and failed jobs.

We want to scale back the jobs so that the metadata jobs don't overwhelm the database.

- Only refresh metadata when importing all schools

I think we put this in as a bit of a safety net, but looking at the hooks in `School` we only refresh the metadata on `create`, so refreshing all metadata at the end of an update/import is overkill.

Switch to refreshing metadata inline, which will only happen when the `GIAS::Importer` adds a new school - presumably an infrequent event.

Retain the async metadata refresh for the scenario where there are no `GIAS::School` records in the database and we are importing them all for the first time, as this will be more efficient.

- Rework the queues/workers

We are finding in production that the volume of workers is overloading the database; we currently have a maximum of 3 processes * 3 threads * 10 workers, so 90 concurrent jobs.

As we have an `*` in one of our workers it will end up picking up any pending job potentially blocking the more critical/time-sensitive queues such as `mailer`.

Scale the number of physical workers down to 2.

Ensure we have a dedicated workers for mailers/default/metadata/everything else and scale the metadata workers down to a maximum of 2 at any given time (1 thread, 1 process and 2 workers). This is also enforced in `limits_concurrency` on the job in case we scale the workers up from 2 in the future.

Only run parity check/migration/seed jobs in relevant environments.

### Guidance to review

It's the school handler metadata jobs specifically that seem to cause the high CPU usage rather than the other handlers.

I tried various process/thread settings for the metadata refresh, at around 4 total concurrent jobs it was still hovering at ~95% max CPU. Dropping to 2 concurrent jobs brings the CPU usage down to a steady 70% - note they go into a new `Blocked Jobs` tab due to the concurrency restrictions:

Screenshot 2026-04-28 at 13 57 48

Screenshot 2026-04-28 at 14 55 02

The UI remained responsive during the full sync on the parity check environment. The refresh took a little longer than previously, but its not time-sensitive so still fine.

### DIFF
--- a/app/jobs/refresh_metadata_job.rb
+++ b/app/jobs/refresh_metadata_job.rb
@@ -1,6 +1,8 @@
 class RefreshMetadataJob < ApplicationJob
   queue_as :metadata
 
+  limits_concurrency to: 2, key: :refresh_metadata
+
   def perform(object_type:, object_ids:, track_changes: false)
     objects = object_type.where(id: object_ids)
     Metadata::Manager.new.refresh_metadata!(objects, track_changes:)
--- a/app/services/gias/importer.rb
+++ b/app/services/gias/importer.rb
@@ -15,10 +15,7 @@ module GIAS
     end
 
     def fetch
-      DeclarativeUpdates.skip(:metadata) do
-        import_only? ? fetch_and_import_only : fetch_and_update
-      end
-      Metadata::Handlers::School.refresh_all_metadata!(async: true)
+      import_only? ? fetch_and_import_only : fetch_and_update
     end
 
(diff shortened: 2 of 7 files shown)


tonyheadford (approved):
LGTM

mooktakim (approved):

leandroalemao (approved):

peteryates (approved):
Nice, seems like a sensible move.

PR #2792, opened 2026-04-28
Fix flaky uplifts spec by pinning declarations to a known-good contract period

mooktakim:
### Context

Fix flaky test when creating declarations contract period with uplift

* Create contract period with uplift and use that for the tests

### DIFF
--- a/spec/services/payment_calculator/banded/uplifts_spec.rb
+++ b/spec/services/payment_calculator/banded/uplifts_spec.rb
@@ -1,22 +1,25 @@
 RSpec.describe PaymentCalculator::Banded::Uplifts do
+  let(:contract_period) { FactoryBot.create(:contract_period, year: 2024, uplift_fees_enabled: true).tap { it.update!(uplift_fees_enabled: true) } }
+  let(:active_lead_provider) { FactoryBot.create(:active_lead_provider, contract_period:) }
+
   let(:instance) do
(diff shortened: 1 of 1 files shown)


spec/services/payment_calculator/banded/uplifts_spec.rb
```diff
@@ -1,22 +1,25 @@
 RSpec.describe PaymentCalculator::Banded::Uplifts do
+  let(:contract_period) { FactoryBot.create(:contract_period, year: 2024).tap { it.update!(uplift_fees_enabled: true) } }
```
ethax-ross: Did you get to the bottom of why this was `false` sometimes? It looks like its `true` by default:

https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/spec/factories/contract_period_factory.rb#L29

mooktakim: The issue was the schools factory would create `uplift_fees_enabled: false` for 2025 and the sequence would pick 2025.

ethax-ross: ah ok, makes sense, thanks -- just a minor suggestion in that case:

```
FactoryBot.create(:contract_period, year: 2024, uplift_fees_enabled: true)
```

ethax-ross (approved):

leandroalemao (approved):

PR #2795, opened 2026-04-28
[3872] Add rails_semantic_logger for JSON logs in production

mooktakim:
### Context

Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/3872

* Using `rails_semantic_logger` for JSON logs
* Rails 8.1 Structured Event we would need to create our own JSON formatter, STDOUT appender and parameter filtering.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -72,7 +72,7 @@ group :development do
   gem "pg_query"
 end
 
-group :production, :review, :staging, :sandbox do
+group :production, :review, :staging, :sandbox, :paritycheck do
   gem "rails_semantic_logger"
 end
 
--- a/Gemfile
+++ b/Gemfile
@@ -72,6 +72,10 @@ group :development do
   gem "pg_query"
 end
 
+group :production, :review, :staging, :sandbox do
+  gem "rails_semantic_logger"
+end
+
 group :test do
   gem "capybara"
   gem "playwright-ruby-client"
--- a/Gemfile
+++ b/Gemfile
@@ -72,7 +72,7 @@ group :development do
   gem "pg_query"
 end
 
-group :production do
+group :production, :review, :staging, :sandbox, :paritycheck, :migration do
   gem "rails_semantic_logger"
 end
 
--- a/Gemfile
+++ b/Gemfile
@@ -72,7 +72,7 @@ group :development do
   gem "pg_query"
 end
 
-group :production, :review, :staging, :sandbox, :paritycheck do
+group :production do
   gem "rails_semantic_logger"
 end
 
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -22,4 +22,12 @@ class ApplicationController < ActionController::Base
   def set_sentry_user
     Sentry.set_user(email: current_user&.email)
   end
+
+private
+
+  def append_info_to_payload(payload)
+    super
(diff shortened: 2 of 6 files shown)


leandroalemao (approved):

config/environments/production.rb
```diff
@@ -58,13 +58,14 @@
...
-  config.logger = ActiveSupport::Logger.new($stdout)
-    .tap  { |logger| logger.formatter = ::Logger::Formatter.new }
-    .then { |logger| ActiveSupport::TaggedLogging.new(logger) }
-
-  # Prepend all log lines with the following tags.
+  # Log to STDOUT in JSON for logit ingestion. rails_semantic_logger replaces
+  # the Rails logger and honours config.log_tags + config.filter_parameters.
   config.log_tags = [:request_id]
+  config.semantic_logger.application = "" # No need to send the application name as logstash reads it from Cloud Foundry log tags
+  config.rails_semantic_logger.format = :json # Logit expects JSON-formatted logs
```
ethax-ross: I think this is redundant as we're adding a JSON appender below?

Gemfile
```diff
@@ -72,6 +72,10 @@ group :development do
   gem "pg_query"
 end
 
+group :production, :review, :staging, :sandbox do
```
ethax-ross: Do we need migration/parity check here? tbh I think we just build `production` for all of them so I'm not sure the others are necessary 🤔

ethax-ross (approved):
Looks good just some minor thoughts; would you mind doing a review of our `filter_parameters` against our models to make sure we're not missing any PII attributes before this goes out:

https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/config/initializers/filter_parameter_logging.rb#L11-L38

Gemfile
mooktakim: i'm not sure tbh, we have these envs, thought i'd add it to them

mooktakim:
> Looks good just some minor thoughts; would you mind doing a review of our `filter_parameters` against our models to make sure we're not missing any PII attributes before this goes out:
> 
> https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/config/initializers/filter_parameter_logging.rb#L11-L38

Couldn't find anything else to add other than `code` for the OTP

Gemfile
ethax-ross: 👍 we can maybe have another ticket to clean up the gemfile

app/controllers/application_controller.rb
```diff
@@ -22,4 +22,12 @@ def require_admin
   def set_sentry_user
     Sentry.set_user(email: current_user&.email)
   end
+
+private
+
+  def append_info_to_payload(payload)
+    super
+    payload[:current_user_class] = current_user&.class&.name
+    payload[:current_user_id] = current_user.try(:id)
```
ethax-ross, line 31: haven't come across `try` before, looks nicer than optional chaining 😅

ethax-ross (approved):

Gemfile
mooktakim: i just tried removing it and seems we do need them, added in paritycheck too

app/controllers/application_controller.rb
mooktakim: ah thats the old way lol

ethax-ross (approved):

PR #2798, opened 2026-04-29
Preserve previous contract period for provider-led transfers

ericaporter:
[Ticket](https://github.com/DFE-Digital/register-ects-project-board/issues/3875)

### Context
When registering an ECT who is continuing provider-led training from a previous school, we should keep the original contract period where appropriate rather than defaulting to the current one.

- Update contract period selection to preserve the previous training period's contract period (unless reassignment applies or payments are frozen)
- Introduce a `preserve: option` on `ContractPeriods::ForECTRegistration` to separate wizard and registration concerns
- Align schedule selection with the resolved contract period by passing preserved_schedule explicitly to - `TrainingPeriods::Create`
- Ensure school partnership matching is scoped to the preserved contract period, falling back to an expression of interest when no matching partnership exists

### Guidance to review
The main thing to sense check is the intended behaviour around contract period selection:

- When continuing provider-led training in a non-frozen contract period, we prioritise the previous contract period over the current one
- preserve: false is passed in two places — `Queries#registration_contract_period` and `RegisterECT#registration_contract_period` — both of which need the date-based CP rather than the preserved one
- Schedule selection always aligns with the chosen contract period via preserved_schedule
- Worth checking the Schedules::Find logic to confirm we are not accidentally overriding the preserved contract period in edge cases

Also worth checking this against the logic in Schedules::Find, as it previously “pinned” the contract period based on dates — want to make sure we’re not accidentally overriding the preserved contract period in edge cases.

### DIFF
--- a/app/services/schools/register_ect.rb
+++ b/app/services/schools/register_ect.rb
@@ -98,6 +98,7 @@ module Schools
             started_on: training_period_started_on,
             school_partnership:,
             expression_of_interest:,
+            schedule: preserved_schedule,
             author:
           ).call
         end
@@ -211,6 +212,16 @@ module Schools
 
     def previous_training_period
       store&.previous_training_period
+    end
+
+    def preserved_schedule
+      return nil unless preserving_previous_contract_period?
+
+      previous_training_period&.schedule
+    end
+
+    def preserving_previous_contract_period?
+      contract_period == previous_training_period&.contract_period
     end
   end
 end
--- a/app/services/schools/register_ect.rb
+++ b/app/services/schools/register_ect.rb
@@ -98,7 +98,7 @@ module Schools
             started_on: training_period_started_on,
             school_partnership:,
             expression_of_interest:,
-            schedule: preserved_schedule,
+            contract_period: preserved_contract_period,
             author:
           ).call
         end
@@ -218,14 +218,10 @@ module Schools
       store&.previous_training_period
     end
 
-    def preserved_schedule
-      return nil unless preserving_previous_contract_period?
+    def preserved_contract_period
+      return nil unless contract_period == previous_training_period&.contract_period
 
-      previous_training_period&.schedule
-    end
-
-    def preserving_previous_contract_period?
-      contract_period == previous_training_period&.contract_period
+      contract_period
     end
   end
 end
--- a/app/services/schools/register_ect.rb
+++ b/app/services/schools/register_ect.rb
@@ -206,7 +206,8 @@ module Schools
     def registration_contract_period
       @registration_contract_period ||= ContractPeriods::ForECTRegistration.new(
         started_on:,
-        previous_training_period:
+        previous_training_period:,
+        preserve: false
       ).call
     end
 
--- a/app/services/schools/register_ect.rb
+++ b/app/services/schools/register_ect.rb
@@ -187,13 +187,13 @@ module Schools
           previous_school_partnership_id: previous_id,
           school:,
           author:,
-          current_contract_period_year: registration_contract_period.year
+          current_contract_period_year: contract_period.year
         )
     end
 
     def training_period_started_on
       if ContractPeriods::Reassignment.new(training_period: previous_training_period).required?
-        [ect_at_school_period.started_on, registration_contract_period.started_on].max
+        [ect_at_school_period.started_on, contract_period.started_on].max
       else
         [ect_at_school_period.started_on, Date.current].max
       end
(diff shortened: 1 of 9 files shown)


app/services/schools/register_ect.rb
```diff
@@ -98,6 +98,7 @@ def create_training_period!
             started_on: training_period_started_on,
             school_partnership:,
             expression_of_interest:,
+            schedule: preserved_schedule,
```
thenapking: My question here is whether we need to pass in the reserved schedule.  The `Schedules::Find` service was selecting the right schedule identifier, but in the current contract period, rather than in the contract period from the previous training period.  If we could ensure the school_partnership or expression_of_interest has the correct contract period, from the previous training period, we might be able to do away with this override.  The contract period for the expression of interest and school partnership is set via the TrainingPeriodSources mixin, which uses the (overriden) contract_period method.  I *think* but I am not 100% sure that if that contract period is set correctly to the previous this should work without having to pass in the schedule.

app/services/schools/register_ect.rb
```diff
@@ -199,18 +200,32 @@ def training_period_started_on
...
     def contract_period
-      @contract_period ||= registration_contract_period
+      @contract_period ||= ContractPeriods::ForECTRegistration.new(
+        started_on:,
+        previous_training_period:
+      ).call
     end
 
     def registration_contract_period
       @registration_contract_period ||= ContractPeriods::ForECTRegistration.new(
```
thenapking: My feeling here is that calling `ContractPeriods::ForECTRegistration` twice is confusing.  This call forces the class to ignore the `previous_training_period` by passing `preserve: false`.  Which means that some of the logic for determinining which contract period is returned by  `ContractPeriods::ForECTRegistration` is in this class.  Instead the class should be fullt decoupled from this class, and only be called once.  In simple terms that class should return either:
1) the reassigned contract period, if it is required
2) the previous contract period, if a previous provider led training period in a non frozen year exists
3) the current contract period
I think it can do that fully on its own without this coupling.

thenapking (commented):
Tested locally and this looks like it works.  I think that it could be simplified / decoupled slightly, but that might not be possible.

app/services/schools/register_ect.rb
ericaporter: The issue is that `Schedules::Find` derives the contract period from the registration date `ContractPeriod.containing_date(latest_start_date)` rather than the school partnership or EOI. So even though both of those correctly reference the previous CP `Schedules::Find` still resolves a schedule from the current CP causing a mismatch.

Fixing Schedules::Find to derive the CP from the partnership/EOI instead would be the cleaner long-term solution but it seemed like beyond the scope for this fix - it probably needs to be looked at separately

ericaporter: Just to check - did you mean to try something like:
```
def initialize(period:, training_programme:, started_on:, period_type_key:, mentee:, contract_period: nil)
  @period = period
  @training_programme = training_programme
  @started_on = started_on
  @period_type_key = period_type_key
  @mentee = mentee
  @contract_period = contract_period
end
```
Then update `def contract_period_year` to use `@contract_period` when present, falling back on the date based look up when not?

This might simplify things...

thenapking: Yes, that was what I meant.  This is much simpler now there's only one contract_period method!

spec/wizards/schools/register_ect_wizard/registration_store/queries_spec.rb
```diff
@@ -187,8 +187,8 @@ def build_previous_training_period_double(provider_led:, contract_period:, lead_
 
       before { stub_previous_training(previous_training_period, previous_ect_period:) }
 
-      it "returns the normal registration contract period" do
-        expect(queries.registration_contract_period).to eq(contract_period_2025)
+      it "returns the previous training period's contract period" do
+        expect(queries.registration_contract_period).to eq(contract_period_2021)
```
thenapking: This is a small thing, but it could cause confusion in the future.  2021 is a contract_period we're reassigning, so  I'd create the previous_training_period in a later contract_period such as 2023.

thenapking (approved):
LGTM.

spec/wizards/schools/register_ect_wizard/registration_store/queries_spec.rb
ericaporter: Yeah good point, now updated 👍

PR #2799, opened 2026-04-29
Update production URLs in guidance and comms

ethax-ross:
We want to ensure all URLs include the `www` subdomain - especially for references to the API URLs - as some lead provider API clients are not following the redirects from `https://` to `https://www.` correctly.


avinhurry (approved):
Looks good.

Shall we update the healthcheck URL here too?

https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/config/terraform/application/config/production.tfvars.json#L7

ethax-ross:
> Looks good.
> 
> Shall we update the healthcheck URL here too?
> 
> https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/config/terraform/application/config/production.tfvars.json#L7

@avinhurry I almost updated that, but then I figured its checking more of the application health by flexing the redirect 😅

avinhurry:
> I almost updated that, but then I figured its checking more of the application health by flexing the redirect

@ethax-ross oooh, that was a great call! That can be our test for the redirect.

cpjmcquillan (approved):

claire-hughez:
Looks good to me, thanks Ross

PR #2805, opened 2026-04-29
TRNless Wonders

peterdavidhamilton:
### Context

TRN-less teachers will raise 500s when passed to the TRS client

Add scopes, guards and specs for the rare occasion a teacher has no TRN

Currently `4` in the wild.

### DIFF
--- a/app/jobs/teachers/schedule_trs_sync_job.rb
+++ b/app/jobs/teachers/schedule_trs_sync_job.rb
@@ -5,7 +5,13 @@ module Teachers
     BATCH_SIZE = 50
 
     def perform
-      teachers = Teacher.with_trn.found_in_trs.active_in_trs.ordered_by_trs_data_last_refreshed_at_nulls_first.limit(BATCH_SIZE)
+      teachers =
+        Teacher
+          .with_trn
+          .found_in_trs
+          .active_in_trs
+          .ordered_by_trs_data_last_refreshed_at_nulls_first
+          .limit(BATCH_SIZE)
 
       teachers.each_with_index do |teacher, i|
         Teachers::SyncTeacherWithTRSJob.set(wait: i.seconds).perform_later(teacher:)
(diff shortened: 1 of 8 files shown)


app/jobs/teachers/schedule_trs_sync_job.rb
```diff
@@ -5,7 +5,7 @@ class ScheduleTRSSyncJob < ApplicationJob
     BATCH_SIZE = 50
 
     def perform
-      teachers = Teacher.found_in_trs.active_in_trs.ordered_by_trs_data_last_refreshed_at_nulls_first.limit(BATCH_SIZE)
+      teachers = Teacher.with_trn.found_in_trs.active_in_trs.ordered_by_trs_data_last_refreshed_at_nulls_first.limit(BATCH_SIZE)
```
peteryates: Now might be the time to indent the calls, the line's quite tricky to follow.

peterdavidhamilton: Gladly

app/models/teacher.rb
```diff
@@ -120,6 +120,8 @@ class Teacher < ApplicationRecord
     order(arel_table[:trs_data_last_refreshed_at].asc.nulls_first)
   }
 
+  scope :with_trn, -> { where.not(trn: nil) }
+  scope :trnless, -> { where(trnless: true) }
```
peteryates: I wondered about this pair too, it feels a bit weird that they're not the inverse. But, then `not_trnless` is a double negative, which is a different kind of bad 😅

peteryates (approved):

app/models/teacher.rb
peteryates: What about

```suggestion
  scope :with_trn, -> { where(trnless: false) }
  scope :without_trn, -> { where(trnless: true) }
```

peterdavidhamilton: I'd prefer that, `with/without` is how I prefer to read spec contexts. I'll update it.

PR #2807, opened 2026-04-29
[#3878] Bug: Error when registering a mentor 

thenapking:
### Context

When a school registered a mentor who was ineligable for funding an error was raised and the mentor was not registered.  Instead, the mentor should have been registered but a training period shouldn't have been created for them.

### Guidance to review
The following teachers have been set to be ineligable for funding with the 3 possible reasons and dates in the past.  None are registered as mentors or ECTs in the system so you should be able to register them as new mentors in any school.

  { trn: "3002577", first_name: "Jonas",    last_name: "Bloggs" },
  { trn: "3002578", first_name: "Cynthia",  last_name: "Parks" },
  { trn: "3002579", first_name: "Taylor",   last_name: "Hawkins" },


spec/services/schools/register_mentor_spec.rb
```diff
@@ -259,10 +259,12 @@
       context "when the mentor is ineligible for funding" do
         let!(:teacher) { FactoryBot.create(:teacher, :ineligible_for_mentor_funding, trn:) }
 
-        it "raises a MentorIneligibleForTraining error" do
-          expect { service.register! }
-            .to raise_error(Schools::RegisterMentor::MentorIneligibleForTraining,
-                            /Mentor #{teacher.id} is not eligible for funded training/)
+        it "creates an associated MentorATSchoolPeriod record" do
```
cpjmcquillan:
```suggestion
        it "creates an associated MentorAtSchoolPeriod record" do
```

cpjmcquillan (approved):

lib/tasks/product_review/3878.rake
```diff
@@ -0,0 +1,26 @@
...
+  desc "Set up teachers who are ineligable for mentor funding (#3878)"
+  task "3878" => :environment do
+    Rails.logger.info "Setting up teachers who are ineligible for mentor funding"
+
+    reasons = %w[completed_declaration_received completed_during_early_roll_out started_not_completed]
+
+    CANDIDATE_TEACHERS.each_with_index do |teacher_attrs, i|
+      teacher = Teacher.find_or_initialize_by(trn: teacher_attrs[:trn])
+
+      teacher_attrs[:started_on] || Date.new(2024, 8 + i, 1)
```
avinhurry: This looks like it isn't assigned or used. Can we remove it?

lib/tasks/product_review/3878.rake
```diff
@@ -0,0 +1,26 @@
+namespace :product_review do
+  desc "Set up teachers who are ineligable for mentor funding (#3878)"
```
avinhurry:
```suggestion
  desc "Set up teachers who are ineligible for mentor funding (#3878)"
```

avinhurry (approved):
Looks good.

Had some minor comments about the rake task. Do we need to merge that task, or could we run it in the review app and remove it before merging?

cpjmcquillan:
> Looks good.
> 
> Had some minor comments about the rake task. Do we need to merge that task, or could we run it in the review app and remove it before merging?

I assumed that's what we'd be doing, but glad you've raised it explicitly! 💚

lib/tasks/product_review/3878.rake
thenapking: I'm going to remove the rake task - it was copied over from another one I was working with, but which I also won't merge.

PR #2809, opened 2026-04-29
Mitigate `Authentication failure! csrf_detected` errors

cpjmcquillan:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/3885

Before, we didn't handle [this error] when it was raised by omniauth and requests were redirected to `/auth/failure`.

This meant users were redirected to "Page not found" (which also happens quietly).

This is problematic, because it doesn't give users a clear idea of what happened, nor what they can do to remedy the issue.

This routes requests to `/auth/failure` to `SessionsController#failure` and adds some explicit logging to Sentry so we can track issues more easily. The session is ended too, which should give users a clean slate from which to sign in.
    
The page also prompts users to "try again", which resolves the issue in our experience.

[this error]: https://github.com/omniauth/omniauth_openid_connect/blob/4847685183147ff80697079481a72c27d3ad378d/lib/omniauth/strategies/openid_connect.rb#L133

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -38,6 +38,11 @@ class SessionsController < ApplicationController
     redirect_to post_login_redirect_path
   end
 
+  def failure
+    Sentry.capture_message("[Authentication failure] #{params[:message]}")
+    session_manager.end_session!
+  end
+
 private
 
   delegate :id_token, to: :user_builder
--- a/app/views/sessions/failure.md.erb
+++ b/app/views/sessions/failure.md.erb
@@ -1,0 +1,9 @@
+---
+title: There was a problem signing in
+---
+
+We were unable to sign you in.
+
+Please try again.
+
+<%= school_sign_in_cta("Continue to DfE Sign-in") %>
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -28,6 +28,7 @@ Rails.application.routes.draw do
   # omniauth sign-in
   get "/auth/:provider/callback", to: "sessions#create"
   post "/auth/:provider/callback", to: "sessions#create"
+  get "/auth/failure", to: "sessions#failure"
   get "/sign-in", to: "sessions#new"
   get "/sign-out", to: "sessions#destroy"
 
--- a/spec/requests/sessions_spec.rb
+++ b/spec/requests/sessions_spec.rb
@@ -223,4 +223,53 @@ RSpec.describe "Sessions", type: :request do
       end
     end
   end
+
+  describe "GET /auth/failure" do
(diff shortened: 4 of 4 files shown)


cpjmcquillan:
~~Hoping to add a feature test that mocks Omniauth/DfE sign in (and simulates the `csrf_detected` error), but will see how that goes!~~

I couldn't get one working, so we're going without!

app/controllers/sessions_controller.rb
```diff
@@ -38,6 +38,15 @@ def update
     redirect_to post_login_redirect_path
   end
 
+  def failure
+    if authenticated?
+      Sentry.capture_message("[Authentication failure] already_authenticated")
+      redirect_to post_login_redirect_path
```
thenapking: I'm worried that redirecting on failure when authenticated could cause an infinite loop or regression.  Could we add some sort of protection against that? Eg add a failure counter to the session and only redirect if we are under 3 or so failures?

cpjmcquillan: If the user ends up here, it's because `omniauth` has raised an error. 

Currently, we don't do anything, the application returns `not_found` and the user sees "Page not found".

The intention of this change is to show this failure page instead when that happens.

If the user is authenticated, then this should redirect them to the relevant page. I'm not sure I understand how redirecting to the relevant page would result in the user being redirected infinitely 🤔

At the minute what happens when the user attempts to sign in again after encountering this issue, is they are authenticated already and get straight to the relevant page.

app/controllers/sessions_controller.rb
```diff
@@ -38,6 +38,15 @@ def update
     redirect_to post_login_redirect_path
   end
 
+  def failure
+    if authenticated?
+      Sentry.capture_message("[Authentication failure] already_authenticated")
+      redirect_to post_login_redirect_path and return
```
ethax-ross: Not sure this `and return` is needed?

ethax-ross (approved):
Makes sense to me 👍 as discussed in Teams a feature spec would be ideal but it can come later as this is pressing

app/controllers/sessions_controller.rb
peteryates: I'm not sure either. If they hit this error is it worth us explicitly clearing their session so they have to make a fresh start?

cpjmcquillan: For now I've removed the redirect and we're clearing the session instead. We can monitor and revisit if necessary!

cpjmcquillan:
Tested in staging ✅

https://github.com/user-attachments/assets/c54a8ffe-a299-4591-838e-2c1b9fd6070e

PR #2814, opened 2026-04-29
Disable incident banner

cpjmcquillan:
We're confident this is no longer a P1 incident, so we can stop showing the incident banner.


thenapking (approved):
LGTM

avinhurry (approved):

peteryates (approved):

PR #2816, opened 2026-04-29
Remove `ENABLE_SCHOOLS_INTERFACE` flag

cpjmcquillan:
Since we migrated from ECF to RECT, there is no going back!

If we're not doing to disable this flag, then we may as well remove it.


peterdavidhamilton (approved):
That must've felt good

leandroalemao (approved):

PR #2822, opened 2026-04-30
Add lead provider short name

ethax-ross:
Adds a short name to lead providers; we often refer to lead providers internally with this terminology and it will let us tag our request logs with the lead provider somewhat obfuscated.

Initially I made short name unique, but its going to require fixing a whole bunch of specs as `name` isn't unique, so I've back-peddled on that for now -- we can pick it up separately for both `name` and `short_name`.

### DIFF
--- a/db/migrate/20260430085049_add_short_name_to_lead_providers.rb
+++ b/db/migrate/20260430085049_add_short_name_to_lead_providers.rb
@@ -1,0 +1,30 @@
+class AddShortNameToLeadProviders < ActiveRecord::Migration[8.1]
+  def up
+    add_column :lead_providers, :short_name, :string, null: true
+
+    short_names = {
+      "Ambition Institute" => "ambition",
+      "Best Practice Network" => "bpn",
+      "Capita" => "capita",
+      "Education Development Trust" => "edt",
+      "National Institute of Teaching" => "niot",
+      "Teach First" => "tf",
+      "UCL Institute of Education" => "ucl",
+    }
+
+    unknown_short_names = LeadProvider.where.not(name: short_names.keys).pluck(:name)
+
+    if unknown_short_names.any?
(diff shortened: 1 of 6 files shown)


mooktakim:
@ethax-ross do we want to also include this short name in `append_info_to_payload` for logit?

Maybe a `payload[:current_user_name]` field

craigmdavidson:
We backfill at the point of migration run, but don't assign anywhere going forward from this?

Am I missing something?

ethax-ross:
> @ethax-ross do we want to also include this short name in `append_info_to_payload` for logit?
> 
> Maybe a `payload[:current_user_name]` field

I have a follow up PR for this for after the field is out 👍

ethax-ross:
> We backfill at the point of migration run, but don't assign anywhere going forward from this?
> 
> Am I missing something?

Good point I may need to update some seeds, but I don't think in the application we create lead providers outside of factories/seeds

db/migrate/20260430085049_add_short_name_to_lead_providers.rb
```diff
@@ -0,0 +1,15 @@
+class AddShortNameToLeadProviders < ActiveRecord::Migration[8.1]
+  def up
+    add_column :lead_providers, :short_name, :string
+
+    LeadProvider.find_each do |lead_provider|
+      lead_provider.update!(short_name: lead_provider.name.split.map(&:first).join)
```
peteryates: wonder if we should hardcode these to start with 🤔 

* Ambition Institute: `ambition`
* Best Practice Network: `bpn`
* Capita: `capita`
* Education Development Trust: `edt`
* National Institute of Teaching: `niot`
* Teach First: `tf`
* UCL Institute of Education: `ucl`

ethax-ross:
Review app is in a bad state so I'm going to re-open this.

PR #2823, opened 2026-04-30
[3892] Fix empty `training_record_id` bug for some participants

leandroalemao:
### Context

Ticket: [3892](https://github.com.mcas.ms/DFE-Digital/register-ects-project-board/issues/3892)

Some lead providers shared with us that a training record ID is NULL for a participant over the API. This shouldn't happen.

- set the `api_ect_training_record_id` and `api_mentor_training_record_id` by default in the DB/remove explicitly setting them in register ECT/mentor services and backfill existing records where they are null;

### DIFF
--- a/db/migrate/20260430103643_backfill_empty_teachers_training_record_ids.rb
+++ b/db/migrate/20260430103643_backfill_empty_teachers_training_record_ids.rb
@@ -1,0 +1,15 @@
+class BackfillEmptyTeachersTrainingRecordIds < ActiveRecord::Migration[8.1]
+  def up
+    Teacher.where(api_ect_training_record_id: nil).find_each do |teacher|
+      teacher.update_column(:api_ect_training_record_id, SecureRandom.uuid)
+    end
+
+    Teacher.where(api_mentor_training_record_id: nil).find_each do |teacher|
+      teacher.update_column(:api_mentor_training_record_id, SecureRandom.uuid)
+    end
+  end
+
+  def down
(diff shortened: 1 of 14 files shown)


db/migrate/20260430103643_backfill_empty_teachers_training_record_ids.rb
```diff
@@ -0,0 +1,13 @@
+class BackfillEmptyTeachersTrainingRecordIds < ActiveRecord::Migration[8.1]
+  def up
+    Teacher.where(api_ect_training_record_id: nil).or(Teacher.where(api_mentor_training_record_id: nil)).find_each do |teacher|
```
ethax-ross: Can we maybe do these as two separate queries:

```
Teacher.where(api_ect_training_record_id: nil).find_each do |teacher|
  teacher.update!(api_ect_trianing_record_id: SecureRandom.uuid
end

Teacher.where(api_mentor_training_record_id: nil).find_each do |teacher|
  teacher.update!(api_mentor_training_record_id: SecureRandom.uuid
end
```

ethax-ross (commented):
Thanks Leandro, looks good - can we hold off on merging as we're checking with Nathan for any knock-on effects/issues from defaulting these to have a value

ethax-ross (approved):

peterdavidhamilton (approved):

ethax-ross (approved):

PR #2826, opened 2026-04-30
Redirect old ECF routes

avinhurry:
### Summary

- Redirect legacy ECF1 URLs that appear in early Google results for "Register early career teachers" to the new service root

- Default pages to `X-Robots-Tag: none`, and explicitly allow indexing for `/` and `/appropriate-body` when indexing is enabled

We have marked the Google results as outdated and requested a refresh. Once Google no longer indexes these URLs, [we should remove the legacy redirect routes and their request specs](https://github.com/DFE-Digital/register-early-career-teachers-public/issues/2827).

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -3,8 +3,6 @@
 # 2. Sessions::Manager#current_user -> Sessions::User.from_session
 #
 class SessionsController < ApplicationController
-  after_action :prevent_search_engine_indexing, only: :new
-
   def new
     render :new
   end
@@ -48,10 +46,6 @@ class SessionsController < ApplicationController
 private
 
   delegate :id_token, to: :user_builder
-
-  def prevent_search_engine_indexing
-    response.headers["X-Robots-Tag"] = "none"
-  end
 
   def record_school_user_signs_in_event
     return unless session_user.is_a?(Sessions::Users::SchoolUser)
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -3,6 +3,8 @@
 # 2. Sessions::Manager#current_user -> Sessions::User.from_session
 #
 class SessionsController < ApplicationController
+  after_action :prevent_search_engine_indexing, only: :new
+
   def new
     render :new
   end
@@ -46,6 +48,10 @@ class SessionsController < ApplicationController
 private
 
   delegate :id_token, to: :user_builder
+
+  def prevent_search_engine_indexing
+    response.headers["X-Robots-Tag"] = "none"
+  end
 
   def record_school_user_signs_in_event
     return unless session_user.is_a?(Sessions::Users::SchoolUser)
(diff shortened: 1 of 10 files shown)


config/routes.rb
```diff
@@ -18,6 +18,11 @@
   get "/access-denied", to: "pages#access_denied"
   get "/healthcheck", to: "health_check#show", as: :rails_health_check
 
+  # Redirect old indexed ECF1 URLs
+  get "/users/sign_in", to: redirect("/")
+  get "/check-account", to: redirect("/")
+  get "/nominations/resend-email", to: redirect("/")
```
avinhurry: Any more?

app/controllers/sessions_controller.rb
```diff
@@ -3,6 +3,8 @@
 # 2. Sessions::Manager#current_user -> Sessions::User.from_session
 #
 class SessionsController < ApplicationController
+  after_action :prevent_search_engine_indexing, only: :new
```
peteryates: I wonder if it'd make more sense to invert this default it to "none" everywhere, and only set it to "all" (or "nofollow") on the handful of landing pages?

peteryates: Then the `allow_indexing` config setting (which only allows it in production) could only override the default if the setting is enabled.

Feels easier to deal with the exception rather than the rule.

avinhurry: Yeah, good shout. If we only want the home and landing pages indexed, this is the better approach.

I've changed it so pages default to `X-Robots-Tag: none`, and only the opted in landing pages set `X-Robots-Tag: all` when indexing is enabled.

All new changes are in the latest commit.

peteryates (approved):
🚢

spec/requests/search_engine_indexing_spec.rb
```diff
@@ -1,9 +1,43 @@
...
+    it "allows the home page to be indexed" do
+      get("/")
+
+      expect(response.headers["X-Robots-Tag"]).to eq("all")
+    end
+
+    it "allows the appropriate body landing page to be indexed" do
+      get("/appropriate-body")
+
+      expect(response.headers["X-Robots-Tag"]).to eq("all")
```
avinhurry, line 16: I've opted in `/` and `/appropriate-body` for indexing only.

 Can you please confirm if:
- `/appropriate-body` should remain indexable (@peterdavidhamilton)

- `/api/guidance` and/or `/api/docs/v3` should also be indexable? (@ethax-ross)

Or any other pages?

peteryates: I think both of those pages should be indexable, they're the 3 landing pages for our 3 groups of users.

It _probably_ makes sense to allow it for the content under `/api/guidance` too, but that's less of a worry for now and one to discuss with the team later.

avinhurry: Thank you, I've updated `/api/guidance` to be indexed too.

Going to merge this now to try and improve the user journey from search engines but please shout if there any issues @peterdavidhamilton / @ethax-ross.

PR #2833, opened 2026-04-30
[#3836] Bug: Mentor reported leaving at one school finishes at all schools

thenapking:
### Context

When a mentor is mentoring across multiple schools, when one school reports them as leaving, this is populating their end date for their at_school periods at all other schools and training periods (where the training wasn't associated with that school).

The `MentorAtSchoolPeriods::Finish` class explicitly looks for all mentor_at_school/training/mentorship periods associated with the teacher and closes them.  I have created a new method that closes only those from the reporting school, and renamed the existing method so the difference between the two is clear.  Other classes should continue to use the method which closes all periods.

### DIFF
--- a/app/services/schools/register_mentor.rb
+++ b/app/services/schools/register_mentor.rb
@@ -45,7 +45,7 @@ module Schools
     def register!
       ActiveRecord::Base.transaction do
         create_teacher!
-        finish_existing_at_school_periods! if finish_existing_at_school_periods
+        finish_periods_at_all_schools! if finish_existing_at_school_periods
         start_at_school!
         create_training_period! unless mentoring_at_several_schools?
         set_eligibility_for_funding!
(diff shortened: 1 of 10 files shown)


app/services/schools/register_mentor.rb
ericaporter: Does this return need to be renamed too?

spec/services/mentor_at_school_periods/finish_spec.rb
```diff
@@ -191,14 +191,136 @@
...
+          teacher:,
+          school: mentor_at_school_period.school,
+          happened_at: finished_on
+        ).once
+
+        subject.finish_periods_at_reported_school!
+      end
+
+      context "when the mentor is training at another school" do
+        let!(:training_period) { FactoryBot.create(:training_period, :for_mentor, :ongoing, mentor_at_school_period: other_mentor_at_school_period, started_on:) }
```
ericaporter: Possibly this should be `other_training_period`

ericaporter (approved):
Looks good!! 👍

app/services/schools/register_mentor.rb
thenapking: Yes, in fact it allows us to clean up the naming in this class and the registration store, which had already been commented on.  We can just pass `mentoring_at_new_school_only` straight from the store without renaming it.

PR #2840, opened 2026-05-01
[3870] Fix replacement schedule logic

rwrrll:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/3870

A new mentor should be assigned a `replacement` schedule (instead of a `standard` one) when they're taking over from a previous mentor who already started training this ECT, otherwise the new mentor would "start fresh" on a standard schedule and we'd double-count training milestones and funding.

As per the ticket description / comments, "started training" here has been clarified to mean: the previous mentor has at least one non-voided declaration whose `declaration_date` falls inside their mentorship period for this ECT. The date bounds matter because a mentor's training period can outlive (or precede) any individual mentorship — we only want declarations that landed while they were mentoring this mentee.

Two fixes in this PR to achieve this:
- The lookup now spans all of the mentee's `ECTAtSchoolPeriods`, not just their current one. Before this PR, if the ECT moved schools, the new mentor at school B could be given a fresh standard schedule, even though the mentor at school A had already declared training.
- Voided declarations are now excluded — they shouldn't count as evidence that training started.

### DIFF
--- a/app/services/schedules/find.rb
+++ b/app/services/schedules/find.rb
@@ -90,7 +90,7 @@ module Schedules
     def mentee_has_declared_training_with_previous_mentor?
       MentorshipPeriod
         .joins(:mentee, mentor: { training_periods: :declarations })
-        .where(ect_at_school_periods: { teacher_id: mentee.teacher_id })
+        .where(ect_at_school_periods: { teacher_id: mentee.teacher_id }) # all periods at _any_ school
         .where.not(mentor_at_school_period_id: period.id)
         .where.not(declarations: { payment_status: :voided })
         .where("declarations.declaration_date >= mentorship_periods.started_on")
--- a/app/services/schedules/find.rb
+++ b/app/services/schedules/find.rb
@@ -85,15 +85,19 @@ module Schedules
       end
     end
 
-    def mentorship_periods_for_mentee_with_different_mentor
-      MentorAtSchoolPeriod
-        .joins(:mentorship_periods)
-        .merge(MentorshipPeriod.for_mentee(mentee.id))
-        .where.not(id: period.id)
-    end
-
     def previous_mentor_started_training?
-      mentorship_periods_for_mentee_with_different_mentor.joins(:declarations).exists?
+      MentorshipPeriod
+        .joins(:mentee)
+        .joins("INNER JOIN training_periods AS prev_tp
+                  ON prev_tp.mentor_at_school_period_id = mentorship_periods.mentor_at_school_period_id")
+        .joins("INNER JOIN declarations
+                  ON declarations.training_period_id = prev_tp.id")
+        .where(ect_at_school_periods: { teacher_id: mentee.teacher_id })
+        .where.not(mentorship_periods: { mentor_at_school_period_id: period.id })
+        .where.not(declarations: { payment_status: :voided })
+        .where("declarations.declaration_date >= mentorship_periods.started_on")
+        .where("declarations.declaration_date <= COALESCE(mentorship_periods.finished_on, CURRENT_DATE)")
+        .exists?
     end
 
     def replacement_schedule?
--- a/app/services/schedules/find.rb
+++ b/app/services/schedules/find.rb
@@ -85,7 +85,9 @@ module Schedules
       end
     end
 
-    def previous_mentor_started_training?
+    # A non-voided declaration dated within the mentorship period signals that the
+    # previous mentor started training (i.e. they weren't only assigned)
+    def mentee_has_declared_training_with_previous_mentor?
       MentorshipPeriod
         .joins(:mentee, mentor: { training_periods: :declarations })
         .where(ect_at_school_periods: { teacher_id: mentee.teacher_id })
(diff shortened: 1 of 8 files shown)


app/services/schedules/find.rb
```diff
@@ -85,15 +85,15 @@ def identifier_type
...
-        .joins(:mentorship_periods)
-        .merge(MentorshipPeriod.for_mentee(mentee.id))
-        .where.not(id: period.id)
-    end
-
     def previous_mentor_started_training?
-      mentorship_periods_for_mentee_with_different_mentor.joins(:declarations).exists?
+      MentorshipPeriod
+        .joins(:mentee, mentor: { training_periods: :declarations })
+        .where(ect_at_school_periods: { teacher_id: mentee.teacher_id })
```
thenapking: This is now quite a big join to have added to the `Schedules::Find` class.  I *think* but I am not totally sure that `.where(ect_at_school_periods: { teacher_id: mentee.teacher_id })` is redundant, because it is already covered by `.joins(:mentee)`.    

Is there a way we could abstract the logic for non-voided declarations between two dates into a scope in the Declaration class (and pick which is earlier of the end date and current date in Ruby?  I think logic to pick Declarations between two dates would be useful in general anyway.

thenapking (commented):
Looks good.  One comment about the SQL join which might not be possible to address.  I wanted to test this myself, but I think we need a couple of mentors setup with the correct declarations in order to see it working.

app/services/schedules/find.rb
rwrrll: Yeah not a mega fan of the big query, but the joins and desired structure here are novel, I don't think we there's anything we can really borrow from anywhere else. Sadly `joins(:mentee)` won't do the trick, the `.where(ect_at_school_periods: { teacher_id: mentee.teacher_id })` is what lets it span all of the teacher's ECTAtSchoolPeriods.

We could theoretically add to Declaration:

```
  scope :not_voided, -> { where.not(payment_status: :voided) }
  scope :with_declaration_date_between, ->(from, to) { where(declaration_date: from..to) }
```

...but the date comparison stuff needs to happen in SQL because the bounds reference mentorship_periods columns, not literal Ruby values, so only a minor readability improvement here while OTOH I'm hesitant to add scopes that are specific to a single query, I can't see anywhere else either of those would be used. Alas!!

thenapking: Ah, right that explains why you have `.where(ect_at_school_periods: { teacher_id: mentee.teacher_id })`.  Perhaps we could make that a named scope, or a method in this class so it's clear that is fetching all the teacher's ECT periods?  The `not_voided` scope is pretty vanilla so I think we could add that to the Declarations model, even though it's only used here.  But perhaps adding the date scope is too much, as you say.  But with that at least the query would be a little easier to read.

app/services/schedules/find.rb
```diff
@@ -85,15 +85,15 @@ def identifier_type
...
     end
 
-    def mentorship_periods_for_mentee_with_different_mentor
-      MentorAtSchoolPeriod
-        .joins(:mentorship_periods)
-        .merge(MentorshipPeriod.for_mentee(mentee.id))
-        .where.not(id: period.id)
-    end
-
     def previous_mentor_started_training?
```
cpjmcquillan: I wonder if we could rename this method to make the intent a little clearer? Or whether comments or test descriptions could help too?  💭

I _think_ we want to see whether a mentorship period exists for the given mentee teacher at **any** school, with a non-voided declaration associated to any of the mentor teacher's training periods declared after the mentorship period started and before it finished (or before today if it hasn't finished). Is that correct?

I'm not sure I completely understand how the declarations and declaration dates tie in to things.

cpjmcquillan (approved):
I read the associated ticket and things are a bit clearer to me now.

Should we capture some of that context in the commit message or PR description so there is a bit less digging to do for future us? 

What do you think? 🤔

app/services/schedules/find.rb
rwrrll: Yes good shout on both counts! Updated PR description and this method... does it make more sense now, or still a bit unclear?

thenapking: I still think we need some sort of naming on this bit of the join at least `.where(ect_at_school_periods: { teacher_id: mentee.teacher_id })`, because it is really unclear from the code why that is required, and joins(:mentee) doesn't do it.  If we can't name it in a method/scope could we add a comment?  I just think we're going to forget.

cpjmcquillan: @thenapking I think it's because we finish/start mentorship periods when a teacher moves schools. We want to retrieve all the mentorship periods associated with the mentee teacher at **any** school. `.joins(:mentee)` would just join on the single `ECTAtSchoolPeriod`, but we care about **every** `ECTAtSchoolPeriod` where the `teacher_id` matches `mentee.teacher_id`.

Does that make sense? 🤔

cpjmcquillan (approved):
@rwrrll the business logic still doesn't make much sense to me 😆, but I think the renaming of the method, comment, and updated PR description communicate the _why_ much more clearly now! 💚

app/services/schedules/find.rb
rwrrll: I've added an extra inline comment specifically on that line, hopefully an extra breadcrumb :)

PR #2841, opened 2026-05-01
Add lead provider name to logs

ethax-ross:
It's a pain having to map between the `id` and the lead provider to find out who is making which requests; adding the LP short name to each log entry.

No easy way to test this that I can find but other request specs should cover it such that we know it won't blow up.


leandroalemao (approved):

craigmdavidson (approved):
LGTM

mooktakim (approved):

PR #2842, opened 2026-05-01
[3874] Fix TRS name updates not bumping api_updated_at

mooktakim:
### Context

Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/3874

* Normalise `trs_first_name` and `trs_last_name` with squish.

* `DeclarativeUpdates.touch` now uses `after_create`, `after_update` and `after_destroy` instead of `after_commit` so that we run touch on every save within a single transaction.

### DIFF
--- a/app/models/concerns/declarative_updates.rb
+++ b/app/models/concerns/declarative_updates.rb
@@ -21,30 +21,34 @@ module DeclarativeUpdates
 
     def touch(target, when_changing: [], on_event: %i[update], timestamp_attribute: :updated_at, if: nil)
       condition = binding.local_variable_get(:if)
+      events = Array(on_event)
+      options = { target:, when_changing:, condition:, timestamp_attribute: }
 
-      after_commit(on: on_event) do
-        next if DeclarativeUpdates.skip?(:touch)
+      after_create { DeclarativeUpdates.perform_touch(self, **options) } if events.include?(:create)
+      after_update { DeclarativeUpdates.perform_touch(self, **options) } if events.include?(:update)
+      after_destroy { DeclarativeUpdates.perform_touch(self, **options) } if events.include?(:destroy)
+    end
+  end
 
-        should_touch_based_on_changes = destroyed? || when_changing.blank? || when_changing.any? do |attr|
-          saved_change_to_attribute?(attr)
-        end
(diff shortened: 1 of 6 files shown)


leandroalemao (approved):

spec/models/teacher_spec.rb
```diff
@@ -38,6 +38,46 @@
                     timestamp_attribute: :api_unfunded_mentor_updated_at
   end
 
+  describe "touch rollback", :with_touches do
+    it "rolls back the api_updated_at touch when the surrounding transaction rolls back" do
```
ethax-ross, line 42: I see you covered it already in the TRS sync specs, but is it worth an explicit test here as well for multiple saves inside the same transaction update correctly?

app/models/concerns/declarative_updates.rb
```diff
@@ -21,30 +21,34 @@ def refresh_metadata(target, when_changing: [], on_event: %i[update])
 
     def touch(target, when_changing: [], on_event: %i[update], timestamp_attribute: :updated_at, if: nil)
       condition = binding.local_variable_get(:if)
+      events = Array(on_event)
+      options = { target:, when_changing:, condition:, timestamp_attribute: }
 
-      after_commit(on: on_event) do
-        next if DeclarativeUpdates.skip?(:touch)
+      after_create { DeclarativeUpdates.perform_touch(self, **options) } if events.include?(:create)
```
ethax-ross, line 27: Do we need to check any potential side-effects given these will now run even if the transaction is later rolled back? I know the `api_updated_at` will be rolled back and you have a test for that, but is there anything else that could run and cause issues? I don't think there is, but might be worth a quick check unless you have already? 🤔

ethax-ross (approved):
Looks good just a couple of minor thoughts, thanks Mook 👍

app/models/concerns/declarative_updates.rb
mooktakim: Looked into it
Touch only does `update_column`, no validations or callbacks, so no cascading saves or recursion. Every touch in the repo targets a timestamp column, and the SQL runs inside the transaction so rolls back with it.

PR #2845, opened 2026-05-01
[#3898] Redirect users to access denied page on error fetching roles

thenapking:
Test redirection and errors

### Context
Errors in [Sentry](https://dfe-teacher-services.sentry.io/issues/7446855386/?project=4508369974788096&query=is%3Aunresolved&referrer=issue-stream)

When a user has not been assigned to an organisation in DfE Sign In the endpoint we use to fetch roles can return errors, as per the [documentation](https://github.com/DFE-Digital/login.dfe.public-api#get-user-access-to-service).

Currently we do not catch this error, and so a generic `RuntimeError` is raised and the user is redirected to the standard 500 error page.  Instead we want to redirect the user to the access denied page, so they know there's a problem with their account/access.

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -19,13 +19,13 @@ class SessionsController < ApplicationController
     end
   rescue Sessions::Users::Builder::UnknownOrganisation,
          Sessions::Users::Builder::UnknownPersonaType,
-         Sessions::Users::Builder::UnknownProvider,
-         DfESignIn::APIClient::AccessLevelNotFound,
-         DfESignIn::APIClient::UsersNotFound,
-         DfESignIn::APIClient::OrganisationNotFound,
-         DfESignIn::APIClient::RolesNotFound
+         Sessions::Users::Builder::UnknownProvider
 
-    session[:invalid_user_organisation_name] = user_builder.organisation_name if user_builder&.organisation_name
+    session[:invalid_user_organisation_name] = user_builder.organisation_name if user_builder.organisation_name
+    redirect_to access_denied_path
+  rescue DfESignIn::APIClient::NotFoundError => e
+    Sentry.capture_exception(e)
+
     redirect_to access_denied_path
   end
 
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -19,9 +19,10 @@ class SessionsController < ApplicationController
     end
   rescue Sessions::Users::Builder::UnknownOrganisation,
          Sessions::Users::Builder::UnknownPersonaType,
-         Sessions::Users::Builder::UnknownProvider
+         Sessions::Users::Builder::UnknownProvider,
+         DfESignIn::APIClient::AccessLevelNotFound
 
-    session[:invalid_user_organisation_name] = user_builder.organisation_name
+    session[:invalid_user_organisation_name] = user_builder.organisation_name if user_builder&.organisation_name
     redirect_to access_denied_path
   end
 
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -24,7 +24,7 @@ class SessionsController < ApplicationController
     session[:invalid_user_organisation_name] = user_builder.organisation_name
     redirect_to access_denied_path
   rescue DfESignIn::APIClient::NotFoundError => e
-    Sentry.capture_exception(e)
+    Sentry.capture_message(e.message, level: :info)
 
     redirect_to access_denied_path
   end
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -21,7 +21,7 @@ class SessionsController < ApplicationController
          Sessions::Users::Builder::UnknownPersonaType,
          Sessions::Users::Builder::UnknownProvider
 
-    session[:invalid_user_organisation_name] = user_builder.organisation_name if user_builder.organisation_name
+    session[:invalid_user_organisation_name] = user_builder.organisation_name
     redirect_to access_denied_path
   rescue DfESignIn::APIClient::NotFoundError => e
     Sentry.capture_exception(e)
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -21,9 +21,9 @@ class SessionsController < ApplicationController
          Sessions::Users::Builder::UnknownPersonaType,
          Sessions::Users::Builder::UnknownProvider,
          DfESignIn::APIClient::AccessLevelNotFound,
-         DfESignIn::APIClient::UserNotFound,
+         DfESignIn::APIClient::UsersNotFound,
          DfESignIn::APIClient::OrganisationNotFound,
-         DfESignIn::APIClient::RoleNotFound
+         DfESignIn::APIClient::RolesNotFound
 
     session[:invalid_user_organisation_name] = user_builder.organisation_name if user_builder&.organisation_name
     redirect_to access_denied_path
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -20,7 +20,10 @@ class SessionsController < ApplicationController
   rescue Sessions::Users::Builder::UnknownOrganisation,
          Sessions::Users::Builder::UnknownPersonaType,
          Sessions::Users::Builder::UnknownProvider,
-         DfESignIn::APIClient::AccessLevelNotFound
+         DfESignIn::APIClient::AccessLevelNotFound,
+         DfESignIn::APIClient::UserNotFound,
+         DfESignIn::APIClient::OrganisationNotFound,
+         DfESignIn::APIClient::RoleNotFound
 
     session[:invalid_user_organisation_name] = user_builder.organisation_name if user_builder&.organisation_name
     redirect_to access_denied_path
--- a/lib/dfe_sign_in/api_client.rb
+++ b/lib/dfe_sign_in/api_client.rb
@@ -8,8 +8,8 @@ module DfESignIn
     class DfESignInDisabled < StandardError; end
     class OrganisationNotFound < StandardError; end
     class AccessLevelNotFound < StandardError; end
-    class RoleNotFound < StandardError; end
-    class UserNotFound < StandardError; end
+    class RolesNotFound < StandardError; end
+    class UsersNotFound < StandardError; end
 
     attr_reader :connection
 
@@ -61,7 +61,7 @@ module DfESignIn
       if response.success?
         response.body
       else
-        raise UserNotFound, "Could not retrieve user: #{response.status} #{response.body}"
+        raise UsersNotFound, "Could not retrieve users: #{response.status} #{response.body}"
       end
     end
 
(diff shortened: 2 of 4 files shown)


app/services/organisation/access.rb
```diff
@@ -31,7 +33,12 @@ def roles
     # @see https://github.com/DFE-Digital/login.dfe.public-api
     # @return [DfESignIn::AccessLevel]
     def access_level
-      @access_level ||= DfESignIn::APIClient.new.access_levels(organisation_id:, user_id:)
+      @access_level ||= DfESignIn::APIClient.new.access_levels(
+        organisation_id:,
+        user_id:
+      )
+    rescue RuntimeError
```
thenapking: Obviously this is a bad idea to rescue such a generic error.  The errors which come out of the `APIClient` are all given this generic class, and a description which may include some more information.  However filtering on that description, which ultimately comes from DfESignIn also feels brittle.  Rewriting the lib to produce better error classes feels beyond the scope of this bug fix.  I am open to suggestions to a better path on this!

app/services/sessions/users/builder.rb
```diff
@@ -111,6 +111,8 @@ def dfe_sign_in?
       # @return [Array<String>] SchoolUser, AppropriateBodyUser, DfEUser
       def dfe_sign_in_roles
         @dfe_sign_in_roles ||= ::Organisation::Access.new(user_id: uid, organisation_id: organisation.id).roles
+      rescue Organisation::Access::NotAssignedError
+        raise(UnknownOrganisation, organisation)
```
thenapking: I've reused the existing error class because as I've understood this, the error is happening when the user has not been assigned to the organisation, but I am not fully sure that is the case.  I'm happy to add a specific error class for this.  Please suggest a name!

spec/requests/sessions_spec.rb
```diff
@@ -130,6 +167,20 @@
...
+
+      context "when the school user has not been assigned to the organisation" do
+        before do
+          allow(client).to receive(:access_levels).and_raise(
+            RuntimeError,
+            "API request failed: 404 RuntimeError"
+          )
+        end
+
+        it "redirects to the access denied page" do
```
thenapking: This is an integration test an might not really be necessary given we test redirection on error handling for this error above.  However it is the specific cause of the bug I am fixing so I thought adding a test was not unreasonable.  Happy to remove though.

lib/dfe_sign_in/api_client.rb
peteryates: Probably worth adding custom errors here too, and in the one below.

peteryates: Or, maybe rename the one above to `APIClient::RequestFailed` and remove the "API request failed:" prefix, and start the message with "could not retrieve access_levels" or "could not retrieve users".

lib/dfe_sign_in/api_client.rb
```diff
@@ -6,6 +6,10 @@ module DfESignIn
   #
   class APIClient
     class DfESignInDisabled < StandardError; end
+    class OrganisationNotFound < StandardError; end
+    class AccessLevelNotFound < StandardError; end
+    class RoleNotFound < StandardError; end
+    class UserNotFound < StandardError; end
```
peterdavidhamilton: These two errors might give the wrong impression, that they are about a specific user or role; rather than query users and roles on the service.

lib/dfe_sign_in/api_client.rb
```diff
@@ -43,7 +47,7 @@ def access_levels(organisation_id:, user_id:, service_id: client_id)
       if response.success?
         AccessLevel.from_response_body(response.body)
       else
-        raise "API request failed: #{response.status} #{response.body}"
+        raise AccessLevelNotFound, "Could not retrieve access level: #{response.status} #{response.body}"
```
rwrrll:
```suggestion
        raise AccessLevelNotFound, "Could not retrieve access levels: #{response.status} #{response.body}"
```

app/controllers/sessions_controller.rb
```diff
@@ -19,9 +19,13 @@ def create
     end
   rescue Sessions::Users::Builder::UnknownOrganisation,
          Sessions::Users::Builder::UnknownPersonaType,
-         Sessions::Users::Builder::UnknownProvider
+         Sessions::Users::Builder::UnknownProvider,
+         DfESignIn::APIClient::AccessLevelNotFound,
+         DfESignIn::APIClient::UsersNotFound,
+         DfESignIn::APIClient::OrganisationNotFound,
+         DfESignIn::APIClient::RolesNotFound
```
rwrrll: If we ever add a new DfE Sign In endpoint, we'd have to update this list... but if these exceptions inherited from a common ancestor (e.g. `DfESignIn::APIClient::RequestFailed` or something) we get it for free and this cleans up to:

```suggestion
         DfESignIn::APIClient::RequestFailed
```

lib/dfe_sign_in/api_client.rb
```diff
@@ -6,6 +6,10 @@ module DfESignIn
   #
   class APIClient
     class DfESignInDisabled < StandardError; end
+    class OrganisationNotFound < StandardError; end
+    class AccessLevelNotFound < StandardError; end
+    class RolesNotFound < StandardError; end
+    class UsersNotFound < StandardError; end
```
rwrrll: Assuming we need all 4 different exception classes, and the different messages aren't enough...

As above it would be good if they inherited from a common ancestor to consolidate rescues, also there's a mix of plural and singular here (AFAICT all return arrays so I think they should all be plural), and lastly to be precise it's not that they're not found (404), it's that the requests were not successful (non-2xx), so maybe e.g.:

```suggestion
    class RequestFailed < StandardError; end
    class OrganisationsRequestFailed < RequestFailed; end
    class AccessLevelsRequestFailed < RequestFailed; end
    class RolesRequestFailed < RequestFailed; end
    class UsersRequestFailed < RequestFailed; end
```

spec/requests/sessions_spec.rb
```diff
@@ -41,11 +41,95 @@
...
+        expect(response).to redirect_to(access_denied_path)
+      end
+    end
+
+    [
+      DfESignIn::APIClient::OrganisationNotFound,
+      DfESignIn::APIClient::AccessLevelNotFound,
+      DfESignIn::APIClient::UsersNotFound,
+      DfESignIn::APIClient::RolesNotFound
+    ].each do |error_class|
```
rwrrll: If you do introduce a common ancestor exception class and just rescue that, then the spec here can probably cover that one exception and the metaprogramming isn't needed

app/controllers/sessions_controller.rb
```diff
@@ -19,9 +19,13 @@ def create
...
 
-    session[:invalid_user_organisation_name] = user_builder.organisation_name
+    session[:invalid_user_organisation_name] = user_builder.organisation_name if user_builder&.organisation_name
```
rwrrll: Safe navigation is suspicious here because `user_builder` is a memoized class instantiation, it can't return `nil`, only an instance of `Sessions::Users::Builder` or raise. If it did raise, calling `user_builder` inside the rescue would re-raise.

Although I don't love accessing the ivar outside the reader method since it implicitly depends on the current call flow, you could get what you're after with `@user_builder&.organisation_name` (the ivar is only set if a builder was successfully created earlier in create).

spec/requests/sessions_spec.rb
```diff
@@ -41,11 +41,95 @@
...
+        let(:user_builder) { instance_double(Sessions::Users::Builder) }
+
+        before do
+          FactoryBot.create(:appropriate_body_period, dfe_sign_in_organisation_id:)
+
+          allow(Sessions::Users::Builder).to receive(:new).and_return(user_builder)
+          allow(user_builder).to receive(:organisation_name).and_return(nil)
+          allow(user_builder).to receive(:session_user).and_raise(Sessions::Users::Builder::UnknownOrganisation, dfe_sign_in_organisation_id)
+
+          allow(client).to receive(:access_levels).and_raise(error_class, "Error message")
```
rwrrll: Looks like the `access_levels` stub will never execute because of the `session_user` stub which is called first, so I think the test only proves `UnknownOrganisation` redirects (which the existing rescue already covered before this PR). To exercise the new rescue branches, we should drop the Builder stub and let the real builder hit the stubbed client.

spec/requests/sessions_spec.rb
```diff
@@ -41,11 +41,95 @@
...
+            dfe_sign_in_organisation_id:,
+            dfe_sign_in_user_id:,
+            dfe_sign_in_roles: %w[SchoolUser],
+          }
+        end
+
+        let(:user_builder) { instance_double(Sessions::Users::Builder) }
+
+        before do
+          FactoryBot.create(:appropriate_body_period, dfe_sign_in_organisation_id:)
```
rwrrll: Is this factory call needed?

rwrrll (commented):
A few aesthetic thoughts 💅  but a couple of important bits to check re: safe navigation and double stubbing masking the scenario we want to test here

spec/requests/sessions_spec.rb
thenapking: That's a good spot.  I've removed this stub as it is also no longer needed.

app/controllers/sessions_controller.rb
thenapking: I was calling this more as a caution.  Since I am now rescuing separately this is not required.  The issue I was expecting was that no organisation had been found and I worried this would also error, but couldn't precisely prove it.

spec/requests/sessions_spec.rb
thenapking: Having looked into the DfE Sign In library more closely and its use within this project, I've concluded that the only error we should rescue is `NotFound`.  Other errors, including 500 errors indicate either a widespread problem with the service, or a configuration error, not something unique to the user, and I do not want to rescue those in a way that would suggest the user doesn't have access.  The other endpoints in this library are as far as I can tell all unused.  We only use the access level, and that returns a 404 when the user is not assigned to an organisation or service, and therefore can't have any access level, which is the precise error this bug caused.  I've rewritten the error handling for the DfESignIn class to improve it, and clear up some tech debt.  If in the future we want to rescue all errors we can by rescuing `DfESignIn::APIClient::ResponseError`.

rwrrll (approved):
:shipit:

app/controllers/sessions_controller.rb
```diff
@@ -22,6 +22,10 @@ def create
          Sessions::Users::Builder::UnknownProvider
 
     session[:invalid_user_organisation_name] = user_builder.organisation_name
+    redirect_to access_denied_path
+  rescue DfESignIn::APIClient::NotFoundError => e
+    Sentry.capture_exception(e)
```
cpjmcquillan: Just wondering if we want to capture an exception here, or whether a message would satisfy our requirements? 🤔

I'd expect an error in Sentry to be some unexpected application exception that we ought to resolve, but I don't think that's the case here.

What do you think?

cpjmcquillan (approved):

app/controllers/sessions_controller.rb
thenapking: What I was intending here was that the message is built in the RequestError class from the status and body of the API response:
`"DfE Sign In API request failed with status #{response.status}: #{response.body}"`
I don't want to do that formatting etc here, but in the error class.  I've not used Sentry for error logging before so if the correct way to do this is with `Sentry.capture_message(e.message)` then I'll change it.  But I assumed then we'd loose the whole backtrace and other things?

cpjmcquillan: So you can do something like `Sentry.capture_message("[DfE Sign In] Not found", level: :info)` which categorises it as **not** an error in Sentry, but lets us log it still. We should still get all the backtrace etc.

PR #2854, opened 2026-05-05
Make mentorship ranges inclusive end

peteryates:
WIP - changing the range in `mentorship_periods` so it's inclusive at both ends.

### DIFF
--- a/spec/requests/schools/ects/change_mentor_wizard_spec.rb
+++ b/spec/requests/schools/ects/change_mentor_wizard_spec.rb
@@ -257,7 +257,7 @@ describe "Schools::ECTs::ChangeMentorWizardController" do
 
             ect_at_school_period.reload
             expect(ect_at_school_period.current_or_next_mentorship_period.mentor)
-            .to eq(other_mentor_at_school_period)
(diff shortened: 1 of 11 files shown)


spec/requests/schools/ects/change_mentor_wizard_spec.rb
```diff
@@ -257,10 +257,10 @@
 
             ect_at_school_period.reload
             expect(ect_at_school_period.current_or_next_mentorship_period.mentor)
-              .to eq(other_mentor_at_school_period)
+            .to eq(other_mentor_at_school_period)
```
peteryates: I think the indentation gone a bit awry here.

peteryates:
Thanks @thenapking - changes look good to me :ok_hand: 

There's one record we'd need to fix in prod before we merge this. I'll do that now.

spec/requests/schools/ects/change_mentor_wizard_spec.rb
thenapking: Fixed

thenapking (approved):
LGTM

PR #2856, opened 2026-05-05
[3515] Prevent mentor_no_longer_being_mentor withdrawal reason for ECTs

mooktakim:
### Context

Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/3515

* New validation in `API::Teachers::Withdraw`
* New validation in `TrainingPeriod` to reject the mentor-only withdrawal reason for ECTs
* Placeholder release note

### DIFF
--- a/app/services/api/teachers/withdraw.rb
+++ b/app/services/api/teachers/withdraw.rb
@@ -3,7 +3,7 @@ module API::Teachers
     include API::Concerns::Teachers::SharedAction
 
     WITHDRAWAL_REASONS = TrainingPeriod.withdrawal_reasons.values.map(&:dasherize).freeze
-    MENTOR_ONLY_WITHDRAWAL_REASONS = %w[mentor-no-longer-being-mentor].freeze
+    MENTOR_ONLY_WITHDRAWAL_REASONS = TrainingPeriod::MENTOR_ONLY_WITHDRAWAL_REASONS.map(&:dasherize).freeze
 
     attribute :reason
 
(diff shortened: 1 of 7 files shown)


app/services/api/teachers/withdraw.rb
```diff
@@ -3,6 +3,7 @@ class Withdraw
     include API::Concerns::Teachers::SharedAction
 
     WITHDRAWAL_REASONS = TrainingPeriod.withdrawal_reasons.values.map(&:dasherize).freeze
+    MENTOR_ONLY_WITHDRAWAL_REASONS = %w[mentor-no-longer-being-mentor].freeze
```
ethax-ross: Is it worth pushing this down to the `TrainingPeriod`?

spec/factories/training_period_factory.rb
```diff
@@ -142,7 +142,11 @@
       started_on { period_start_date }
       finished_on { period_end_date  }
       withdrawn_at { Faker::Date.between(from: started_on, to: (period_end_date)) }
-      withdrawal_reason { TrainingPeriod.withdrawal_reasons.values.sample }
+      withdrawal_reason do
+        reasons = TrainingPeriod.withdrawal_reasons.values
+        reasons -= %w[mentor_no_longer_being_mentor] if ect_at_school_period.present?
```
ethax-ross: If we push `MENTOR_ONLY_WITHDRAWAL_REASONS` to the `TrainingPeriod` model I think it would make sense to reference the constant here?

ethax-ross (commented):
Looks good just a few minor suggestions; did you check to see if we have any of these shapes on production that will need fixing before this goes out?

leandroalemao (approved):

PR #2853, opened 2026-05-05
[3903] Remove mentor name link from ECT summary card

avinhurry:
### Context

Some schools have been changing the current mentor's details instead of assigning a new mentor to an ECT. This PR removes the mentor hyperlink from the ECT listing summary card to reduce that confusion.

### Changes

- Remove the mentor name link from the ECT listing summary card
- Keep the assign mentor link when an ECT has no mentor

| Before | After |
|--------|-------|
|image|image|

PR #2857, opened 2026-05-05
[3913] Remove mentor name links from ECT summaries

avinhurry:
### Summary

Following on from #2853, we also want to remove the links to the mentor from the ECT show page.

### Context

Some schools have been changing the current mentor's details instead of assigning a new mentor to an ECT. This PR removes the mentor hyperlink from the ECT show page to reduce that confusion.

| Before | After |
|--------|-------|
|image|image|

### DIFF
--- a/app/helpers/ect_helper.rb
+++ b/app/helpers/ect_helper.rb
@@ -95,7 +95,7 @@ module ECTHelper
   end
 
   # @param ect [ECTAtSchoolPeriod]
-  def ect_mentor_details(ect)
+  def ect_mentor_name_or_assign_mentor_link(ect)
     mentorship = ECTAtSchoolPeriods::Mentorship.new(ect)
 
     return link_to_assign_mentor(ect) if mentorship.current_mentor.blank?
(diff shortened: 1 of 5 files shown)


app/helpers/ect_helper.rb
```diff
@@ -95,13 +95,12 @@ def ect_start_date(ect)
   end
 
   # @param ect [ECTAtSchoolPeriod]
-  def ect_mentor_details(ect, link_to_mentor: true)
+  def ect_mentor_details(ect)
```
thenapking: Now this is never a link to the mentor, is a more descriptive name better? Something like `assign_mentor_or_show_current_mentor` ?

thenapking (approved):
LGTM.  One small optional comment about naming,

app/helpers/ect_helper.rb
avinhurry: Yeah, I `agree ect_mentor_details` was too vague. I've renamed it to `ect_mentor_name_or_assign_mentor_link`.

I was trying to keep the old shape easy to recover (because there's a good chance we will revert this), but that left unnecessary dead code/comments. I've removed that now. The commit history is enough if we need to revert.

PR #2858, opened 2026-05-05
Link directly to swagger page with agreed copy above API heading

craigmdavidson:
### Context

Resolves https://github.com.mcas.ms/DFE-Digital/register-ects-project-board/issues/3677

swagger

From issue acceptance criteria:
* The "Swagger API documentation" nav link should link directly to [Swagger](https://sandbox.register-early-career-teachers.education.gov.uk/api/docs/v3), removing the current [interim page](https://sandbox.register-early-career-teachers.education.gov.uk/api/guidance/swagger-api-documentation) out of the journey.
* The Swagger documentation page should have the same header nav as the rest of the sandbox pages (see mockup screenshot above)
> * The Swagger documentation page should have the agreed copy (above) placed directly above the "Find out more about Swagger". If that's not easy (due to swagger-ui limitations) we can put it above the main header (and maybe tweak the heading to a h1).
* The PR paces the copy _above_ the _"Lead Provider API" title, along with a heading that matches the nav link name. Placing the copy within the API (I.e. below the "Lead Provider API" heading) *is hard*. While it can be done (by effectively embedding this in the API itself, i.e. adding adding a `description` to the info block of the "Lead Provider API" in `swagger.yaml` but at a cost of styling. See screenshot below:

**Embedding the additional copy in the API docs, at a cost of losing the styling**
embed

### Guidance to review 

* Single commit, the old URL has a permanent redirect.


app/views/api/documentation/index.html.erb
```diff
@@ -1 +1,12 @@
+<div class="govuk-width-container">
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <h1 class="govuk-heading-l">Swagger API documentation</h1>
```
ethax-ross, line 4: I wonder if it's possible to change the `Lead Provider API` heading to be a `h2` to avoid having two `h1` tags on a page, which I don't think is ideal accessibility-wise?

ethax-ross (approved):

app/views/api/documentation/index.html.erb
craigmdavidson: As this is generated, it would require javascript manipulation.

PR #2859, opened 2026-05-05
Align admin timeline H1 with the other admin tabs

avinhurry:
### Summary

Align the timeline H1 with the other admin tabs.

| Before | After |
|--------|-------|
|image|image|

### DIFF
--- a/app/views/admin/teachers/timeline/show.html.erb
+++ b/app/views/admin/teachers/timeline/show.html.erb
@@ -1,6 +1,6 @@
 <%
   page_data(
-    title: smart_quotes(teacher_full_name(@teacher) + "'s timeline"),
+    title: teacher_full_name(@teacher),
     caption: teacher_trn(@teacher),
     caption_size: 'm',
     error: false,


peteryates (approved):
From an accessibility point of view I don't think we have this quite right.

The `<h1>` will be the first thing read out by a screen reader, and it should [describe the contents of the page](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/Heading_Elements#avoid_using_multiple_h1_elements_on_one_page).

Here we have 6 sibling pages with the same `<h1>`.

image


I wonder if it should be more like this:

```html
<h1 class="govuk-heading-l">
  Dave Robinson<span class="govuk-visually-hidden">: Timeline</span>
</h1>
```

I'll approve as this is a problem regardless of this one header, but can we raise a backlog ticket with the accessibility label to delve into this a bit deeper?

avinhurry:
> From an accessibility point of view I don't think we have this quite right.
> 
> The `<h1>` will be the first thing read out by a screen reader, and it should [describe the contents of the page](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/Heading_Elements#avoid_using_multiple_h1_elements_on_one_page).
> 
> Here we have 6 sibling pages with the same `<h1>`.
> 
> image
> I wonder if it should be more like this:
> 
> ```
> <h1 class="govuk-heading-l">
>   Dave Robinson<span class="govuk-visually-hidden">: Timeline</span>
> </h1>
> ```
> 
> I'll approve as this is a problem regardless of this one header, but can we raise a backlog ticket with the accessibility label to delve into this a bit deeper?

Yeah thanks @peteryates, you're right. It's slightly lower priority because this is admin facing, but I agree we should address it across the teacher tabs.

I've added a backlog ticket for it here: https://github.com/DFE-Digital/register-early-career-teachers-public/issues/2860

Feel free to tweak or edit it as needed.

PR #2865, opened 2026-05-06
Make contract period ranges inclusive end

peteryates:
### Context

This follows on from #2854 and #2834 and this time makes the ranges on the `contract_periods` table have an inclusive (`[]`) rather than an exlcusive (`[)`) one.

It's more intuitive for dates.

Fixes DFE-Digital/register-ects-project-board#3413

- **Replace contract_periods `[)` range with `[]`**
- **Standardise on inclusive range for contract periods**


thenapking (approved):
Looks good.  Swapping out the methods has worked without any failures, which is great.

cpjmcquillan (approved):

PR #2866, opened 2026-05-06
[3918] Only show ECTs from current mentorship periods

avinhurry:
## Summary

This PR fixes the mentor list showing ECTs from previous mentorship periods.

`current_or_future_ects` now filters through current or future mentorship periods, so an ECT only appears under a mentor when that mentorship period is current or upcoming.

### DIFF
--- a/app/models/mentor_at_school_period.rb
+++ b/app/models/mentor_at_school_period.rb
@@ -6,12 +6,15 @@ class MentorAtSchoolPeriod < ApplicationRecord
   belongs_to :school, inverse_of: :mentor_at_school_periods
   belongs_to :teacher, inverse_of: :mentor_at_school_periods
   has_many :mentorship_periods, inverse_of: :mentor, dependent: :destroy
+  has_many :current_or_future_mentorship_periods,
+           -> { current_or_future },
+           class_name: "MentorshipPeriod"
   has_many :training_periods, inverse_of: :mentor_at_school_period, dependent: :destroy
   has_many :declarations, through: :training_periods
   has_many :events
   has_many :current_or_future_ects,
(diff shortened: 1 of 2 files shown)


app/models/mentor_at_school_period.rb
```diff
@@ -6,12 +6,15 @@ class MentorAtSchoolPeriod < ApplicationRecord
...
   belongs_to :teacher, inverse_of: :mentor_at_school_periods
   has_many :mentorship_periods, inverse_of: :mentor, dependent: :destroy
+  has_many :current_or_future_mentorship_periods,
+           -> { current_or_future },
+           class_name: "MentorshipPeriod"
   has_many :training_periods, inverse_of: :mentor_at_school_period, dependent: :destroy
   has_many :declarations, through: :training_periods
   has_many :events
   has_many :current_or_future_ects,
            -> { current_or_future.induction_not_completed.includes(:teacher) },
```
peteryates, line 16: ECT at school periods are closed off once someone completes induction, is this condition necessary?

peteryates (approved):
Looks good. Left a comment on a scope that looks a bit odd.

cpjmcquillan (approved):

app/models/mentor_at_school_period.rb
avinhurry: Hmm good question. I don't think we need it. If completing induction always closes the ect at school period, then a passed or failed teacher shouldn't have a current/future ect at school anyway so `current_or_future` would already exclude them.

I think it's only defensive against inconsistent data, e.g.

```ruby
teacher.trs_induction_status = "Passed"
ect_at_school_period.finished_on = nil
```
Happy to remove it separately if we think that defensiveness is unnecessary.

PR #2869, opened 2026-05-07
[3612] Add validation rule to ensure `deadline_date` has passed for `payable` and `paid` statements

leandroalemao:
### Context

Ticket: [3612](https://github.com.mcas.ms/DFE-Digital/register-ects-project-board/issues/3612).

We have seed data in a bad shape at the moment, which means we have some payable statements where the deadline_date has not yet passed. This results in new declaration being attached to statements that are not in an open state. This should never be the case in production, but we should add validation rules and make the intention more explicit.

1- Add a validation rule to ensure a statement cannot move to the `payable/paid` state unless the `deadline_date` is in the past.
2- Update the declarations Create service to filter output fee statements to only those that are open;
3- Make sure there are no time-related edge cases around marking statements as payable/the logic here still holds;
4- Make sure our seed data is all generating correctly and not falling foul of the new validation rule.
5- Existing data checked (sandbox/prod): there's no payable or paid statements records with deadline date in the future.

### Guidance to review

Review app

### DIFF
--- a/app/models/statement.rb
+++ b/app/models/statement.rb
@@ -3,6 +3,7 @@ class Statement < ApplicationRecord
 
   VALID_FEE_TYPES = %w[output service].freeze
 
+  # Associations
   belongs_to :contract
   has_many :adjustments
   has_many :payment_declarations, inverse_of: :payment_statement, class_name: "Declaration"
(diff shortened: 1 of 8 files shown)


app/models/statement.rb
```diff
@@ -11,24 +12,45 @@ class Statement < ApplicationRecord
...
 
+  # Validations
   validates :contract, presence: { message: "Contract is required" }
   validates :fee_type,
             presence: { message: "Enter a fee type" },
-            inclusion: { in: VALID_FEE_TYPES, message: "Fee type must be output or service" }
+            inclusion: {
+              in: Statement.fee_types.keys,
+              message: "Fee type must be output or service"
+            }
```
cpjmcquillan: Do we need this validation if we have `validate:` set on the enum? 💭

app/models/statement.rb
```diff
@@ -11,24 +12,45 @@ class Statement < ApplicationRecord
...
   validates :year, numericality: { greater_than_or_equal_to: 2020, is_less_than_or_equal_to: :maximum_year, only_integer: true, message: "Year must be on or after 2020 and on or before #{maximum_year}" }
   validates :api_id, uniqueness: { case_sensitive: false, message: "API id already exists for another statement" }
   validate :unique_lead_provider_month_year
-
+  validates :status,
+            presence: { message: "Enter a status" },
+            inclusion: {
+              in: Statement.statuses.keys,
+              message: "Choose a valid status"
+            }
```
cpjmcquillan: Similarly, do we need this if we're validating the enum too?

app/models/statement.rb
```diff
@@ -81,4 +103,11 @@ def unique_lead_provider_month_year
 
     errors.add(:base, "Statement with the same month and year already exists for this active lead provider")
   end
+
+  def deadline_date_in_the_past
```
ethax-ross, line 97: I think we're missing an explicit test for this/the new error message?

app/services/statements/mark_as_payable.rb
```diff
@@ -5,7 +5,12 @@ class MarkAsPayable
     attr_reader :statement
 
     def self.mark_all_eligible!
-      deadline_past_statements = Statement.where(status: :open).where(deadline_date: ...Date.current)
+      deadline_past_statements = Statements::Search.new(
+        fee_type: "output",
```
ethax-ross: I'm not sure we want to restrict to only output fee types here? I check ECF and I think it grabs all of them?

https://github.com/DFE-Digital/early-careers-framework/blob/main/app/jobs/finance/statements/mark_as_payable.rb#L7

ethax-ross (commented):
Thanks Leandro just a few minor comments/questions

app/models/statement.rb
```diff
@@ -81,4 +93,11 @@ def unique_lead_provider_month_year
...
+    return unless output_fee? && (payable? || paid?)
```
ethax-ross: Do we need to remove the `output_fee?` check here as well? 🤔

ethax-ross (approved):
Looks good, I think we just need to drop the output fee check from the validation as well though

PR #2872, opened 2026-05-07
Remove the beach umbrella from sandbox

peteryates:
It makes the label wrap and doesn't look great. Boo.

image


claire-hughez (approved):
if this breaks the service blame @peteryates

PR #2874, opened 2026-05-07
[3731] Change how we show uplift fee clawbacks

rwrrll:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/3731

Don't count uplift clawbacks as part of billable uplifts, treat them separately instead

- Payment overview
  - Uplifts amount increases because we no longer include clawbacks here
  - Clawbacks amount increases (i.e. more negative) because we now include uplift clawbacks here
  - Same total, other rows unaffected
- Uplift fees
  - Number of participants increases because we no longer subtract clawbacks here (aligns with overview)
- Clawbacks table
  - Now includes a "Uplifts" row if there are uplift clawbacks (aligns with overview)
- All other sections are unchanged

| Before | After |
|--------|--------|
| localhost_3000_admin_finance_statements_650 | localhost_3000_admin_finance_statements_650 (1) | 

### Guidance to review

I've run the product review rake task to make sure there's a statement showing uplift clawbacks on this review app: 
https://cpd-ec2-review-2874-web.test.teacherservices.cloud/admin/finance/statements/1304

When we're ready, we can also deploy this to migration to compare production vs this branch for a known problematic statement

### DIFF
--- a/app/components/admin/statements/clawbacks_component.html.erb
+++ b/app/components/admin/statements/clawbacks_component.html.erb
@@ -13,12 +13,12 @@
         end
 
         table.with_body do |body|
-          declaration_type_outputs_with_clawbacks(calculator).each do |declaration_type_output|
+          rows(calculator).each do |clawback|
             body.with_row do |row|
-              row.with_cell { payment_type(calculator, declaration_type_output) }
-              row.with_cell { declaration_type_output.refundable_count.to_s }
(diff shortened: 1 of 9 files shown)


cpjmcquillan (approved):
Is it worth deploying this to migration so we can review against ECF1 statements?

PR #2876, opened 2026-05-08
Reduce authentication failure level in Sentry

cpjmcquillan:
We're not planning on fixing these issues right now, so it felt sensible to specify the "level" in Sentry as `info` rather than `error`.


avinhurry (approved):
Makes sense

PR #2877, opened 2026-05-08
Preserving previous contract period on re-registration

ericaporter:
### Context
When a mentor is re-registered, new provider-led training periods were being created using the contract period derived from the new registration start date. This could move the mentor from their original contract period into the current one.

- Add `ContractPeriods::ForMentorRegistration` to resolve the contract period for mentor registration
- Preserve the previous provider-led contract period where possible
- Derive the previous contract period from the training period, EOI, school partnership, or schedule
- Pass the resolved contract period into provider-led mentor training period creation
- Add feature and unit coverage for mentor re-registration preserving the original contract period

### Guidance to review
Alongside checking that the CP is preserved on re-registration - also sense check is that mentor behaviour intentionally differs from ECT reassignment logic. Mentors are not moved from closed 2021/2022 periods into 2024 here; instead, we preserve the previous non-frozen provider-led contract period when creating the new training period.

### DIFF
--- a/app/services/contract_periods/for_mentor_registration.rb
+++ b/app/services/contract_periods/for_mentor_registration.rb
@@ -34,9 +34,7 @@ module ContractPeriods
     def previous_contract_period
       @previous_contract_period ||=
         @previous_training_period.contract_period ||
-        @previous_training_period.expression_of_interest&.contract_period ||
-        @previous_training_period.school_partnership&.active_lead_provider&.contract_period ||
-        @previous_training_period.schedule&.contract_period
+        @previous_training_period.expression_of_interest&.contract_period
     end
   end
 end
--- a/app/services/contract_periods/for_mentor_registration.rb
+++ b/app/services/contract_periods/for_mentor_registration.rb
@@ -1,0 +1,42 @@
+module ContractPeriods
+  class ForMentorRegistration
+    # NOTE: unlike ECTs, mentors in 2021/2022 contract periods are not migrated
+    # to 2024 on re-registration. Those mentors have a funding end date that
+    # prevents re-registration, so no reassignment logic is needed here.
+    class NoContractPeriodFoundForStartedOnDate < StandardError; end
+
+    def initialize(started_on:, previous_training_period: nil)
+      @started_on = started_on
+      @previous_training_period = previous_training_period
+    end
+
+    def call
+      return previous_contract_period if preserve_previous_contract_period?
+
+      ContractPeriod.for_registration_start_date(@started_on) ||
+        raise(
+          NoContractPeriodFoundForStartedOnDate,
+          "No contract period found for started_on=#{@started_on}"
+        )
+    end
+
+  private
+
+    def preserve_previous_contract_period?
+      return false unless @previous_training_period
+      return false unless @previous_training_period.provider_led_training_programme?
+      return false unless previous_contract_period
+      return false if previous_contract_period.payments_frozen?
+
+      ContractPeriod.for_registration_start_date(@started_on).present?
+    end
+
+    def previous_contract_period
+      @previous_contract_period ||=
+        @previous_training_period.contract_period ||
+        @previous_training_period.expression_of_interest&.contract_period ||
+        @previous_training_period.school_partnership&.active_lead_provider&.contract_period ||
+        @previous_training_period.schedule&.contract_period
+    end
+  end
+end
--- a/app/services/schools/register_mentor.rb
+++ b/app/services/schools/register_mentor.rb
@@ -84,6 +84,7 @@ module Schools
       earliest_matching_school_partnership if lead_provider.present?
     end
 
+    # Override TrainingPeriodSources default of ContractPeriod.current
     def contract_period
       resolved_contract_period
     end
@@ -91,15 +92,15 @@ module Schools
     def resolved_contract_period
       @resolved_contract_period ||= ContractPeriods::ForMentorRegistration.new(
         started_on: mentor_at_school_period.started_on,
-        previous_training_period:
+        previous_training_period: latest_mentor_training_period
       ).call
     end
 
-    def previous_training_period
-      @previous_training_period ||= TrainingPeriod
-                                      .for_mentor_trn(teacher.trn)
-                                      .order(started_on: :desc)
-                                      .first
+    def latest_mentor_training_period
+      @latest_mentor_training_period ||= TrainingPeriod
+                                           .for_mentor_trn(teacher.trn)
+                                           .order(started_on: :desc)
+                                           .first
     end
 
     def create_teacher!
--- a/app/services/schools/register_mentor.rb
+++ b/app/services/schools/register_mentor.rb
@@ -72,7 +72,8 @@ module Schools
                                                                 school_partnership:,
                                                                 expression_of_interest:,
                                                                 mentee:,
-                                                                author: @author).call
+                                                                author: @author,
+                                                                contract_period: resolved_contract_period).call
     end
 
     def mentor_ineligible_for_funding?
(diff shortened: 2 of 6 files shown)


app/services/contract_periods/for_mentor_registration.rb
```diff
@@ -0,0 +1,42 @@
+module ContractPeriods
+  class ForMentorRegistration
+    # NOTE: unlike ECTs, mentors in 2021/2022 contract periods are not migrated
+    # to 2024 on re-registration. Those mentors have a funding end date that
+    # prevents re-registration, so no reassignment logic is needed here.
+    class NoContractPeriodFoundForStartedOnDate < StandardError; end
```
ericaporter, line 6: I am happy to remove this comment if anyone objects but it's a big difference between how ECT and Mentor CPs are resolved and it's hard to make it explicit from tests alone

app/services/schools/register_mentor.rb
```diff
@@ -83,6 +84,24 @@ def school_partnership
       earliest_matching_school_partnership if lead_provider.present?
     end
 
+    def contract_period
+      resolved_contract_period
+    end
```
cpjmcquillan, line 90: It doesn't look like we're using this anywhere, do we need it?

spec/services/contract_periods/for_mentor_registration_spec.rb
```diff
@@ -0,0 +1,207 @@
...
+          TrainingPeriod,
+          contract_period: frozen_contract,
+          expression_of_interest: nil,
+          school_partnership: nil,
+          schedule: nil,
+          provider_led_training_programme?: true
+        )
+      end
+
+      it "falls back to the registration contract period if the previous period is frozen" do
```
cpjmcquillan: We're capturing the "in a payments frozen contract period" above ⤴️, so I reckon we don't need to repeat it in the test description.

```suggestion
      it "falls back to the registration contract period" do
```

What do you think? 🤔

spec/services/contract_periods/for_mentor_registration_spec.rb
```diff
@@ -0,0 +1,207 @@
...
+          provider_led_training_programme?: true
+        )
+      end
+
+      before { allow(contract_2024).to receive(:payments_frozen?).and_return(true) }
+
+      it "returns the registration contract period" do
+        expect(resolver.call).to eq(contract_2025)
+      end
+    end
```
cpjmcquillan, line 145: Is this any different from the payments frozen test above? 🤔

app/services/contract_periods/for_mentor_registration.rb
```diff
@@ -0,0 +1,42 @@
...
+      ContractPeriod.for_registration_start_date(@started_on).present?
+    end
+
+    def previous_contract_period
+      @previous_contract_period ||=
+        @previous_training_period.contract_period ||
+        @previous_training_period.expression_of_interest&.contract_period ||
+        @previous_training_period.school_partnership&.active_lead_provider&.contract_period ||
+        @previous_training_period.schedule&.contract_period
+    end
```
cpjmcquillan, line 38: What's the difference between `training_period.contract_period` and `training_period.school_partnership.active_lead_provider.contract_period`? 🤔

As far as I can tell, the associations are both defined through the `active_lead_provider`, via the `lead_provider_delivery_partnership` on the `school_partnership`.

They should always resolve to the same thing, unless I'm missing something!

app/services/schools/register_mentor.rb
```diff
@@ -83,6 +84,24 @@ def school_partnership
...
+
+    def resolved_contract_period
+      @resolved_contract_period ||= ContractPeriods::ForMentorRegistration.new(
+        started_on: mentor_at_school_period.started_on,
+        previous_training_period:
+      ).call
+    end
+
+    def previous_training_period
```
cpjmcquillan: Splitting hairs a bit, I wonder if we can add more meaning here by choosing a name that's a bit more specific than "previous"? 💭

Maybe `historic_training_period` or `existing_training_period_for_mentor`?

cpjmcquillan (approved):
Couple minor comments about an unused method and a repeated test.

Also a question about whether we need to traverse the school partnership explicitly.

Other than that I think this is good to go! I think the comment and the PR description are really helpful 👍🏼

app/services/schools/register_mentor.rb
thenapking: As I understood it, this is overriding the definition from `TrainingPeriodSources`.  The other methods in that module use it.

app/services/contract_periods/for_mentor_registration.rb
thenapking: I think we should keep it.  It's very useful context that could be forgotten.

app/services/contract_periods/for_mentor_registration.rb
thenapking: I think that once the training_period is created, it should also have the same contract_period as its schedule.

app/services/schools/register_mentor.rb
thenapking: Perhaps `previous_school_training_period`?

spec/features/schools/mentors/register_mentor/edge_cases/preserves_contract_period_on_re_registration_spec.rb
```diff
@@ -0,0 +1,194 @@
...
+  include_context "test TRS API returns a teacher"
+  include SchoolPartnershipHelpers
+
+  let(:trn) { "3002586" }
+
+  scenario "contract period is preserved when a mentor re-registers at a new school" do
+    given_there_is_a_school_in_the_service
+    and_the_school_is_in_a_partnership_with_a_lead_provider
+    and_there_is_an_ect_with_no_mentor_registered_at_the_school
+    and_there_is_a_mentor_previously_registered_in_an_older_contract_period
```
thenapking: To make it even more explicit could we call this `and_there_is_a_mentor_previously_registered_at_another_school_in_an_older_contract_period`?

thenapking (approved):
LGTM.  I think there's one thing to address around selecting the contract period from the previous training period, and perhaps a few changes on naming.

app/services/schools/register_mentor.rb
cpjmcquillan: Ok 👌🏼 (but scary!) 👻

app/services/contract_periods/for_mentor_registration.rb
ericaporter: Looking at this again - I think `@previous_training_period.school_partnership&.active_lead_provider&.contract_period` is redundant and can be removed alongside `@previous_training_period.schedule&.contract_period` - these were included because of test cases not having the right associations

app/services/schools/register_mentor.rb
ericaporter: We have the same pattern in `RegisterECT#contract_period` where it overides the TPS contract_period. I've added a comment to make it a bit less scary. 

We could add a ticket to the backlog to look at replacing `TrainingPeriodSources#contract_period` - perhaps making the dependency explicit rather than relying on the override pattern...

spec/services/contract_periods/for_mentor_registration_spec.rb
ericaporter: I've removed this one now `def previous_contract_period` has been simplified

app/services/schools/register_mentor.rb
ericaporter: Yeah it's not perfect. 

I think the issue with `historic_training_period` is that it implies it's old/archived, but it could be a very recent period that just happened to finish and `existing_training_period_for_mentor` possibly `existing_training_period` is all we would need...

If the lookup were scoped to school I think `previous_school_training_period` would work but it might be confusing since a mentor can have multiple non-overlapping training periods at the same school.

Looking at the codebase, `latest_mentor_training_period` might be the better option as it better describes what is happening in the method...?

app/services/schools/register_mentor.rb
thenapking: That would be a good idea, because it's used in a lot of places and often overridden now so it isn't quite fit for purpose.

app/services/schools/register_mentor.rb
```diff
@@ -83,6 +84,25 @@ def school_partnership
...
+      @resolved_contract_period ||= ContractPeriods::ForMentorRegistration.new(
+        started_on: mentor_at_school_period.started_on,
+        previous_training_period: latest_mentor_training_period
+      ).call
+    end
+
+    def latest_mentor_training_period
+      @latest_mentor_training_period ||= TrainingPeriod
+                                           .for_mentor_trn(teacher.trn)
+                                           .order(started_on: :desc)
```
thenapking, line 102: There's a scope in `Interval` for that order query which is `latest_first`.  Since you are doing this on teacher you can actually do:
```
teacher.mentor_training_periods.latest_first&.first
```

PR #2878, opened 2026-05-08
Display the training periods in the admin pages from newest to oldest

leoapost:
### Context
Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/3727

The admin teacher training page should show training history from newest to oldest.

  - fetch training periods without grouping so they can be displayed in the correct order
  - identify in the controller which training summaries should display the API row
  - replace the `index` argument in `TrainingSummaryComponent` with an explicit `show_api_row` flag

### DIFF
--- a/app/controllers/admin/teachers/trainings_controller.rb
+++ b/app/controllers/admin/teachers/trainings_controller.rb
@@ -8,7 +8,8 @@ module Admin
         @navigation_items = helpers.admin_teacher_navigation_items(@teacher, :training)
         @breadcrumbs = teacher_breadcrumbs
         @ect_at_school_periods = @teacher.ect_at_school_periods
-        @ect_training_periods = @teacher.ect_training_periods.includes(ect_at_school_period: :teacher).order(started_on: :desc).group_by(&:lead_provider_delivery_partnership)
+        @ect_training_periods = @teacher.ect_training_periods.includes(ect_at_school_period: :teacher).order(started_on: :desc)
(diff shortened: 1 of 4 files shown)


app/controllers/admin/teachers/trainings_controller.rb
```diff
@@ -8,7 +8,8 @@ def show
         @navigation_items = helpers.admin_teacher_navigation_items(@teacher, :training)
         @breadcrumbs = teacher_breadcrumbs
         @ect_at_school_periods = @teacher.ect_at_school_periods
-        @ect_training_periods = @teacher.ect_training_periods.includes(ect_at_school_period: :teacher).order(started_on: :desc).group_by(&:lead_provider_delivery_partnership)
+        @ect_training_periods = @teacher.ect_training_periods.includes(ect_at_school_period: :teacher).order(started_on: :desc)
```
peteryates: We have the scope `#latest_first` (in `Interval`) for this.

leoapost: I've replaced it with the `#latest_first` scope. Thanks.

peteryates (approved):
Nice, all looks sensible to me.

I don't think it's really too important here as it's clear what's going on, but ViewComponent [supports collections](https://viewcomponent.org/guide/collections.html#collection-iteration-context) so instead of passing `show_api_row` around it could use `iteration.first?` (see the example).

PR #2879, opened 2026-05-08
Update Swagger schema to allow joining to be null in transfers.

craigmdavidson:
### Context

Resolves: https://github.com/DFE-Digital/register-early-career-teachers-public/pull/2871

In a transfer joining is optional. However, in Swagger OpenAPI 3.0, a property may be optional and nullable, or optional but not nullable. Instead of omitting the joining property when there is none, we want to return this as null (to be in line with ECF1).

As such we need to update the Swagger schema for transfers to allow joining to be null.

Updates the Swagger schema to allow joining to be null.

### Guidance to review

Simple change

### DIFF
--- a/public/api/docs/v3/swagger.yaml
+++ b/public/api/docs/v3/swagger.yaml
@@ -3129,6 +3129,7 @@ components:
               example: '2021-05-31'
         joining:
           type: object
+          nullable: true
           required:
           - school_urn
           properties:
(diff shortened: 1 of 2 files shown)


cpjmcquillan (approved):

ethax-ross (approved):
Thanks Craig! While I was reviewing this I also noticed we should have `date` in the required list alongside `school_urn` (only `provider` is nullable in `joining` and `leaving`). We can raise this as a separate PR if need be though.

leandroalemao (approved):

PR #2881, opened 2026-05-08
Add validation to ASPs range and extract concern `at_school_period`

ltello:

- Validate the range of ASPs to cover the inner periods (training periods and mentorship periods)
- Extract `at_school_period` concern from `ect_at_school_period` and `mentor_at_school_period`

### DIFF
--- a/app/models/concerns/at_school_period.rb
+++ b/app/models/concerns/at_school_period.rb
@@ -31,7 +31,8 @@ module AtSchoolPeriod
     validates :teacher_id,
               presence: true
 
-    validate :covering_inner_periods
+    validate :covering_training_periods, if: -> { persisted? && training_periods.any? }
+    validate :covering_mentorship_periods, if: -> { persisted? && mentorship_periods.any? }
 
     # Scopes
     scope :for_school, ->(school_id) { where(school_id:) }
@@ -59,20 +60,18 @@ module AtSchoolPeriod
   end
 
   # Validations
-  def covering_inner_periods
-    return if started_on.blank?
-    return unless persisted?
+  def covering_training_periods
+    current_range = (started_on..finished_on)
+    return if training_periods.all? { current_range.cover?(it.range) }
 
-    inner_periods = mentorship_periods + training_periods
-    return if inner_periods.empty?
+    errors.add(:base, "Date range does not cover all the training periods")
+  end
 
-    starts = inner_periods.map(&:started_on)
-    ends = inner_periods.map(&:finished_on)
-    earliest_started_on = starts.min unless starts.any?(&:nil?)
-    latest_finished_on = ends.max unless ends.any?(&:nil?)
-    return if (started_on..finished_on).cover?(earliest_started_on..latest_finished_on)
+  def covering_mentorship_periods
+    current_range = (started_on..finished_on)
+    return if mentorship_periods.all? { current_range.cover?(it.range) }
 
-    errors.add(:base, "Date range does not cover all the inner periods")
+    errors.add(:base, "Date range does not cover all the mentorship periods")
   end
 
   # Methods
--- a/app/models/concerns/at_school_period.rb
+++ b/app/models/concerns/at_school_period.rb
@@ -62,14 +62,14 @@ module AtSchoolPeriod
   # Validations
   def covering_training_periods
     current_range = (started_on..finished_on)
-    return if training_periods.all? { current_range.cover?(it.range) }
+    return if training_periods.all? { current_range.cover?(it.started_on..it.finished_on) }
 
     errors.add(:base, "Date range does not cover all the training periods")
   end
 
   def covering_mentorship_periods
     current_range = (started_on..finished_on)
-    return if mentorship_periods.all? { current_range.cover?(it.range) }
+    return if mentorship_periods.all? { current_range.cover?(it.started_on..it.finished_on) }
 
     errors.add(:base, "Date range does not cover all the mentorship periods")
   end
--- a/app/models/concerns/at_school_period.rb
+++ b/app/models/concerns/at_school_period.rb
@@ -1,0 +1,83 @@
+module AtSchoolPeriod
+  extend ActiveSupport::Concern
+
+  included do
+    # Associations
+    has_many :events
+
+    has_one :current_or_next_training_period, -> { current_or_future.earliest_first }, class_name: "TrainingPeriod"
+    has_one :earliest_training_period, -> { earliest_first }, class_name: "TrainingPeriod"
+    has_one :latest_training_period, -> { latest_first }, class_name: "TrainingPeriod"
+
+    # Callbacks
+    touch -> { teacher }, on_event: %i[create destroy update], when_changing: %i[email], timestamp_attribute: :api_updated_at
+    refresh_metadata -> { school }, on_event: %i[create destroy update]
+    refresh_metadata -> { teacher }, on_event: %i[create destroy update]
+
+    # Validations
+    validates :email,
+              notify_email: true,
+              allow_nil: true
+
+    validates :school_id,
+              presence: true
+
+    validates :started_on,
+              presence: true
+
+    validates :teacher_id,
+              presence: true
+
+    validate :covering_inner_periods
+
+    # Scopes
+    scope :for_school, ->(school_id) { where(school_id:) }
+    scope :for_teacher, ->(teacher_id) { where(teacher_id:) }
+    scope :with_school, -> { includes(school: :gias_school) }
+    scope :with_teacher, -> { includes(:teacher) }
+
+    scope :with_partnerships_for_contract_period, ->(year) {
+      joins(training_periods: {
+        active_lead_provider: :contract_period
+      }).where(contract_periods: { year: })
+    }
+
+    scope :with_expressions_of_interest_for_contract_period, ->(year) {
+      joins(training_periods: {
+        expression_of_interest: :contract_period
+      })
+        .where(contract_periods: { year: })
+    }
+
+    scope :with_expressions_of_interest_for_lead_provider_and_contract_period, ->(year, lead_provider_id) {
+      with_expressions_of_interest_for_contract_period(year)
+        .where(expression_of_interest: { lead_provider_id: })
+    }
+  end
+
+  # Validations
+  def covering_inner_periods
+    inner_periods = mentorship_periods + training_periods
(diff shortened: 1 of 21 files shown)


tonyheadford (approved):
Looks sensible, as suggested would be good to get a review from a schools dev

app/models/concerns/at_school_period.rb
```diff
@@ -0,0 +1,89 @@
...
+    inner_periods = mentorship_periods + training_periods
+    return if inner_periods.empty?
+
+    starts = inner_periods.map(&:started_on)
+    ends = inner_periods.map(&:finished_on)
+    earliest_started_on = starts.min unless starts.any?(&:nil?)
+    latest_finished_on = ends.max unless ends.any?(&:nil?)
+    return if (started_on..finished_on).cover?(earliest_started_on..latest_finished_on)
+
+    errors.add(:base, "Date range does not cover all the inner periods")
```
peteryates: I think the error message is a bit vague here, not knowing whether there's an offending training or mentorship period would mean there's some detective work needed to identify the problem. Worth splitting this into `covering_training_periods` and `covering_mentorship_periods`? The logic could still of couse be shared in a private method.

app/models/concerns/at_school_period.rb
```diff
@@ -0,0 +1,89 @@
...
+    return unless persisted?
+
+    inner_periods = mentorship_periods + training_periods
+    return if inner_periods.empty?
+
+    starts = inner_periods.map(&:started_on)
+    ends = inner_periods.map(&:finished_on)
+    earliest_started_on = starts.min unless starts.any?(&:nil?)
+    latest_finished_on = ends.max unless ends.any?(&:nil?)
+    return if (started_on..finished_on).cover?(earliest_started_on..latest_finished_on)
```
peteryates: Would it be easier to just check all of them, rather than pull out the starts/ends?

```ruby
training_periods.all? { range.cover?(it.range) }
```

cpjmcquillan: I agree this would be a lot clearer 👍🏼

I might be tempted to be slightly more explicit though just so it's more obvious which range is which.

```ruby
training_periods.all? do |training_period|
  range.cover?(training_period.range)
end
```

app/services/mentor_at_school_periods/finish.rb
```diff
@@ -31,7 +31,7 @@ def finish_periods_at_reported_school!
   def finish!(period)
     finish_mentorship_periods!(period)
     finish_training_periods!(period)
-    finish_mentor_at_school_period!(period)
+    finish_mentor_at_school_period!(period.reload)
```
cpjmcquillan: I don't think it's obvious _why_ we need to reload the `MentorAtSchoolPeriod` and the `ECTAtSchoolPeriod` here and in `ECTAtSchoolPeriods::Finish`.

Do you think it's worth breaking this change into a separate commit and giving some more context in the commit message, or perhaps adding a comment to the code? 🤔

app/models/concerns/at_school_period.rb
ltello: I can try and check all of them but if the at school period's `started_on` or `finished_on` dates are changing, `range` won't be updated yet before saving at validation time because it's a Postgress generated column, so the comparisons would not be correct and the validation result wouldn't be reliable.

peteryates: Ah yeah, that's a fair point.

So instead of `range`, maybe set a `new_range` variable?

app/models/concerns/at_school_period.rb
```diff
@@ -0,0 +1,88 @@
...
+    validates :school_id,
+              presence: true
+
+    validates :started_on,
+              presence: true
+
+    validates :teacher_id,
+              presence: true
+
+    validate :covering_training_periods, if: -> { persisted? && training_periods.any? }
```
avinhurry, line 34: This new parent side validation is useful, but I think the invariant is still not fully protected from the child side.

[TrainingPeriod#enveloped_by_trainee_at_school_period](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/models/training_period.rb#L209-L210) currently skips the containment validation when `finished_on` is blank, so an ongoing `TrainingPeriod` can still be created or updated outside a finished ECT/Mentor `at_school_period`.

This validation would catch that only when the parent ASP is saved, not when `TrainingPeriod` itself is saved.

Should this PR also update the `TrainingPeriod` validation so ongoing training is only valid when the parent ASP is ongoing?

app/services/mentor_at_school_periods/finish.rb
```diff
@@ -31,7 +31,7 @@ def finish_periods_at_reported_school!
   def finish!(period)
     finish_mentorship_periods!(period)
     finish_training_periods!(period)
```
avinhurry, line 33: This only closes training periods that are ongoing on `finished_on`, but it doesn't handle future/unstarted training periods.

With the new `AtSchoolPeriod` validation, a future mentor training period would sit outside the newly finished `mentor_at_school_period` range and could prevent `finish_mentor_at_school_period!` from saving.

Should we mirror [ECTAtSchoolPeriods::Finish#destroy_unstarted_training_periods!](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/services/ect_at_school_periods/finish.rb#L84-L89) here and delete training periods that start on or after `finished_on` before closing the mentor ASP?

app/services/mentor_at_school_periods/finish.rb
ltello: Commit with comments added to highlight why reloading the at school period on `Finish` SO needs to happen for validation purposes.

app/models/concerns/at_school_period.rb
ltello: exactly

app/models/concerns/at_school_period.rb
ltello: Validation split into 2 more explicit ones.

avinhurry (commented):
The intent looks good, and the concern extraction makes sense. 

I've left a couple of comments on edge cases that may be worth tightening while introducing the new ASP range invariant. 

Hope it's useful!

app/services/ect_at_school_periods/finish.rb
```diff
@@ -12,9 +12,14 @@ def initialize(ect_at_school_period:, finished_on:, author:, reported_by_school_
 
     def finish!
       ActiveRecord::Base.transaction do
-        finish_ect_at_school_period!
         finish_mentorship_periods!
         finish_training_period!
```
avinhurry, line 17: Moving `finish_training_period!` before `finish_ect_at_school_period!` changes which training period gets attached to the `teacher_left_school_as_ect` event.

`event_params[:training_period]` is resolved through `current_or_next_training_period`, but after `finish_training_period!` runs, a leave date of today or in the past means the previously active training period is no longer current/future. The transaction doesn't help with this.

Should we snapshot the event training period before closing the inner training period, or pass the finished training period through explicitly?

app/services/mentor_at_school_periods/finish.rb
ltello: I think this is out of the scope of this PR. Let's discuss if we create a new one to fix this issue.

app/models/concerns/at_school_period.rb
ltello: I think this is out of the scope of this PR. Let's discuss if we create a new one to fix this issue.

app/services/ect_at_school_periods/finish.rb
ltello: I think you are right Avin. I have added a new commit to keep using the right training period for the event.

app/services/mentor_at_school_periods/finish.rb
ltello: Added a commit to destroy unstarted training periods for each `mentor_at_school_period` finished

avinhurry: Thanks for adding this. This matches the ECT finish behaviour and should avoid the mentor leaving flow failing on the new ASP validation.

app/models/concerns/at_school_period.rb
avinhurry: That makes sense. Happy for this one to be handled as a follow up, especially now the mentor finish issue has been covered in this PR.

I think the rule we want is probably: ongoing training is valid only when the parent ASP is also ongoing.

I can pick up the follow up if helpful, let me know.

app/services/mentor_at_school_periods/finish.rb
```diff
@@ -31,7 +31,12 @@ def finish_periods_at_reported_school!
...
-    finish_mentor_at_school_period!(period)
+
+    # period validation depends on inner periods.
+    # We need it to get updated with the latest inner period changes happening above.
+    period.reload
+
+    finish_mentor_at_school_period!(period.reload)
```
avinhurry: Non blocking nit: this reloads `period` twice now. 

Could just reload once before passing it into `finish_mentor_at_school_period!`?

avinhurry (approved):
This looks good to go to me.

Thanks for addressing the mentor finish and ECT event cases. Happy for the remaining child side `TrainingPeriod` validation to be handled as a follow up.

Nice work!

app/services/ect_at_school_periods/finish.rb
```diff
@@ -8,18 +8,26 @@ def initialize(ect_at_school_period:, finished_on:, author:, reported_by_school_
       @author = author
       @reported_by_school_id = reported_by_school_id
       @record_event = record_event
+      @training_period = ect_at_school_period.current_or_next_training_period if unstarted_training_periods.none?
```
peteryates, line 11: Are `unstarted_training_periods` guaranteed to not be started? They're selected using the `finished_on` rather than `Date.current`, so it seems like there's a possibility that if we finish someone in the past, the method could return records that are yet to start.

peteryates (commented):
Looks good to me, left a question on `unstarted_training_periods`.

app/services/ect_at_school_periods/finish.rb
ltello: Well, I'm just reusing the logic already in place in this service.
`finished_on` is used because one of the consumers of this service is the school journey to tell us an ECT is leaving. That leaving date can be other than the current day. 

Similarly the ECT registration journey can set `finished_on` to a date other than today.

PR #2882, opened 2026-05-08
[3377] Add validation rule to prevent mentors being moved to reduced schedules

leandroalemao:
### Context

Ticket: [3377](https://github.com.mcas.ms/DFE-Digital/register-ects-project-board/issues/3377)

Mentors should not be placed on reduced schedules in any cohort. There are a single digit number of mentors on reduced schedules currently. We should prevent this in future.

- Add validation rule to prevent mentors being moved to reduced schedules;
- Update the seeds accordingly;
- Add script to fix bad existing data;

### Guidance to review

[Review app](https://cpd-ec2-review-2882-web.test.teacherservices.cloud/admin)

### DIFF
--- a/app/models/schedule.rb
+++ b/app/models/schedule.rb
@@ -1,26 +1,23 @@
 class Schedule < ApplicationRecord
-  REPLACEMENT_SCHEDULE_IDENTIFIERS = %w[
-    ecf-replacement-april
-    ecf-replacement-january
-    ecf-replacement-september
-  ].freeze
-
   enum :identifier,
-       {
-         "ecf-extended-april" => "ecf-extended-april",
-         "ecf-extended-january" => "ecf-extended-january",
-         "ecf-extended-september" => "ecf-extended-september",
-         "ecf-reduced-april" => "ecf-reduced-april",
-         "ecf-reduced-january" => "ecf-reduced-january",
-         "ecf-reduced-september" => "ecf-reduced-september",
-         "ecf-replacement-april" => "ecf-replacement-april",
-         "ecf-replacement-january" => "ecf-replacement-january",
-         "ecf-replacement-september" => "ecf-replacement-september",
-         "ecf-standard-april" => "ecf-standard-april",
-         "ecf-standard-january" => "ecf-standard-january",
-         "ecf-standard-september" => "ecf-standard-september"
-       },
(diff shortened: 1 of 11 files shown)


peterdavidhamilton (approved):
Looks good, I'm certainly happy to ignore SonarCloud for the time being. It feels like any time we attempt to treat ECTs and mentors the same we ultimately need to uncouple them again.

app/models/schedule.rb
```diff
@@ -5,6 +5,12 @@ class Schedule < ApplicationRecord
     ecf-replacement-september
   ].freeze
 
+  REDUCED_SCHEDULE_IDENTIFIERS = %w[
```
ethax-ross: I wonder if its worth making this less explicit in case we add a new reduced schedule in the future; I expect checking it includes `reduced` in the identifier would be enough? 🤔

db/scripts/fix_mentors_training_period_schedules.rb
```diff
@@ -0,0 +1,20 @@
+# This script finds training periods for mentors that are on a reduced schedule and updates them to the corresponding standard schedule.
+# This is necessary because mentors should not be on reduced schedules, which are only applicable for ECTs.
+
+REDUCED_SCHEDULE_IDENTIFIERS = {
```
ethax-ross: `SCHEDULE_MAPPINGS`?

db/scripts/fix_mentors_training_period_schedules.rb
```diff
@@ -0,0 +1,20 @@
...
+  "ecf-reduced-april": "ecf-standard-april",
+  "ecf-reduced-january": "ecf-standard-january",
+  "ecf-reduced-september": "ecf-standard-september"
+}.freeze
+
+TrainingPeriod.where.not(mentor_at_school_period_id: nil).where(schedule_id: Schedule.where(identifier: REDUCED_SCHEDULE_IDENTIFIERS.keys).select(:id)).find_each do |training_period|
```
ethax-ross: Do we want to spit out the IDs of the teachers this will change so we can keep a record?

ethax-ross (approved):
Looks good Leandro 👍 couple of minor comments but very much take it or leave it

PR #2888, opened 2026-05-11
[3947] Change AB on school view

thenapking:
### Context
Currently, an ECT's appropriate body is recorded through the appropriate body period, which has a name attribute.  This is due to be refactored.  Until then a mismatch can exist between an ECT's induction period and the appropriate body period.  Therefore, if an ECT has an induction period, we want to display the AB associated with that in the school summary page of the admin console, and the ECT induction summary for the schools.  If the ECT has more than one induction period, we take the details from the latest period.

If the teacher is both a Mentor and an ECT these details do not show up in components relative to the `mentor_at_school_period` so no changes are needed.

As I understand it there is or will be a ticket in the backlog to provide a more thorough fix to this and potentially clear up the issues with AB's name being recorded on the period etc etc.  This PR is to fix the presentation of the data as it currently stands, in order to stop confusion for users.

Admin view
image

School view
image

### Guidance to review
Take an ECT with no induction periods and create an induction period with a different AB from the one reported.  You should see this present on the above screen now.  Find an ECT who has not had the AB reported.  Create a new induction period.  You should see the AB from the induction period is present.

### DIFF
--- a/app/services/teachers/induction.rb
+++ b/app/services/teachers/induction.rb
@@ -11,6 +11,10 @@ module Teachers
       teacher.ongoing_induction_period
     end
 
+    def latest_induction_period
+      induction_periods.latest_first&.first
+    end
+
     def past_induction_periods
       @past_induction_periods ||= induction_periods.finished.latest_first
     end
--- a/app/services/teachers/induction.rb
+++ b/app/services/teachers/induction.rb
@@ -12,7 +12,7 @@ module Teachers
     end
 
     def latest_induction_period
-      induction_periods.latest_first&.first
+      @latest_induction_period ||= induction_periods.latest_first&.first
     end
 
     def past_induction_periods
--- a/spec/components/schools/ect_induction_details_component_spec.rb
+++ b/spec/components/schools/ect_induction_details_component_spec.rb
@@ -8,7 +8,12 @@ RSpec.describe Schools::ECTInductionDetailsComponent, type: :component do
                       started_on: Date.new(2023, 9, 1))
   end
 
+  let(:latest_induction_period) { nil }
+  let(:past_induction_period) { nil }
+
   before do
+    past_induction_period
+    latest_induction_period
     render_inline(described_class.new(ect))
   end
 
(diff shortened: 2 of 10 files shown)


app/services/teachers/induction.rb
```diff
@@ -11,6 +11,10 @@ def current_induction_period
       teacher.ongoing_induction_period
     end
 
+    def latest_induction_period
+      induction_periods.latest_first&.first
+    end
```
cpjmcquillan, line 16: I haven't checked, but I imagine this will lead to some N+1 queries.

Perhaps we should define a relationship between the teacher and their latest induction period, that way we can `.include` it and get ahead of these before they become a problem.

```ruby
has_one :latest_induction_period, -> { latest_first }, class_name: "InductionPeriod"
```

We do something similar for ongoing periods [here](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/197f8b68627ee7eb6cc20af3be4e9df502d661fd/app/models/teacher.rb#L38).

spec/components/schools/ect_induction_details_component_spec.rb
```diff
@@ -45,30 +50,77 @@
...
+      expect(page).to have_selector(".govuk-summary-list__value", text: "Current AB")
+      expect(page).not_to have_selector(".govuk-summary-list__value", text: "Alpha Teaching School Hub")
+      expect(page).not_to have_selector(".govuk-summary-list__value", text: "Not reported")
     end
 
-    it "renders the induction start date fallback" do
+    it "renders the induction start date from the induction period, with a hint" do
       expect(page).to have_selector(".govuk-summary-list__key", text: "Induction start date")
-      expect(page).to have_selector(".govuk-summary-list__value", text: "Yet to be reported by the appropriate body")
+      expect(page).to have_selector(".govuk-summary-list__value", text: "1 September 2023This has been reported by an appropriate body")
```
cpjmcquillan, line 97: This is a bit awkward to read. And the test makes it seem like the markup is weird.

I wonder whether we could be a bit more specific and assert against the two elements, so we don't have something like `September 2023This` in the test? 🤔

cpjmcquillan (approved):

spec/components/schools/ect_induction_details_component_spec.rb
thenapking: Yeah, that's a fair point. The text overflows, so the markup isn't weird, but I'll see if I can make that more readable.

PR #2889, opened 2026-05-11
Make training period ranges have an inclusive end

peteryates:
### Context

This follows on from #2357, #2834 and #2854,  where we convert ranges from `[)` ranges to `[]`.

Here we're doing `training_periods`.

The `[)` range allows a new period to begin on the day the previous one finishes.

This can cause a bit of confusion, because the ranges `2025-01-03..2025-02-03` and `2025-02-03..2025-03-03` can coexist, but where was the trainee on `2025-02-03`?

Using `[]` means the later period must start on the day following the closure of the previous one at the earliest.

### What's changed

The main changes here are:

* updating the functionality so when we open a 'replacement' period today, the previous one finishes yesterday
* fixing the specs so they're not building test scenarios with 'overlapping' periods

### DIFF
--- a/app/services/declarations/mentor_completion.rb
+++ b/app/services/declarations/mentor_completion.rb
@@ -37,7 +37,7 @@ module Declarations
 
       TrainingPeriods::Create.provider_led(
         period:,
-        started_on: latest_training_period.finished_on,
+        started_on: latest_training_period.finished_on + 1.day,
         finished_on: period.finished_on,
         school_partnership: latest_training_period.school_partnership,
         expression_of_interest: nil,
--- a/app/services/api_seed_data/ect_participant_action_scenarios.rb
+++ b/app/services/api_seed_data/ect_participant_action_scenarios.rb
@@ -66,7 +66,7 @@ module APISeedData
 
         # Active training period at the same school but with a different lead provider
         school_partnership = school_partnerships(excluding_active_lead_provider: active_lead_provider, year: contract_period.year, school:).sample
-        training_time_period = { started_on: training_period_at_first_school.finished_on, finished_on: nil }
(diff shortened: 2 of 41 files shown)


app/services/declarations/mentor_completion.rb
```diff
@@ -37,7 +37,7 @@ def create_training_period!
 
       TrainingPeriods::Create.provider_led(
         period:,
-        started_on: latest_training_period.finished_on,
+        started_on: latest_training_period.finished_on + 1.day,
```
peterdavidhamilton, line 40: Have we weighed the differeces between `+ 1.day`, `advance(days: 1)`, `next_day`? Thinking about month end behaviour or leap years for example.

peteryates: TIL about `#next_day`. I assume it's an alias for `#tomorrow`, which I opted not to use because the word is relative to today, which is a bit confusing.

I like `#next_day` though - but I think adding `1.day` to a date will work in the same way.

peteryates: Let's get this merged and then if we want to switch `+ 1.day` to `#next_day` we'll do it as a follow up.

avinhurry (commented):
I may have missed this in another PR, but have we covered changing the `training_periods` [check constraint](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/db/schema.rb#L924) from `finished_on > started_on` to `finished_on >= started_on` somewhere?

This PR can create a one day training period when an existing period started yesterday and is replaced today, so I think the DB constraint would need to allow that too.

cpjmcquillan (approved):

PR #2891, opened 2026-05-11
[3952] Scope unfunded mentor email to the requesting lead provider - Part 1

mooktakim:
### Context

Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/3952

Split into 3 PR's.

PR 1:
- Migration to add nullable latest_mentor_at_school_period_id + index.
- belongs_to :latest_mentor_at_school_period on Metadata::TeacherLeadProvider (+ spec).
- Handler stamps the new column alongside api_mentor_id (+ spec assertions in the existing populated and nil contexts).
- Fixed the N+1 you flagged on handlers/teacher.rb:33 by extending the existing includes chain to preload ect_at_school_period.latest_mentorship_period.mentor.teacher.

PR 2: https://github.com/DFE-Digital/register-early-career-teachers-public/pull/2912
- switch UnfundedMentorSerializer + UnfundedMentors::Query to read via the new field; add has_many :mentor_lead_provider_metadata on Teacher (renamed per your suggestion); restore the cross-LP behaviour tests; update RemoveTeacher to skip associations with a non-default primary_key.

PR 3: https://github.com/DFE-Digital/register-early-career-teachers-public/pull/2914
- drop api_mentor_id, derive everywhere from latest_mentor_at_school_period.teacher.api_id.

- `GET /api/v3/unfunded-mentors` always returns the latest mentor at school period, ignoring the calling lead provider.
- Now the serializer returns the latest mentor at school period filter by the calling lead provider, with fallback to the latest.


app/serializers/api/teachers/unfunded_mentor_serializer.rb
```diff
@@ -3,8 +3,16 @@ class AttributesSerializer < Blueprinter::Base
...
+    field(:email) do |teacher, options|
+      lead_provider_id = options[:lead_provider_id]
+
+      matching_periods = teacher.mentor_at_school_periods.select do |masp|
+        masp.mentorship_periods.any? do |msp|
+          msp.mentee.training_periods.any? { |tp| tp.active_lead_provider&.lead_provider_id == lead_provider_id }
+        end
+      end
+
+      (matching_periods.max_by(&:started_on) || teacher.latest_mentor_at_school_period).email
```
ethax-ross: Could this ever be `nil` given how we populate `api_mentor_id`?

https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/services/metadata/handlers/teacher.rb#L30-L32

app/services/api/teachers/unfunded_mentors/query.rb
```diff
@@ -38,7 +38,10 @@ def preload_associations(results)
       results
         .strict_loading
         .includes(
-          :latest_mentor_at_school_period
+          :latest_mentor_at_school_period,
+          mentor_at_school_periods: {
+            mentorship_periods: { mentee: { training_periods: :active_lead_provider } }
```
ethax-ross: Can we do a few tests against snapshot to see what impact this has on the response time? if it has a big impact I think it would be a relatively small change to stamp the `latest_menor_at_school_period_id` instead of the `api_mentor_id` in the handler, then we could get both via that without too much joining?

```
metadata.latest_mentor_at_school_period.teacher.api_id
metadata.latest_mentor_at_school_period.mentor.email
```

spec/requests/api/v3/unfunded_mentors_spec.rb
```diff
@@ -75,4 +75,63 @@ def transform_sort_attribute(sort_attribute)
     it_behaves_like "a does not filter by updated_since endpoint"
     it_behaves_like "a N+1 queries free endpoint", :get
   end
+
+  describe "email scoped to the requesting lead provider" do
```
ethax-ross: I think we should rely on the shared contexts here rather than introducing a one-off request spec/associated setup 🤔 if we want to go this way I’d prefer we make that a deliberate, consistent change across the suite rather than introducing bits here and there?

ethax-ross (approved):
Thanks Mook, just left a few thoughts

app/services/api/teachers/unfunded_mentors/query.rb
mooktakim: Index becomes ~4.6x slower. Will stamp on metadata

ethax-ross: 😞 ok, I think we'll need to coordinate it in a few PRs/deploys to avoid errors:

- Add the new metadata field, let the nightly task populate it
- Switch over the application logic to reference the new field
- Drop the old field from the metadata now its no longer referenced

app/models/metadata/teacher_lead_provider.rb
```diff
@@ -11,6 +11,7 @@ class TeacherLeadProvider < Metadata::Base
     belongs_to :latest_mentor_training_period, optional: true, class_name: "TrainingPeriod"
     belongs_to :latest_ect_contract_period, optional: true, class_name: "ContractPeriod", foreign_key: :latest_ect_contract_period_year
     belongs_to :latest_mentor_contract_period, optional: true, class_name: "ContractPeriod", foreign_key: :latest_mentor_contract_period_year
+    belongs_to :latest_mentor_at_school_period, optional: true, class_name: "MentorAtSchoolPeriod"
```
ethax-ross: Can we add a spec for this

app/services/metadata/handlers/teacher.rb
```diff
@@ -27,9 +27,10 @@ def upsert_lead_provider_metadata!
         latest_mentor_training_period = latest_mentor_training_period_by_lead_provider(teacher:)[lead_provider_id]
         latest_ect_contract_period = latest_ect_training_period&.contract_period
         latest_mentor_contract_period = latest_mentor_training_period&.contract_period
-        api_mentor_id = if latest_ect_training_period&.at_school_period&.latest_mentorship_period&.ongoing_today?
-                          latest_ect_training_period.at_school_period.latest_mentorship_period.mentor&.teacher&.api_id
-                        end
+        ongoing_latest_mentorship_period = latest_ect_training_period&.at_school_period&.latest_mentorship_period
+        ongoing_latest_mentorship_period = nil unless ongoing_latest_mentorship_period&.ongoing_today?
+        latest_mentor_at_school_period = ongoing_latest_mentorship_period&.mentor
+        api_mentor_id = latest_mentor_at_school_period&.teacher&.api_id
```
ethax-ross: Do we need to add some extra `includes` here to avoid N+1s?

https://github.com/DFE-Digital/register-early-career-teachers-public/blob/bb045b55e456f49c93aedd40ce10463b2b3f14a5/app/services/metadata/handlers/teacher.rb#L61C26-L61C62

app/services/metadata/handlers/teacher.rb
```diff
@@ -40,6 +41,7 @@ def upsert_lead_provider_metadata!
           latest_mentor_training_period_id: latest_mentor_training_period&.id,
           latest_ect_contract_period_year: latest_ect_contract_period&.year,
           latest_mentor_contract_period_year: latest_mentor_contract_period&.year,
+          latest_mentor_at_school_period_id: latest_mentor_at_school_period&.id,
```
ethax-ross: Can we cover this with a spec?

lib/remove_teacher.rb
```diff
@@ -25,7 +25,9 @@ def call
 
   # @return [Array<Symbol>]
   def has_many_associations
-    Teacher.reflect_on_all_associations(:has_many).map(&:name).sort
+    Teacher.reflect_on_all_associations(:has_many)
+      .reject { |r| r.options[:primary_key].present? }
```
ethax-ross: Why did this need to change? 🤔

app/models/teacher.rb
```diff
@@ -27,6 +27,11 @@ class Teacher < ApplicationRecord
   has_many :induction_extensions, inverse_of: :teacher
   has_many :teacher_id_changes, inverse_of: :teacher, dependent: :destroy
   has_many :lead_provider_metadata, class_name: "Metadata::TeacherLeadProvider", dependent: :destroy
+  has_many :mentored_lead_provider_metadata,
```
ethax-ross: `mentor_lead_provider_metadata`?

app/serializers/api/teachers/unfunded_mentor_serializer.rb
```diff
@@ -3,8 +3,14 @@ class AttributesSerializer < Blueprinter::Base
...
 
     field(:full_name) { |teacher| Teachers::Name.new(teacher).full_name }
-    field(:email) do |teacher, _options|
-      teacher.latest_mentor_at_school_period.email
+    field(:email) do |teacher, options|
+      lead_provider_id = options[:lead_provider_id]
+
+      teacher.mentored_lead_provider_metadata
+        .select { |m| m.lead_provider_id == lead_provider_id }
+        .max_by { |m| m.latest_mentor_at_school_period.started_on }
```
ethax-ross: Do we need this `max_by`? also we need to coordinate this change a bit, otherwise it will be `nil` until the metadata refreshes in prod. So we need to ship the change to add the new field, let the metadata refresh and then update the serializer? Shall we drop `api_mentor_id` as well since we can get it via the `latest_mentor_at_school_period.teacher.api_id`?

lib/remove_teacher.rb
mooktakim: this was to avoid deleting records that didn't belong to that Teacher.
`mentored_lead_provider_metadata` references the mentees training record, so this was a way to avoid deleting the record that didn't belong to as we set `primary_key`

app/serializers/api/teachers/unfunded_mentor_serializer.rb
mooktakim: Will break this up into 3 PRs

mooktakim: `max_by` is more to double check, as we're loading all `mentored_lead_provider_metadata` and they might not be in order.

app/services/metadata/handlers/teacher.rb
ethax-ross: I think we need a clearer name for what this field is; looking at the surrounding fields you would expect this to be the teacher's own latest `MentorAtSchoolPeriod` (similar to how we have `latest_mentor_training_period`) but really its the teacher's mentor's latest `MentorAtSchoolPeriod` 🤯 

`ongoing_mentorship_latest_mentor_at_school_period`
`assigned_mentor_latest_school_period`
`ect_assigned_mentor_latest_school_period`

Open to suggestions 😬

app/services/metadata/handlers/teacher.rb
```diff
@@ -27,9 +27,10 @@ def upsert_lead_provider_metadata!
         latest_mentor_training_period = latest_mentor_training_period_by_lead_provider(teacher:)[lead_provider_id]
         latest_ect_contract_period = latest_ect_training_period&.contract_period
         latest_mentor_contract_period = latest_mentor_training_period&.contract_period
-        api_mentor_id = if latest_ect_training_period&.at_school_period&.latest_mentorship_period&.ongoing_today?
-                          latest_ect_training_period.at_school_period.latest_mentorship_period.mentor&.teacher&.api_id
-                        end
+        ongoing_latest_mentorship_period = latest_ect_training_period&.at_school_period&.latest_mentorship_period
```
ethax-ross: Not sure if its any more readable, take it or leave it:

```
latest_mentorship_period = latest_ect_training_period&.at_school_period&.latest_mentorship_period
latest_mentor_at_school_period = latest_mentorship_period.mentor if latest_mentorship_period&.ongoing_today?
api_mentor_id = latest_mentor_at_school_period&.teacher&.api_id
```

ethax-ross (commented):
Looks good, I think we just need to think about the naming of that field

app/services/metadata/handlers/teacher.rb
mooktakim: Went with `ect_assigned_mentor_latest_school_period` 👍

ethax-ross (approved):

PR #2893, opened 2026-05-11
Admin users can search by training record IDs

cpjmcquillan:
Now we're live, there has been feedback from Users that it would be useful if they could search for Teachers by their training record ID too.

Currently we only search by participant ID, but this _casts_ that net a bit more widely.

Now, we partially match against a teacher's `api_ect_training_record_id` and their `api_mentor_training_record_id` too.

### DIFF
--- a/app/services/admin/teachers/search.rb
+++ b/app/services/admin/teachers/search.rb
@@ -40,10 +40,16 @@ module Admin
         trns = query_string.scan(/\b\d{7}\b/)
         return scope.where(trn: trns) if trns.any?
 
-        full_text_scope = full_text_matches(scope)
-        return full_text_scope unless api_participant_id_query?
+        api_ids = query_string.scan(/\b[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}\b/i)
+        return api_id_matches(scope, api_ids) if api_ids.any?
 
-        full_text_scope.or(api_participant_id_matches(scope))
+        full_text_matches(scope)
+      end
+
+      def api_id_matches(scope, ids)
+        scope.where(api_id: ids)
+          .or(scope.where(api_ect_training_record_id: ids))
+          .or(scope.where(api_mentor_training_record_id: ids))
       end
 
       def full_text_matches(scope)
@@ -57,16 +63,6 @@ module Admin
 
       def normalized_full_text_query
         @normalized_full_text_query ||= query_string.scan(/[[:alnum:]]+/).join(" ")
-      end
-
-      def api_participant_id_query?
-        query_string.match?(/\A(?=.*[\d-])[a-f0-9-]{8,}\z/i)
-      end
-
-      def api_participant_id_matches(scope)
-        escaped_query = ActiveRecord::Base.sanitize_sql_like(query_string)
-
-        scope.where("CAST(teachers.api_id AS text) ILIKE ?", "%#{escaped_query}%")
       end
 
       def filter_teachers_by_role(scope)
--- a/app/views/admin/teachers/index.html.erb
+++ b/app/views/admin/teachers/index.html.erb
@@ -5,7 +5,7 @@
         :q,
         value: params[:q],
         label: { text: "Search for teacher", size: "s" },
-        hint: { text: "Name, TRN or API participant ID" },
+        hint: { text: "Name, TRN, API participant ID, or API training record ID" },
         form_group: { class: "app-teachers-filters__search" }
       ) %>
 
@@ -65,4 +65,3 @@
 <%= govuk_button_link_to("Import a record from TRS", new_admin_import_ect_find_path, secondary: true) %>
 
 <%= render Shared::PaginationSummaryComponent.new(pagy: @pagy, record_name: "teachers") %>
-
--- a/spec/services/admin/teachers/search_spec.rb
+++ b/spec/services/admin/teachers/search_spec.rb
@@ -1,17 +1,18 @@
 RSpec.describe Admin::Teachers::Search do
-  subject { described_class.new(query_string:, role:, contract_period:) }
+  subject(:search) { described_class.new(query_string:, role:, contract_period:) }
 
   let(:query_string) { nil }
   let(:role) { nil }
   let(:contract_period) { nil }
 
   describe "#teacher_scope" do
+    subject(:teacher_scope) { search.teacher_scope }
+
     context "when the query is blank" do
       let!(:teacher) { FactoryBot.create(:teacher) }
+      let!(:other_teacher) { FactoryBot.create(:teacher) }
 
-      it "returns all teachers" do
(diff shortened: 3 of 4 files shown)


app/services/admin/teachers/search.rb
```diff
@@ -41,9 +41,9 @@ def matching(scope)
         return scope.where(trn: trns) if trns.any?
 
         full_text_scope = full_text_matches(scope)
-        return full_text_scope unless api_participant_id_query?
+        return full_text_scope unless api_ids_query?
```
cpjmcquillan: Not convinced about the ambiguity in this renaming, but `api_participant_id_or_training_record_id_query?` felt a bit too verbose even for me!

thenapking: If I understand this correctly, we're matching `query_string` on a regex which tells us whether that string probably has some form of ID in it.  But "probably" makes its naming ambiguous.  If we negate it and call the method something like `query_omits_token_string`, the we can have:
```
return full_text_scope if query_omits_token_string
```
I know that violates negative naming.  But it could be more generic.  Perhaps just: `full_text_query?`

thenapking (approved):
LGTM.  Hopefully my suggestion on naming helps!

app/services/admin/teachers/search.rb
```diff
@@ -59,14 +59,16 @@ def normalized_full_text_query
...
         query_string.match?(/\A(?=.*[\d-])[a-f0-9-]{8,}\z/i)
       end
 
-      def api_participant_id_matches(scope)
+      def api_ids_matches(scope)
         escaped_query = ActiveRecord::Base.sanitize_sql_like(query_string)
 
         scope.where("CAST(teachers.api_id AS text) ILIKE ?", "%#{escaped_query}%")
+         .or(scope.where("CAST(teachers.api_ect_training_record_id AS text) ILIKE ?", "%#{escaped_query}%"))
+         .or(scope.where("CAST(teachers.api_mentor_training_record_id AS text) ILIKE ?", "%#{escaped_query}%"))
```
avinhurry: I checked the query plan locally against current prod numbers (~262k teachers).

This does fallback to a parallel seq scan because the substring UUID matching can't use the existing indexes, but performance still seemed acceptable for this low volume admin search for now. 

Just worth keeping in mind if the dataset or usage grows.

app/views/admin/teachers/index.html.erb
```diff
@@ -5,7 +5,7 @@
         :q,
         value: params[:q],
         label: { text: "Search for teacher", size: "s" },
-        hint: { text: "Name, TRN or API participant ID" },
+        hint: { text: "Name, TRN, API participant ID, or API training record ID" },
```
avinhurry, line 8: This feels a little verbose, how about:

`Name, TRN, API participant or training record ID`

Tradeoff is that it's slightly less explicit, but it avoids spilling onto the next line.

spec/services/admin/teachers/search_spec.rb
```diff
@@ -44,6 +44,26 @@
       end
     end
 
+    context "when it is a partial API ECT training record ID" do
+      let(:query_string) { "5766141740" }
+      let!(:teacher) { FactoryBot.create(:teacher, api_ect_training_record_id: "123e4567-e89b-12d3-a456-576614174000") }
+      let!(:other_teacher) { FactoryBot.create(:teacher, api_ect_training_record_id: "999e4567-e89b-12d3-a456-426614174999") }
+
+      it "matches the API participant ID" do
```
avinhurry:
```suggestion
      it "matches the API ECT training record ID" do
```

spec/services/admin/teachers/search_spec.rb
```diff
@@ -44,6 +44,26 @@
...
+        expect(subject.teacher_scope).to contain_exactly(teacher)
+      end
+    end
+
+    context "when it is a partial API mentor training record ID" do
+      let(:query_string) { "5766141740" }
+      let!(:teacher) { FactoryBot.create(:teacher, api_mentor_training_record_id: "123e4567-e89b-12d3-a456-576614174000") }
+      let!(:other_teacher) { FactoryBot.create(:teacher, api_mentor_training_record_id: "999e4567-e89b-12d3-a456-426614174999") }
+
+      it "matches the API participant ID" do
```
avinhurry:
```suggestion
      it "matches the API mentor training record ID" do
```

app/services/admin/teachers/search.rb
avinhurry: My vote would be `api_id_fragment_query?` that's generic enough to cover participant ID and training record IDs, but still specific about this being the UUID/API ID substring path rather than the full text match.

avinhurry (approved):

Rmcnaney:
Product review plssss @mason-emily

peteryates:
Just had a quick play after standup and if we support searching by the the prefix, we can make it fast. We just need to tell PostgreSQL to index it like it's a `varchar` even though it's numeric.

Build some sample data:

```sql
create table sample (uuid uuid default uuid_generate_v1(), number integer not null);

insert into sample (number) values (select generate_series(1,10000000));
-- INSERT 0 10000000
```

If we add an index to `uuid` using `text_pattern_ops`, it should support
indexing like we would a `varchar`.

```sql
create index idx_uuid_as_text on sample ((uuid::text) text_pattern_ops);
```

Now we can do indexed prefix searching:

```sql
explain select * from sample where uuid::text like '3cf3aeb0%';

-- ┌────────────────────────────────────────────────────────────────────────────────────────────────────┐
-- │                                             QUERY PLAN                                             │
-- ├────────────────────────────────────────────────────────────────────────────────────────────────────┤
-- │ Bitmap Heap Scan on sample  (cost=1957.57..65644.53 rows=50050 width=20)                           │
-- │   Filter: ((uuid)::text ~~ '3cf3aeb0%'::text)                                                      │
-- │   ->  Bitmap Index Scan on idx_uuid_as_text  (cost=0.00..1945.06 rows=50050 width=0)               │
-- │         Index Cond: (((uuid)::text ~>=~ '3cf3aeb0'::text) AND ((uuid)::text ~<~ '3cf3aeb1'::text)) │
-- └────────────────────────────────────────────────────────────────────────────────────────────────────┘
```

Which is much faster than suffix searching:

```sql
explain select * from sample where uuid::text like '%c87f5456cc89';

-- ┌─────────────────────────────────────────────────────────────────────────────┐
-- │                                 QUERY PLAN                                  │
-- ├─────────────────────────────────────────────────────────────────────────────┤
-- │ Gather  (cost=1000.00..137847.68 rows=1001 width=20)                        │
-- │   Workers Planned: 2                                                        │
-- │   ->  Parallel Seq Scan on sample  (cost=0.00..136747.58 rows=417 width=20) │
-- │         Filter: ((uuid)::text ~~ '%c87f5456cc89'::text)                     │
-- └─────────────────────────────────────────────────────────────────────────────┘
```

peteryates (approved):

PR #2894, opened 2026-05-11
Add ability for admin users to create new delivery partners

ericaporter:
### Context
Admins currently have no way to create new delivery partners through the admin console. This ticket adds the ability to create a new delivery partner via a simple name entry form, generating an API ID automatically on creation.

- New/create actions added to Admin::DeliveryPartnersController with breadcrumb and backlink support
- Admin::DeliveryPartners::Create service wraps the save and event recording in a transaction
delivery_partner_created added to Event::EVENT_TYPES and a corresponding - Events::Record class method added
"Add a new delivery partner" button added to the delivery partners index page
- New delivery partner name form view with GOV.UK Design System components, presence and case-insensitive uniqueness validation
- Success banner on the show page via the existing alert: flash convention

### Guidance to review
1. Log in as an admin user
2. Navigate to Organisations → Delivery partners
3. Click "Add a new delivery partner"
4. Attempt to submit with a blank name — verify validation error appears
5. Attempt to submit with the name of an existing delivery partner — verify duplicate validation error appears
6. Submit a valid name — verify you are redirected to the new delivery partner's show page with the success banner and API ID visible
7. Verify the new delivery partner appears in the index search


app/views/admin/delivery_partners/index.html.erb
```diff
@@ -5,6 +5,10 @@
   hint_text: "Search by name",
 ) %>
 
+<%= govuk_link_to "Add a new delivery partner",
```
cpjmcquillan: There is a [built-in helper](https://govuk-components.x-govuk.org/helpers/button/#link-styled-like-a-button) for this FYI

```suggestion
<%= govuk_button_link_to "Add a new delivery partner",
```

spec/features/admin/organisations/delivery_partners_spec.rb
cpjmcquillan, line 1: Is there some nuance about using `page.locator("main.govuk-main-wrapper")` here that is significant? 🤔

If it is, should we add a comment about it or talk about it in the commit message? If it isn't, do we need it?

cpjmcquillan (approved):

app/services/admin/delivery_partners/create.rb
```diff
@@ -0,0 +1,27 @@
+module Admin
+  module DeliveryPartners
+    class Create
```
thenapking, line 3: I'm not sure this should be in the components directory.  It looks more like a service than a view component, and doesn't include `Component` in the name.  I've also just noticed your test is in the `spec/services` directory, so I think this was a mistake!

app/controllers/admin/delivery_partners_controller.rb
```diff
@@ -4,7 +4,7 @@ class DeliveryPartnersController < AdminController
 
     layout "full"
     before_action :set_delivery_partner, only: %i[show edit update]
-    before_action :assign_backlink_params, only: %i[index show edit update]
+    before_action :assign_backlink_params, only: %i[index show edit update new create]
```
thenapking: I don't think we need `only` here, as it lists all the actions we have in the controller.
```suggestion
    before_action :assign_backlink_params
```

config/routes/admin.rb
```diff
@@ -29,7 +29,7 @@
       end
 
       resources :lead_providers, only: %i[index], path: "lead-providers"
-      resources :delivery_partners, only: %i[index show edit update], path: "delivery-partners" do
+      resources :delivery_partners, only: %i[index show new create edit update], path: "delivery-partners" do
```
thenapking, line 32:
```suggestion
      resources :delivery_partners, except: :destroy, path: "delivery-partners" do
```

thenapking (approved):
Looks good.  I think a couple of minor things to clear up with the controller actions and a mislocated file, but is otherwise fine.

app/views/admin/delivery_partners/index.html.erb
ericaporter: thanks 💡

config/routes/admin.rb
ericaporter: Good spot - full house!

app/services/admin/delivery_partners/create.rb
ericaporter: Yeah great spot! 🙈

spec/features/admin/organisations/delivery_partners_spec.rb
ericaporter: Copied from adjacent file - which I have now updated too 😆

spec/models/delivery_partner_spec.rb
```diff
@@ -16,9 +16,14 @@
...
 
-    it { is_expected.to validate_presence_of(:name) }
     it { is_expected.to validate_uniqueness_of(:name).case_insensitive.with_message("A delivery partner with this name already exists") }
     it { is_expected.to validate_uniqueness_of(:api_id).case_insensitive.with_message("API id already exists for another delivery partner") }
+
+    it "has the correct blank name error message" do
+      subject.name = ""
+      subject.valid?
+      expect(subject.errors[:name]).to include("Enter a delivery partner name")
+    end
```
cpjmcquillan: Is there a reason the shoulda-matcher didn't work here?

I would have thought we could do this

```ruby
it { is_expected.to validate_presence_of(:name).with_message("Enter a delivery partner name") }
```

spec/helpers/flash_helper_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+      end
+    end
+
+    context "when flash[:alert] is a hash" do
+      it "returns the body value" do
+        expect(helper.flash_alert_body({ "heading" => "Delivery partner added", "body" => "You can now add lead providers working with them." })).to eq("You can now add lead providers working with them.")
+      end
+    end
+  end
+end
```
cpjmcquillan: Missing `\n` here

app/helpers/flash_helper.rb
cpjmcquillan, line 1: I think this is fine for now, but I wonder whether it's worth us building an abstract `FlashAlertComponent` at some point since there is a little more complexity here 🤔

cpjmcquillan (approved):
LGTM!

app/helpers/flash_helper.rb
ericaporter: I'll add a ticket for this in issues 👍

PR #2895, opened 2026-05-11
Migration data fixing processor

tonyheadford:
### Context

To enable the ability to surgically update/fix data that was migrated from ECF1, we're adding the ability to use a CSV to describe the changes at the model level.

The patching is comprised of a script `db/scripts/migration_data_fixes.rb` that reads a CSV file from the same directory `db/scripts/migration_data_fixes.csv` which describes the changes that should be made.

The CSV has the following columns:

- `object_type`: the Ruby class name of the record to modify e.g. `ECTAtSchoolPeriod` or `TrainingPeriod`
- `object_id`: the ID of the record (not needed for `create` actions)
- `action`: the type of change to be made, one of `create`, `update` or `delete`
- `attributes`: a comma separated list of attribute name/value pairs required for `create` and `update` actions

To enable us to test the changes to real data in specs, I've added a spec generator that can take a `Teacher` record and generate a basic spec, building the teacher's child models from factories with the correct relationships and with non-PII values.

Generating a spec can be done from the Rails console, e.g.

```
MigrationFixes::SpecGenerator.new(teacher: example_teacher).save!
```

This will create a spec file in the `spec/migration_fixes/real_examples/` folder with the filename `teacher_<teacher.id>_spec.rb`

In addition to the specs, the changes should always be thoroughly tested in the migration environment before applying them to production.

Adds initial migration fixes and processing scripts, migration fixes process service and spec generator functionality


db/scripts/migration_data_fixes.rb
```diff
@@ -0,0 +1,31 @@
...
+  csv_log << %w[object_type,object_id,action,attributes,errors]
+  processor = MigrationFixes::Processor.new
+
+  CSV.foreach(csv_file, headers: true, header_converters: :symbol) do |row|
+    object = processor.process!(data_change: row.to_h)
+    errors = object.presence&.errors&.to_json
+
+    csv_log << [row[:object_type], row[:object_id], row[:action], row[:attributes], errors]
+  rescue StandardError => e
+    Rails.logger.warn("ERROR: TrainingPeriod ID: #{tp_id} - #{e.message}")
```
leoapost: If the intention is to log the object type and ID along with the error message, then it should probably look something like this:
```suggestion
    Rails.logger.warn("ERROR processing #{row[:object_type]} ID #{row[:object_id]}: #{e.class} - #{e.message}")
```

app/services/migration_fixes/processor.rb
```diff
@@ -0,0 +1,52 @@
...
+
+    case action
+    when "create"
+      target_object = create!(object_type, extract_attributes(data_change[:attributes]))
+    when "update"
+      update!(target_object, extract_attributes(data_change[:attributes]))
+    when "delete"
+      delete!(target_object)
+    else
+      raise "Unknown action '#{action}'"
```
leoapost: This should address the SonarCloud warning:
```suggestion
      raise ArgumentError, "Unknown action '#{action}'"
```

db/scripts/migration_data_fixes.rb
```diff
@@ -0,0 +1,31 @@
...
+# attributes  - a comma separated list of attribute name and value pairs in the format "attrname_1,value_1,attrname_2,value_2".
+#               Ensure the list is enclosed in quotes so the CSV sees just a single data field
+#
+# $ kubectl exec -it <pod-name> -- bin/rails runner db/scripts/migration_data_fixes.rb
+csv_log = nil
+
+begin
+  csv_file = Rails.root.join("db/scripts/migration_data_fixes.csv")
+  csv_log = CSV.open(Rails.root.join("tmp/migration_data_fixes_log-#{Time.zone.now.to_fs(:iso8601)}.csv"), "w")
+  csv_log << %w[object_type,object_id,action,attributes,errors]
```
leoapost: This will produce a single string containing commas, not five separate header cells. Replacing the commas with spaces should fix it.

```suggestion
  csv_log << %w[object_type object_id action attributes errors]
```

leoapost (changes requested):

db/scripts/migration_data_fixes.rb
tonyheadford: Yes well spotted. I noticed them earlier and fixed them in the later PR

leoapost (approved):
LGTM

PR #2900, opened 2026-05-12
[3744] Add validation for leaving date and new school start date when teacher is moving schools

thenapking:
### Context
NB please note that the error message has changed slightly from the card with agreement from @mason-emily.   
ACs
1. Validation on new school start date added to block a start date earlier than the latest training period start date at the previous school.

2. Error message for this scenario will say: Our records show that {{teacher name}} started their latest training at {{school}} on {{latest training period start date}}. Enter a start date after {{earliest end date of latest training period}}.

3. The existing school start date error message will take priority over this new one if there are no training periods with a later start date.

4. Validation on school date leaving date to block a date that is earlier than the latest training period start date at the current school.

Error message for leaving school scenario should say: Our records show that {{teacher name}} started their latest training on {{latest training period start date}}. Enter a date after {{earliest end date of latest training period}}.

When we were discussing this we thought there might be hidden complexity. The first issue we discovered was to do with backdated training periods.  The second issue was to do with the interplay of `ect_at_school_periods` and `training_periods`.

image

Here we are trying to create a new ECT period at a separate school.  The existing ECT and training period may not start on the same day.  There may be several training periods.  We need to ensure that the latest start date of either period is used for validation.  In addition if the entered date is before the start of both the ECT and training period we need to prioritise errors from the ECT period.

Since any open periods need to be finished, and can't be zero length we need to take this into account when giving the earliest date the ECT period can be created at a new school.  For instance, consider an ECT with a training period that starts on 1st Jan.  This cannot be zero length so the earliest it can finish is 2nd Jan.  Therefore the earliest valid start date at a new school is 3rd Jan.

Additionally there may be several training periods, of which we only consider the latest.  If that training period starts in the future it will be deleted by the ECT registration process and we therefore don't let it affect validation.  However if the ECT was registered today, and their first training period started today, then that ECT period would block registration at another school today.  The earliest registration date would be the day after tomorrow.  Similarly we now ensure that if a training period has started today, the earliest registration date at another school is the day after tomorrow.  

Due to the variety of combinations present to make this validation work, I have introduced a custom validation class, which exhaustively tests these combinations.  The wizard steps then handle error messages given the period which overlaps and the earliest valid date.

In the tests for the service class there are shared examples which include the logic above that states an entered date is invalid yesterday, today, and tomorrow, but valid the day after tomorrow, if it clashes with a period that starts today.  I've tried to make that as readable as possible.  The easiest way I found to think of it was this four day period around the entered date, with the last day being a valid date.

We have recently updated the logic around backdated ECT periods, in this [PR](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/2688).  This means that when registering an ECT in the past their first training period starts today.  Previously we allowed the niche case where two schools do that on the same day.  This validation blocks that, and insists that the second school cannot back date their registration.

Please note that the validator class also works for Mentor periods.

### Guidance to review

I found that for some reason many of the listed teachers could not be registered, and therefore you will have to use these very specific ones listed below.

Please note that in order to test the validation applies to the mentors you will need to select "Yes, they will be mentoring at our school only".  You should also test on the check answers page that the validations are triggered when you attempt to change the start date.

| TRN     | Name           | Type                     | Period Start | Training Period 1 | Training Period 2 | Training Period 3 |
|---------|----------------|--------------------------|--------------|-------------------|-------------------|-------------------|
| 3002577 | Jonas Bloggs   | ect_at_school_period     | 2025-09-01   | provider_led (2025-09-01 → ongoing) |  |  |
| 3002578 | Cynthia Parks  | ect_at_school_period     | 2025-09-01   | provider_led (2025-10-01 → ongoing) |  |  |
| 3002579 | Taylor Hawkins | ect_at_school_period     | 2025-09-01   | provider_led (2025-10-01 → 2025-10-31) | provider_led / ambition (2025-11-01 → ongoing) |  |
| 3002580 | Muhammed Ali   | ect_at_school_period     | 2025-09-01   | school_led (2025-10-01 → 2025-10-31) | provider_led / ambition (2025-11-01 → ongoing) |  |
| 3002582 | Robson Scottie | ect_at_school_period     | 2025-09-01   | provider_led (2025-10-01 → 2025-10-31) | school_led (2025-11-01 → 2025-11-30) | provider_led (2025-12-01 → ongoing) |
| 3012858 | Dave Teacher   | mentor_at_school_period  | 2025-09-01   | provider_led (2025-09-01 → ongoing) |  |  |
| 3003943 | Donna Msa      | mentor_at_school_period  | 2025-09-01   | provider_led (2025-10-01 → ongoing) |  |  |
| 3012235 | Claire Cool    | mentor_at_school_period  | 2025-09-01   | provider_led (2025-10-01 → 2025-10-31) | provider_led / ambition (2025-11-01 → ongoing) |  |


ericaporter (approved):
You might want to check the SonarCloud warning in period_boundary.rb - I think it wants an else within the case statement. I don't know if this will cause an issue when you go to deploy or not.

PR #2901, opened 2026-05-12
Ensure ECTs can transfer the day a previous 2-day school period ended

cpjmcquillan:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/3972

Before, we ran into issues when Users tried to register a new ECT the same day
a 2-day `ECTAtSchoolPeriod` ended at another school.

If the teacher has an `ECTAtSchoolPeriod` at another school that started on or
before the registration date and is either:

- ongoing (i.e has no `finished_on`), or
- has a `finished_on` on or after the registration date

then we "finish" that `ECTAtSchoolPeriod` on the day before the registration
date. This means the ECT can be registered on the given registration date
without any of the periods overlapping.

However we ran into [issues] for these short school periods. Since we try to
"finish" the period the day before the registration date, that meant we tried to
change the previous period so it started and finished on the same day.

Once upon a time (pre https://github.com/DFE-Digital/register-early-career-teachers-public/pull/2357), that would have been a problem. But now that
represents a period with a duration of 1 day.

To make it possible, though, we need to update the validations and database
constraints to reflect that.

This updates the inteval validation to allow `finished_on` to be the same as
`started_on`, and changes the database constraint to allow period lengths that
**are at least zero**.

Now, ECTs can be transfered without running into errors in these scenarios.

---

Note: it might seem strange that a teacher could be at a school for a single
day. It is strange! These are mostly a result of migration, where we have some
"stub" records from the economy migrator.

[issues]: https://dfe-teacher-services.sentry.io/issues/7462387892

### DIFF
--- a/spec/services/schools/register_ect_spec.rb
+++ b/spec/services/schools/register_ect_spec.rb
@@ -773,13 +773,13 @@ RSpec.describe Schools::RegisterECT do
       end
     end
 
-    context "when ECT is transferring from another school on the same date as the previously reported leaving date" do
+    context "when ECT is transferring from another school on the same date " \
+            "as the previously reported leaving date" do
       let(:training_programme) { "school_led" }
       let(:lead_provider) { nil }
       let(:other_school) { FactoryBot.create(:school) }
       let!(:teacher) { FactoryBot.create(:teacher, trn:) }
       let(:started_on) { Date.new(2026, 3, 18) }
-      let(:expected_finished_on) { started_on.yesterday }
 
       let!(:existing_period) do
         FactoryBot.create(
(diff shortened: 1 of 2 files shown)


db/migrate/20260512143605_change_ect_at_school_periods_period_length_constraint.rb
```diff
@@ -0,0 +1,11 @@
+class ChangeECTAtSchoolPeriodsPeriodLengthConstraint < ActiveRecord::Migration[8.1]
+  def up
+    remove_check_constraint :ect_at_school_periods, name: "period_length_greater_than_zero", if_exists: true
+    add_check_constraint :ect_at_school_periods, "finished_on >= started_on", name: "period_length_greater_than_zero", validate: true
```
cpjmcquillan: I decided to keep the name of the constraint the same, since I figured we are going to update the other constraints for other periods too.

It made sense to redefine what a period length of "1" represents, rather than requiring some mental gymnastics to figure out what a length of "0" might mean.

app/models/concerns/interval.rb
```diff
@@ -52,7 +52,9 @@ module Interval
   def period_dates_validation
     return if incomplete?
 
-    errors.add(:finished_on, "The end date must be later than the start date (#{started_on.to_fs(:govuk)})") if invalid_date_order?
+    if invalid_date_order?
+      errors.add(:finished_on, "The end date must be later than the start date (#{started_on.to_fs(:govuk)})")
+    end
```
cpjmcquillan, line 57: I switched this round because it felt way easier to read the important bit, but happy for people to disagree!

spec/services/schools/register_ect_spec.rb
```diff
@@ -793,15 +793,76 @@
...
+      end
+
+      # TODO: we want to handle this more gracefully (i.e communicate to Users)
+      # that they need to choose a different start date
+      it "raises an error" do
+        expect { service.register! }
+          .to raise_error(ActiveRecord::RecordInvalid)
+          .with_message(/The end date must be later than the start date/)
+      end
+    end
```
cpjmcquillan, line 864: @Aaminap would you be able to write up a ticket for this in the backlog please?

Essentially the way we finish/manage previous school periods when dates overlap doesn't work with 1-day periods (we can't set the finished on before the started on), so we need to bubble these errors up to the user on the check answers page (or catch them earlier on the start date step).

Let me know if you have any questions! 💚

thenapking: That is what this PR addresses: https://github.com/DFE-Digital/register-early-career-teachers-public/pull/2900

app/models/concerns/interval.rb
```diff
@@ -82,7 +84,7 @@ def successors = finished_on ? siblings.started_on_or_after(finished_on) : self.
   def valid_date_order?
     return true if incomplete?
 
-    started_on < finished_on
+    started_on <= finished_on
```
peteryates, line 87: Do we need to wait for the inclusive ends PRs (only training and induction periods left) to be made before we switch this?

db/migrate/20260512143605_change_ect_at_school_periods_period_length_constraint.rb
peteryates: Yeah this is tricky. Maybe `finished_on_not_before_started_on`?

peteryates (approved):
Looks great. Just one question on the order we need to merge these related date changes (this, Tom's #2900 and mine #2889).

app/models/concerns/interval.rb
thenapking: I've just realised that this PR makes a change to _all_ intervals, but it only then updates the code for finishing `ect_at_school_periods`.  We would want to ensure that this also applies to `mentor_at_school_periods` and `training_periods`.  I know that's not in the ACs of this card, but they are also affected in the same way and would be affected by the change to the `Interval` class. 
I think this means it may interact badly with #2900.  That adds a validation for ECTs and Mentors moving school or leaving, which raises a validation error if the user enters a date on the same day as the end of those periods.  It asks them to enter the next day.  I think it might be right that we do want them to be able to enter the same day.  But as @peteryates points out, we need at least to have his fix applied to training periods before this will work. May change will work now, and prevent issues and we can refactor the dates in the validations once #2889  has been merged and before this is.

PR #2902, opened 2026-05-12
[3929] Add finance contract periods index page

leandroalemao:
### Context

Ticket: [3929](https://github.com.mcas.ms/DFE-Digital/register-ects-project-board/issues/3929)

For the opening registration journey we are going to be adding a set of views, all of which will stem from a new page that lists our contract periods.

- Add finance contract periods index page

### Guidance to review

Review app


app/controllers/admin/finance/contract_periods_controller.rb
```diff
@@ -0,0 +1,14 @@
...
+  class ContractPeriodsController < Admin::Finance::BaseController
+    layout "full"
+
+    def index
+      @breadcrumbs = {
+        "Finance" => admin_finance_path,
+        "Contract periods" => nil,
+      }
+      page = params[:page].presence
+      @pagy, @contract_periods = pagy(::ContractPeriods::Search.new(params[:q]).search, page:)
```
ethax-ross: Shall we scrap `params[:q]`? 🤔 I can't imagine we'd need to search contract periods given their number... I find a parameter called `q` pretty unhelpful/unclear in general but appreciate this seems to be an established pattern.

I'd question whether or not we want/need a `Search` service at all tbh; if we're following an established pattern maybe its worth it to be consistent, but it does feel a bit overkill -- lets see what others think.

config/routes/admin.rb
```diff
@@ -106,6 +106,10 @@
         end
 
         resource :search_declarations, only: :show, path: "search-declarations"
+
+        constraints -> { !Rails.env.production? } do
+          resources :contract_periods, only: %i[index show], path: "contract-periods"
```
ethax-ross: Shall we keep this to `index` here and Mook can add `show` as part of his PR? maybe we should add it to the various .envs and have a proper per-environment config for it as well? 🤔

app/views/admin/finance/contract_periods/_add_contract_period_button.html.erb
```diff
@@ -0,0 +1,5 @@
+<div class="govuk-grid-row">
```
ethax-ross: I would just stick this in the `index` view but again I'm not familiar with established patterns on the front end stuff

spec/services/contract_periods/search_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  let!(:contract_period) { FactoryBot.create(:contract_period, year: 2022) }
+  let!(:other_contract_period) { FactoryBot.create(:contract_period, year: 2024) }
+
+  it "finds contract periods by year" do
+    expect(ContractPeriods::Search.new(2022).search).to contain_exactly(contract_period)
+  end
+
+  it "excludes non-matching contract periods" do
+    expect(ContractPeriods::Search.new(2024).search).to contain_exactly(other_contract_period)
```
ethax-ross: I think this test is redundant/the same as the one above?

config/routes/admin.rb
```diff
@@ -106,6 +106,10 @@
         end
 
         resource :search_declarations, only: :show, path: "search-declarations"
+
+        constraints -> { !Rails.env.production? } do
```
ethax-ross: Can we add a request spec and some coverage that this isn't accessible outside of enabled environments?

ethax-ross (commented):
Thanks Leandro just a few questions, mostly from my general lack of knowledge around anything in our frontend 😅

mooktakim (approved):
I don't think we need search for contract periods as there won't be many anyway.
Other than comments from Ross lgtm

app/views/admin/finance/show.html.erb
```diff
@@ -30,5 +30,22 @@
     <p class="govuk-body">
       Search for declarations by declaration ID.
     </p>
+
+    <% if !Rails.env.production? %>
+      <hr class="govuk-section-break govuk-section-break--l govuk-section-break--visible">
```
peterdavidhamilton: there is a helper for this one

peterdavidhamilton (approved):

app/views/admin/finance/show.html.erb
```diff
@@ -30,5 +30,22 @@
...
+        <%= govuk_link_to(
+              "Contract periods",
+              admin_contract_periods_path,
+              no_visited_state: true,
+              no_underline: true
+            ) %>
+      </h2>
+
+      <p class="govuk-body">
+        View and manage contract periods for opening registration.
```
StephenC-V1: Very minor point/question, but I'm not sure we need 'for opening registrations'. Perhaps just 'View and manage contract periods'.

StephenC-V1 (commented):
Reviewed, single comment/suggestion.

PR #2903, opened 2026-05-12
[3559] Show training API responses per lead provider

avinhurry:
### Context

Fixes the teacher training history API response rows so they show on the latest training period per lead provider, separately for ECT and mentor histories.

- Group API response rows by lead provider instead of partnership
- Apply the same API response row logic to mentor training periods.

### Guidance to review

Log in as an admin user and find a teacher with multiple training periods across different lead providers.

Check the Training tab. The API response row should show on the latest card for each lead provider, and not on older cards for the same lead provider.

### DIFF
--- a/app/controllers/admin/teachers/trainings_controller.rb
+++ b/app/controllers/admin/teachers/trainings_controller.rb
@@ -24,8 +24,7 @@ module Admin
           training_periods
             .select(&:provider_led_training_programme?)
             .group_by { |training_period| lead_provider_id_for(training_period) }
-            .values
-            .map { it.first.id }
+            .map { |_lead_provider_id, training_periods| training_periods.first.id }
         end
       end
 
--- a/app/controllers/admin/teachers/trainings_controller.rb
+++ b/app/controllers/admin/teachers/trainings_controller.rb
@@ -8,12 +8,32 @@ module Admin
         @navigation_items = helpers.admin_teacher_navigation_items(@teacher, :training)
         @breadcrumbs = teacher_breadcrumbs
         @ect_at_school_periods = @teacher.ect_at_school_periods
-        @ect_training_periods = @teacher.ect_training_periods.includes(ect_at_school_period: :teacher).latest_first
-        @period_ids_to_show_api_row = @ect_training_periods.group_by(&:lead_provider_delivery_partnership).values.map { it.first.id }
-        @mentor_training_periods = @teacher.mentor_training_periods.includes(mentor_at_school_period: :teacher).latest_first
+        @ect_training_periods = @teacher.ect_training_periods
+          .includes(:active_lead_provider, :expression_of_interest, ect_at_school_period: :teacher)
+          .latest_first
+        @mentor_training_periods = @teacher.mentor_training_periods
+          .includes(:active_lead_provider, :expression_of_interest, mentor_at_school_period: :teacher)
+          .latest_first
+        @training_period_ids_with_api_response = training_period_ids_with_api_response
       end
 
     private
+
+      def training_period_ids_with_api_response
+        [@ect_training_periods, @mentor_training_periods].flat_map do |training_periods|
+          training_periods
+            .select(&:provider_led_training_programme?)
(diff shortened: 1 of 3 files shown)


app/controllers/admin/teachers/trainings_controller.rb
```diff
@@ -8,13 +8,33 @@ def show
...
       end
 
     private
 
+      def training_period_ids_with_api_response
+        [@ect_training_periods, @mentor_training_periods].flat_map do |training_periods|
+          training_periods
+            .select(&:provider_led_training_programme?)
+            .group_by { |training_period| lead_provider_id_for(training_period) }
+            .except(nil)
```
cpjmcquillan: Do we need to be this defensive? 💭

I'd expect all provider led training programmes to have a lead provider ID.

Which scenarios does this cover? 🤔

app/controllers/admin/teachers/trainings_controller.rb
```diff
@@ -8,13 +8,33 @@ def show
...
+        @training_period_ids_with_api_response = training_period_ids_with_api_response
       end
 
     private
 
+      def training_period_ids_with_api_response
+        [@ect_training_periods, @mentor_training_periods].flat_map do |training_periods|
+          training_periods
+            .select(&:provider_led_training_programme?)
+            .group_by { |training_period| lead_provider_id_for(training_period) }
```
thenapking, line 26: I think you might be able to simplify this with `filter_map`.  Then you wouldn't need `except`, `values` or `map`.  I've used semantic naming for the keys, but it could just be `key, values` or something, since all we're really trying to achieve is to get the first from the list of grouped training periods, if I've understood it correctly.
```suggestion
       .group_by { lead_provider_id_for(it) }
      .filter_map { |lead_provider_id, training_periods| training_periods.first.id }
```

cpjmcquillan (approved):

app/controllers/admin/teachers/trainings_controller.rb
```diff
@@ -8,13 +8,33 @@ def show
...
+          .latest_first
+        @mentor_training_periods = @teacher.mentor_training_periods
+          .includes(:active_lead_provider, :expression_of_interest, mentor_at_school_period: :teacher)
+          .latest_first
+        @training_period_ids_with_api_response = training_period_ids_with_api_response
       end
 
     private
 
+      def training_period_ids_with_api_response
```
thenapking: The querying here is a bit tricky and I've been trying to think of other ways to do it.  I've seen SQL `COALESCE` used elsewhere in the code but I don't think it really makes it that more simple.  Perhaps a better name for the method would make the query easier to understand? How about `lead_provider_ids_for_latest_training_periods`? Or `latest_training_periods_with_api_response`?

app/controllers/admin/teachers/trainings_controller.rb
avinhurry: Yeah, you're right. We don't need to be this defensive, so I've removed that check.

Thanks!

app/controllers/admin/teachers/trainings_controller.rb
avinhurry: Good shout. I've renamed this to `latest_training_period_ids_with_api_response` so it's clearer that the method returns training period IDs rather than lead provider IDs or records.

I've left this in ruby rather than moving it into SQL/`COALESCE`, as I agree that would not make this easier to follow for a small page level display rule.

app/controllers/admin/teachers/trainings_controller.rb
avinhurry: Good shout. I removed the `nil` check based on @cpjmcquillan's feedback, so there's no need for `filter_map` now. 

I've adapted your suggestion and removed the `values` call.

thenapking: Great, yes this is more clear to read now, and I do agree that there's nothing to be gained from using `COALESCE` etc in terms of clarity.

thenapking (approved):

PR #2911, opened 2026-05-13
Fix flaky `Teachers::Manage` spec

ethax-ross:
The test to ensure the `:new` method is private fails consistently when running the whole test suite locally.

I haven't been able to figure out how/why it fails, but reworking the test to check `private_methods` directly instead of asserting an error is raised when calling `new` seems to resolve the flakyness for me.


peterdavidhamilton (approved):

leandroalemao (approved):

peteryates (commented):
Looks more robust 👍

PR #2912, opened 2026-05-13
[3952] Scope unfunded mentor email to the requesting lead provider - Part 2

mooktakim:
### Context

Ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/3952

Split into 3 PRs:

PR 1 (merged): https://github.com/DFE-Digital/register-early-career-teachers-public/pull/2891
- Added `ect_assigned_mentor_latest_school_period_id` column on `Metadata::TeacherLeadProvider`, populated via the metadata refresh.

PR 2 (this PR):
- Switches `GET /api/v3/unfunded-mentors` to read the email via the new metadata column, so each lead provider sees the email from the school where the mentor is mentoring *their* ECT.
- Removes every remaining reference to `api_mentor_id` in app code so PR 3's column drop can ship safely.

PR 3: https://github.com/DFE-Digital/register-early-career-teachers-public/pull/2914
- Drops the `api_mentor_id` column from the metadata table.

- `UnfundedMentorSerializer` — `email` field filters `teacher.mentor_lead_provider_metadata` by the requesting `lead_provider_id`, picks the row with the most recent `ect_assigned_mentor_latest_school_period.started_on`, and returns that period's email. `max_by` is retained for correctness when a mentor mentors ECTs at different schools for the same LP.
- `UnfundedMentors::Query` — preload chain set to `mentor_lead_provider_metadata: :ect_assigned_mentor_latest_school_period`, keeping the endpoint N+1-free under `strict_loading`. `where_lead_provider_is` filter switched from `api_mentor_id` to a join through `ect_assigned_mentor_latest_school_period → teacher`.
- `Teacher` — adds `has_many :mentor_lead_provider_metadata, through: :mentor_at_school_periods`.
- `MentorAtSchoolPeriod` — adds the inverse `has_many :mentor_lead_provider_metadata` with `dependent: :nullify`, so destroying a MASP nullifies the metadata back-reference rather than dangling.
- `API::TeacherSerializer` — `mentor_id` field now reads `metadata.ect_assigned_mentor_latest_school_period&.teacher&.api_id`. `API::Teachers::Query` preload extended to load that chain.
- `Metadata::TeacherLeadProvider` — `touch` on `api_updated_at` now fires when `ect_assigned_mentor_latest_school_period_id` changes (instead of `api_mentor_id`).
- `Metadata::Handlers::Teacher` — stops computing/stamping `api_mentor_id`; uses `latest_mentorship_period.mentor_at_school_period_id` directly so the includes simplify back to `:latest_mentorship_period`.
- Restored cross-LP behaviour tests in the serializer spec (two LPs receive two different emails for the same shared mentor); dropped `api_mentor_id` assertions from the handler spec.

### Guidance to review

PR 3 depends on this. Merge order: PR 2 → PR 3 (so the app stops referencing `api_mentor_id` before the column is dropped).

### DIFF
--- a/app/models/teacher.rb
+++ b/app/models/teacher.rb
@@ -27,6 +27,11 @@ class Teacher < ApplicationRecord
   has_many :induction_extensions, inverse_of: :teacher
   has_many :teacher_id_changes, inverse_of: :teacher, dependent: :destroy
   has_many :lead_provider_metadata, class_name: "Metadata::TeacherLeadProvider", dependent: :destroy
+  has_many :mentor_lead_provider_metadata,
+           class_name: "Metadata::TeacherLeadProvider",
+           foreign_key: :api_mentor_id,
+           primary_key: :api_id,
+           inverse_of: false
   has_many :induction_periods
   has_many :appropriate_body_periods, through: :induction_periods
   has_many :events
--- a/app/models/teacher.rb
+++ b/app/models/teacher.rb
@@ -27,7 +27,7 @@ class Teacher < ApplicationRecord
   has_many :induction_extensions, inverse_of: :teacher
   has_many :teacher_id_changes, inverse_of: :teacher, dependent: :destroy
   has_many :lead_provider_metadata, class_name: "Metadata::TeacherLeadProvider", dependent: :destroy
-  has_many :mentor_lead_provider_metadata, through: :mentor_at_school_periods
+  has_many :lead_provider_metadata_for_mentees, through: :mentor_at_school_periods
   has_many :induction_periods
   has_many :appropriate_body_periods, through: :induction_periods
   has_many :events
--- a/app/models/teacher.rb
+++ b/app/models/teacher.rb
@@ -27,11 +27,7 @@ class Teacher < ApplicationRecord
   has_many :induction_extensions, inverse_of: :teacher
   has_many :teacher_id_changes, inverse_of: :teacher, dependent: :destroy
   has_many :lead_provider_metadata, class_name: "Metadata::TeacherLeadProvider", dependent: :destroy
-  has_many :mentor_lead_provider_metadata,
-           class_name: "Metadata::TeacherLeadProvider",
-           foreign_key: :api_mentor_id,
-           primary_key: :api_id,
-           inverse_of: false
+  has_many :mentor_lead_provider_metadata, through: :mentor_at_school_periods
   has_many :induction_periods
   has_many :appropriate_body_periods, through: :induction_periods
   has_many :events
--- a/app/serializers/api/teachers/unfunded_mentor_serializer.rb
+++ b/app/serializers/api/teachers/unfunded_mentor_serializer.rb
@@ -3,8 +3,14 @@ class API::Teachers::UnfundedMentorSerializer < Blueprinter::Base
     exclude :id
 
     field(:full_name) { |teacher| Teachers::Name.new(teacher).full_name }
-    field(:email) do |teacher, _options|
-      teacher.latest_mentor_at_school_period.email
+    field(:email) do |teacher, options|
+      lead_provider_id = options[:lead_provider_id]
+
+      teacher.mentor_lead_provider_metadata
+        .select { |m| m.lead_provider_id == lead_provider_id }
+        .max_by { |m| m.latest_mentor_at_school_period.started_on }
+        .latest_mentor_at_school_period
+        .email
     end
     field(:trn, name: :teacher_reference_number)
     field :created_at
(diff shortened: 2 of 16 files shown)


app/models/teacher.rb
```diff
@@ -27,6 +27,11 @@ class Teacher < ApplicationRecord
   has_many :induction_extensions, inverse_of: :teacher
   has_many :teacher_id_changes, inverse_of: :teacher, dependent: :destroy
   has_many :lead_provider_metadata, class_name: "Metadata::TeacherLeadProvider", dependent: :destroy
+  has_many :mentor_lead_provider_metadata,
```
leandroalemao: Hi @mooktakim any spec to add for this new association? tkss

leandroalemao (approved):

app/serializers/api/teachers/unfunded_mentor_serializer.rb
```diff
@@ -3,8 +3,14 @@ class AttributesSerializer < Blueprinter::Base
     exclude :id
 
     field(:full_name) { |teacher| Teachers::Name.new(teacher).full_name }
-    field(:email) do |teacher, _options|
-      teacher.latest_mentor_at_school_period.email
+    field(:email) do |teacher, options|
```
ethax-ross, line 6: I was expecting this to just go through the existing metadata's new association? 🤔 so:

```
teacher
  .lead_provider_metadata
  .find { |m| m.lead_provider_id == lead_provider_id }
  .ect_assigned_mentor_latest_school_period
  .email
```

ethax-ross (commented):
I think we want to bring across this change into this PR as well?

https://github.com.mcas.ms/DFE-Digital/register-early-career-teachers-public/pull/2914/changes#diff-2d01e3a9878ce428be6d570b96bada371e938ac0883de57374c7b0c23f0010c3R28

Otherwise the migration will remove the `api_mentor_id` but the application will still reference it until the deploy finishes, which can result in some errors?

app/serializers/api/teachers/unfunded_mentor_serializer.rb
mooktakim: This won't work because `Metadata::Handlers::Teacher` stamps `ect_assigned_mentor_latest_school_period_id` only from the ECT's point of view. For an unfunded mentor's own record the column is `nil`, so the chain hits
  `nil.email`.

The `MentorAtSchoolPeriod` we want is stamped on the *ECT's* metadata row (pointing back at the mentor's `MentorAtSchoolPeriod`), which is why we need `mentor_lead_provider_metadata` to reach it.

mooktakim: I've moved over the other changes from part 3 for `api_mentor_id`

ethax-ross: > This won't work because `Metadata::Handlers::Teacher` stamps `ect_assigned_mentor_latest_school_period_id` only from the ECT's point of view. For an unfunded mentor's own record the column is `nil`, so the chain hits `nil.email`.
> 
> The `MentorAtSchoolPeriod` we want is stamped on the _ECT's_ metadata row (pointing back at the mentor's `MentorAtSchoolPeriod`), which is why we need `mentor_lead_provider_metadata` to reach it.

Ah I understand now, thanks for explaining. I wonder if there's a clearer name for the `mentor_lead_provider_metadata` relationship, as its not really the `lead_provider_metadata` for this teacher as a mentor? 🤔 is it more `lead_provider_metadata_for_mentees`? it hurts my head thinking about it 😅

mooktakim: definitely giving me a headache lol 
Will rename to `lead_provider_metadata_for_mentees`

app/serializers/api/teachers/unfunded_mentor_serializer.rb
```diff
@@ -3,8 +3,14 @@ class AttributesSerializer < Blueprinter::Base
...
+      lead_provider_id = options[:lead_provider_id]
+
+      teacher.lead_provider_metadata_for_mentees
```
ethax-ross, line 9: I think there should only ever be 1 of these for each LP/teacher, so could use:

```
teacher.lead_provider_metadata_for_mentees
       .sole { |m| m.lead_provider_id == lead_provider_id }
       .ect_assigned_mentor_latest_school_period
       .email
```

ethax-ross (approved):
Looks good just a minor suggestion. Can we hold off on merging this just now though as we've had a high priority issue around mentor ids I want to look into/understand what's going on first https://github.com.mcas.ms/DFE-Digital/register-ects-project-board/issues/3986

app/serializers/api/teachers/unfunded_mentor_serializer.rb
mooktakim: I'm not sure if it would be one result. Can the Mentor have multiple mentees at the same LP?
Currently we'd take the latest mentor at school period and use that for email.

ethax-ross: Yeah I think you're right, lets leave it as-is 👍

PR #2913, opened 2026-05-13
[3954] Service to generate schedules for a new contract period

peterdavidhamilton:
### Context

[Epic](https://github.com/DFE-Digital/register-ects-project-board/issues/3927) opening registration journey

Closes https://github.com/DFE-Digital/register-ects-project-board/issues/3954

```
ContractPeriods::SeedFromPrevious
  #schedule!
    with a previous contract period
      when the contract period starts today
        is expected to raise ContractPeriods::SeedFromPrevious::ContractPeriodStartedError with "Contract periods cannot be scheduled after they have started"
      when the contract period has not started yet
        advances dates by one year if set
        raises an error if called repeatedly
        copies schedules without milestones
        clones from the previous contract period
        copies milestone declaration types
        copies schedule identifiers
        rolls back changes if an error occurs
        returns :scheduled
      when the contract period has already started
        is expected to raise ContractPeriods::SeedFromPrevious::ContractPeriodStartedError with "Contract periods cannot be scheduled after they have started"
      when schedules already exist for the contract period
        is expected to raise ContractPeriods::SeedFromPrevious::AlreadyScheduledError with "The contract period already has schedules"
    without a previous contract period
      is expected to raise ContractPeriods::SeedFromPrevious::NoPreviousContractPeriodError with "No previous contract period found"
  #initialize
    when :contract_period is nil
      is expected to raise ArgumentError with "Contract period is required"
```


app/services/contract_periods/default_schedule_generator.rb
```diff
@@ -0,0 +1,119 @@
...
+      return :already_scheduled if already_scheduled?
+
+      ActiveRecord::Base.transaction do
+        if cloneable?
+          clone_previous_schedules_and_milestones
+        else
+          create_default_schedules_and_milestones
+        end
+      end
+    end
```
peterdavidhamilton: Did we only want an error raised when already scheduled or also when already started?

Should we rescue errors and re-raise too?

What should our errors and messages be?

app/services/contract_periods/default_schedule_generator.rb
```diff
@@ -0,0 +1,119 @@
...
+          )
+        end
+      end
+      :created
+    end
+
+    # @return [Symbol] :cloned
+    def clone_previous_schedules_and_milestones
+      previous_contract_period.schedules.includes(:milestones).find_each do |previous_schedule|
+        next if previous_schedule.milestones.none?
```
ethax-ross: I don't think we have schedules without milestones, but even if we do I think having them brought across might be more expected? 🤔

app/services/contract_periods/default_schedule_generator.rb
ethax-ross: I think we should raise here instead of returning, something like:

```
AlreadyScheduledError, "The contract period already has schedules"
ContractPeriodStartedError, "Contract periods cannot be scheduled after they have started"
```

I don't think we should catch these errors; the 'already scheduled' error shouldn't be hit because we will only call this on creation and we can put validation in place to ensure only contract periods that start in the future can be created.

app/services/contract_periods/default_schedule_generator.rb
```diff
@@ -0,0 +1,119 @@
...
+    end
+
+    # @return [Boolean]
+    def already_started?
+      contract_period.started_on_or_before_today?
+    end
+
+    # @return [Boolean]
+    def cloneable?
+      previous_contract_period&.schedules&.any? { |schedule| schedule.milestones.any? }
```
ethax-ross: Would making it `cloneable?` only if the `previous_contract_period` exists be simpler and more expected? There will always be a schedule as it stands, but if there's not in the future I expect the default assumption would be 'we'll do that next year as well'?

app/services/contract_periods/default_schedule_generator.rb
```diff
@@ -0,0 +1,119 @@
...
+      :cloned
+    end
+
+    # @param identifier [String] e.g. "ecf-standard-september"
+    # @return [Date]
+    def schedule_start_date(identifier)
+      case identifier
+      when /september/ then Date.new(contract_period.year, 6, 1)
+      when /january/ then Date.new(next_period_year, 1, 1)
+      when /april/ then Date.new(next_period_year, 4, 1)
```
ethax-ross: Worth a `raise` on `else` if we encounter a new/unexpected identifier?

app/services/contract_periods/default_schedule_generator.rb
```diff
@@ -0,0 +1,119 @@
...
+    # @return [Symbol] :created, :cloned, :already_started, :already_scheduled
+    def schedule!
+      return :already_started if already_started?
+      return :already_scheduled if already_scheduled?
+
+      ActiveRecord::Base.transaction do
+        if cloneable?
+          clone_previous_schedules_and_milestones
+        else
+          create_default_schedules_and_milestones
```
ethax-ross: I'm on the fence if we should include creating defaults 🤔 I wouldn't expect a new contract period to be spun up with a gap/no previous contract period in the system. I'm leaning towards removing this just to avoid the extra complexity but open to opinions (if we do it here it stands that we should also do it for contracts, statements and delivery partnerships which might all be more complicated than schedules).

ethax-ross (commented):
Thanks PDH, looking good I've just left some thoughts

app/services/contract_periods/default_schedule_generator.rb
```diff
@@ -0,0 +1,119 @@
+module ContractPeriods
+  # Service creates Schedule and Milestone records and duplicates from the previous year if they exist.
+  #
+  # Milestone#milestone_date acts as the deadline for declarations
+  #
+  class DefaultScheduleGenerator
```
craigmdavidson: I ended up going with **SeedFromPrevious** (i.e. in this case **ContractPeriods::SeedFromPrevious**) which I think reveals intent slightly better?

app/services/contract_periods/default_schedule_generator.rb
```diff
@@ -0,0 +1,119 @@
...
+
+    # @return [Boolean]
+    def cloneable?
+      previous_contract_period&.schedules&.any? { |schedule| schedule.milestones.any? }
+    end
+
+    # @return [ContractPeriod, nil]
+    def previous_contract_period
+      @previous_contract_period ||= ContractPeriod.find_by(year: previous_period_year)
+    end
```
craigmdavidson: Maybe avoid proliferation of tiny methods? consider:
```
  def previous_contract_period
    return if current_contract_period&.started_on.blank?

    @previous_contract_period ||= ContractPeriod.containing_date(current_contract_period.started_on - 1.day)
  end
```

app/services/contract_periods/default_schedule_generator.rb
```diff
@@ -0,0 +1,119 @@
...
+
+    def initialize(contract_period:)
+      raise ArgumentError, "Contract period is required" if contract_period.nil?
+
+      @contract_period = contract_period
+    end
+
+    # @return [Symbol] :created, :cloned, :already_started, :already_scheduled
+    def schedule!
+      return :already_started if already_started?
```
craigmdavidson: Do we need micro-methods when the implementation that micro-method implementation is `contract_period.schedules.any?` it doesn't express intent any more clearly?

app/services/contract_periods/default_schedule_generator.rb
```diff
@@ -0,0 +1,119 @@
...
+    end
+
+    # @return [Integer]
+    def next_period_year
+      contract_period.year + 1
+    end
+
+    # @param date [Date]
+    # @return [Date, nil]
+    def advance_year(date)
```
craigmdavidson: Does this method buy us anything over inlining it - I'm not sure there is any extra intention revealed?

app/services/contract_periods/default_schedule_generator.rb
```diff
@@ -0,0 +1,119 @@
...
+    def initialize(contract_period:)
+      raise ArgumentError, "Contract period is required" if contract_period.nil?
+
+      @contract_period = contract_period
+    end
+
+    # @return [Symbol] :created, :cloned, :already_started, :already_scheduled
+    def schedule!
+      return :already_started if already_started?
+      return :already_scheduled if already_scheduled?
```
craigmdavidson: > If a Schedule has already been created for the ContractPeriod we should raise an error if service the action is attempted again.

🤔 Do we not want to raise rather than just return here?

craigmdavidson (commented):
Looks good so far. A few nitpicks.

app/services/contract_periods/default_schedule_generator.rb
peterdavidhamilton: Happy to rename.

app/services/contract_periods/default_schedule_generator.rb
peterdavidhamilton: That's what I wanted suggested names for 👍🏼

app/models/schedule.rb
```diff
@@ -16,6 +16,7 @@ class Schedule < ApplicationRecord
        ].index_by(&:itself),
        validate: { message: "Choose an identifier from the list" }
 
+  STANDARD_SCHEDULE_IDENTIFIERS = identifiers.keys.grep(/standard/).freeze
```
craigmdavidson: This isn't used anymore?

app/services/contract_periods/seed_from_previous.rb
```diff
@@ -0,0 +1,75 @@
+module ContractPeriods
+  # Service creates Schedule and Milestone records and duplicates from the previous year if they exist.
+  #
+  # Milestone#milestone_date acts as the deadline for declarations
+  #
+  class SeedFromPrevious
+    DECLARATION_TYPES_FOR_STANDARD = %w[
```
craigmdavidson: This isn't used anymore?

app/services/contract_periods/seed_from_previous.rb
```diff
@@ -0,0 +1,75 @@
...
+
+    # @return [Integer]
+    def next_period_year
+      contract_period.year + 1
+    end
+
+    # @param identifier [String] e.g. "ecf-standard-september"
+    # @raise [UnknownIdentifierError]
+    # @return [Date]
+    def schedule_start_date(identifier)
```
craigmdavidson: This doesn't seem to be used anymore?

app/services/contract_periods/seed_from_previous.rb
```diff
@@ -0,0 +1,75 @@
...
+  # Milestone#milestone_date acts as the deadline for declarations
+  #
+  class SeedFromPrevious
+    DECLARATION_TYPES_FOR_STANDARD = %w[
+      started retained-1 retained-2 retained-3 retained-4 completed
+    ].freeze
+
+    class AlreadyScheduledError < StandardError; end
+    class ContractPeriodStartedError < StandardError; end
+    class UnknownIdentifierError < StandardError; end
```
craigmdavidson: This is only ever called by `schedule_start_date` which is not called anymore?

app/services/contract_periods/seed_from_previous.rb
```diff
@@ -0,0 +1,75 @@
...
+
+  private
+
+    # @return [ContractPeriod, nil]
+    def previous_contract_period
+      @previous_contract_period ||= ContractPeriod.find_by(year: contract_period.year - 1)
+    end
+
+    # @return [Integer]
+    def next_period_year
```
craigmdavidson: This is only ever called by schedule_start_date which is not called anymore?

spec/services/contract_periods/seed_from_previous_spec.rb
```diff
@@ -0,0 +1,144 @@
+RSpec.describe ContractPeriods::SeedFromPrevious, type: :model do
```
craigmdavidson: This isn't a `:model`?

app/services/contract_periods/seed_from_previous.rb
```diff
@@ -0,0 +1,75 @@
...
+      @contract_period = contract_period
+    end
+
+    # @raise [AlreadyScheduledError, ContractPeriodStartedError]
+    # @return [Symbol] :no_previous_contract_period, :scheduled
+    def schedule!
+      raise AlreadyScheduledError, "The contract period already has schedules" if contract_period.schedules.any?
+      raise ContractPeriodStartedError, "Contract periods cannot be scheduled after they have started" if contract_period.started_on_or_before_today?
+
+      return :no_previous_contract_period if previous_contract_period.blank?
```
craigmdavidson: Why would we do return rather than raise here? This means the client has to both handle errors _and_ check a response code?  

If we raised we would be consistent enough not to worry about a return code?

craigmdavidson (changes requested):
Looks good! 
There is a bit of dead code following the recent changes?

app/services/contract_periods/seed_from_previous.rb
```diff
@@ -0,0 +1,55 @@
+module ContractPeriods
+  # Service creates Schedule and Milestone records and duplicates from the previous year if they exist.
+  #
+  # Milestone#milestone_date acts as the deadline for declarations
+  #
+  class SeedFromPrevious
+    class AlreadyScheduledError < StandardError; end
+    class ContractPeriodStartedError < StandardError; end
+    class NoPreviousContractPeriodError < StandardError; end
+    class UnknownIdentifierError < StandardError; end
```
craigmdavidson: Stray unused error, I think?

craigmdavidson (approved):
LGTM!

spec/services/contract_periods/seed_from_previous_spec.rb
```diff
@@ -0,0 +1,147 @@
...
+      end
+    end
+
+    context "with a previous contract period" do
+      let(:previous_contract_period) do
+        FactoryBot.create(:contract_period,
+                          year: previous_year)
+      end
+
+      let!(:previous_schedule_september) do
```
ethax-ross, line 39: Worth adding a second schedule to cover migrating multiple?

ethax-ross (approved):
Looks good 👍 lets align on the method for finding the previous contract period but that can maybe be picked up in Craig's PR

spec/services/contract_periods/seed_from_previous_spec.rb
peterdavidhamilton: I've deleted too much :)

ethax-ross (approved):

PR #2918, opened 2026-05-14
Add ActiveLeadProvider::SeedFromPrevious service

craigmdavidson:
### Context

resolves https://github.com.mcas.ms/DFE-Digital/register-ects-project-board/issues/3938

Service to seed an ActiveLeadProvider from the previous contract_period's relevant ActiveLeadProvider.

### DIFF
--- a/app/services/active_lead_providers/seed_from_previous.rb
+++ b/app/services/active_lead_providers/seed_from_previous.rb
@@ -114,8 +114,8 @@ private
         year: new_year,
         # Statements run on a fixed monthly cycle: the deadline is the last day
         # of the preceding month, and payment is made on the 25th of the month.
-        deadline_date: Date.new(new_year, previous_statement.month, 1).prev_day,
-        payment_date: Date.new(new_year, previous_statement.month, 25),
+        deadline_date: previous_statement.deadline_date.years_since(year_offset),
+        payment_date: previous_statement.payment_date.years_since(year_offset),
         fee_type: previous_statement.fee_type,
         status: :open
       )
--- a/app/services/active_lead_providers/seed_from_previous.rb
+++ b/app/services/active_lead_providers/seed_from_previous.rb
@@ -1,0 +1,124 @@
+# Seeds a newly-created active lead provider with a copy of its previous
+# contract period's setup: delivery partnerships, plus a single new contract
+# (inc. fee structures, bands and statements) — based on the previous contract
+# that owned the latest statement, and carrying every previous statement
+# rolled forward to the new period's year.
+#
+# It Errors if it cannot do this.
+class ActiveLeadProviders::SeedFromPrevious
+  class PreviousActiveLeadProviderError < StandardError; end
+  class AlreadyPopulatedError < StandardError; end
+
+  attr_reader :active_lead_provider
+
+  delegate :lead_provider, to: :active_lead_provider
+  delegate :contract_period, to: :active_lead_provider, prefix: :current
+  delegate :lead_provider_delivery_partnerships, :contracts, :statements,
+           to: :previous_activation, prefix: :previous
+
+  def initialize(active_lead_provider:)
+    raise ArgumentError, "active_lead_provider is required" if active_lead_provider.blank?
+
+    @active_lead_provider = active_lead_provider
+  end
+
+  def call
+    raise PreviousActiveLeadProviderError, "no previous active_lead_provider found for #{active_lead_provider.id}" if previous_activation.blank?
+    if previous_activation_empty?
+      raise PreviousActiveLeadProviderError,
+            "active_lead_provider for #{active_lead_provider.id} is missing previous delivery partnerships, contracts or statements"
+    end
+    raise AlreadyPopulatedError, "active_lead_provider #{active_lead_provider.id} already has data" if active_lead_provider_populated?
+
+    # This is a large graph, so let's make all or nothing...
+    ActiveRecord::Base.transaction do
+      create_new_delivery_partnerships
+      create_new_contract
+    end
+  end
+
+  def previous_activation
+    @previous_activation ||= previous_contract_period&.active_lead_providers&.find_by(lead_provider:)
+  end
+
+  def previous_contract_period
+    return if current_contract_period.blank?
+
+    @previous_contract_period ||= ContractPeriod.find_by(year: current_contract_period.year - 1)
+  end
+
+  def previous_latest_contract
+    @previous_latest_contract ||= previous_statements.order(year: :desc, month: :desc).first.contract
+  end
+
+private
+
+  def previous_activation_empty?
+    previous_lead_provider_delivery_partnerships.empty? ||
+      previous_contracts.empty? || previous_statements.empty?
+  end
+
+  def active_lead_provider_populated?
+    active_lead_provider.lead_provider_delivery_partnerships.any? ||
+      active_lead_provider.contracts.any?
+  end
+
+  def create_new_delivery_partnerships
+    previous_lead_provider_delivery_partnerships.each do |previous_partnership|
+      active_lead_provider.lead_provider_delivery_partnerships.create!(delivery_partner: previous_partnership.delivery_partner)
+    end
+  end
+
+  def create_new_contract
+    # Bands and fee structures validate by querying the database for their
+    # already-persisted siblings (see create_banded_fee_structure), so the graph
+    # has to be created as it's built — assembling it in memory and saving once
+    # would run those validations before the siblings exist.
+    previous_contract = previous_latest_contract
+
+    active_lead_provider.contracts.create!(
+      contract_type: previous_contract.contract_type,
+      ecf_contract_version: previous_contract.ecf_contract_version,
+      ecf_mentor_contract_version: previous_contract.ecf_mentor_contract_version,
+      banded_fee_structure: create_banded_fee_structure(previous_contract.banded_fee_structure),
+      flat_rate_fee_structure: create_flat_rate_fee_structure(previous_contract.flat_rate_fee_structure),
+      statements: build_new_statements,
+      vat_rate: previous_contract.vat_rate
+    )
+  end
+
+  def create_banded_fee_structure(previous_fee_structure)
+    return unless previous_fee_structure
+
+    # Band's band_consistency_across_active_lead_provider calls .reload on its parent
+    # banded_fee_structure, so the parent must be inserted before any newly created band validates.
+    previous_fee_structure.dup.tap do |new_fee_structure|
+      new_fee_structure.save!
+      previous_fee_structure.bands.each { |band| new_fee_structure.bands << band.dup }
+    end
+  end
+
+  def create_flat_rate_fee_structure(previous_fee_structure)
+    return unless previous_fee_structure
+
+    previous_fee_structure.dup.tap(&:save!)
+  end
+
+  def build_new_statements
+    year_offset = current_contract_period.year - previous_contract_period.year
+
+    previous_statements.map do |previous_statement|
+      new_year = previous_statement.year + year_offset
+      Statement.new(
+        month: previous_statement.month,
+        year: new_year,
+        # Statements run on a fixed monthly cycle: the deadline is the last day
+        # of the preceding month, and payment is made on the 25th of the month.
+        deadline_date: Date.new(new_year, previous_statement.month, 1).prev_day,
+        payment_date: Date.new(new_year, previous_statement.month, 25),
+        fee_type: previous_statement.fee_type,
+        status: :open
+      )
+    end
+  end
+end
--- a/app/services/active_lead_providers/seed_from_previous.rb
+++ b/app/services/active_lead_providers/seed_from_previous.rb
@@ -37,6 +37,8 @@ class ActiveLeadProviders::SeedFromPrevious
     end
   end
 
+private
+
   def previous_activation
     @previous_activation ||= previous_contract_period&.active_lead_providers&.find_by(lead_provider:)
   end
@@ -50,8 +52,6 @@ class ActiveLeadProviders::SeedFromPrevious
   def previous_latest_contract
     @previous_latest_contract ||= previous_statements.order(year: :desc, month: :desc).first.contract
   end
-
-private
 
   def previous_activation_empty?
     previous_lead_provider_delivery_partnerships.empty? ||
--- a/app/services/active_lead_providers/seed_from_previous.rb
+++ b/app/services/active_lead_providers/seed_from_previous.rb
@@ -112,8 +112,6 @@ private
       Statement.new(
         month: previous_statement.month,
         year: new_year,
-        # Statements run on a fixed monthly cycle: the deadline is the last day
-        # of the preceding month, and payment is made on the 25th of the month.
         deadline_date: previous_statement.deadline_date.years_since(year_offset),
         payment_date: previous_statement.payment_date.years_since(year_offset),
         fee_type: previous_statement.fee_type,
--- a/spec/services/active_lead_providers/seed_from_previous_spec.rb
+++ b/spec/services/active_lead_providers/seed_from_previous_spec.rb
@@ -1,0 +1,210 @@
+RSpec.describe ActiveLeadProviders::SeedFromPrevious do
+  subject(:service) { described_class.new(active_lead_provider: teach_first_activation_2026) }
+
+  # given a new contract_period...
+  let!(:contract_period_2026) { FactoryBot.create(:contract_period, :with_schedules, year: 2026, mentor_funding_enabled: true) }
+  # and an old contract_period...
+  let!(:contract_period_2025) { FactoryBot.create(:contract_period, :with_schedules, year: 2025) }
+
+  # given a lead provider, active in both periods...
+  let(:teach_first) { FactoryBot.create(:lead_provider, name: "Teach First") }
+  let(:teach_first_activation_2026) { FactoryBot.create(:active_lead_provider, lead_provider: teach_first, contract_period: contract_period_2026) }
+  let!(:teach_first_activation_2025) { FactoryBot.create(:active_lead_provider, lead_provider: teach_first, contract_period: contract_period_2025) }
+
+  describe "finding the appropriate previous contract_period and active_lead_provider" do
+    # We don't want to find earlier contract_periods..
+    let!(:contract_period_2024) { FactoryBot.create(:contract_period, :with_schedules, year: 2024) }
+    let!(:contract_period_2023) { FactoryBot.create(:contract_period, :with_schedules, year: 2023) }
+
+    # We don't want to find earlier activations..
+    let!(:teach_first_activation_2024) { FactoryBot.create(:active_lead_provider, lead_provider: teach_first, contract_period: contract_period_2024) }
+
+    # We don't want to find other lead_providers..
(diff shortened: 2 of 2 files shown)


spec/services/active_lead_providers/seed_from_previous_spec.rb
```diff
@@ -0,0 +1,180 @@
...
+  subject(:service) { described_class.new(active_lead_provider: teach_first_activation_2026) }
+
+  # given a new contract_period...
+  let!(:contract_period_2026) { FactoryBot.create(:contract_period, :with_schedules, year: 2026, mentor_funding_enabled: true) }
+  # and an old contract_period...
+  let!(:contract_period_2025) { FactoryBot.create(:contract_period, :with_schedules, year: 2025) }
+
+  # given a lead provider, active in both periods...
+  let(:teach_first) { FactoryBot.create(:lead_provider, name: "Teach First") }
+  let(:teach_first_activation_2026) { FactoryBot.create(:active_lead_provider, lead_provider: teach_first, contract_period: contract_period_2026) }
```
craigmdavidson, line 11: Following on from the comment above about model naming, for clarity of intent, I use **activation** for the noun in the spec.

app/services/active_lead_providers/seed_from_previous.rb
```diff
@@ -0,0 +1,116 @@
...
+    raise AlreadyPopulatedError, "active_lead_provider #{active_lead_provider.id} already has data" if active_lead_provider_populated?
+
+    # This is a large graph, so let's make all or nothing...
+    ActiveRecord::Base.transaction do
+      create_new_delivery_partnerships
+      create_new_contracts
+    end
+  end
+
+  def previous_activation
```
craigmdavidson, line 42: I use `previous_activation` because `previous_active_lead_provider` doesn't really make sense. 
**ActiveLeadProvider** implies an `is_a` relationship, but the **LeadProvider** is always the same, it may be active or not for a given **ContractPeriod**.  The relationship is really an **Activation**. 

**ActiveLeadProvider** is a poorly named model but that is out of scope of this PR.

app/services/active_lead_providers/seed_from_previous.rb
```diff
@@ -0,0 +1,116 @@
...
+      active_lead_provider.lead_provider_delivery_partnerships.create!(delivery_partner: previous_partnership.delivery_partner)
+    end
+  end
+
+  def create_new_contracts
+    # Bands and fee structures validate by querying the database for their
+    # already-persisted siblings (see create_banded_fee_structure), so the graph
+    # has to be created as it's built — assembling it in memory and saving once
+    # would run those validations before the siblings exist.
+    previous_contracts.each do |previous_contract|
```
craigmdavidson: https://github.com.mcas.ms/DFE-Digital/register-early-career-teachers-public/blob/main/app/models/contract/banded_fee_structure/band.rb#L99

👀 see the `reload`!

```
    unique_bands_by_index = bands_for_active_lead_provider
      .group_by { it.banded_fee_structure.reload.bands.index(it) || it.banded_fee_structure.bands.count }
      .transform_values { |bands| bands.map { it.attributes.except(*attributes_to_ignore) }.uniq }
```

in `validate :band_consistency_across_active_lead_provider` 😱

app/services/active_lead_providers/seed_from_previous.rb
```diff
@@ -0,0 +1,116 @@
...
+
+  def initialize(active_lead_provider:)
+    raise ArgumentError, "active_lead_provider is required" if active_lead_provider.blank?
+
+    @active_lead_provider = active_lead_provider
+  end
+
+  def call
+    raise PreviousActiveLeadProviderError, "no previous active_lead_provider found for #{active_lead_provider.id}" if previous_activation.blank?
+    raise PreviousActiveLeadProviderError, "previous active_lead_provider for #{active_lead_provider.id} is empty" if previous_activation_empty?
```
ethax-ross: It took me a second to realise what this meant; is it worth being more specific? Something like:

"active_lead_provider for #{active_lead_provider.id} is missing previous delivery partnerships, contracts or statements"

app/services/active_lead_providers/seed_from_previous.rb
ethax-ross: We only want to bring across the _latest_ contract for each lead provider; so the one attached to the `Statement` with the latest `deadline_date` or `payment_date`. Typically each lead provider will start the registration year with a single contract spanning all their statements and somewhere down the road changes are introduced and a chunk of statements move onto a new contract.

app/services/active_lead_providers/seed_from_previous.rb
```diff
@@ -0,0 +1,116 @@
...
+
+  def create_new_contracts
+    # Bands and fee structures validate by querying the database for their
+    # already-persisted siblings (see create_banded_fee_structure), so the graph
+    # has to be created as it's built — assembling it in memory and saving once
+    # would run those validations before the siblings exist.
+    previous_contracts.each do |previous_contract|
+      active_lead_provider.contracts.create!(
+        contract_type: previous_contract.contract_type,
+        ecf_contract_version: previous_contract.ecf_contract_version,
```
ethax-ross: These `ecf_*` attributes should be removed with the work to pull out all the migration code; I'll check in with the DM team to check its not been missed

app/services/active_lead_providers/seed_from_previous.rb
ethax-ross: Yeah this method is a bit of a shocker 😞 some of the rules we encountered around contracts only transpired after we had finished modelling it (we had to try and infer everything from ECF). We can hopefully revisit this when we get a bit of breathing room.

app/services/active_lead_providers/seed_from_previous.rb
```diff
@@ -0,0 +1,116 @@
...
+    year_offset = current_contract_period.year - previous_contract_period.year
+
+    previous_contract.statements.map do |previous_statement|
+      new_year = previous_statement.year + year_offset
+      Statement.new(
+        month: previous_statement.month,
+        year: new_year,
+        # Statements run on a fixed monthly cycle: the deadline is the last day
+        # of the preceding month, and payment is made on the 25th of the month.
+        deadline_date: Date.new(new_year, previous_statement.month, 1).prev_day,
```
ethax-ross: Would it be simpler/more robust to use the `year_offset` here? I guess the question is if someone tweaks a payment/deadline date during a contract period to _not_ adhere to this logic, would we want carry that over 🤔 

`deadline_date: previous_statement.deadline_date.years_since(year_offset)`

app/services/active_lead_providers/seed_from_previous.rb
```diff
@@ -0,0 +1,116 @@
...
+      create_new_delivery_partnerships
+      create_new_contracts
+    end
+  end
+
+  def previous_activation
+    @previous_activation ||= previous_contract_period&.active_lead_providers&.find_by(lead_provider:)
+  end
+
+  def previous_contract_period
```
ethax-ross, line 46: Having `previous_contract_period` and `previous_activation` public solely for the tests is a bit of a smell

app/services/active_lead_providers/seed_from_previous.rb
ethax-ross: That's a good shout; would you mind raising that as an issue on the repo?

spec/services/active_lead_providers/seed_from_previous_spec.rb
```diff
@@ -0,0 +1,180 @@
...
+      end
+    end
+
+    describe "contracts" do
+      let(:previous_contracts) { teach_first_activation_2025.contracts }
+      let(:new_contracts) { teach_first_activation_2026.contracts }
+
+      it "builds new contracts based on previous contracts" do
+        expect(new_contracts.size).to eq previous_contracts.size
+        expect(new_contracts.map(&:lead_provider).uniq).to eq [teach_first]
```
ethax-ross: Take it or leave it, its just my preference 😅 

`expect(new_contracts.map(&:lead_provider)).to all(eq(teach_first))`

spec/services/active_lead_providers/seed_from_previous_spec.rb
```diff
@@ -0,0 +1,180 @@
...
+
+    it "raises an error rather than duplicating data" do
+      expect { service.call }
+        .to raise_error(described_class::AlreadyPopulatedError, /active_lead_provider \d+ already has data/)
+    end
+  end
+
+private
+
+  def create_subordinate_records(active_lead_provider)
```
ethax-ross, line 177: Is it worth shifting any of this into traits?

app/services/active_lead_providers/seed_from_previous.rb
```diff
@@ -0,0 +1,116 @@
...
+    return if current_contract_period&.started_on.blank?
+
+    @previous_contract_period ||= ContractPeriod.containing_date(current_contract_period.started_on - 1.day)
```
ethax-ross: This should be fine, but looking at `ContractPeriod` we have validation/enforce `no_overlaps` at the DB level -- I wonder if we should add a `no_gaps` validation rule for belt and braces?

spec/services/active_lead_providers/seed_from_previous_spec.rb
```diff
@@ -0,0 +1,180 @@
...
+
+  context "when the previous activation has no subordinate data" do
+    it "raises an error" do
+      expect { service.call }
+        .to raise_error(described_class::PreviousActiveLeadProviderError, /previous active_lead_provider for \d+ is empty/)
+    end
+  end
+
+  context "when the previous activation is only partially populated" do
+    before { FactoryBot.create_list(:lead_provider_delivery_partnership, 3, active_lead_provider: teach_first_activation_2025) }
```
ethax-ross: Worth expanding this to cover all the branches of the condition (so contracts and statements as well)?

ethax-ross (commented):
Looks great 👏 just some minor thoughts/suggestions

app/services/active_lead_providers/seed_from_previous.rb
craigmdavidson: A smell potentially, but not something bad (I would argue)... 

These are methods that should really be public methods on the models (`contract_period.previous`, and `active_lead_provider.previous`). But that is something we do not want to do?

We could use `send(&:previous_contract_period)`, etc, but this is horrible, if we need to see these values they should be public - and these are service objects, not actually methods, so that isn't a problem.

We could extract micro-service objects (**ContractPeriod::PreviousContractPeriod** etc), but that is an express path to spaghetti / procedural programming.

We could skip the tests, as having the correct seeded data implicitly asserts the right source for the seeds. Which could be a reasonable position, but it is nice to ascertain that we identify the correct contract_period and correct activation?

Or have additional public finder methods on the service class. Which I think with the correct option off the table (public methods on the model) this is the best option remaining?

app/services/active_lead_providers/seed_from_previous.rb
craigmdavidson: :+1: I can remove this mapping if you think there is no risk?

app/services/active_lead_providers/seed_from_previous.rb
craigmdavidson: The current implementation self corrects.

> would we want carry that over 🤔

Would that ever happen? I'm assuming not❓ 🤷 -

spec/services/active_lead_providers/seed_from_previous_spec.rb
craigmdavidson: I wouldn't as I would look to do this when there are multiple specs requiring this? Otherwise we can pollute the factory files with piles of one off situations. 

Kent Beck's three strikes rule is useful here I think?

Full disclosure though: I find rSpec with FactoryBot a poorly thought out, slow, convoluted mess compared to Minitest with fixtures - so I can't always trust my instincts! 😆  

So if you want trait's, I can do that?

app/services/active_lead_providers/seed_from_previous.rb
ethax-ross: > These are methods that should really be public methods on the models 

Agreed. The issue is we have very slim models as an intentional design choice and small service objects as a bit of an established pattern ([Teachers::Name](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/services/teachers/name.rb) is an example). Branching out from that is going to involve a wider discussion, but maybe its something we need to broach?

> if we need to see these values they should be public

Why do we need to see them, though, aside from the tests? this is my main sticking point; the tests shouldn't be dictating the public API of the service.

> We could skip the tests, as having the correct seeded data implicitly asserts the right source for the seeds.

This would have been my approach; the contract periods we expect to be involved in the service are defined and used in the assertions already.

app/services/active_lead_providers/seed_from_previous.rb
ethax-ross: I caught up with the DM team and it sounds like they may be removed at a later date/the link back to ECF is still useful until they are confident there are no more data issues that resulted from the migration. Lets leave this as-is for now, mainly because it looks like these are mandatory fields so making them nullable would be extra work for short-lived code 👍

app/services/active_lead_providers/seed_from_previous.rb
ethax-ross: It depends really; it only self corrects if that's the desired behaviour 😅 I think this is one for @andycrawshaw to weigh in on -- Andy, if a statement date is changed in 2025 so that it no longer follows the established pattern (last day of previous month/25th of statement month) would we want to retain those dates in 2026 or reset them? have statement dates ever been changed in the past? (we can run a query on snapshot to check this if you're not sure)?

app/services/active_lead_providers/seed_from_previous.rb
craigmdavidson: https://github.com.mcas.ms/DFE-Digital/register-early-career-teachers-public/issues/2921

app/services/active_lead_providers/seed_from_previous.rb
craigmdavidson: > Looks good 👍 lets align on the method for finding the previous contract period but that can maybe be picked up in Craig's PR

:+1: Okay, so if I implement this in a similar way to @peterdavidhamilton in his PR (https://github.com.mcas.ms/DFE-Digital/register-early-career-teachers-public/pull/2913), the "no gaps" problem goes away (from this PR at least).

```
  def previous_contract_period
    return if current_contract_period.blank?

    @previous_contract_period ||= ContractPeriod.find_by(year: current_contract_period.year - 1)
  end
```

spec/services/active_lead_providers/seed_from_previous_spec.rb
craigmdavidson: Done

app/services/active_lead_providers/seed_from_previous.rb
craigmdavidson: Done

spec/services/active_lead_providers/seed_from_previous_spec.rb
craigmdavidson: Left as is for now, we can do this if we have other test specs requiring the fixture data.

spec/services/active_lead_providers/seed_from_previous_spec.rb
craigmdavidson: This hasn't become moot, as the change to a single contract has changed the assertion to `expect(new_contract).to eq teach_first`.

app/services/active_lead_providers/seed_from_previous.rb
craigmdavidson: Done

app/services/active_lead_providers/seed_from_previous.rb
craigmdavidson: Done.

app/services/active_lead_providers/seed_from_previous.rb
craigmdavidson: Leaving as is.

app/services/active_lead_providers/seed_from_previous.rb
```diff
@@ -0,0 +1,124 @@
...
+
+  def build_new_statements
+    year_offset = current_contract_period.year - previous_contract_period.year
+
+    previous_statements.map do |previous_statement|
+      new_year = previous_statement.year + year_offset
+      Statement.new(
+        month: previous_statement.month,
+        year: new_year,
+        # Statements run on a fixed monthly cycle: the deadline is the last day
```
ethax-ross: Is this comment a bit irrelevant now?

app/services/active_lead_providers/seed_from_previous.rb
```diff
@@ -0,0 +1,124 @@
...
+  delegate :lead_provider_delivery_partnerships, :contracts, :statements,
+           to: :previous_activation, prefix: :previous
+
+  def initialize(active_lead_provider:)
+    raise ArgumentError, "active_lead_provider is required" if active_lead_provider.blank?
+
+    @active_lead_provider = active_lead_provider
+  end
+
+  def call
```
ethax-ross, line 25: I don't feel all that strongly - and I know `call` is fairly idiomatic in Ruby service objects that only have one public method - but IMO method names should reflect what they do, so `seed!` or `seed_from_previous!`.

My main gripe is when a service object is inevitably assigned to a poorly named instance variable and you end up once removed from the meaning, e.g. `service.call` whereas `service.seed_from_previous!` still gives you a clue/alludes to the behaviour.

ethax-ross (approved):
Looks great! 👍

app/services/active_lead_providers/seed_from_previous.rb
craigmdavidson: I would push back _lightly_ on this one...

> My main gripe is when a service object is inevitably assigned to a poorly named instance variable and you end up once removed from the meaning, e.g. service.call whereas service.seed_from_previous! still gives you a clue/alludes to the behaviour.

A service object, isn't really an object in the normal sense - it is in essence a method (or rather a **Method** object) itself (it is an implementation of Kent Beck's Method Object pattern). It shouldn't be assigned to a variable - so the naming of the non-assignment is generally moot:
I.e. `ActiveLeadProviders::SeedFromPrevious.new(active_lead_provider:).call`

But the `call` name specifically: 
1. As the standard for invoking ruby **Method** objects, highlights to the reader that this is a service object and not an OO model poRo; and 
2. leaves the door open to additional affordances (e.g. shortening calls to  `ActiveLeadProviders::SeedFromPrevious.(active_lead_provider:)` (i.e. when we realise that doing this means it get a whole load of extracts by having ApplicationService or something as a superclass).

FWIW: I am _not a fan_ of the extensive use of service objects as an alternative to properly designed objects, and think it is anti-pattern. But if we are going to use the pattern, when it is used it should be obvious.

But as I say, just pushing back lightly, if you are really not bothered I would prefer to leave it, if you are actually bothered, I am happy to change! 😄 

?

app/services/active_lead_providers/seed_from_previous.rb
craigmdavidson: Agreed. Done

app/services/active_lead_providers/seed_from_previous.rb
ethax-ross: I don't feel strongly about it; I agree that if it's not assigned to a variable it's not a problem, but in my experience it will inevitably happen 🙃 

Happy to leave it as-is 👍

PR #2920, opened 2026-05-14
Widen training card so JSON is less cramped

peteryates:
This summary list has no actions so we can disable the actions column which [allows the content to stretch right across](https://govuk-components.x-govuk.org/components/summary-list/#summary-lists-without-actions).

| Before | After |
| ------ | ------ |
| Screenshot From 2026-05-14 18-44-31 | Screenshot From 2026-05-14 18-44-17 |


avinhurry (approved):
Nice! I did not realise `actions: false` removed the reserved actions column 💡

PR #2923, opened 2026-05-15
[3843] Block registrations between 1st and 15th June

rwrrll:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/3843

- Introduce a `RegistrationWindow` class that defines a period during which registration is closed
- Add a concern to the appropriate controllers that redirects to an information page if the current date is within that window

### Guidance to review

> [!CAUTION]
> ~**Remove the second commit d0e3fc4f480f4f9c3eca7b7d570c1568ddc7020c before merging.** It's only there to make this easily testable on a review app, piggybacking on the existing method for testing API responses with a special header. It is not part of the work, doesn't need to be reviewed, and should not be merged!~

☝️  removed

### DIFF
--- a/app/controllers/schools/registration_window_closed_controller.rb
+++ b/app/controllers/schools/registration_window_closed_controller.rb
@@ -1,0 +1,7 @@
+module Schools
+  class RegistrationWindowClosedController < SchoolsController
+    def show
+      @reopens_on = RegistrationWindow.reopens_on.strftime("%-d %B")
+    end
+  end
+end
--- a/app/models/schools/registration_window.rb
+++ b/app/models/schools/registration_window.rb
@@ -1,0 +1,13 @@
+module Schools
+  class RegistrationWindow
+    CLOSED_PERIOD = Date.new(2026, 6, 1)..Date.new(2026, 6, 14)
+
+    def self.closed?
+      CLOSED_PERIOD.cover?(Date.current)
+    end
+
+    def self.reopens_on
+      CLOSED_PERIOD.end + 1.day
+    end
+  end
+end
--- a/app/views/schools/registration_window_closed/show.html.erb
+++ b/app/views/schools/registration_window_closed/show.html.erb
@@ -1,0 +1,17 @@
+<% page_data(
+     title: "You cannot register early career teachers (ECTs) or mentors until #{@reopens_on}"
+   ) %>
+
+<p class="govuk-body">Until <%= @reopens_on %> you cannot:</p>
+<ul class="govuk-list govuk-list--bullet">
+  <li>register new early career teachers (ECTs) and mentors</li>
+  <li>change ECT training programmes</li>
+</ul>
+
+<p class="govuk-body">If you want to register an ECT who's already started their induction you should:</p>
+<ol class="govuk-list govuk-list--number">
+  <li>Contact the ECT's lead provider so that they can set up the ECT's training programme.</li>
+  <li>Register the ECT in this service when registration opens.</li>
+</ol>
+
+<p class="govuk-body"><%= govuk_link_to("Back to ECTs", schools_ects_home_path) %></p>
--- a/app/views/schools/registration_window_closed/show.html.erb
+++ b/app/views/schools/registration_window_closed/show.html.erb
@@ -3,15 +3,21 @@
    ) %>
 
 <p class="govuk-body">Until <%= @reopens_on %> you cannot:</p>
-<ul class="govuk-list govuk-list--bullet">
-  <li>register new early career teachers (ECTs) and mentors</li>
-  <li>change ECT training programmes</li>
-</ul>
+<%= govuk_list(
+  [
+    "register new early career teachers (ECTs) and mentors",
+    "change ECT training programmes"
+  ],
+  type: :bullet
+) %>
 
 <p class="govuk-body">If you want to register an ECT who’s already started their induction you should:</p>
-<ol class="govuk-list govuk-list--number">
-  <li>Contact the ECT’s lead provider so that they can set up the ECT's training programme.</li>
-  <li>Register the ECT in this service when registration opens.</li>
-</ol>
+<%= govuk_list(
+  [
+    "Contact the ECT’s lead provider so that they can set up the ECT's training programme.",
+    "Register the ECT in this service when registration opens."
+  ],
+  type: :number
+) %>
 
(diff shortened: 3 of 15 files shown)


cpjmcquillan (commented):
I think this is a great start, but one thing we might want to address as part of this change (or later).

**Not hardcoding registration close/open dates**

Ideally we don't want to hardcode these dates, especially as we look towards a day (one day...) where the service doesn't require any developer input to maintain.

At the minute I'm spiking out how many things break when the `ContractPeriod` dates aren't continuous. Once that's done (and I've fixed anything that needs fixing) we should be able to do something like this:

```ruby
redirect_to schools_registration_window_closed_path unless ContractPeriod.current
```

and

```ruby
@reopens_on = ContractPeriod.closest_to(Date.current).merge(ContractPeriod.current_or_future).finished_on.to_fs(:govuk)
```

although something like `ContractPeriod.upcoming` might be a nice abstraction.

What do you think?

app/controllers/schools/registration_window_closed_controller.rb
```diff
@@ -0,0 +1,7 @@
+module Schools
+  class RegistrationWindowClosedController < SchoolsController
+    def show
+      @reopens_on = RegistrationWindow.reopens_on.strftime("%-d %B")
```
cpjmcquillan, line 4: Should we use `.to_fs(:govuk)` here? 🤔

app/models/schools/registration_window.rb
```diff
@@ -0,0 +1,13 @@
...
+  class RegistrationWindow
+    CLOSED_PERIOD = Date.new(2026, 6, 1)..Date.new(2026, 6, 14)
+
+    def self.closed?
+      CLOSED_PERIOD.cover?(Date.current)
+    end
+
+    def self.reopens_on
+      CLOSED_PERIOD.end + 1.day
+    end
```
cpjmcquillan, line 11: Added https://github.com/DFE-Digital/register-ects-project-board/issues/3994 to revisit this 👌🏼

app/views/schools/registration_window_closed/show.html.erb
```diff
@@ -0,0 +1,17 @@
...
+<p class="govuk-body">Until <%= @reopens_on %> you cannot:</p>
+<ul class="govuk-list govuk-list--bullet">
+  <li>register new early career teachers (ECTs) and mentors</li>
+  <li>change ECT training programmes</li>
+</ul>
+
+<p class="govuk-body">If you want to register an ECT who's already started their induction you should:</p>
+<ol class="govuk-list govuk-list--number">
+  <li>Contact the ECT's lead provider so that they can set up the ECT's training programme.</li>
+  <li>Register the ECT in this service when registration opens.</li>
```
cpjmcquillan: Should we use [`govuk_list`](https://govuk-components.x-govuk.org/helpers/list/) here?

app/views/schools/registration_window_closed/show.html.erb
```diff
@@ -0,0 +1,17 @@
...
+
+<p class="govuk-body">Until <%= @reopens_on %> you cannot:</p>
+<ul class="govuk-list govuk-list--bullet">
+  <li>register new early career teachers (ECTs) and mentors</li>
+  <li>change ECT training programmes</li>
+</ul>
+
+<p class="govuk-body">If you want to register an ECT who's already started their induction you should:</p>
+<ol class="govuk-list govuk-list--number">
+  <li>Contact the ECT's lead provider so that they can set up the ECT's training programme.</li>
```
cpjmcquillan:
```suggestion
  <li>Contact the ECT’s lead provider so that they can set up the ECT's training programme.</li>
```

app/views/schools/registration_window_closed/show.html.erb
```diff
@@ -0,0 +1,17 @@
...
+     title: "You cannot register early career teachers (ECTs) or mentors until #{@reopens_on}"
+   ) %>
+
+<p class="govuk-body">Until <%= @reopens_on %> you cannot:</p>
+<ul class="govuk-list govuk-list--bullet">
+  <li>register new early career teachers (ECTs) and mentors</li>
+  <li>change ECT training programmes</li>
+</ul>
+
+<p class="govuk-body">If you want to register an ECT who's already started their induction you should:</p>
```
cpjmcquillan:
```suggestion
<p class="govuk-body">If you want to register an ECT who’s already started their induction you should:</p>
```

spec/models/schools/registration_window_spec.rb
```diff
@@ -0,0 +1,52 @@
+describe Schools::RegistrationWindow do
+  let(:first_day_of_closure) { Date.new(2026, 6, 1) }
+  let(:last_day_of_closure) { Date.new(2026, 6, 14) }
+
+  describe ".closed?" do
+    subject { described_class.closed? }
```
cpjmcquillan: I think these tell a really clear story, but I wonder if we could make them even clearer to read if we did 

```ruby
subject { described_class }
```

and then things like

```ruby
it { is_expected.to be_closed }
```

What do you think? 💭

spec/support/shared_examples/registration_window_redirect.rb
```diff
@@ -0,0 +1,30 @@
+RSpec.shared_examples "a route blocked when the registration window is closed" do
+  around do |example|
+    travel_back
+    travel_to(date)
+    example.run
+  ensure
+    travel_back
+  end
```
cpjmcquillan, line 5: Do we need all this travelling or could we just do this?

```ruby
travel_to(date)
example.run
```

cpjmcquillan (approved):
LGTM!

app/controllers/schools/registration_window_closed_controller.rb
rwrrll: The designs specify without the year alas!

spec/models/schools/registration_window_spec.rb
rwrrll: Yeah nice!

app/views/schools/registration_window_closed/show.html.erb
rwrrll: Can do! Updated

spec/support/shared_examples/registration_window_redirect.rb
rwrrll: It was needed because of some interaction with other time travels... but no longer necessary it seems! 🥳

PR #2924, opened 2026-05-15
Fix Pagy error when navigating back to delivery partners index from lead provider partnerships page

ericaporter:
[Ticket ](https://github.com/DFE-Digital/register-early-career-teachers-public/issues/2917)

### Context
Navigating back to the delivery partners index using breadcrumbs after adding a lead provider partnership is erroring.

- Use compact_blank on backlink params in DeliveryPartnershipsController
- Add feature spec to assert no empty page params in redirect URL

### Guidance to review
Log in as an admin user
Navigate to Organisations → Delivery partners
Click on a delivery partner
Click "Add" next to a contract period year
Select a lead provider and click Confirm
Click the "Delivery partners" breadcrumb
Verify you land on the delivery partners index without a 500 error and the URL does not contain page= or q=


avinhurry (approved):
Looks good 👍🏽

cpjmcquillan (approved):

PR #2927, opened 2026-05-18
[3948] Add finance contract period create update pages

leandroalemao:
### Context

Ticket: [3948](https://github.com.mcas.ms/DFE-Digital/register-ects-project-board/issues/3948)

Finance user wants to manage contract periods for opening registration;

- Add pages to allow creating and updating contract periods. 

### Guidance to review

cpd-ec2-review-2927-web test teacherservices cloud_admin_finance_contract-periods_new_


app/controllers/admin/finance/contract_periods_controller.rb
```diff
@@ -12,19 +16,75 @@ def index
...
+    def new
+      @contract_period = ContractPeriod.new
+    end
+
+    def create
+      @contract_period = ContractPeriod.new(contract_period_params)
+
+      @contract_period.transaction do
+        if @contract_period.save
+          Events::Record.record_contract_period_added_event!(author: current_user, contract_period: @contract_period)
```
ethax-ross: I know we agreed a service was overkill for the querying, but I think we should probably have services for the various `Create`, `Update` and `Destroy` actions we're going to support in these screens to send the event and create/update/destroy the model 🤔

app/controllers/admin/finance/contract_periods_controller.rb
```diff
@@ -12,19 +16,80 @@ def index
...
+          finished_on
+          detailed_evidence_types_enabled
+          mentor_funding_enabled
+          uplift_fees_enabled
+        ]
+      )
+    end
 
+    def set_contract_period_flags
       @editable = !@contract_period.started_on_or_before_today?
```
ethax-ross: I think based on some recent discussions we may move some of these into methods on the model, but we can maybe address that at the end as a refactor

app/controllers/admin/finance/contract_periods_controller.rb
```diff
@@ -12,19 +16,80 @@ def index
...
 
-      @breadcrumbs = {
-        "Finance" => admin_finance_path,
-        "Contract periods" => admin_contract_periods_path,
-        @contract_period.year.to_s => nil,
-      }
+    def redirect_if_contract_period_not_editable
+      return if @editable
+
+      redirect_to(request.referer || admin_contract_periods_path)
```
ethax-ross: I wonder if its worth adding a quick flash message in to flag why they are redirected?

app/services/contract_periods/create.rb
```diff
@@ -0,0 +1,31 @@
...
+        record_event! or raise ActiveRecord::Rollback
+      end
+
+      contract_period
+    end
+
+  private
+
+    def record_event!
+      return false unless contract_period.persisted?
```
ethax-ross: Do we need this and the `or raise` given we're `save!` with a bang anyway?

app/services/contract_periods/update.rb
```diff
@@ -0,0 +1,33 @@
...
+    attr_reader :contract_period, :author, :params
+
+    def initialize(author:, contract_period:, params:)
+      @params = params
+      @contract_period = contract_period
+      @author = author
+    end
+
+    def update!
+      contract_period.assign_attributes(params)
```
ethax-ross, line 14: I think we should have the check in here to prevent editing a contract period that has started as well

ethax-ross (commented):
Thanks Leandro this is looking good, I haven't had chance to do a full review but wanted to give some initial thoughts/suggestions

app/services/contract_periods/create.rb
```diff
@@ -0,0 +1,31 @@
...
+      @author = author
+      @contract_period = ContractPeriod.new(params)
+    end
+
+    def create!
+      raise ActiveRecord::RecordInvalid, contract_period unless contract_period.valid?
+
+      ActiveRecord::Base.transaction do
+        contract_period.save!
+        record_event! or raise ActiveRecord::Rollback
```
ethax-ross: We have the `SeedFromPrevious` service in `main` now so we should be able to call it here: https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/services/contract_periods/seed_from_previous.rb

app/services/contract_periods/create.rb
```diff
@@ -0,0 +1,31 @@
...
+    def initialize(author:, params:)
+      @author = author
+      @contract_period = ContractPeriod.new(params)
+    end
+
+    def create!
+      raise ActiveRecord::RecordInvalid, contract_period unless contract_period.valid?
+
+      ActiveRecord::Base.transaction do
+        contract_period.save!
```
ethax-ross, line 14: We'll want newly created contract periods to be `enabled` by default I think -- the schema defaults it to `false`, but I expect all newly created contract periods will always be `enabeld`

peterdavidhamilton (approved):
👍🏼

spec/models/contract_period_spec.rb
```diff
@@ -242,4 +242,30 @@
       it { is_expected.to be_payments_frozen }
     end
   end
+
+  describe "#editable?" do
+    context "when the contract period has not started yet" do
+      let(:contract_period) { FactoryBot.create(:contract_period, started_on: 1.month.from_now, finished_on: 2.months.from_now) }
```
ethax-ross: Can we have a single `subject(:contract_period)` with only the `started_on` changing for each `context` and using `it { is_expected.to be_editable }`

app/services/contract_periods/create.rb
```diff
@@ -0,0 +1,32 @@
...
+      @contract_period = ContractPeriod.new(params.merge(enabled: true))
+    end
+
+    def create!
+      return false unless contract_period.valid?
+
+      ActiveRecord::Base.transaction do
+        contract_period.save!
+        record_event!
+        add_schedules_and_milestones!
```
ethax-ross: `seed_from_previous!`

spec/requests/admin/finance/contract_periods_spec.rb
```diff
@@ -1,79 +0,0 @@
-RSpec.describe "Admin finance contract periods", type: :request do
```
ethax-ross, line 1: Not sure why this was renamed? 🤔

spec/services/contract_periods/create_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+      it "creates an event with the expected values" do
+        freeze_time
+
+        subject.create!
+
+        perform_enqueued_jobs
+
+        last_event = Event.find_by(event_type: "contract_period_added")
+        contract_period = last_event.contract_period
+        expect(contract_period.year).to eql(1.year.from_now.year)
```
ethax-ross: `expect(contract_period).to have_attributes(...)`

Can we also explicitly cover that the contract period is created with `enabled: true`

spec/services/contract_periods/create_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+        expect(contract_period).to be_persisted
+      end
+
+      it "records a `contract_period_added` event" do
+        subject.create!
+
+        perform_enqueued_jobs
+
+        expect(Event.all.map(&:event_type)).to match_array(%w[contract_period_added])
+      end
```
ethax-ross, line 45: Can we add a spec to show it calls the `SeedFromPrevious` service?

app/controllers/admin/finance/contract_periods_controller.rb
```diff
@@ -12,19 +16,80 @@ def index
...
+        "Contract periods" => admin_contract_periods_path,
+        @contract_period.year.to_s => nil,
+      }
+    end
+
+    def new
+      @contract_period = ContractPeriod.new
+    end
+
+    def create
```
ethax-ross, line 30: I hit some errors when I was testing this in the UI; we need to handle the errors the `SeedFromPrevious` service can raise here gracefully, similar to here:

https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/controllers/admin/finance/active_lead_providers_controller.rb#L40-L43

ethax-ross (commented):
Thanks for picking up the changes PDH 🙏  it's looking good, I just hit some errors when testing we need to handle.

I'm not sure why but when I submit this contract period the logs show a `Contract periods cannot be scheduled after they have started` 500 error, could you take a look to see what's going on as I wouldn't expect it from the dates I'm using?

Image

PR #2935, opened 2026-05-18
[3939] Add schedule index for a contract period

peterdavidhamilton:
### Context

[Epic](https://github.com/DFE-Digital/register-ects-project-board/issues/3927) opening registration journey

Closes https://github.com/DFE-Digital/register-ects-project-board/issues/3939

We are not including milestones on this page yet, just a simple list of schedules

- [x] controller and view with supporting helper
- [x] request and view spec
- [x] inject feature flag config into request specs
- [x] link from the contract period page task list

### Guidance to review

cpd-ec2-review-2935-web test teacherservices cloud_admin_finance_contract-periods_2026_schedules (1)

```
Admin finance schedules
  GET /admin/finance/contract-periods/:id/schedules
    when authenticated as a non-finance DfE user
      requires authorisation with finance privileges
    with an authenticated non-DfE user
      requires authorisation
    when authenticated as a finance DfE user
      and the contract period is not yet open
        has an enabled button
      and the contract period is closed
        has a disabled button
        displays the schedules index page
        displays the schedules for the contract period
    when not authenticated
      redirects to sign in page
```

```
admin/finance/schedules/index.html.erb
  renders breadcrumbs
  has intro text
  when the current contract period has all possible schedules
    has a disabled add schedule button
  when the contract period has schedules
    lists their names
  when a contract period is open to amendments
    has an enabled add a schedule button
  when a new contract period is opened
    has no schedules
  when a contract period is closed to amendments
    has a disabled add schedule button
```

### DIFF
--- a/app/controllers/admin/finance/schedules_controller.rb
+++ b/app/controllers/admin/finance/schedules_controller.rb
@@ -2,6 +2,12 @@ module Admin::Finance
   class SchedulesController < Admin::Finance::BaseController
     before_action :set_contract_period
     def index
+      @breadcrumbs = {
+        "Finance" => admin_finance_path,
+        "Contract periods" => admin_contract_periods_path,
+        @contract_period.year.to_s => admin_contract_period_path(@contract_period),
+        "Schedules" => nil
+      }
     end
 
   private
--- a/app/controllers/admin/finance/schedules_controller.rb
+++ b/app/controllers/admin/finance/schedules_controller.rb
@@ -1,6 +1,12 @@
 module Admin::Finance
   class SchedulesController < Admin::Finance::BaseController
+    before_action :set_contract_period
     def index
+    end
+
+  private
+
+    def set_contract_period
       @contract_period = ContractPeriod.find(params[:contract_period_id])
     end
   end
--- a/app/controllers/admin/finance/schedules_controller.rb
+++ b/app/controllers/admin/finance/schedules_controller.rb
@@ -1,0 +1,7 @@
+module Admin::Finance
+  class SchedulesController < Admin::Finance::BaseController
+    def index
+      @contract_period = ContractPeriod.find(params[:contract_period_id])
+    end
+  end
+end
--- a/app/helpers/admin/schedules_helper.rb
+++ b/app/helpers/admin/schedules_helper.rb
@@ -2,4 +2,8 @@ module Admin::SchedulesHelper
   def schedule_name(identifier)
     identifier.delete_prefix("ecf-").titleize
   end
+
+  def sort_schedules(contract_period)
+    contract_period.schedules.sort_by { |s| ::Schedule.identifiers.keys.index(s.identifier) }
+  end
 end
--- a/app/helpers/admin/schedules_helper.rb
+++ b/app/helpers/admin/schedules_helper.rb
@@ -1,0 +1,24 @@
+module Admin::SchedulesHelper
+  SCHEDULE_MAPPINGS = {
+    # standard
+    "ecf-standard-april" => "Standard April",
+    "ecf-standard-january" => "Standard January",
+    "ecf-standard-september" => "Standard September",
+    # extended
+    "ecf-extended-april" => "Extended April",
+    "ecf-extended-january" => "Extended January",
+    "ecf-extended-september" => "Extended September",
+    # reduced
+    "ecf-reduced-april" => "Reduced April",
+    "ecf-reduced-january" => "Reduced January",
+    "ecf-reduced-september" => "Reduced September",
+    # replacement
+    "ecf-replacement-april" => "Replacement April",
+    "ecf-replacement-january" => "Replacement January",
+    "ecf-replacement-september" => "Replacement September",
+  }.freeze
+
+  def schedule_name(identifier)
+    SCHEDULE_MAPPINGS.fetch(identifier, identifier.humanize)
+  end
+end
--- a/app/helpers/admin/schedules_helper.rb
+++ b/app/helpers/admin/schedules_helper.rb
@@ -1,24 +1,5 @@
 module Admin::SchedulesHelper
-  SCHEDULE_MAPPINGS = {
-    # standard
-    "ecf-standard-april" => "Standard April",
-    "ecf-standard-january" => "Standard January",
-    "ecf-standard-september" => "Standard September",
-    # extended
-    "ecf-extended-april" => "Extended April",
-    "ecf-extended-january" => "Extended January",
-    "ecf-extended-september" => "Extended September",
-    # reduced
-    "ecf-reduced-april" => "Reduced April",
-    "ecf-reduced-january" => "Reduced January",
-    "ecf-reduced-september" => "Reduced September",
-    # replacement
-    "ecf-replacement-april" => "Replacement April",
-    "ecf-replacement-january" => "Replacement January",
-    "ecf-replacement-september" => "Replacement September",
-  }.freeze
-
   def schedule_name(identifier)
-    SCHEDULE_MAPPINGS.fetch(identifier, identifier.humanize)
+    identifier.delete_prefix("ecf-").titleize
   end
 end
--- a/app/models/contract_period.rb
+++ b/app/models/contract_period.rb
@@ -65,6 +65,10 @@ class ContractPeriod < ApplicationRecord
     payments_frozen_at.present? && payments_frozen_at <= Time.zone.now
   end
 
+  def all_schedules?
+    schedules.count.eql?(Schedule.identifiers.size)
+  end
+
 private
 
   def siblings
--- a/app/views/admin/finance/schedules/index.html.erb
+++ b/app/views/admin/finance/schedules/index.html.erb
@@ -25,9 +25,6 @@
 <% end %>
 
 <%=
-  disabled =
-    @contract_period.schedules.count.eql?(Schedule.identifiers.size) ||
-    @contract_period.started_on_or_before_today?
-
-  govuk_button_to("Add schedule", "#", disabled:)
+govuk_button_to("Add schedule", "#",
+  disabled: (@contract_period.all_schedules? || @contract_period.started_on_or_before_today?))
 %>
--- a/app/views/admin/finance/schedules/index.html.erb
+++ b/app/views/admin/finance/schedules/index.html.erb
@@ -3,7 +3,7 @@
     title: "Schedules",
     caption: @contract_period.year,
     caption_size: "l",
-    breadcrumbs: ["Contract periods", @contract_period.year]
+    breadcrumbs: @breadcrumbs
   )
 %>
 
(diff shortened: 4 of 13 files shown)


app/helpers/admin/schedules_helper.rb
```diff
@@ -0,0 +1,24 @@
...
+    "ecf-reduced-september" => "Reduced September",
+    # replacement
+    "ecf-replacement-april" => "Replacement April",
+    "ecf-replacement-january" => "Replacement January",
+    "ecf-replacement-september" => "Replacement September",
+  }.freeze
+
+  def schedule_name(identifier)
+    SCHEDULE_MAPPINGS.fetch(identifier, identifier.humanize)
+  end
```
peterdavidhamilton, line 4: Humanising the schedule identifier could move to a view component but for now the simplest solution is a global helper, asserted through request/view specs

app/views/admin/finance/schedules/index.html.erb
```diff
@@ -0,0 +1,33 @@
...
+    )
+  %>
+<% end %>
+
+<%=
+  disabled =
+    @contract_period.schedules.count.eql?(Schedule.identifiers.size) ||
+    @contract_period.started_on_or_before_today?
+
+  govuk_button_to("Add schedule", "#", disabled:)
```
peterdavidhamilton: I have intentionally kept this logic out of the controller for this MVP.

config/application.rb
```diff
@@ -63,6 +63,7 @@ class Application < Rails::Application
     config.sentry_dsn = ENV["SENTRY_DSN"]
     config.enable_request_specs_timeout = ActiveModel::Type::Boolean.new.cast(ENV.fetch("CI", false))
     config.enable_trs_teacher_refresh = ActiveModel::Type::Boolean.new.cast(ENV.fetch("ENABLE_TRS_TEACHER_REFRESH", true))
+    config.enable_finance_contract_periods = ActiveModel::Type::Boolean.new.cast(ENV.fetch("ENABLE_FINANCE_CONTRACT_PERIODS", false))
```
peterdavidhamilton, line 66: Copied from an earlier stage and will be rebased away

spec/rails_helper.rb
```diff
@@ -34,6 +34,11 @@
   config.infer_spec_type_from_file_location!
   config.filter_rails_from_backtrace!
 
+  # Opening registration contract periods
+  config.before(:each, :enable_finance_contract_periods) do
```
peterdavidhamilton, line 38: Recommend we use this setup

spec/views/admin/finance/schedules/index.html.erb_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+      Schedule.identifiers { |identifier| FactoryBot.create(:schedule, identifier:, contract_period:) }
+    end
+
+    it "has a disabled add schedule button" do
+      render
+      expect(rendered).to have_button("Add schedule", disabled: true)
+    end
+  end
+
+  context "when a contract period is opened" do
```
ethax-ross: `is open`?

spec/requests/admin/finance/schedules_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+          expect(response.body).to include("Manage schedules for the 2025 contract period.")
+        end
+
+        it "displays the schedules for the contract period" do
+          expect(response.body).to include("Standard January")
+          expect(response.body).to include("Standard April")
+          expect(response.body).not_to include("Extended January")
+        end
+
+        it "has a disabled button" do
```
ethax-ross, line 55: I'd be inclined to omit these here as they're already covered in the view specs (and would be better suited to feature specs maybe when we can test the whole journey)?

app/views/admin/finance/schedules/index.html.erb
```diff
@@ -0,0 +1,33 @@
...
+        [
+          govuk_link_to(schedule_name(schedule.identifier), "#"),
+        ]
+      end
+    )
+  %>
+<% end %>
+
+<%=
+  disabled =
```
ethax-ross: Worth moving this logic into a helper?

spec/rails_helper.rb
ethax-ross: I wonder if its worth having a conscious/wider refactoring of this behaviour? 🤔 my worry is if its in `test.rb` as it is in the other PR (which seems to be the 'standard' at the moment) as well as here in this PR it might be misleading/confusing?

https://github.com/DFE-Digital/register-early-career-teachers-public/pull/2902/changes#diff-77f322f5ab0c54b1f2793b339574333dc55889645d910a93ede0fd7aa13b217aR79

I agree this is a more robust approach, though 🙂

ethax-ross (commented):
Looking great 👍 just a few minor questions/suggestions

app/helpers/admin/schedules_helper.rb
ethax-ross: I wonder if we should avoid redefining the schedule identifiers and just have something like:

```
def schedule_name(identifier)
  identifier.delete_prefix("ecf-").tr("-", " ").titleize
end
```

I could argue it either way tbf but if we are going explicit I'd opt for failure when it encounters one it doesn't have a mapping for. I think its worth having dedicated tests for this helper too?

spec/requests/admin/finance/schedules_spec.rb
peterdavidhamilton: I can do that. I laid down the requests before the views

app/views/admin/finance/schedules/index.html.erb
peterdavidhamilton: If we are favouring view components I'd suggest moving it straight there when we are ready.

app/helpers/admin/schedules_helper.rb
```diff
@@ -0,0 +1,24 @@
+module Admin::SchedulesHelper
```
craigmdavidson, line 1: Should this be **Admin::Finance::SchedulesHelper**?

app/helpers/admin/schedules_helper.rb
craigmdavidson: +1 - I would go for the function over the explicit mapping. This is admin so less is more:

Note: we don't need the `.tr("-", " ")` we can just have: `identifier.delete_prefix("ecf-").titleize` for the same results.

spec/requests/admin/finance/schedules_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+      include_context "sign in as finance DfE user"
+
+      before do
+        FactoryBot.create(:schedule, identifier: "ecf-standard-january", contract_period:)
+        FactoryBot.create(:schedule, identifier: "ecf-standard-april", contract_period:)
+        FactoryBot.create(:schedule, identifier: "ecf-extended-january")
+        get "/admin/finance/contract-periods/#{contract_period.id}/schedules"
+      end
+
+      context "and the contract period is closed" do
```
craigmdavidson, line 41: The contract_period is not so much "closed" as "started" is it not?

app/views/admin/finance/schedules/index.html.erb
```diff
@@ -0,0 +1,33 @@
...
+    @contract_period.schedules.count.eql?(Schedule.identifiers.size) ||
```
craigmdavidson: We have already loaded `@contract_period.schedules` so using `size` would avoid an additional query.

app/controllers/admin/finance/schedules_controller.rb
```diff
@@ -0,0 +1,7 @@
+module Admin::Finance
+  class SchedulesController < Admin::Finance::BaseController
+    def index
+      @contract_period = ContractPeriod.find(params[:contract_period_id])
```
craigmdavidson, line 16: We know what's coming so maybe go for an idiomatic `before_action :set_contract_period`?

app/views/admin/finance/schedules/index.html.erb
craigmdavidson: `@contract_period.fully_scheduled?` :sob:

craigmdavidson (approved):
LGTM - couple of nitpicks.

spec/requests/admin/finance/schedules_spec.rb
peterdavidhamilton: Closed/Opened to amendnents is the intent. I'll reword it

app/views/admin/finance/schedules/index.html.erb
craigmdavidson: FWIW: 
```
govuk_button_to(
  "Add schedule", 
  "#", 
  disabled: (
    @contract_period.schedules.size.eql?(Schedule.idenfiers.size) ||  
    @contract_period.started_on_or_before_today?
   )
)
```

Is probably clearer than assigning a variable? It would be even nicer as:
  `disabled: @contract_period.frozen?` but that's a discussion for another day. 😄

app/helpers/admin/schedules_helper.rb
```diff
@@ -0,0 +1,5 @@
...
+  def schedule_name(identifier)
```
ethax-ross, line 2: Worth a dedicated test for this?

ethax-ross (approved):
Looks great 👍

app/models/contract_period.rb
```diff
@@ -65,6 +65,10 @@ def payments_frozen?
     payments_frozen_at.present? && payments_frozen_at <= Time.zone.now
   end
 
+  def all_schedules?
```
craigmdavidson, line 68: I'm not completely sure what `contract_period.all_schedules?` means? 

Is there a way to make it more intention revealing?
Maybe, say is this `fully_scheduled?` or `all_schedules_added?` or something else?

spec/models/contract_period_spec.rb
```diff
@@ -242,4 +242,29 @@
...
+    let(:schedule_identifiers) { Schedule.identifiers.values }
+
+    context "when all possible schedules have been allocated to the contract period" do
+      before do
+        schedule_identifiers.each do |identifier|
+          FactoryBot.create(:schedule, identifier:, contract_period:)
+        end
+      end
+
+      it { expect(contract_period).to be_all_schedules }
```
craigmdavidson, line 257: So, `be_fully_scheduled` (`fully_scheduled?`) or `be_all_schedules_allocated` (`all_schedules_allocated?`) would both work here, and better reveal intent?

app/models/contract_period.rb
peterdavidhamilton: All possible schedules

spec/models/contract_period_spec.rb
peterdavidhamilton: It depends on a few things I think. Is there a minimum number of schedules? This the max number. It depends on the domain language but it is nice when the matcher reads well.

app/helpers/admin/schedules_helper.rb
peterdavidhamilton: I've added one but the name format was covered in view the spec. It depends whether we intend to move toward view components.

PR #2938, opened 2026-05-18
[3942] Add admin statements index for contract period and lead provider

mooktakim:
### Context

resolves https://github.com/DFE-Digital/register-ects-project-board/issues/3942

* Statements index page listing all statements for contract period and lead provider

Screenshot 2026-05-22 at 11 27 51

### Guidance to review

* Simple relatively standard Rails controller with index action.
* Implements link from ActiveLeadProviders#index to statements.
* Introduces namespace collisions that require namespace modifications to ActiveLeadProvidersController

### DIFF
--- a/app/controllers/admin/finance/active_lead_providers/statements_controller.rb
+++ b/app/controllers/admin/finance/active_lead_providers/statements_controller.rb
@@ -10,7 +10,7 @@ module Admin::Finance::ActiveLeadProviders
       @breadcrumbs = {
         "Finance" => admin_finance_path,
         "Contract periods" => admin_contract_periods_path,
-        contract_period.year.to_s => admin_contract_period_path(contract_period),
+        @active_lead_provider.contract_period_year.to_s => admin_contract_period_path(contract_period),
         @active_lead_provider.lead_provider_name => admin_contract_period_active_lead_providers_path(contract_period),
       }
       @pagy, @statements = pagy(@active_lead_provider.statements.order(year: :asc, month: :asc))
--- a/app/controllers/admin/finance/active_lead_providers/statements_controller.rb
+++ b/app/controllers/admin/finance/active_lead_providers/statements_controller.rb
@@ -1,0 +1,25 @@
+module Admin::Finance::ActiveLeadProviders
+  class StatementsController < Admin::Finance::BaseController
+    layout "full"
+
+    before_action :set_active_lead_provider
+
+    def index
+      @breadcrumbs = {
+        "Finance" => admin_finance_path,
+        "Contract periods" => admin_contract_periods_path,
+        @active_lead_provider.contract_period_year.to_s => admin_contract_period_path(@active_lead_provider.contract_period),
+        @active_lead_provider.lead_provider_name => nil,
+      }
+      @pagy, @statements = pagy(@active_lead_provider.statements.order(year: :asc, month: :asc))
+    end
+
+  private
+
+    def set_active_lead_provider
+      @active_lead_provider = ActiveLeadProvider
+        .includes(:contract_period, :lead_provider)
+        .find(params[:active_lead_provider_id])
+    end
+  end
+end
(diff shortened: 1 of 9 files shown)


peterdavidhamilton (commented):
If we could pop a minimal view spec in now it would set us up for the eventual end to end refactor. It'll duplicate an assertion in the request spec but I think that should remain for now.

app/controllers/admin/finance/contract_period_statements_controller.rb
```diff
@@ -0,0 +1,32 @@
+module Admin::Finance
+  class ContractPeriodStatementsController < Admin::Finance::BaseController
+    layout "full"
+
+    before_action :set_contract_period_and_lead_provider
```
craigmdavidson: Since **Statements** are subordinate to **ActiveLeadProvider**, and **ActiveLeadProvidersController** is the parent controller, `set_active_lead_provider` would be simpler, more idiomatic, feeds happily into the **::Statements::Search**?

app/controllers/admin/finance/contract_period_statements_controller.rb
```diff
@@ -0,0 +1,32 @@
+module Admin::Finance
+  class ContractPeriodStatementsController < Admin::Finance::BaseController
```
craigmdavidson: This is not really a **ContractPeriodStatementsController**? I.e. it doesn't show **Statements** for all **ContractPeriods**, is it not **Admin::Finance::ActiveLeadProviders::StatementsController**?

I.e. it shows **Statements** for a given **ActiveLeadProvider** (which is a **LeadProvider's** participation in a given **ContractPeriod**)?

craigmdavidson (dismissed):
Looks good, but it looks like another PR has leaked in to this (ContractPeriods CRUD actions ❓ 🤷 )? 

Ignoring those bits, just two recommendations:

1. Tighten the controller class naming - it is not _all_ statements for any given **ContractPeriod** (as would be implied by the current name **ContractPeriodStatementsController**).
2. Instead, use **ActiveLeadProvider** rather than **ContractPeriod** and **LeadProvider** as the controller scope, and name the controller and route accordingly.

This would also address the ghost route added to routes.

tonyheadford (commented):
I think it looks OK to me (excepting the comments made by the others). The direct link in the Guidance section isn't working for me, I get a page not found error after signing in as a finance person (if that's correct).

app/controllers/admin/finance/active_lead_provider_statements_controller.rb
```diff
@@ -0,0 +1,28 @@
+module Admin::Finance
+  class ActiveLeadProviderStatementsController < Admin::Finance::BaseController
```
ethax-ross: Do we want to namespace this?

```
Admin::Finance::ActiveLeadProviders::StatementsController
```

app/controllers/admin/finance/active_lead_providers/statements_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+        @contract_period.year.to_s => admin_contract_period_path(@contract_period),
+        @lead_provider.name => nil,
+      }
+
+      @pagy, @statements = pagy(@active_lead_provider.statements.order(year: :asc, month: :asc))
+    end
+
+  private
+
+    def set_active_lead_provider
```
ethax-ross, line 21: Worth splitting/renaming this as the method is doing more than just setting the active lead provider?

spec/requests/admin/finance/active_lead_providers/statements_spec.rb
```diff
@@ -0,0 +1,141 @@
...
+
+  let(:path) { admin_active_lead_provider_statements_path(active_lead_provider) }
+
+  before do
+    allow(Rails.application.config).to receive(:enable_finance_contract_periods).and_return(env_var_value)
+  end
+
+  describe "GET /admin/finance/active-lead-providers/:active_lead_provider_id/statements" do
+    context "when enabled" do
+      let(:env_var_value) { true }
```
ethax-ross: `enable_finance_contract_periods`

PR #2939, opened 2026-05-18
Add ActiveLeadProvidersController

craigmdavidson:
### Context

Resolves https://github.com.mcas.ms/DFE-Digital/register-ects-project-board/issues/3935
and resolves https://github.com.mcas.ms/DFE-Digital/register-ects-project-board/issues/3971

#### Fully populated

Screenshot 2026-05-20 at 11 38 42

Notes: 
* Use a secondary CTA to keep the page from being to "shouty".

**Note: The older images below are missing the index copy, e.g. These lead providers are active for the 2026 contract period (1 June 2026 to 31 May 2027).**

#### Blank slate

Given we have just created a Contract Period, when we navigate to the list of ActiveLeadProvider's, it will be empty.
Notes: 
* Right aligned CTA button.

Screenshot 2026-05-19 at 13 19 25

#### Adding an activation

When we click the add button, we only need to select the Lead Provider we want to Activate.

Notes: 
* Since there is a single field, is there any value in the "Lead Provider" label?
* Radio buttons work better than the blank typeahead for the small number of Lead Providers available.

Screenshot 2026-05-19 at 13 19 35

#### Confirming addition

When we activate a Lead Provider, we call the SeedFromPrevious service to seed Contracts, Statements and Delivery Partners.

Notes: 
* We expect to have maybe 5 - 10 ActiveLeadProviders in the table. Having a column for Contracts, Statements and Delivery Partners limits the vertical sprawl.
* We provide a full text description rather than a number to provide a more affordant clickable area (i.e. "1 contract" as the link rather than "1").
* We use a secondary button for the _remove_ button as to be less "shouty" this really becomes apparent when there are a handful of rows.

Screenshot 2026-05-19 at 13 19 42

#### Multiple activations

This will be the standard state of the screen.

Notes: 
* As activations are added, the simple tabular structure allows us to see the activation we have just added without scrolling.
* We use a secondary button for the _remove_ button as to be less "shouty" this really becomes apparent when there are a handful of rows. (again)

Screenshot 2026-05-19 at 13 19 59

#### Remove activation 

Removing an activation happens prior to the ContractPeriod.

Notes: 
* I think this is the correct confirmation style, the Contract Period hasn't started, so this isn't an OMG type action?

Screenshot 2026-05-19 at 13 20 11

#### Locked because ContractPeriod has started

After a Contract Period has started the page is no longer editable.

Notes: 
* Small `govuk-hint` text indicating why the ActiveLeadProvider list is not editable.

Screenshot 2026-05-19 at 13 23 32

#### Issue with seeding

We allow the ActiveLeadProvider to be created, but show the error.

Notes: 
* Have changed the error wording in the **SeedFromPrevious** service to better suit being surfaced the UI - concretely we refer to the year and lead provider name rather than the object id.
* We don't get too excited about surfacing service error messages as this is for internal (admin) users.
 
Screenshot 2026-05-19 at 16 18 01

### Guidance to review

This is pretty much a standard Rails controller with the following idiosyncrasies:
- Instead of relying on `dependent: :destroy` relationships to cascade deletes we use an explicit CascadeDelete service.
- We call a SeedFromPrevious service to "seed" after we create the ActiveLeadProvider activation. While the SeedFromPrevious creates subordinates in a transaction (so will rollback all if any fail), the ActiveLeadProvider is created outwith the transaction - so we can still have the ActiveLeadProvider; this is useful if we do not have a previous Activation from which to seed.
- In our index we include the subordinates to avoid an N+1 situation, for the number of activations we have this is potentially overkill.
- We add dead (`"#"`) links to the contracts, statements, and delivery partners (lead provider delivery partners) in lieu of those controllers.
- See additional notes associated with the UI screens.

#### Acceptance criteria

✅  We link to this step from the contract period page task list.
✅ We have a short description of what the form is for (ask for copy).
✅ When ActiveLeadProvider records exist for the contract period already we should add a heading, delete button (placeholder) and a task list for them as per the UI.
✅ We should NOT allow deleting active lead providers on contract periods that have started/are open, so disable the button (but we can still display the task list).
✅ We should display a list of lead providers that can be added to the contract period - this should only include lead providers not already added and those that were active in the previous contract period.
✅ On adding a new lead provider we should create the ActiveLeadProvider, run it through the service to generate the data for the contract period and refresh the page.

### DIFF
--- a/app/controllers/admin/finance/active_lead_providers_controller.rb
+++ b/app/controllers/admin/finance/active_lead_providers_controller.rb
@@ -46,6 +46,9 @@ module Admin::Finance
       ActiveLeadProviders::CascadeDelete.new(active_lead_provider:).call
       flash[:notice] = "#{lead_provider_name} removed"
       redirect_to admin_contract_period_active_lead_providers_path(@contract_period)
+    rescue ActiveLeadProviders::CascadeDelete::CascadeDeleteError => e
+      flash[:error] = "Cannot remove #{lead_provider_name}: #{e.message}"
+      redirect_to admin_contract_period_active_lead_providers_path(@contract_period)
     end
 
   private
--- a/app/controllers/admin/finance/active_lead_providers_controller.rb
+++ b/app/controllers/admin/finance/active_lead_providers_controller.rb
@@ -24,10 +24,12 @@ module Admin::Finance
     end
 
     def create
-      @active_lead_provider = @contract_period.active_lead_providers.build(active_lead_provider_params)
+      @active_lead_provider = ActiveLeadProviders::Create.new(
+        contract_period: @contract_period,
+        lead_provider_id: active_lead_provider_params[:lead_provider_id]
+      ).call
 
-      if @active_lead_provider.save
-        ActiveLeadProviders::SeedFromPrevious.new(active_lead_provider: @active_lead_provider).call
+      if @active_lead_provider.persisted?
         flash[:notice] = "#{@active_lead_provider.lead_provider.name} added"
         redirect_to admin_contract_period_active_lead_providers_path(@contract_period)
       else
--- a/app/controllers/admin/finance/active_lead_providers_controller.rb
+++ b/app/controllers/admin/finance/active_lead_providers_controller.rb
@@ -46,7 +46,7 @@ module Admin::Finance
     def destroy
       active_lead_provider = @contract_period.active_lead_providers.find(params[:id])
       lead_provider_name = active_lead_provider.lead_provider.name
-      ActiveLeadProviders::CascadeDelete.new(active_lead_provider:).call
+      ActiveLeadProviders::CascadeDelete.new(active_lead_provider:, author: current_user).call
       flash[:notice] = "#{lead_provider_name} removed"
       redirect_to admin_contract_period_active_lead_providers_path(@contract_period)
     rescue ActiveLeadProviders::CascadeDelete::CascadeDeleteError => e
--- a/app/controllers/admin/finance/active_lead_providers_controller.rb
+++ b/app/controllers/admin/finance/active_lead_providers_controller.rb
@@ -25,6 +25,7 @@ module Admin::Finance
 
     def create
       @active_lead_provider = ActiveLeadProviders::Create.new(
+        author: current_user,
         contract_period: @contract_period,
         lead_provider_id: active_lead_provider_params[:lead_provider_id]
       ).call
--- a/app/controllers/admin/finance/active_lead_providers_controller.rb
+++ b/app/controllers/admin/finance/active_lead_providers_controller.rb
@@ -1,0 +1,78 @@
+module Admin::Finance
+  class ActiveLeadProvidersController < Admin::Finance::BaseController
+    layout "full"
+
+    before_action :set_contract_period
+    before_action :reject_if_contract_period_started, only: %i[new create destroy]
+
+    def index
+      @breadcrumbs = {
+        "Finance" => admin_finance_path,
+        "Contract periods" => admin_contract_periods_path,
+        @contract_period.year.to_s => admin_contract_period_path(@contract_period),
+      }
+      @editable = editable?
+      @active_lead_providers = @contract_period
+        .active_lead_providers
+        .with_lead_provider_ordered_by_name
+        .includes(:contracts, :statements, :delivery_partners)
+    end
+
+    def new
+      @active_lead_provider = @contract_period.active_lead_providers.build
+      @available_lead_providers = available_lead_providers
+    end
+
+    def create
+      @active_lead_provider = @contract_period.active_lead_providers.build(active_lead_provider_params)
+
+      if @active_lead_provider.save
+        ActiveLeadProviders::SeedFromPrevious.new(active_lead_provider: @active_lead_provider).call
+        flash[:notice] = "#{@active_lead_provider.lead_provider.name} added"
+        redirect_to admin_contract_period_active_lead_providers_path(@contract_period)
+      else
+        @available_lead_providers = available_lead_providers
+        render :new, status: :unprocessable_entity
+      end
+    rescue ActiveLeadProviders::SeedFromPrevious::PreviousActiveLeadProviderError,
+           ActiveLeadProviders::SeedFromPrevious::AlreadyPopulatedError => e
+      flash[:error] = "Cannot seed: #{e.message}"
+      redirect_to admin_contract_period_active_lead_providers_path(@contract_period)
+    end
+
+    def destroy
+      active_lead_provider = @contract_period.active_lead_providers.find(params[:id])
+      lead_provider_name = active_lead_provider.lead_provider.name
+      ActiveLeadProviders::CascadeDelete.new(active_lead_provider:).call
+      flash[:notice] = "#{lead_provider_name} removed"
+      redirect_to admin_contract_period_active_lead_providers_path(@contract_period)
+    end
+
+  private
+
+    def set_contract_period
+      @contract_period = ContractPeriod.find(params[:contract_period_id])
+    end
+
+    def reject_if_contract_period_started
+      return if editable?
+
+      flash[:error] = "Active lead providers cannot be changed once the contract period has started"
+      redirect_to admin_contract_period_active_lead_providers_path(@contract_period)
+    end
+
+    def available_lead_providers
+      LeadProvider
+        .where.not(id: @contract_period.active_lead_providers.select(:lead_provider_id))
+        .alphabetical
+    end
+
+    def active_lead_provider_params
+      params.expect(active_lead_provider: [:lead_provider_id])
+    end
+
+    def editable?
+      !@contract_period.started_on_or_before_today?
+    end
+  end
+end
--- a/app/services/active_lead_providers/cascade_delete.rb
+++ b/app/services/active_lead_providers/cascade_delete.rb
@@ -1,9 +1,12 @@
 module ActiveLeadProviders
   class CascadeDelete
-    # Defensive nullification/deletion below assumes the active lead provider has
-    # not yet been used: active lead providers should not be deleted once their
-    # contract period has started. Without this defence the FK constraints would
-    # block deletion if any usage data ever existed.
+    # Active lead providers should only be deleted before their contract period
+    # has started, i.e. while still unused. If any usage data references this
+    # active lead provider we refuse to delete it and raise CascadeDeleteError for
+    # the controller to handle. The destruction below can then safely assume there
+    # is no usage data depending on the records it removes.
+
+    class CascadeDeleteError < StandardError; end
 
     attr_reader :active_lead_provider
 
@@ -12,21 +15,35 @@ module ActiveLeadProviders
     end
 
     def call
+      reject_if_in_use!
+
       ActiveRecord::Base.transaction do
         destroy_statements!
         destroy_contracts_and_fee_structures!
         destroy_lead_provider_delivery_partnerships!
-        nullify_expressions_of_interest!
         active_lead_provider.destroy!
       end
     end
 
   private
 
+    def reject_if_in_use!
+      raise CascadeDeleteError, "Declarations are present" if declarations.exists?
+      raise CascadeDeleteError, "Training periods are present" if training_periods.exists?
+      raise CascadeDeleteError, "Expressions of interest are present" if active_lead_provider.expressions_of_interest.exists?
+    end
+
+    def declarations
+      statement_ids = active_lead_provider.statements.ids
+      Declaration.where(payment_statement_id: statement_ids).or(Declaration.where(clawback_statement_id: statement_ids))
+    end
+
+    def training_periods
+      TrainingPeriod.where(school_partnership_id: active_lead_provider.school_partnerships.ids)
+    end
+
     def destroy_statements!
       statement_ids = active_lead_provider.statements.ids
-      Declaration.where(payment_statement_id: statement_ids).update_all(payment_statement_id: nil)
-      Declaration.where(clawback_statement_id: statement_ids).update_all(clawback_statement_id: nil)
       Statement::Adjustment.where(statement_id: statement_ids).delete_all
       Statement.where(id: statement_ids).destroy_all
     end
@@ -47,13 +64,8 @@ module ActiveLeadProviders
     def destroy_lead_provider_delivery_partnerships!
       lpdp_ids = active_lead_provider.lead_provider_delivery_partnerships.ids
       school_partnership_ids = SchoolPartnership.where(lead_provider_delivery_partnership_id: lpdp_ids).ids
-      TrainingPeriod.where(school_partnership_id: school_partnership_ids).update_all(school_partnership_id: nil)
       SchoolPartnership.where(id: school_partnership_ids).destroy_all
       LeadProviderDeliveryPartnership.where(id: lpdp_ids).destroy_all
-    end
-
-    def nullify_expressions_of_interest!
-      active_lead_provider.expressions_of_interest.update_all(expression_of_interest_id: nil)
     end
   end
 end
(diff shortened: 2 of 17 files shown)


app/controllers/admin/finance/active_lead_providers_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+      if @active_lead_provider.save
+        ActiveLeadProviders::SeedFromPrevious.new(active_lead_provider: @active_lead_provider).call
+        flash[:notice] = "#{@active_lead_provider.lead_provider.name} added"
+        redirect_to admin_contract_period_active_lead_providers_path(@contract_period)
+      else
+        @available_lead_providers = available_lead_providers
+        render :new, status: :unprocessable_entity
+      end
+    rescue ActiveLeadProviders::SeedFromPrevious::PreviousActiveLeadProviderError,
+           ActiveLeadProviders::SeedFromPrevious::AlreadyPopulatedError => e
```
craigmdavidson, line 41: It's highly improbable that we would hit an **ActiveLeadProviders::SeedFromPrevious::AlreadyPopulatedError** immediately after creating a new **ActiveLeadProvider**, but not _impossible_. So for belt and braces, and the minimal cognitive cost let's leave it in.

app/controllers/admin/finance/active_lead_providers_controller.rb
```diff
@@ -0,0 +1,74 @@
+module Admin
+  class ActiveLeadProvidersController < AdminController
+    layout "full"
+
+    before_action :set_contract_period
+    before_action :reject_if_contract_period_started, only: %i[new create destroy]
```
craigmdavidson, line 6: The UI doesn't show buttons in this case, so this is just for belt and braces.

app/services/active_lead_providers/cascade_delete.rb
```diff
@@ -0,0 +1,59 @@
+module ActiveLeadProviders
+  class CascadeDelete
+    # Defensive nullification/deletion below assumes the active lead provider has
+    # not yet been used: active lead providers should not be deleted once their
+    # contract period has started. Without this defence the FK constraints would
+    # block deletion if any usage data ever existed.
```
craigmdavidson: We purposefully do not actively block the **CascadeDelete** here if the `contract_period` has started, since we are essentially recreating the `destroy! => dependent: :destroy` functionality of Rails. 

The job of deciding when it is appropriate to delete or not is the job higher up the call hierarchy (In this case this is done in the controller). Instead we handle the presence of declarations and expressions of interest safely.

app/services/active_lead_providers/seed_from_previous.rb
```diff
@@ -23,12 +24,12 @@ def initialize(active_lead_provider:)
...
   def call
-    raise PreviousActiveLeadProviderError, "no previous active_lead_provider found for #{active_lead_provider.id}" if previous_activation.blank?
+    raise PreviousActiveLeadProviderError, "No previous activation found in #{previous_contract_period.year} for #{lead_provider_name}" if previous_activation.blank?
     if previous_activation_empty?
       raise PreviousActiveLeadProviderError,
-            "active_lead_provider for #{active_lead_provider.id} is missing previous delivery partnerships, contracts or statements"
+            "Key info for #{lead_provider_name} is missing previous delivery partnerships, contracts or statements."
     end
-    raise AlreadyPopulatedError, "active_lead_provider #{active_lead_provider.id} already has data" if active_lead_provider_populated?
+    raise AlreadyPopulatedError, "#{lead_provider_name} already has data for #{current_contract_period.year}" if active_lead_provider_populated?
```
craigmdavidson, line 32: These message changes just make surfacing the error messages more understandable.

This is internal (Admin) UI.

spec/services/active_lead_providers/cascade_delete_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    expect(Contract::FlatRateFeeStructure).not_to exist(flat_rate_fee_structure_id)
+    expect(Contract::BandedFeeStructure).not_to exist(banded_fee_structure_id)
+    expect(Contract::BandedFeeStructure::Band.where(id: band_ids)).not_to exist
+    expect(Statement).not_to exist(statement.id)
+    expect(LeadProviderDeliveryPartnership).not_to exist(lead_provider_delivery_partnership.id)
+    expect(DeliveryPartner).to exist(delivery_partner.id)
+  end
+
+  it "wraps the deletions in a transaction" do
+    allow(active_lead_provider).to receive(:destroy!).and_raise(ActiveRecord::RecordNotDestroyed)
```
craigmdavidson, line 39: `active_lead_provider.destroy!` is the last thing in the cascade chain (so it's really a reverse cascade).

spec/views/admin/active_lead_providers/index.html.erb_spec.rb
```diff
@@ -0,0 +1,78 @@
+RSpec.describe "admin/active_lead_providers/index.html.erb" do
```
craigmdavidson: 🤔 Is this spec overkill for an Admin page?

app/controllers/admin/finance/active_lead_providers_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+      @breadcrumbs = {
+        "Finance" => admin_finance_path,
+        "Contract periods" => admin_contract_periods_path,
+        @contract_period.year.to_s => admin_contract_period_path(@contract_period),
+      }
+      @editable = editable?
+      @active_lead_providers = @contract_period
+        .active_lead_providers
+        .with_lead_provider_ordered_by_name
+        .includes(:contracts, :statements, :delivery_partners)
```
craigmdavidson, line 18: We want to include the subordinates to avoid N+1 on subordinate counts.

app/services/active_lead_providers/cascade_delete.rb
```diff
@@ -0,0 +1,59 @@
...
+      # destroyed before the fee structures it references. Fee structures are not
+      # shared between contracts (uniqueness validation on Contract enforces this).
+      active_lead_provider.contracts.find_each do |contract|
+        flat_rate_fee_structure = contract.flat_rate_fee_structure
+        banded_fee_structure = contract.banded_fee_structure
+        contract.destroy!
+        flat_rate_fee_structure&.destroy!
+        banded_fee_structure&.destroy!
+      end
+    end
```
craigmdavidson, line 67: This dance (which is also reflected in the SeedFromPrevious) is a smell that the foreign keys to **FlatRateFeeStructure** and **BandedFeeStructure** point the wrong way round.  

These are subordinate to the contract (as implied even by their modulized names), as such these should `belong_to` the **Contract** rather than `have_one :contract`.

app/services/active_lead_providers/cascade_delete.rb
peterdavidhamilton: So if this PR seeks to do the work of https://github.com/DFE-Digital/register-ects-project-board/issues/3971.

It should also encorporate the event logging elements built from https://github.com/DFE-Digital/register-ects-project-board/issues/3948.

I can park my work and fill that gap later if you'd rather @craigmdavidson @ethax-ross

app/views/admin/finance/active_lead_providers/index.html.erb
```diff
@@ -0,0 +1,61 @@
...
+            <%# TODO: link to admin delivery partners page once it exists %>
+            <%= govuk_link_to(pluralize(alp.delivery_partners.size, "delivery partner"), "#") %>
+          <% end %>
+          <% if @editable %>
+            <% row.with_cell do %>
+              <%= govuk_button_to "Remove",
+                    admin_contract_period_active_lead_provider_path(@contract_period, alp),
+                    method: :delete,
+                    secondary: true,
+                    class: "govuk-!-margin-bottom-0" %>
```
tonyheadford, line 48: This is not accessibility friendly, having a table with a "Remove" button per row, but no context of what that applies to - not sure if that's already been considered?
Same applies to the links

spec/views/admin/active_lead_providers/index.html.erb_spec.rb
peterdavidhamilton: If we can get a smoke test in for heading/subheading, and any key business text, that sort of thing, then I'm happy.

app/controllers/admin/finance/active_lead_providers_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+
+    def reject_if_contract_period_started
+      return if editable?
+
+      flash[:error] = "Active lead providers cannot be changed once the contract period has started"
+      redirect_to admin_contract_period_active_lead_providers_path(@contract_period)
+    end
+
+    def available_lead_providers
+      LeadProvider
```
ethax-ross, line 71: I think we want to restrict this to only lead providers that were active in the previous contract period, or it could blow up when it comes to seed from previous?

app/services/active_lead_providers/cascade_delete.rb
```diff
@@ -0,0 +1,59 @@
...
+
+    attr_reader :active_lead_provider
+
+    def initialize(active_lead_provider:)
+      @active_lead_provider = active_lead_provider
+    end
+
+    def call
+      ActiveRecord::Base.transaction do
+        destroy_statements!
```
ethax-ross, line 24: Can we send an event for the deletion of an active lead provider

app/services/active_lead_providers/cascade_delete.rb
```diff
@@ -0,0 +1,59 @@
...
+        nullify_expressions_of_interest!
+        active_lead_provider.destroy!
+      end
+    end
+
+  private
+
+    def destroy_statements!
+      statement_ids = active_lead_provider.statements.ids
+      Declaration.where(payment_statement_id: statement_ids).update_all(payment_statement_id: nil)
```
ethax-ross: If a declaration exists against the statement we should prevent the deletion; removing the statement from a billable declaration would raise an error anyway, so it can probably just be omitted and covered with a test?

app/services/active_lead_providers/cascade_delete.rb
```diff
@@ -0,0 +1,59 @@
...
+        contract.destroy!
+        flat_rate_fee_structure&.destroy!
+        banded_fee_structure&.destroy!
+      end
+    end
+
+    def destroy_lead_provider_delivery_partnerships!
+      lpdp_ids = active_lead_provider.lead_provider_delivery_partnerships.ids
+      school_partnership_ids = SchoolPartnership.where(lead_provider_delivery_partnership_id: lpdp_ids).ids
+      TrainingPeriod.where(school_partnership_id: school_partnership_ids).update_all(school_partnership_id: nil)
```
ethax-ross: Same here, if a training period/expression of interest exists lets just block the deletion

ethax-ross (commented):
Looking great, just a few minor suggestions 👍

app/controllers/admin/finance/active_lead_providers_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+        .includes(:contracts, :statements, :delivery_partners)
+    end
+
+    def new
+      @active_lead_provider = @contract_period.active_lead_providers.build
+      @available_lead_providers = available_lead_providers
+    end
+
+    def create
+      @active_lead_provider = @contract_period.active_lead_providers.build(active_lead_provider_params)
```
ethax-ross: Can we stick this in a service to be consistent with elsewhere in the app, similar to https://github.com/DFE-Digital/register-early-career-teachers-public/blob/3ebb3d8d17013d8d2ac5cb6ae6640fc39f5ddfd1/app/services/contract_periods/create.rb

So the `Create` service will save the active lead provider, kick off the seed from previous and emit an event

app/controllers/admin/finance/active_lead_providers_controller.rb
craigmdavidson: 🤔 I have made it blow up cleanly. I.e. you can include the Lead Provider but will have to add subordinate records yourself?

But if we are happy to restrict it that's great! 😄

app/controllers/admin/finance/active_lead_providers_controller.rb
craigmdavidson: 🤢 😭  😆 👍  No problem 😄

app/controllers/admin/finance/active_lead_providers_controller.rb
ethax-ross: Ah I had missed the `rescue` - that's fine, as long as its handled gracefully I'm happy 👍

app/controllers/admin/finance/active_lead_providers_controller.rb
ethax-ross: Consistency is king 😅 

But in all seriousness, I'm happy to entertain other approaches if you have opinions around how to improve/restructure things; it would need to be a wider discussion though 🙂

app/services/active_lead_providers/cascade_delete.rb
ethax-ross: Agreed; originally when modelling this I believed a fee structure could be applied/shared across multiple contracts, hence the direction of travel in the foreign key just now. This isn't how it ended up, though, so we should put the `contract_id` on the fee structure 👍 thanks for flagging, I'll raise a ticket.

https://github.com/DFE-Digital/register-ects-project-board/issues/3997

app/controllers/admin/finance/active_lead_providers_controller.rb
craigmdavidson: > Consistency is king 😅

I agree. :sob:

> But in all seriousness, I'm happy to entertain other approaches if you have opinions around how to improve/restructure things; it would need to be a wider discussion though 🙂

It comes down to let Rails be Rails, 
when we are looking to deviate from what DHH would do, 
*then we should stop and think extremely carefully,*
as we _will_ be wrong.

Consider in **ContractPeriod::Create**:

```
  def create!
      raise ActiveRecord::RecordInvalid, contract_period unless contract_period.valid?

        contract_period.save!
        record_event! or raise ActiveRecord::Rollback
      end

      contract_period
    end
```

This is a standard controller action moved into a method object, for no real reason? 
This bit `raise ActiveRecord::RecordInvalid, contract_period unless contract_period.valid?` is just nuts! 
There is no complex no orchestration here. 
Just a record_event! that might be better as a generic ApplicationRecord callback.

That is not good code.

app/services/active_lead_providers/cascade_delete.rb
peterdavidhamilton: [I'll not park it then](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/2939#discussion_r3273659695) 🫠

Can we link that this PR closes 2 tickets in the description so it is reflected on the board please.

app/views/admin/finance/active_lead_providers/index.html.erb
craigmdavidson: Added `visually_hidden_suffix: alp.lead_provider.name` for buttons and similar for links.

app/services/active_lead_providers/cascade_delete.rb
craigmdavidson: Done

app/services/active_lead_providers/cascade_delete.rb
craigmdavidson: Done

app/services/active_lead_providers/cascade_delete.rb
craigmdavidson: Done

app/controllers/admin/finance/active_lead_providers_controller.rb
craigmdavidson: Done

ethax-ross (approved):
Nice work! 👏 

It may be worth offloading the seeding to a job in the future as it can take a good few seconds to create all the data. Not a big deal as this is the admin UI, but maybe worth considering as an improvement down the line.

tonyheadford (approved):
Looks great!

PR #2947, opened 2026-05-19
Fix overlapping training periods

peteryates:
Since changing the training period range inclusivity, these training periods are now invalid. They should finish the day before the next one starts, not on the same day.

This script nudges their `finished_on` back by one day.

As we're now allowing periods that start and finish on the same day, the validation has been updated to reflect that.

### DIFF
--- a/db/migrate/20260519161732_change_period_length_check_constraints.rb
+++ b/db/migrate/20260519161732_change_period_length_check_constraints.rb
@@ -2,44 +2,44 @@ class ChangePeriodLengthCheckConstraints < ActiveRecord::Migration[8.1]
   def up
     # ECTAtSchoolPeriods
     remove_check_constraint :ect_at_school_periods, name: "period_length_greater_than_zero", if_exists: true
-    add_check_constraint :ect_at_school_periods, "finished_on >= started_on", name: "finished_on_not_before_started_on", if_exists: true, validate: true
+    add_check_constraint :ect_at_school_periods, "finished_on >= started_on", name: "finished_on_not_before_started_on", if_not_exists: true, validate: true
 
     # MentorAtSchoolPeriods
     remove_check_constraint :mentor_at_school_periods, name: "period_length_greater_than_zero", if_exists: true
-    add_check_constraint :mentor_at_school_periods, "finished_on >= started_on", name: "finished_on_not_before_started_on", if_exists: true, validate: true
+    add_check_constraint :mentor_at_school_periods, "finished_on >= started_on", name: "finished_on_not_before_started_on", if_not_exists: true, validate: true
 
     # TrainingPeriods
     remove_check_constraint :training_periods, name: "period_length_greater_than_zero", if_exists: true
-    add_check_constraint :training_periods, "finished_on >= started_on", name: "finished_on_not_before_started_on", if_exists: true, validate: true
+    add_check_constraint :training_periods, "finished_on >= started_on", name: "finished_on_not_before_started_on", if_not_exists: true, validate: true
 
     # MentorshipPeriods
     remove_check_constraint :mentorship_periods, name: "period_length_greater_than_zero", if_exists: true
-    add_check_constraint :mentorship_periods, "finished_on >= started_on", name: "finished_on_not_before_started_on", if_exists: true, validate: true
+    add_check_constraint :mentorship_periods, "finished_on >= started_on", name: "finished_on_not_before_started_on", if_not_exists: true, validate: true
 
     # ContractPeriods
     remove_check_constraint :contract_periods, name: "period_length_greater_than_zero", if_exists: true
-    add_check_constraint :contract_periods, "finished_on >= started_on", name: "finished_on_not_before_started_on", if_exists: true, validate: true
+    add_check_constraint :contract_periods, "finished_on >= started_on", name: "finished_on_not_before_started_on", if_not_exists: true, validate: true
   end
 
   def down
     # ECTAtSchoolPeriods
     remove_check_constraint :ect_at_school_periods, name: "finished_on_not_before_started_on", if_exists: true
-    add_check_constraint :ect_at_school_periods, "finished_on >= started_on", name: "period_length_greater_than_zero", if_exists: true, validate: true
+    add_check_constraint :ect_at_school_periods, "finished_on > started_on", name: "period_length_greater_than_zero", if_not_exists: true, validate: true
 
     # MentorAtSchoolPeriods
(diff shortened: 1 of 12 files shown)


cpjmcquillan (approved):

db/scripts/fix_overlapping_training_periods.rb
```diff
@@ -0,0 +1,45 @@
...
+  with
+    ect_data as (
+      select easp.teacher_id, 'ect' as type, tp.id, tp.started_on, tp.finished_on
+      from training_periods tp
+      inner join ect_at_school_periods easp on tp.ect_at_school_period_id = easp.id
+    ),
+    mentor_data as (
+      select masp.teacher_id, 'mentor' as type, tp.id, tp.started_on, tp.finished_on
+      from training_periods tp
+      inner join mentor_at_school_periods masp on tp.ect_at_school_period_id = masp.id
```
rwrrll: Is this supposed to join to ECT periods?

db/migrate/20260519161732_change_period_length_check_constraints.rb
```diff
@@ -0,0 +1,45 @@
...
+
+    # ContractPeriods
+    remove_check_constraint :contract_periods, name: "period_length_greater_than_zero", if_exists: true
+    add_check_constraint :contract_periods, "finished_on >= started_on", name: "finished_on_not_before_started_on", if_exists: true, validate: true
+  end
+
+  def down
+    # ECTAtSchoolPeriods
+    remove_check_constraint :ect_at_school_periods, name: "finished_on_not_before_started_on", if_exists: true
+    add_check_constraint :ect_at_school_periods, "finished_on >= started_on", name: "period_length_greater_than_zero", if_exists: true, validate: true
```
rwrrll: I think all these are the same constraint just with a different name, is that intended, or should it change the constraint?

db/migrate/20260519161732_change_period_length_check_constraints.rb
```diff
@@ -0,0 +1,45 @@
+class ChangePeriodLengthCheckConstraints < ActiveRecord::Migration[8.1]
+  def up
+    # ECTAtSchoolPeriods
+    remove_check_constraint :ect_at_school_periods, name: "period_length_greater_than_zero", if_exists: true
+    add_check_constraint :ect_at_school_periods, "finished_on >= started_on", name: "finished_on_not_before_started_on", if_exists: true, validate: true
```
rwrrll: When adding a constraint do we want `if_not_exists` instead?

db/scripts/fix_overlapping_training_periods.rb
ericaporter: Should this be `tp.mentor_at_school_period_id` instead of `tp.ect_at_school_period_id` ?

db/scripts/fix_overlapping_training_periods.rb
```diff
@@ -0,0 +1,45 @@
...
+# training period ranges
+#
+# Here, the query joins training periods to themselves using the teacher_id
+# and gets earlier/later pairs by joining one record's started_on to the other
+# record's finished_on
+#
+# Note, we're adding the 'type' field so we don't join ECT training periods to
+# mentor training periods
+#
+# For each one of the results we want to push the finshed_on date one day earlier
```
rwrrll:
```suggestion
# For each one of the results we want to push the finished_on date one day earlier
```

spec/models/contract_period_spec.rb
```diff
@@ -148,10 +148,10 @@
   end
 
   describe "check constraints" do
-    subject { FactoryBot.build(:contract_period, started_on: Date.current, finished_on: Date.current) }
+    subject { FactoryBot.build(:contract_period, started_on: Date.current, finished_on: Date.yesterday) }
 
-    it "prevents 0 day periods from being written to the database" do
-      expect { subject.save(validate: false) }.to raise_error(ActiveRecord::StatementInvalid, /PG::CheckViolation/)
+    it "prevents periods with a duration less than 1 day from being written to the database" do
+      expect { subject.save(validate: false) }.to raise_error(ActiveRecord::StatementInvalid, /PG::DataException/)
```
rwrrll, line 154: This suggests maybe we don't need the constraint if the data type rejects the value, I guess no harm in having the constraint too although it does have a performance penalty, haven't properly thought it through so just mentioning in case it's a thing!

rwrrll (commented):
Had a quick look!

db/migrate/20260519161732_change_period_length_check_constraints.rb
```diff
@@ -0,0 +1,45 @@
...
+    remove_check_constraint :ect_at_school_periods, name: "finished_on_not_before_started_on", if_exists: true
+    add_check_constraint :ect_at_school_periods, "finished_on >= started_on", name: "period_length_greater_than_zero", if_exists: true, validate: true
+
+    # MentorAtSchoolPeriods
+    remove_check_constraint :mentor_at_school_periods, name: "finished_on_not_before_started_on", if_exists: true
+    add_check_constraint :mentor_at_school_periods, "finished_on >= started_on", name: "period_length_greater_than_zero", if_exists: true, validate: true
+
+    # TrainingPeriods
+    remove_check_constraint :training_periods, name: "finished_on_not_before_started_on", if_exists: true
+    add_check_constraint :training_periods, "finished_on >= started_on", name: "period_length_greater_than_zero", if_exists: true, validate: true
```
ericaporter: Should your down migrations restore the previous `finished_on > started_on`?

spec/models/contract_period_spec.rb
```diff
@@ -148,10 +148,10 @@
...
     end
```
ericaporter, line 155: Might be worth asserting that started_on can eq finished_on and be valid, to be super explicit

db/scripts/fix_overlapping_training_periods.rb
peteryates: :facepalm: Yes it should, thank you!

db/migrate/20260519161732_change_period_length_check_constraints.rb
cpjmcquillan: Good spot!

db/migrate/20260519161732_change_period_length_check_constraints.rb
cpjmcquillan: Good spot!

db/migrate/20260519161732_change_period_length_check_constraints.rb
cpjmcquillan: Good spot!

PR #2948, opened 2026-05-19
Schools can assign induction tutor details to the upcoming contract period

cpjmcquillan:
### Context

Driven by https://github.com/DFE-Digital/register-early-career-teachers-public/pull/2922

Before, we only ever cared about the current contract period.

In https://github.com/DFE-Digital/register-early-career-teachers-public/pull/2922, the comparison between the last updated year for a school's
induction tutor details and `ContractPeriod.current` raised errors when there
was no "current" contract period (i.e in the gap).

Between the end of the 2025 contract period and the start of the 2026 contract
period, we still need users to be able to access the service (although lots of
functionality is restricted).

This reworks the induction tutor details checks so that they look to the future
in the absence of a "current" contract period.

Now, users are prompted to assign induction tutor details for the "upcoming"
contract period when there is no "current" contract period (i.e in "the gap").

This seemed pragmatic, since we would have asked schools to submit this
information as soon as the next contract period opened. This gives schools an
opportunity to fulfil that obligation sooner.

Crucially, no errors are raised either. If there is neither a current or an
upcoming contract period, then no update to induction tutor details is required.
The user is able to access the service, but registration is blocked.

### Guidance to review

We might want to wait for #2772 to land and temporarily edit the CP dates (as in #2922) to review this

### DIFF
--- a/app/models/contract_period.rb
+++ b/app/models/contract_period.rb
@@ -34,12 +34,11 @@ class ContractPeriod < ApplicationRecord
   end
 
   def self.current = containing_date(Date.current)
+  def self.upcoming = closest_to(Date.current).starting_tomorrow_or_after.first
+  def self.current_or_upcoming = current.presence || upcoming
 
-  def self.upcoming
-    closest_to(Date.current)
-      .merge(current_or_future)
-      .excluding(current)
-      .first
+  def self.current_or_upcoming!
+    current_or_upcoming || raise(StandardError, "No current or upcoming contract period")
   end
 
   def self.earliest_permitted_start_date
--- a/app/models/contract_period.rb
+++ b/app/models/contract_period.rb
@@ -35,6 +35,13 @@ class ContractPeriod < ApplicationRecord
 
   def self.current = containing_date(Date.current)
 
+  def self.upcoming
+    closest_to(Date.current)
+      .merge(current_or_future)
+      .excluding(current)
+      .first
+  end
+
   def self.earliest_permitted_start_date
     return unless current
 
--- a/app/services/schools/induction_tutor_details.rb
+++ b/app/services/schools/induction_tutor_details.rb
@@ -1,6 +1,7 @@
 module Schools
   class InductionTutorDetails
     include Rails.application.routes.url_helpers
+
     attr_reader :user
 
     def initialize(user)
@@ -11,8 +12,9 @@ module Schools
       return unless user&.school_user?
       return if user.dfe_user_impersonating_school_user?
       return unless school
+      return unless closest_contract_period
 
-      last_updated_year.blank? || last_updated_year < ContractPeriod.current.year
+      last_updated_year.blank? || last_updated_year < closest_contract_period.year
     end
 
     def wizard_path
@@ -25,12 +27,11 @@ module Schools
 
   private
 
-    def school
-      @school ||= @user&.school
-    end
+    def school = @school ||= @user&.school
+    def last_updated_year = school.induction_tutor_last_nominated_in&.year
 
-    def last_updated_year
-      school.induction_tutor_last_nominated_in&.year
+    def closest_contract_period
+      @closest_contract_period ||= ContractPeriod.current_or_upcoming
     end
   end
 end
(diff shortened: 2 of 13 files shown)


app/services/schools/induction_tutor_details.rb
```diff
@@ -12,7 +13,7 @@ def update_required?
       return if user.dfe_user_impersonating_school_user?
       return unless school
```
thenapking, line 16: Since this class effectively blocks school users from accessing the site if their induction tutor details are not up to date, and we have this very specific bit of business logic to implement, is it worth raising a meaningful error here if no `contract_period` is found?

app/models/contract_period.rb
```diff
@@ -35,6 +35,13 @@ def self.containing_date(date)
 
   def self.current = containing_date(Date.current)
 
+  def self.upcoming
+    closest_to(Date.current)
+      .merge(current_or_future)
+      .excluding(current)
```
thenapking: It took me a couple of goes at understanding this because of the naming.  `.excluding(current)` excludes the record found by the `self.current` method, if I am understanding this correctly.  But in the `Interval` concern the `current_or_future` scope means `ongoing_today` or `starting_tomorrow_or_after`.  It sort of reads like you just want a `future` scope, because you are excluding the `current` part of it, but that isn't how the logic actually works. I think you can get the same result with `.merge(starting_tomorrow_or_after)` and it's easier to read.

thenapking (commented):
I think this has re-written the wizards for changing the induction tutor well.  There's a couple of things I raised.  I think we need to test the routing to ensure that this change works when there is no current contract period.  There's a shared RSpec example for "an induction redirectable route" which is an integration test that tests the re-routing that the InductionTutorService can cause.  It would be good to include an example there that tests this works when we are in the gap between contract periods.

app/services/schools/induction_tutor_details.rb
```diff
@@ -32,5 +33,9 @@ def school
     def last_updated_year
       school.induction_tutor_last_nominated_in&.year
     end
+
+    def closest_contract_period
+      ContractPeriod.current.presence || ContractPeriod.upcoming
+    end
```
avinhurry, line 35: Had a couple of thoughts here:

1. It could be worth raising a meaningful error when neither a current nor upcoming contract period exists. That should be an invalid setup, but a deliberate error would be clearer than a later `NoMethodError` on `.year`.

2. This lookup feels like it belongs on `ContractPeriod`, since it is now shared business logic. Then this helper could become something like:

```ruby
def closest_contract_period
  ContractPeriod.current_or_upcoming!
end
```

WDYT?

app/services/schools/induction_tutor_details.rb
cpjmcquillan: We've introduced `ContractPeriod.current_or_upcoming!` which raises an error if neither is there 👍🏼

app/services/schools/induction_tutor_details.rb
cpjmcquillan: Both sounded good!

spec/services/schools/induction_tutor_details_spec.rb
```diff
@@ -97,27 +82,57 @@
...
 
-      it "returns true" do
-        expect(service).to be_update_required
+      it { is_expected.not_to be_update_required }
+    end
+
+    context "when the current contract period has finished but the upcoming contract period has not started" do
+      let!(:previous_contract_period) { FactoryBot.create(:contract_period, :previous) }
+      let!(:current_contract_period) do
+        FactoryBot.create(:contract_period, :current, finished_on: Date.yesterday)
```
avinhurry: I think this will fail on 1 June each year because the `:current` trait will set `started_on` to 1 June, but `finished_on: Date.yesterday` will be 31 May.

Probably worth specifying `started_on` here too so the setup works all the time?

avinhurry (approved):

app/models/contract_period.rb
```diff
@@ -34,6 +34,12 @@ def self.containing_date(date)
   end
 
   def self.current = containing_date(Date.current)
+  def self.upcoming = closest_to(Date.current).starting_tomorrow_or_after.first
```
thenapking, line 37: This is much easier to understand now!

PR #2951, opened 2026-05-20
[3923] Add paid and clawed back events to declarations

thenapking:
### Context
We have imported declarations and statements from ECF1.  However their state history is drawn from the events table, and it is missing.

In order to back fill the missing events for declarations imported from ECF1 we need to create two events: a declaration is paid and a  declaration is clawed back.

https://github.com/DFE-Digital/register-ects-project-board/issues/3923

A later PR will add a rake task which will backfill all statements and declarations which are marked as paid, but are missing the events.

This is also a pre-requisite of: https://github.com/DFE-Digital/register-ects-project-board/issues/3397

That will require a service class which will mark each declaration as paid/clawed back, as well as creating the events.  It will be pretty similar to the rake task's code obviously.

### Guidance to review
At this stage there is not much to product review.

### DIFF
--- a/spec/factories/declaration_factory.rb
+++ b/spec/factories/declaration_factory.rb
@@ -152,7 +152,7 @@ FactoryBot.define do
         declaration.clawback_status = clawback_status
 
         if Declaration::BILLABLE_PAYMENT_STATUSES.include?(payment_status) && !evaluator.payment_statement
-          statement_trait = payment_status == "eligable" ? :open : payment_status.to_sym
+          statement_trait = payment_status == "eligible" ? :open : payment_status.to_sym
           declaration.payment_statement = FactoryBot.create(:statement, statement_trait, active_lead_provider:)
         else
           declaration.payment_statement = evaluator.payment_statement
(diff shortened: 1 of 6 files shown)


spec/factories/declaration_factory.rb
```diff
@@ -141,6 +148,22 @@
         declaration.training_period = training_period
         milestone = training_period.schedule.milestones.find { |m| m.declaration_type == declaration.declaration_type }
         declaration.declaration_date = milestone.start_date + 1.day
+        declaration.payment_status = payment_status
+        declaration.clawback_status = clawback_status
+
+        if Declaration::BILLABLE_PAYMENT_STATUSES.include?(payment_status) && !evaluator.payment_statement
+          statement_trait = payment_status == "eligable" ? :open : payment_status.to_sym
```
ericaporter: typo: "eligible"

spec/factories/declaration_factory.rb
```diff
@@ -184,6 +215,22 @@
...
```
ericaporter: "eligible"

ericaporter (approved):
Looks good, just a few small copy updates!

spec/factories/declaration_factory.rb
thenapking: Clearly I cannot spell that word.

PR #2952, opened 2026-05-20
[3943] Add statements CRUD

mooktakim:
### Context

Resolves https://github.com/DFE-Digital/register-ects-project-board/issues/3943

#### Create
Screenshot 2026-05-27 at 16 07 49

#### Show 
Screenshot 2026-05-27 at 16 04 49

#### Edit 
Screenshot 2026-05-27 at 16 04 57

#### Delete
Screenshot 2026-05-27 at 16 05 08

### Guidance to review

This is just a standard (though quite convoluted) Rails controller. The unusual thing is a GET `delete` action to confirm deletion.


app/services/statements/create.rb
```diff
@@ -0,0 +1,26 @@
...
+    # @param params [ActionController::Parameters]
+    def initialize(author:, contract:, params:)
+      @author = author
+      @statement = Statement.new(params.merge(contract:))
+    end
+
+    # @return [Statement]
+    # @raise [ActiveRecord::RecordInvalid]
+    def create_statement!
+      raise ActiveRecord::RecordInvalid, statement unless statement.valid?
```
tonyheadford: This is superfluous isn't it?

app/services/statements/update.rb
```diff
@@ -0,0 +1,30 @@
...
+      @params = params
+    end
+
+    # @return [Statement]
+    # @raise [ActiveRecord::RecordInvalid]
+    def update_statement!
+      statement.assign_attributes(params)
+      modifications = statement.changes
+
+      raise ActiveRecord::RecordInvalid, statement unless statement.valid?
```
tonyheadford: I don't think this is necessary here either is it?

app/services/statements/create.rb
```diff
@@ -0,0 +1,19 @@
...
+    def create_statement!
+      ActiveRecord::Base.transaction do
+        statement.save!
+        Events::Record.record_statement_created_event!(author:, statement:)
+      end
+
+      statement
+    end
+  end
+end
```
craigmdavidson, line 19: 🤯 😆

app/views/admin/finance/active_lead_providers/statements/delete.html.erb
```diff
@@ -0,0 +1,49 @@
+<%
```
craigmdavidson, line 1: 🤔 This is quite an odd pattern?

app/controllers/admin/finance/active_lead_providers/statements_controller.rb
```diff
@@ -16,12 +18,85 @@ def index
...
       @active_lead_provider = ActiveLeadProvider
         .includes(:contract_period, :lead_provider)
         .find(params[:active_lead_provider_id])
     end
+
+    def set_statement
+      @statement = @active_lead_provider.statements.find(params[:id])
+    end
+
+    def redirect_unless_editable
```
craigmdavidson, line 78: I'm not sure if this applies here?

app/controllers/admin/finance/active_lead_providers/statements_controller.rb
```diff
@@ -16,12 +18,85 @@ def index
...
+
+    def edit
+      # This is idiomatic Rails, so we need a comment to keep sonarcube happy.
+    end
+
+    def update
+      ::Statements::Update.new(
+        author: current_user,
+        statement: @statement,
+        params: statement_params.except(:contract_id)
```
craigmdavidson: I am not sure why we restrict editing a statement's contract? I don't see anything about that in the GitHub issue itself - so I assume it is due to a conversation between Mook and @ethax-ross?

app/controllers/admin/finance/active_lead_providers/statements_controller.rb
```diff
@@ -16,12 +18,85 @@ def index
...
+      ).call
+
+      redirect_to statement_path(@statement), notice: "Statement updated"
+    rescue ActiveRecord::RecordInvalid
+      render :edit, status: :unprocessable_content
+    end
+
+    def delete
+      # This is our project specific pattern, so we need a comment to keep sonarcube happy.
+    end
```
craigmdavidson, line 56: ❓❗

app/controllers/admin/finance/active_lead_providers/statements_controller.rb
ethax-ross: No I'm not sure why we'd block that either; I think we want to allow changing the associated contract but not the lead provider (so as long as we scope contracts to the same LP it should be fine). It's quite a common use case that a new contract will be created and existing statements assigned to it.

app/controllers/admin/finance/active_lead_providers/statements_controller.rb
ethax-ross: I think sonarcloud complains about empty controller actions so a comment has been added? I'm not sure if/where we can configure that but I'm all for ignoring it.

app/controllers/admin/finance/active_lead_providers/statements_controller.rb
```diff
@@ -16,12 +18,86 @@ def index
...
         .includes(:contract_period, :lead_provider)
         .find(params[:active_lead_provider_id])
     end
+
+    def set_statement
+      @statement = @active_lead_provider.statements.find(params[:id])
+    end
+
+    def redirect_unless_editable
+      return unless @active_lead_provider.contract_period.started_on_or_before_today?
```
ethax-ross: We also have `editable?` on `ContractPeriod` (which is essentially an alias for `!started_on_or_before_today?`)

app/controllers/admin/finance/active_lead_providers/statements_controller.rb
```diff
@@ -16,12 +18,86 @@ def index
...
+      @statement = e.record
+      render :new, status: :unprocessable_content
+    end
+
+    def edit
+      # This is idiomatic Rails, so we need a comment to keep sonarcube happy.
+    end
+
+    def update
+      ::Statements::Update.new(
```
ethax-ross, line 43: Something I noticed when testing is it looks like we could flip the fee type of a statement from output to service even if it has declarations, which we should block. I've ticketed this up, we can pick it up separately https://github.com/DFE-Digital/register-ects-project-board/issues/4018

app/views/admin/finance/active_lead_providers/statements/_fields.html.erb
```diff
@@ -0,0 +1,36 @@
+<% contract_period = @active_lead_provider.contract_period %>
+
+<%= content_for(:error_summary) { form.govuk_error_summary } %>
+
+<%= form.govuk_collection_select :contract_id,
+    @active_lead_provider.contracts,
+    :id,
+    :contract_type,
```
ethax-ross: Can we maybe humanize and capitalize these (so ITTECF ECTP); I think we'll need a distinguishing element as well for when there are multiple, maybe the date the contract was created would do for now?

spec/models/statement_spec.rb
```diff
@@ -295,4 +295,16 @@
       end
     end
   end
+
+  describe "#referenced_by_declarations?" do
+    it "is false when there are no payment or clawback declarations" do
+      expect(FactoryBot.create(:statement)).not_to be_referenced_by_declarations
+    end
+
+    it "is true when a payment declaration exists" do
```
ethax-ross: Worth covering when only a `clawback_declaration` exists?

app/services/events/record.rb
```diff
@@ -1072,6 +1072,70 @@ def self.record_contract_period_updated_event!(author:, contract_period:, modifi
...
+
+    def self.record_statement_created_event!(author:, statement:)
+      event_type = :statement_created
+
+      active_lead_provider = statement.active_lead_provider
+      lead_provider        = active_lead_provider.lead_provider
+      heading              = "Statement created: #{Statements::Period.for(statement)} #{statement.fee_type} for #{lead_provider.name}"
+
+      metadata = {
+        contract_period_year: active_lead_provider.contract_period_year,
```
ethax-ross: I'm not really sure where/when `metadata` should be used on events 🤔 I thought it was just for things that we don't link directly to the event, so for example when a participant changes schedule we might send their original schedule in the metadata, but the event is linked directly to their new schedule. Maybe one for @peteryates to clarify.

spec/services/statements/create_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+      contract_id: contract.id,
+      month: 11,
+      year: contract_period.year,
+      fee_type: "output",
+      deadline_date: Date.new(contract_period.year, 11, 1),
+      payment_date: Date.new(contract_period.year, 12, 25)
+    }
+  end
+
+  around { |example| perform_enqueued_jobs { example.run } }
```
ethax-ross: I think this is redundant as we're asserting on the `Events::Record` object and not the event that the job creates? ditto for destroy/update.

app/services/statements/destroy.rb
```diff
@@ -0,0 +1,33 @@
...
+      @author = author
+      @statement = statement
+    end
+
+    def call
+      raise DeletionError, "Cannot delete a statement with declarations" if statement.referenced_by_declarations?
+
+      active_lead_provider = statement.active_lead_provider
+      lead_provider = active_lead_provider.lead_provider
+      heading = "Statement deleted: #{Statements::Period.for(statement)} #{statement.fee_type} for #{lead_provider.name}"
```
ethax-ross, line 17: Should this be encapsulated in `record_statement_delete_event!` like it is for create/updatet?

app/services/statements/destroy.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    attr_reader :statement, :author
+
+    def initialize(author:, statement:)
+      @author = author
+      @statement = statement
+    end
+
+    def call
+      raise DeletionError, "Cannot delete a statement with declarations" if statement.referenced_by_declarations?
```
ethax-ross, line 13: I wonder if its worth pushing this rule down into the model as I expect its a constraint we would want to apply universally 🤔

app/views/admin/finance/active_lead_providers/statements/delete.html.erb
```diff
@@ -0,0 +1,49 @@
...
+  statement_path = admin_contract_period_active_lead_provider_statement_path(contract_period, @active_lead_provider, @statement)
+  page_data(
+    title: "Delete statement",
+    caption: @active_lead_provider.lead_provider_name,
+    backlink_href: statement_path
+  )
+%>
+
+<%=
+  govuk_summary_list do |sl|
```
ethax-ross: Is it worth pulling the summary list from `show` into a partial we can reuse here?

app/views/admin/finance/active_lead_providers/statements/show.html.erb
```diff
@@ -0,0 +1,57 @@
+<%
+  contract_period = @active_lead_provider.contract_period
+  page_data(
+    title: "Statement #{Statements::Period.for(@statement)}",
+    caption: @active_lead_provider.lead_provider_name,
+    backlink_href: admin_contract_period_active_lead_provider_statements_path(contract_period, @active_lead_provider)
+  )
+
+  status_colour = { "open" => "blue", "payable" => "yellow", "paid" => "green" }.fetch(@statement.status)
```
ethax-ross: It looks like we have [this presenter](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/presenters/admin/statement_presenter.rb#L12) we could reuse

ethax-ross (commented):
Nicely done Craig! 👏 just a few minor thoughts/suggestions

app/services/events/record.rb
peteryates: There are no hard and fast rules really. It was added originally so when we merged multiple induction periods from DQT into one on RIAB we stashed a copy of the originals just in case we ever needed to refer back to them.

I think the `modifications` column would hold this kind of data in a way more consistent with what we do elsewhere.

app/views/admin/finance/active_lead_providers/statements/edit.html.erb
```diff
@@ -0,0 +1,20 @@
...
+  )
+%>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_with model: @statement,
+                   url: admin_contract_period_active_lead_provider_statement_path(contract_period, @active_lead_provider, @statement),
+                   method: :patch,
+                   local: true do |f| %>
+      <%= render "fields", form: f %>
```
peterdavidhamilton: Could pass in `contract_period` as a local to the partial

app/services/statements/destroy.rb
craigmdavidson: 💯 No! The whole purpose of this "service" is to delete the model.

Either this service should _always_ be used, so the logic should be in the service. Or the service is redundant.

This is my problem with these "services".

(FWIW: Rails already gives us this as: **ActiveRecord::InvalidForeignKey**.)

app/services/statements/destroy.rb
craigmdavidson: The problem is the event is recorded after the statement is deleted, so we shouldn't really be using the statement object to construct the heading?

So `record_statement_delete_event!` doesn't get the deleted statement, to build the heading from.

app/views/admin/finance/active_lead_providers/statements/show.html.erb
craigmdavidson: 😬 😆

app/services/events/record.rb
craigmdavidson: The recording of `create` events generally do not seem to have metadata nor modifications. I think this was just crappy code. We pass in the newly created entity.

I'll remove the metadata.

app/controllers/admin/finance/active_lead_providers/statements_controller.rb
craigmdavidson: done

app/controllers/admin/finance/active_lead_providers/statements_controller.rb
craigmdavidson: Out of scope.

app/views/admin/finance/active_lead_providers/statements/_fields.html.erb
craigmdavidson: Done

spec/models/statement_spec.rb
craigmdavidson: Done

app/services/events/record.rb
craigmdavidson: Done.

spec/services/statements/create_spec.rb
craigmdavidson: Done.

app/services/statements/destroy.rb
craigmdavidson: @ethax-ross - fair? not fair?

app/services/statements/destroy.rb
craigmdavidson: @ethax-ross - fair? not fair?

app/views/admin/finance/active_lead_providers/statements/delete.html.erb
craigmdavidson: Done.

app/views/admin/finance/active_lead_providers/statements/show.html.erb
craigmdavidson: Done.

app/views/admin/finance/active_lead_providers/statements/edit.html.erb
craigmdavidson: Done.

app/views/admin/finance/active_lead_providers/statements/delete.html.erb
```diff
@@ -0,0 +1,30 @@
...
+  contract_period = @active_lead_provider.contract_period
+  statement_path = admin_contract_period_active_lead_provider_statement_path(contract_period, @active_lead_provider, @statement)
+  page_data(
+    title: "Delete statement",
+    caption: @active_lead_provider.lead_provider_name,
+    backlink_href: statement_path
+  )
+%>
+
+<%= render "statement", statement: @statement %>
```
craigmdavidson, line 11: We do this because **Breakman** is not happy when we use `<%= render @statement %>` despite that being the idiomatic style.

app/services/statements/destroy.rb
ethax-ross: > The problem is the event is recorded after the statement is deleted, so we shouldn't really be using the statement object to construct the heading?
> 
> So `record_statement_delete_event!` doesn't get the deleted statement, to build the heading from.

Even though the statement is destroyed I'm sure it remains in-memory, so we could build the heading/modifications from it without issues? we could call it out in the parameter as `destroyed_statement:` to avoid potential pitfalls 🤔 I don't feel super strongly about it, though.

app/services/statements/destroy.rb
ethax-ross: I think the service _should_ always be used, but having the validation in the base model still feels like the more robust approach. That being said, I take your point around the service becoming somewhat redundant 🙂 shall we leave it as-is and I can add it as a point to consider for the ticket we have to spike/refactor the modelling around the opening reg work?

app/views/admin/finance/active_lead_providers/statements/delete.html.erb
ethax-ross: Really?! what issue did `brakeman` have with it? I don't see a security concern 😕

app/views/admin/finance/active_lead_providers/statements/_fields.html.erb
ethax-ross: Thanks, worth changing it here too and adding a unit test? https://github.com/DFE-Digital/register-early-career-teachers-public/pull/2952/changes#diff-8ae7c0c119bc858128fa396675e6d7a1213376c9caefb0ec9827d22a736fedceR10

ethax-ross (approved):
Thanks Craig!

app/views/admin/finance/active_lead_providers/statements/delete.html.erb
craigmdavidson: I know right?

https://github.com/DFE-Digital/register-early-career-teachers-public/actions/runs/26573206049/job/78285470002

app/views/admin/finance/active_lead_providers/statements/_fields.html.erb
craigmdavidson: Done.

PR #2953, opened 2026-05-20
Use GIAS contact email in the API when the school induction tutor email is not set

tonyheadford:
### Context

In the school and school partnerships API endpoints, use the GIAS primary or secondary contact email for the induction tutor email returned in the API response when the induction tutor email has not been set on the School.

Adds a service `Schools::InductionTutorEmail` with the above logic
Updates the school and school partnership serializers to use the new service for the induction tutor email value

### DIFF
--- a/spec/services/schools/induction_tutor_email_spec.rb
+++ b/spec/services/schools/induction_tutor_email_spec.rb
@@ -1,0 +1,57 @@
+describe Schools::InductionTutorEmail do
+  subject(:service) { described_class.new(school:) }
+
+  let(:school) { FactoryBot.create(:school, gias_school:, induction_tutor_email:, induction_tutor_name:, induction_tutor_last_nominated_in: nil) }
+  let(:gias_school) { FactoryBot.create(:gias_school, primary_contact_email:, secondary_contact_email:) }
+
+  let(:induction_tutor_name) { "Griff Walnut" }
+  let(:induction_tutor_email) { "griff@example.com" }
+  let(:primary_contact_email) { "head@example.com" }
+  let(:secondary_contact_email) { "bernard@example.com" }
+
+  describe "#email" do
+    it "returns the school induction tutor email" do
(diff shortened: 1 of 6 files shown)


ethax-ross (commented):
Thanks Tony, can we also update the `SchoolSerializer` with the same behaviour please:

https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/serializers/api/school_serializer.rb#L27

We scoped out this ticket so long ago I also forgot that Claire was going to check with Matt to see if we should take into account DfE sign in SIT details. I'm not sure if she got around to this or not, could you check with Matt if there's anything to do there?

https://github.com/DFE-Digital/register-ects-project-board/issues/2645#issuecomment-4303868814

spec/services/schools/induction_tutor_email_spec.rb
```diff
@@ -0,0 +1,57 @@
+describe Schools::InductionTutorEmail do
+  subject(:service) { described_class.new(school:) }
+
+  let(:school) { FactoryBot.create(:school, gias_school:, induction_tutor_email:, induction_tutor_name:, induction_tutor_last_nominated_in: nil) }
+  let(:gias_school) { FactoryBot.create(:gias_school, primary_contact_email:, secondary_contact_email:) }
+
+  let(:induction_tutor_name) { "Griff Walnut" }
```
ethax-ross, line 7: Can we just omit the name as it doesn't have a bearing? or is it required if email is present?

ethax-ross (approved):

ethax-ross (commented):
Thanks @tonyheadford can we also check with @andycrawshaw or @easeynathan if we want to add a release note for this change

spec/services/schools/induction_tutor_email_spec.rb
tonyheadford: Yes, it's required if email is present

PR #2954, opened 2026-05-20
[3966] Allow opting out of emails for one term

rwrrll:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/3966

Let SITs opt out of emails reminding them to register ECTs and Mentors for the rest of the current term by clicking a link in the reminder email.

A bit of background to a couple of decisions:

1. We don't want to make people sign in when they click the opt out link in the email, so we need the URL to be authenticated. Rather than having to store/manage a secret token per school, we can sign the school ID and include it as a param. Using HMAC here because Postgres can also generate those with `pgcrypto`, so it's compatible with [the forthcoming Blazer query to produce the recipient list / other personalisations](https://github.com/DFE-Digital/register-ects-project-board/issues/3173). I've added an example query to Blazer for this review app, as well as the comment on `ReminderEmailOptOutToken`.

2. We don't want to have to do anything when a new term starts. Setting `opted_out_of_emails_until` to the end of the term makes opt-outs expire implicitly with a query like `WHERE opted_out_of_reminder_emails_until IS NULL OR opted_out_of_reminder_emails_until < CURRENT_DATE()`

### Guidance to review

- I thought the request spec covers enough to negate feature/view specs, anything missing?
- Very tempting to spend more time on `Term`, but it's not clear if / where else it could be useful yet, might be YAGNI

> [!IMPORTANT]
> ### Before merging/deploying:
> - Inform data team of new `opted_out_of_reminder_emails_until` column (notified, waiting to hear back)
> - ~Set `SCHOOL_REMINDER_EMAIL_OPT_OUT_TOKEN_SECRET` in Azure KV for each environment (e.g. `SecureRandom.hex(32)` or whatever) -- I've done review, but not other envs pending code review **Don't re-use the review app token, since that's discoverable from this public PR's review app Blazer queries.**~ done


app/controllers/schools/reminder_email_opt_outs_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  private
+
+    def authenticate_token!
+      @school = School.find(params[:school_id])
+      raise ActiveRecord::RecordNotFound unless
+        ReminderEmailOptOutToken.valid?(school_id: @school.id, token: params[:token])
+    end
+  end
+end
```
avinhurry, line 20: Since this controller inherits from `ApplicationController` it includes the [DfE analytics request tracking](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/controllers/application_controller.rb#L3), should we add `/school/opt-out-of-reminder-emails` to the [excluded path list](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/config/initializers/dfe_analytics.rb#L100) to prevent the token query params from reaching BQ?

app/controllers/schools/reminder_email_opt_outs_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+      @school.update!(opted_out_of_reminder_emails_until: Term.current.last_day)
+    end
+
+  private
+
+    def authenticate_token!
+      @school = School.find(params[:school_id])
+      raise ActiveRecord::RecordNotFound unless
+        ReminderEmailOptOutToken.valid?(school_id: @school.id, token: params[:token])
+    end
```
avinhurry, line 18: Minor hardening: Should we consider checking that `school_id` and `token` are strings before proceeding, so malformed params like `school_id[]=1` or `token[]=x` return `404` instead of potentially raising a `500`?

app/views/schools/reminder_email_opt_outs/create.html.erb
```diff
@@ -0,0 +1,7 @@
+<% page_data(title: "We will not send you reminder emails this term") %>
+
+<p>You've told us that your school does not currently have ECTs or mentors and you do not expect new ECTs or mentors this term.</p>
```
avinhurry:
```suggestion
<p>You’ve told us that your school does not currently have ECTs or mentors and you do not expect new ECTs or mentors this term.</p>
```

app/models/term.rb
```diff
@@ -0,0 +1,19 @@
+class Term < Data.define(:first_day, :last_day)
```
cpjmcquillan: I like this a lot, but I wonder if `Term` is too broad.

Is it worth us being a bit more specific? 🤔

If I was coming to this with fresh eyes I'd probably have a bunch of questions:

- are these school terms?
- are the dates arbitrary?
- why did we choose these dates?
- can I use this value object elsewhere? _should I?_ 

I wonder if we could add a bit more context in the commit message for future us, and perhaps name it slightly more explicitly.

`EmailOptOutTerm`, `OutboundCommsTerm`, `Schools::Term`, `Schools::DfE::Term`, or even `Schools::DfE::EmailOptOutTerm` 😆

What do you think?

config/initializers/rack_attack.rb
```diff
@@ -21,6 +21,7 @@ def remote_ip
     /sign-in
     /auth
     /healthcheck
+    /school/opt-out-of-reminder-emails
```
avinhurry, line 24: Adding the opt out URL here means throttled requests now go through `Rack::Attack` warning logger. That uses `request.fullpath`, so the signed token in the query string is written to the logs.

The spec update to expect `fullpath` (in `rate_limiting_support.rb`) makes the spec pass but also locks in the leak. 

I think we should log `request.path` here, or filter/redact query params before logging.

`request.fullpath` gives more debugging context, but it is too much when the query string contains a bearer token.

config/initializers/rack_attack.rb
avinhurry: Now that the protected routes cover more than just OTP, should we rename this to something broader like just "Protected routes"?

spec/requests/rate_limiting_spec.rb
```diff
@@ -23,6 +23,23 @@ def change_condition
     end
   end
 
+  context "when requesting the school reminder-email opt-out path" do
+    let(:school) { FactoryBot.create(:school) }
+    let(:path)   { new_schools_reminder_email_opt_out_path }
+
+    before { allow(Schools::ReminderEmailOptOutToken).to receive(:valid?).and_return(true) }
+
+    it_behaves_like "a rate limited endpoint", "protected routes (OTP)" do
```
cpjmcquillan: Is it worth changing the description for this `throttle` to reflect it being used for more than just OTP codes? 🤔

app/controllers/schools/reminder_email_opt_outs_controller.rb
```diff
@@ -0,0 +1,21 @@
+module Schools
+  class ReminderEmailOptOutsController < ApplicationController
+    before_action :authenticate_token!
+
+    def new
+      # renders new.html.erb (comment needed for sonarqube)
```
cpjmcquillan: I'd delete the comment and tell sonarqube to take a hike!

avinhurry (commented):
I really like the stateless HMAC approach here. Given the constraints around no sign in, Blazer generated recipient lists, and avoiding per school token storage, it feels like a good fit to me.

I've left a few comments. The main ones are around treating the opt out URL as a bearer token URL and keeping it out of logs/analytics. The rest are minor hardening/cleanup points.

app/controllers/schools/reminder_email_opt_outs_controller.rb
peteryates: Can we just add an exclusion for this in the sonarqube/sonarcloud properties? An empty method is _fine_, especially in controllers which are their own flavour of weird.

> it's a thing that we don't instantiate, it has methods we don't call, those methods don't take parameters, and the return value is ignored

app/services/schools/reminder_email_opt_out_token.rb
cpjmcquillan, line 1: Take this with a pinch of 🧂, I haven't actually tried it...

Could we hack something into [`Blazer.transform_variable`](https://github.com/ankane/blazer#encrypted-data) and lean on Rails and [`signed_id`](https://api.rubyonrails.org/classes/ActiveRecord/SignedId.html#method-i-signed_id)? 🤔

It might be a bit janky, but it would reduce the dependency on `pgcrypto`

Something like this

```ruby
# app/models/school.rb (or, in a service object)

class School < ApplicationRecord
  def self.signed_id_for_email_opt_out(id:)
    find(id).signed_id(expires_in: X.days.from_now, purpose: "school-reminder-email-opt-out")
  end
end
```

```yml
// config/blazer.yml

smart_columns:
  school_opt_out_token: "SELECT id, opt_out_token FROM schools WHERE id IN {value}"
```

```ruby
# config/initializers/blazer.rb

Blazer.transform_variable = lambda do |name, value|
  value = School.signed_id_for_email_opt_out(value) if name == "opt_out_token"
  value
end
```

```ruby
# app/controllers/schools/reminder_email_opt_outs_controller.rb

def authenticate_token!
  @school = School.find_signed!(params[:token])
end
```

I'm not up-to-speed with `blazer`, so I'm not sure that's all the correct syntax. It seems like it should be possible though 🙏🏼

What do you think?

app/services/schools/reminder_email_opt_out_token.rb
```diff
@@ -0,0 +1,46 @@
+module Schools
+  # Generates the HMAC token used in the reminder email opt-out link
+  #
+  # Equivalent SQL for use in Blazer:
+  #
+  #   SELECT
+  #     id,
+  #     '<host>/school/opt-out-of-reminder-emails/new?school_id=' || id ||
+  #       '&token=' || encode(hmac('school-reminder-email-opt-out:' || id::text,
+  #                                '<secret>', 'sha256'), 'hex')
```
cpjmcquillan, line 10: If we do go with this approach, could we do a couple of things to make the user experience slightly better? This is mostly speculation, though, I just _think_ it should work!

- Define the token as a [smart variable](https://github.com/ankane/blazer#smart-variables) and reference the secret there instead (I'm presuming currently we have to paste it into the blazer query)
- Use a [linked column](https://github.com/ankane/blazer#linked-columns) so we don't need to specify the host

app/controllers/schools/reminder_email_opt_outs_controller.rb
rwrrll: Yeah nice, calling `to_s` on both was slightly more succinct so went with that!

config/initializers/rack_attack.rb
rwrrll: Solid knowledge! Changed

app/models/term.rb
rwrrll: Yes 100% agree, obvious now you mention it! I went with `Schools::Term` since the dates are school terms first and foremost, and the opt out periods are intended to align with them (rather than being their own independent thing).

app/services/schools/reminder_email_opt_out_token.rb
rwrrll: Yeah, I actually went down this path as it seems tidier, the lump of concatenation SQL in Blazer is a bit icky, but I found:

`transform_variable` is to transform user input variables (that are filled in at the top of a Blazer query before it runs.) The example in the Blazer README that uses it for encrypted data assumes you're entering one value as a query input and want it encrypted before substitution into the SQL, unfortunately I don't think it transforms results.

`smart_columns` does transform result columns, but the transform is a SQL lookup against a data source (`SELECT id, opt_out_token FROM schools WHERE id IN {value}`). There's no hook to call Ruby (School#signed_id) from there. It's meant for joining display labels (like turning a user_id into a name), and does a funky thing where it kind of renders the "smart" value underneath the raw value in the output, you can see this if you include `urn` in a Blazer query. I thought it did what we want here, but alas.

`linked_columns` is plain string interpolation of {value} into a URL template, I don't _think_ there's a place to inject a computed token.

Because of the intention to bulk-export URLs straight out of a Blazer query result into Notify, any approach that requires Ruby to produce the URL means moving generation out of Blazer entirely (admin page or rake task or something) which is a bigger piece of work, it would scoop out a nice bit of ick but the tradeoff is a much larger liability of other stuff to operate.

At least pgcrypto is a stock PG extension used in lots of Teacher Services apps, rather than a gem or something, and we can rotate the env var to invalidate all outstanding tokens if we do need to.

Very happy to revisit if you can see a Blazer trick I missed, I did really hope to find some Blazer magic along these lines!

app/services/schools/reminder_email_opt_out_token.rb
rwrrll: Yeah again I had a look at these (great minds!!) but they didn't land.

`smart_variables` populate a dropdown of (value, label) pairs to pick from when running a query. They don't compute per-row values for a result set, and they're not really suited to holding a constant like a secret either (the secret would just be visible as the dropdown option). So this one doesn't quite map onto the mechanism.

You're right that we do have to paste the secret into the query definition, but at least it's only once when we create the query in Blazer, and again if we ever rotate the secret (unlikely). The other way I can see to do it is for an initializer or something to read the secret from ENV and write it into the database on boot somewhere that Blazer can JOIN, but this seems more complex and error-prone than the thing we'd like to avoid here :disappointed: 

`linked_columns` interpolate {value} (the current column's value) into a URL template, but only that single column, no access to siblings. So we can't return school_id and reach over to pick up token, and we can't compute the token inside the template either. If we return the full URL from SQL and apply a linked_column to it, Blazer just makes it clickable, the host is still in the SQL so not needing to specify the host doesn't really materialise.

The output also needs to be exportable as a CSV column going into Notify, so a Blazer-side clickable wrapper isn't quite what we need anyway, we need the literal URL string in the result cell.

I'm also definitely not a Blazer expert, the above is just based on trying to achieve the same thing you're looking for, so it may be possible in some way... but I don't know how! Definitely agree the SQL lump doesn't fill me with joy, but once set it's set and no-one should have to see it again :sweat_smile:

app/controllers/schools/reminder_email_opt_outs_controller.rb
rwrrll: TIL you can do this! I've just excluded the one controller for now :boot: but I would vote for disabling this rule for all public controller methods, I do not enjoy it!

rwrrll: TIL you can do this! I've just excluded the one controller for now :boot: but I would vote for disabling this rule for all public controller methods, I do not enjoy it!

config/initializers/dfe_analytics.rb
```diff
@@ -97,7 +97,11 @@
   # Translation missing: en.dfe.analytics.config.excluded_paths.description
   #
   # config.excluded_paths = Translation missing: en.dfe.analytics.config.excluded_paths.default
-  config.excluded_paths = ["/healthcheck", %r{^/admin(/.*)?$}]
+  config.excluded_paths = [
+    "/healthcheck",
+    %r{^/admin(/.*)?$},
+    %r{^/school/opt-out-of-reminder-emails(/.*)?$},
```
avinhurry: This fixes the normal controller analytics path, but i'm not sure throttled requests are covered?

Rack::Attack calls [`API::Request.send_throttled_request`](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/services/api/request.rb#L52), which builds a DfE analytics event with `with_request_details(request)`. That parses `rack_request.query_string` into [`request_query`](https://github.com/DFE-Digital/dfe-analytics/blob/main/lib/dfe/analytics/event.rb#L40), so the opt out `token` param may still be included for throttled requests.

Perhaps we could skip `send_throttled_request` for this path or make `send_throttled_request` respect `DfE::Analytics.config.excluded_paths`.

WDYT?

avinhurry (commented):
Thanks for making those changes.

There's one more issue I spotted relating to throttled analytics requests. Apologies, I didn't catch it in the last review. Other than that, this looks good to me now.

app/services/schools/reminder_email_opt_out_token.rb
cpjmcquillan: Annoying, but at least we tried!

We could create the query on boot from the Rails app, we do something similar in the seeds for non-Production environments. That should save the copy-paste of the secret at least 🙌🏼

config/initializers/dfe_analytics.rb
```diff
@@ -97,7 +97,11 @@
   # Translation missing: en.dfe.analytics.config.excluded_paths.description
   #
   # config.excluded_paths = Translation missing: en.dfe.analytics.config.excluded_paths.default
-  config.excluded_paths = ["/healthcheck", %r{^/admin(/.*)?$}]
+  config.excluded_paths = [
+    "/healthcheck",
+    %r{^/admin(/.*)?$},
+    %r{^/school/opt-out-of-reminder-emails(?:[/?].*)?$},
```
cpjmcquillan, line 103: Why do we want to exclude these email opt-outs from analytics? 🤔

cpjmcquillan (approved):
Looks good!

I just had a question about why we're excluding it from analytics, so might be worth adding some extra context to the commit message.

🚀

config/initializers/dfe_analytics.rb
rwrrll: Comment added to this commit!

PR #2957, opened 2026-05-21
Add contract to fee structure tables (part 1)

tonyheadford:
### Context

Now more is known about the shapes of the contracts and fee structures, it makes sense to reverse the relationship between them so that the `banded_fee_structure` and `flat_rate_fee_structure`  tables `belong_to` a `Contract` and a `Contract` `has_one` `banded_fee_structure` and/or `flat_rate_fee_structure` instead of the reverse.

This is the first step towards this which adds the `contract_id` to the fee structure tables and a script to populate that once the migration has been run.

- DB change to add the `contract_id` to `contract_banded_fee_structures` and `contract_flat_rate_fee_structures` tables
- A script that can be run once deployed to populate the `contract_id` value

### DIFF
--- a/db/scripts/populate_contract_id_on_fee_structures.rb
+++ b/db/scripts/populate_contract_id_on_fee_structures.rb
@@ -1,0 +1,8 @@
+# Populate contract_id value on fee structures so we can reverse the relationship
+#
+# $ kubectl exec -it <pod-name> -- bin/rails runner db/scripts/move_contract_id_to_fee_structures.rb
+
+Contract.find_each do |contract|
+  contract.banded_fee_structure&.update!(contract_id: contract.id)
+  contract.flat_rate_fee_structure&.update!(contract_id: contract.id)
+end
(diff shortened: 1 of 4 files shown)


db/scripts/populate_contract_id_on_fee_structures.rb
```diff
@@ -0,0 +1,8 @@
+# Populate contract_id value on fee structures so we can reverse the relationship
+#
+# $ kubectl exec -it <pod-name> -- bin/rails runner db/scripts/populate_contract_id_on_fee_structures.rb
+
+Contract.find_each do |contract|
```
ethax-ross, line 19: I'm 99% sure its not the case, but is it worth having a quick check to fail if there are any fee structures shared between contracts? something like:

```
shared_banded = Contract
  .where.not(banded_fee_structure_id: nil)
  .group(:banded_fee_structure_id)
  .having("COUNT(*) > 1")
  .exists?

shared_flat_rate = Contract
  .where.not(flat_rate_fee_structure_id: nil)
  .group(:flat_rate_fee_structure_id)
  .having("COUNT(*) > 1")
  .exists?

raise "Fee structures are shared between contracts!" if shared_banded || shared_flat_rate
```

ethax-ross (approved):
Thanks Tony

db/scripts/populate_contract_id_on_fee_structures.rb
tonyheadford: Yes, pretty sure I checked in migration with a similar query but it's a good idea to have it in the script

PR #2961, opened 2026-05-22
Add disaster recovery page

peteryates:
This is a draft copy of our disaster recovery doc.

There's no entire plan because we don't know what might go wrong, so instead there is:

* a list of things that might go wrong
* a list of steps we can take
* some processes which are used in the steps, like deploying the maintenance page or restoring a backup

### DIFF
--- a/documentation/disaster-recovery.md
+++ b/documentation/disaster-recovery.md
@@ -151,6 +151,11 @@ different combination of things in order to resolve it.
 
 ### Full data restore
 
+If we need to replace the database with an older version, we'll need to take the
+service down for multiple hours and inform all of our users.
+
+It should only be done if we suffer from a catastrophic data loss.
+
 If we need to restore the entire database we'll want to do it from a [point in
 time recovery](https://learn.microsoft.com/en-us/azure/postgresql/backup-restore/concepts-backup-restore#point-in-time-recovery) (PITR),
 choosing the latest time where we're sure the data is fully intact.
--- a/documentation/disaster-recovery.md
+++ b/documentation/disaster-recovery.md
@@ -1,0 +1,309 @@
+---
+title: Disaster recovery
+---
+
+Sometimes technology goes wrong.
+
+This document covers some scenarios and provides a list of actions we can take
+to resolve them. It also contains details about informing our users when there's
+a problem.
+
+There's no strict plan, every problem is different and we'll need to do a
+different combination of things in order to resolve it.
+
+## Things that could go wrong
+
+### GitHub goes down
+
+#### Symptoms
+
+- we can't deploy, our Docker images are built with GitHub actions and the
+  container registry we use is on GitHub
+- so long as there's nothing we need to publish imminently, we're ok
+
+#### Actions
+
+- wait for GitHub to come back up
+
+### DfE Sign-in goes down
+
+#### Symptoms
+
+- School users can't log in
+- Appropriate body users can't log in
+- Lead provider API unaffected
+- Admin interface unaffected
+
+#### Actions
+
+- display a [notification banner](#display-a-notification-banner) on the site informing users that there
+  are problems logging in
+- if it's a prolonged outage, hide the 'Sign in' button
+- wait for DfE Sign-in to come back up
+
+#### Contact
+
+- DfE Sign-in will have plans in place to let users know the service is down, us
+  sending extra comms might just confuse matters
+
+### GOV.UK Notify goes down
+
+#### Symptoms
+
+- Admins can't log in
+- Schools unaffected
+- Appropriate bodies unaffected
+- We can't send out any mass comms
+
+#### Actions
+
+- Wait for GOV.UK Notify to come back up
+- inform the support team that Notify's down and they won't be able to log into
+  RECT's admin area
+- if this happens on an important day like registration opening, we might need
+  developers to make the changes directly rather than using the admin UI
+- we can help admins log in by retreiving their OTP from ActiveJob directly, but
+  this is insecure and should be used as a last resort
+
+### TRS API goes down
+
+#### Symptoms
+
+- no teachers can be registered by schools
+- no teachers can be registered by appropriate bodies
+- Lead provider API unaffected
+- Admin interface unaffected
+- teacher syncing will stall, jobs will queue up
+
+#### Actions
+
+- we _probably_ want to temporarily close the app using [maintenance
+  mode](#maintenance-mode) because so much functionality in the app would break
+
+#### Contact
+
+- inform SITs and appropriate bodies they'll be unable to register new teachers
+- inform the support team that they might get an influx of tickets because the
+  app is down or journeys they were mid-way through were interrupted
+
+### Azure goes down
+
+#### Symptoms
+
+- whole app is down
+- databases are inaccessible
+- we're unable to retrieve list of SITs using our database or from DfE Sign-in
+
+#### Actions
+
+- wait for Azure to come back up
+
+#### Contact
+
+- let the support team know there's likely to be an influx of tickets
+
+### We accidentally lose some data
+
+#### Symptoms
+
+- entirely depends which table/tables are affected
+- some parts of the app might not work as expected
+- users might report some records are missing
+
+#### Actions
+
+- this qualifies as an incident, [start the incident process](#start-the-incident-process), the priority depends on
+  * the data that's gone
+  * how easy it is to replace
+  * how many users it will affect
+- [stop the service](#maintenance-mode) do a [full restore](#full-data-restore) if:
+  * important data is missing
+  * we notice it's missing quickly
+- [stop the service](#maintenance-mode) do a [partial restore](#partial-data-restore) if:
+  * important data is missing
+  * we take longer to realise it's missing and want to spend a bit more time
+    carefully inserting the missing records while more new data isn't being
+    added
+- leave the service running and do a [partial restore](#partial-data-restore) if:
+  * less-important data is missing
+  * the data can be added back without being affected by new data being added to
+    the service
+
+### We accidentally delete the database
+
+#### Symptoms
+
+- service is entirely broken
+
+#### Actions
+
+- this is a P1, [start the incident process](#start-the-incident-process)
+- [enable maintenance mode](#maintenance-mode)
+- do a [full data restore](#full-data-restore)
+
+#### Contact
+
+- schools
+- appropriate bodies
+- lead providers
+
+## Actions we can take
+
+### Full data restore
+
+If we need to restore the entire database we'll want to do it from a [point in
+time recovery](https://learn.microsoft.com/en-us/azure/postgresql/backup-restore/concepts-backup-restore#point-in-time-recovery) (PITR),
+choosing the latest time where we're sure the data is fully intact.
+
+1. follow the [PITR process](#point-in-time-restoring-a-database) which will create a new PostgreSQL Server
+   instance in Azure and make a note of the timestamp we want to use
+2. create a commit on a branch that [uncomments production](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/.github/workflows/restore-app-main-db.yml#L18) from the 'Restore database
+   from point in time to new database server' workflow
+3. using the new branch, run the `database-restore-ptr.yml` action against `production`
+   and use the timestamp we recorded in step 1
+4. when it's done, use the temporary maintenance URL to log
+   in and check the data's present
+
+### Partial data restore
+
+If we accidentally lose some data and want to restore it, there's a
+chance that more records have been written since the loss.
+
+This means this is a data insertion task rather than restoring a backup, so
+enabling the maintenance mode is optional here depending on the tables we're
+restoring.
+
+1. follow the [PITR process](#point-in-time-restoring-a-database) which will create a new PostgreSQL Server
+   instance in Azure
+2. we'll probably have to do some manual adjustments so copy what we need to a
+   local backup so we can work with it:
+   ```bash
+   bin/konduit.sh -n cpd-production -s s189p01-cpdec2-pd-pg-pitr -x cpd-ec2-production-web -- pg_dump -F t -E utf8 -f pitr-backup.sql.tar
+   ```
+3. restore it to a local database
+   ```bash
+   createdb pitr-restore
+   tar -x pitr-backup.sql.tar
+   psql pitr-restore < /tmp/pitr-backup.sql
+   ```
+4. work out what we need to restore and how best to do it, it'll probably
+   involve selecting some rows for re-insertion. This will be tricky if we're
+   merging the restored, especially if they span multiple tables.
+5. use the data to build a PR or ad hoc script to re-insert the data
+
+### Maintenance mode
+
+Enabling maintenance mode stops users from accessing the application, but
+leaves it running with an internal URL which will be printed by the [Set maintenance
+mode](https://github.com/DFE-Digital/register-early-career-teachers-public/actions/workflows/maintenance.yml) GitHub Action.
+
+If we want to customise the text on the page, edit the templates in the repo's
+`maintenance_page` directory.
+
+To enable maintenance mode:
+
+1. go to the [Set maintenance mode](https://github.com/DFE-Digital/register-early-career-teachers-public/actions/workflows/maintenance.yml) action in the public repo on GitHub
+2. click 'Run workflow'
+3. set the environment to 'Production' and leave the mode set to 'enable'
+4. run the action
+
+To disable it, follow the same steps as for enabling it but set the mode to
+'disable'.
+
+### Display a notification banner
+
+RECT has two types of banner, incident and maintenance. They both work in the
+same way and are collectively referred to as notification banners.
+
+1. create a new branch with an appropriate name
+2. review the content in the [notification banners partial](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/views/layouts/shared/_notification_banners.html.erb) and change
+   the text if necessary
+3. edit the [environment config](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/db5437cfd05f09d9e10edf8af4c911a852714d13/config/terraform/application/config/production.yml#L10-L11) and change `ENABLE_INCIDENT_BANNER` or
+   `ENABLE_MAINTENANCE_BANNER` to `true`
+4. commit your changes and create a pull request - once your pull request is
+   merged and deployed the notification banner will be enabled
+
+To remove the notification banner, create a new PR with the banner set back to
+`false`.
+
+Additionally, we can create a service banner on DfE Sign-in by logging into [the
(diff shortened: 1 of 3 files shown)


documentation/disaster-recovery.md
```diff
@@ -0,0 +1,265 @@
...
+If the tables are fundamental to the everyday running of the service, like
+`teachers`, then it'll be required. If they're not vital, like `events`
+
+1. follow the [PITR process](#point-in-time-restoring-a-database) which will create a new PostgreSQL Server
+   instance in Azure
+2. we'll probably have to do some manual adjustments so copy what we need to a
+   local backup so we can work with it:
+  ```bash
+   bin/konduit.sh -n cpd-production -s s189p01-cpdec2-pd-pg-pitr -x cpd-ec2-production-web -- pg_dump -F t -E utf8 -f pitr-backup.sql.tar
+    ```
```
peteryates: Restoring locally here isn't really best practice, but if something's on fire and we need to bring the service back to life ASAP by munging data into CSVs with scripts, it's a lot more straightforward than working via `konduit.sh`.

documentation/disaster-recovery.md
```diff
@@ -0,0 +1,266 @@
...
+#### Symptoms
+
+- School users can't log in
+- Appropriate body users can't log in
+- Lead provider API unaffected
+- Admin interface unaffected
+
+#### Actions
+
+- display a [notification banner](#displaying-a-notification-banner) on the site informing users that there
```
peterdavidhamilton: NB: 
We also have the option of a banner on the DfE Sign-In side but that might not be reliable or applicable when they go down. Not sure where it is meant to be used in prod.

documentation/disaster-recovery.md
```diff
@@ -0,0 +1,266 @@
...
+### We accidentally delete the database
+
+#### Symptoms
+
+- nobody can log in
+- service is entirely broken
+
+#### Actions
+
+- this is a P1, [start the incident process
```
avinhurry:
```suggestion
- this is a P1, [start the incident process](#start-the-incident-process)
```

documentation/disaster-recovery.md
```diff
@@ -0,0 +1,266 @@
...
+Less drastic, trickier. If we were to accidentally lose some data and wanted to
+restore it, there's a chance that more records have been written since the
+loss.
+
+This means this is a data insertion task rather than restoring a backup, so
+deploying the maintenance page is optional here depending on the tables we're
+restoring.
+
+If the tables are fundamental to the everyday running of the service, like
+`teachers`, then it'll be required. If they're not vital, like `events`
```
avinhurry: Is this sentence unfinished?

documentation/disaster-recovery.md
```diff
@@ -0,0 +1,266 @@
...
+Any branch can be manually deployed using [the manual deployment workflow](https://github.com/DFE-Digital/register-early-career-teachers-public/actions/workflows/deploy_to_environment.yml).
+
+We can also use the make commands to do it from the command line, providing a
+Docker image has been built and can be pulled from GitHub's container registry.
+
+```bash
+DOCKER_IMAGE=ghcr.io/dfe-digital/abc123 make production terraform-apply
+```
+
+### Contacting started mentoring with a new template and a list of SIT email addresses.
```
avinhurry: `### Contacting users` might be clearer as the heading?

documentation/disaster-recovery.md
```diff
@@ -0,0 +1,266 @@
...
+- I think in this scenario DfE Sign-in will have plans in place to let users
+  know the service is down, us sending extra comms might just confuse matters
+
+### GOV.UK Notify goes down
+
+#### Symptoms
+
+- Admins can't log in
+- Schools unaffected
+- Appropriate bodies unaffacted
```
avinhurry:
```suggestion
- Appropriate bodies  unaffected
```

documentation/disaster-recovery.md
```diff
@@ -0,0 +1,266 @@
...
+
+#### Symptoms
+
+- whole app is down
+- databases are inaccessible
+- we're unable to retrieve list of SITs using our database or from DfE Sign-in
+
+#### Actions
+
+- wait for for Azure to come back up
```
avinhurry:
```suggestion
- wait for Azure to come back up
```

documentation/disaster-recovery.md
```diff
@@ -0,0 +1,266 @@
...
+
+### Point-in-time-restoring a database
+
+A point-in-time-restore creates a copy of the database as it was at the
+specified time. We have a **1 week window**, if data was lost more than a week
+ago restore a database backup instead.
+
+1. find the database in Azure Portal
+2. click the 'Restore' tab at the top of the blade
+3. enter an name for the backup using the established convention with a suffix
```
avinhurry:
```suggestion
3. enter a name for the backup using the established convention with a suffix
```

documentation/disaster-recovery.md
```diff
@@ -0,0 +1,266 @@
...
+
+- no teachers can be registered by schools
+- no teachers can be registered by appropriate bodies
+- Lead provider API unaffected
+- Admin interface unaffected
+- teacher syncing will stall, jobs will queue up
+
+#### Actions
+
+- we _probably_ want to temporarily close the app using the maintenance,
```
avinhurry:
```suggestion
- we _probably_ want to temporarily close the app using maintenance mode,
```

documentation/disaster-recovery.md
```diff
@@ -0,0 +1,266 @@
...
+
+### Full data restore
+
+This is drastic. If we need to restore the entire database we'll want to do it
+from a PITR, choosing the latest time where we're sure the data is fully intact.
+
+1. [enable the maintenance page](#deploying-the-maintenance-page)
+2. follow the [PITR process](#point-in-time-restoring-a-database) which will create a new PostgreSQL Server
+   instance in Azure
+3. create a commit on a branch that [uncomments production](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/.github/workflows/restore-app-main-db.yml#L18) from the 'Restore database
```
avinhurry: I noticed we have two restore workflows: `database-restore-ptr.yml` for point in time restores to a new database server, and `restore-app-main-db.yml` for restoring the app main database from an Azure storage backup.

In the full restore section, we start with the PITR process, then link to `restore-app-main-db.yml`, but the next step says to run `database-restore-ptr.yml`. 

Could we clarify which workflow should be temporarily updated/uncommented for this process?

documentation/disaster-recovery.md
```diff
@@ -0,0 +1,266 @@
...
+   from point in time to new database server' workflow
+4. using the new branch, run the `database-restore-ptr.yml` and specify the name
```
avinhurry: Step 2 already creates the PITR server manually. This step then says to run `database-restore-ptr.yml` and specify that name, but the workflows `new-db-server` input is "Name of the new database server" and is passed to `ptr-postgres` as `new-server`.

Won't this try to create another PITR server with the same name rather than use the one created in step 2? 

Should we skip the manual Azure Portal PITR step and let this workflow create it?

documentation/disaster-recovery.md
```diff
@@ -0,0 +1,266 @@
...
+
+Can be enabled using the [Set maintenance mode](https://github.com/DFE-Digital/register-early-career-teachers-public/actions/workflows/maintenance.yml) GitHub Action.
+
+If we want to customise the text on the page, it's in the `maintenance_page`
+directory in the repo.
+
+### Maintenance banners/notices
+
+Enable them by setting the `ENABLE_MAINTENANCE_BANNER=true` in the
+`production.tfvars.json` file.
```
avinhurry: Should this be pointing at [production.yml ](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/045b202efc986dfb8468021bfb4989e7370d3982/config/terraform/application/config/production.yml)instead of `production.tfvars.json`?

documentation/site/content/index.html
```diff
@@ -49,6 +49,7 @@ <h2 class="govuk-heading-m">Product documentation</h2>
   <li><a class="govuk-link--no-visited-state" href="/product/roadmap/">Roadmap</a></li>
   <li><a class="govuk-link--no-visited-state" href="/product/data-model-progress/">Data model progress</a></li>
   <li><a class="govuk-link--no-visited-state" href="/product/api-versioning/">API Change Management Guidelines</a></li>
+  <li><a class="govuk-link--no-visited-state" href="/product/disaster-recovery/">Disaster recovery</a></li>
```
avinhurry, line 52: Are we comfortable having these docs public with this level of infrastructure details?

Publicly sharing commands like [this one](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/add-disaster-recovery-page/documentation/disaster-recovery.md?plain=1#L170) makes me nervous... It does not expose secrets by itself but it gives a concrete production database access pattern and makes reconnaissance easier. 

I'm not sure having that level of detail public gives us much value but keen to hear your thoughts?

documentation/disaster-recovery.md
```diff
@@ -0,0 +1,266 @@
...
+4. commit your changes and create a pull request - once your pull request is
+   merged and deployed the notification banner will be enabled
+
+To remove the notification banner, create a new PR with the banner set back to
+`false`.
+
+### Start the incident process
+
+1. [Teacher services incident playbook]: https://tech-docs.teacherservices.cloud/operating-a-service/incident-playbook.html#incident-playbook
+2. [Disasters within Teacher services]: https://github.com/DFE-Digital/teacher-services-cloud/blob/main/documentation/disaster-recovery.md
```
avinhurry: Assuming these are meant to be markdown links?

```suggestion
1. [Teacher services incident playbook](https://tech-docs.teacherservices.cloud/operating-a-service/incident-playbook.html#incident-playbook)
2. [Disasters within Teacher services](https://github.com/DFE-Digital/teacher-services-cloud/blob/main/documentation/disaster-recovery.md)
```

documentation/disaster-recovery.md
```diff
@@ -0,0 +1,266 @@
...
+
+### Displaying a notification banner
+
+RECT has two flavours of banner, incident and maintenance. They both work in the
+same way and are collectively referred to as notification banners.
+
+1. create a new branch with an appropriate name
+2. review the content in the [notification banners partial](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/views/layouts/shared/_notification_banners.html.erb) and change
+   the text if necessary
+3. edit the [environment config](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/db5437cfd05f09d9e10edf8af4c911a852714d13/config/terraform/application/config/production.yml#L10-L11) and change `ENABLE_INCIDENT_BANNER` or
```
avinhurry: Just a thought... Do you think linking line numbers is a good idea? They could drift but at least it points people in the general direction I guess.

documentation/disaster-recovery.md
```diff
@@ -0,0 +1,266 @@
...
+#### Symptoms
+
+- nobody can log in
+- service is entirely broken
+
+#### Actions
+
+- this is a P1, [start the incident process
+- [deploy the maintenance page](#maintenance-pages)
+- do a [full database restore](#full-database-restore)
```
avinhurry: The link doesn't match the heading

```suggestion
- do a [full data restore](#full-data-restore)
```

documentation/disaster-recovery.md
avinhurry: Will this work? `pg_dump -F t` would produce a tar format file  intended for  `pg_restore`, but the restore step uses `psql < /tmp/pitr-backup.sql` which expects a plain SQL dump.

Image

https://www.postgresql.org/docs/current/app-pgdump.html

documentation/disaster-recovery.md
```diff
@@ -0,0 +1,266 @@
...
+- Appropriate bodies unaffacted
+- We can't send out any mass comms
+
+#### Actions
+
+- Wait for GOV.UK Notify to come back up
+- inform the support team that Notify's down and they won't be able to log into
+  register early career teachers' admin area
+- if this happens on a big day, like registration opening, we might need
+  developers to make things happen without using the admin UI
```
avinhurry: It's certainly not secure at all but we should be able to log people in by sending them the OTP from active job in a pinch, do you think it could be worth mentioning that?

documentation/disaster-recovery.md
```diff
@@ -0,0 +1,266 @@
...
+
+### Azure goes down
+
+#### Symptoms
+
+- whole app is down
+- databases are inaccessible
+- we're unable to retrieve list of SITs using our database or from DfE Sign-in
+
+#### Actions
```
avinhurry, line 97: If Azure is our biggest single point of failure, do you think we should consider also copying the nightly SQL backup to an off-azure encrypted location? 

That would give us something we can restore locally or into another postgres instance if Azure storage/AKS/postgres are unavailable.

I know we won't be able to run the site but we could at least deal with urgent support queries.

I'm not sure if DfE has a process for this 🤔

avinhurry (commented):
I've read through this. It's definitely worth having, especially for when we're under pressure during an incident.

I've left a few thoughts and observations. I realised I got a bit carried away, so feel free to take what's useful from them and ignore the rest.

documentation/disaster-recovery.md
peteryates: Yeah - it's the line number at that revision, so it shouldn't drift - but it could fall out of date.

documentation/disaster-recovery.md
peteryates: Not even sure what I was doing here 😅

documentation/disaster-recovery.md
peteryates: Ah yeah you're right. This was from my notes from when we lost the events and I manually untarred it with `tar -xvf pitr-backup.sql.tar`. Will fix.

peteryates: Oh  step 3 already covers extracting from the tar with `tar -x pitr-backup.sql.tar`.

documentation/disaster-recovery.md
peteryates: These were originally [footnotes](https://www.markdownguide.org/extended-syntax/#footnotes) that I lifted from another DR doc!

documentation/disaster-recovery.md
peteryates: I think we could do with renaming these jobs and adding descriptions.

In this case we probably wouldn't want to be restoring a PITR straight to prod because usually we'd want to ensure it's the latest version that has its data intact. If the database was accidentally dropped, then it's not so important - but if we dropped `teachers` we wouldn't want to restore too late a version as that would be missing `teachers` too.

I'll change it so we restore to a new server to check the data's all present and correct at the time of the PTIR, and when we're happy, use `database-restore-ptr.yml` to rebuild prod at the timestamp we selected.

documentation/disaster-recovery.md
peteryates: I think this is more a infra thing than an us thing. If Azure goes down we're really limited in what we can do, and so much of DfE will be down too it's a problem at a level above us.

I think under those circumstances we just have to wait for it to come back.

documentation/site/content/index.html
peteryates: All of the constitutent parts are already publicly available. [Konduit](https://github.com/DFE-Digital/teacher-services-cloud/blob/main/scripts/konduit.sh), [cpd-production](https://github.com/DFE-Digital/teacher-services-cloud/blob/ec3603677f0771e42f3fae1295aa4b0faf12f3bf/cluster/terraform_kubernetes/config/production.tfvars.json#L20), [cpd-ec2-production-web](https://github.com/DFE-Digital/teacher-services-cloud/blob/ec3603677f0771e42f3fae1295aa4b0faf12f3bf/cluster/terraform_kubernetes/config/production.tfvars.json#L194).

It's a good shout, but every bit of our infra is open already, so I think we're fine here.

documentation/disaster-recovery.md
peteryates: I think the markdown code blocks are broken in the docs site which is why the `tar` line immediately followed the one before. I'll fix that separately.

documentation/disaster-recovery.md
peteryates: Yep, good shout

documentation/disaster-recovery.md
```diff
@@ -0,0 +1,309 @@
...
+- [enable maintenance mode](#maintenance-mode)
+- do a [full data restore](#full-data-restore)
+
+#### Contact
+
+- schools
+- appropriate bodies
+- lead providers
+
+## Actions we can take
```
peterdavidhamilton, line 150: Is it worth mentioning what data losses could warrant a full vs partial restore here so that when the pressure is on the call can be made more easily? 

I know every situation will be different, but broadly can we say things like, for example: 
- backfilling the events history is low risk and it is best to keep the service running
- financial data can be restored adhoc unless there is an important deadline imminent
- a schools table loss is best just doing a full restore

peteryates: I added an extra paragraph to the top of the full restore section explaining it's a big deal.

documentation/disaster-recovery.md
```diff
@@ -0,0 +1,314 @@
...
+
+If we need to restore the entire database we'll want to do it from a [point in
+time recovery](https://learn.microsoft.com/en-us/azure/postgresql/backup-restore/concepts-backup-restore#point-in-time-recovery) (PITR),
+choosing the latest time where we're sure the data is fully intact.
+
+1. follow the [PITR process](#point-in-time-restoring-a-database) which will create a new PostgreSQL Server
+   instance in Azure and make a note of the timestamp we want to use
+2. create a commit on a branch that [uncomments production](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/.github/workflows/restore-app-main-db.yml#L18) from the 'Restore database
+   from point in time to new database server' workflow
+3. using the new branch, run the `database-restore-ptr.yml` action against `production`
```
avinhurry, line 167: In step 2 we uncomment prod in `restore-app-main-db.yml#L18` but step 3 says to runs the workflow in `database-restore-ptr.yml`.

Am I missing something or should prod be enabled in `database-restore-ptr.yml` instead?

documentation/disaster-recovery.md
avinhurry: I think the issue is slightly different, it isn't whether the tar gets extracted, it's the step 2 uses `pg_dump -F t ` and step 3 restores with `psql` which expects a plain SQL dump.

Should this restore path be using `pg_restore` instead?

documentation/disaster-recovery.md
avinhurry: Yes, that makes sense. We can make the workflow/doc changes to support that two step flow in a follow up.

avinhurry (approved):
This looks good for a first draft and will be nice to have!

It might be worth adding a short note to make it clear the doc is still WIP/draft, but up to you. Just a thought I had 💭

PR #2963, opened 2026-05-22
#2959 - validate ongoing training periods are fully covered by its ASP

ltello:
### Context
Issue [#2959](https://github.com/DFE-Digital/register-early-career-teachers-public/issues/2959).

`TrainingPeriod#enveloped_by_trainee_at_school_period` don't validate for ongoing training periods, so if its parent `ECTAtSchoolPeriod` / `MentorAtSchoolPeriod` was already finished (or started later) it is accepted as valid even though it logically extended past the parent. 

There is a mirror rule on the parent (`AtSchoolPeriod#covering_training_periods`) that already validates this.
This ticket is to close the gap for the child period therefore making the validation symetric and preventing potential exceptions thrown in production.

  - remove the skip on TrainingPeriod
  - adjust specs to create ongoing parents when the child is ongoing.

### DIFF
--- a/spec/factories/training_period_factory.rb
+++ b/spec/factories/training_period_factory.rb
@@ -119,6 +119,21 @@ FactoryBot.define do
 
     trait :ongoing do
       finished_on { nil }
+
+      # An ongoing training period is only valid when its parent at-school period
+      # is also ongoing (otherwise the TP extends past the parent). If the parent
+      # was implicitly created with a finished_on date, make it ongoing so the
+      # envelope validation stays satisfied.
+      after(:build) do |tp|
+        asp = tp.at_school_period
+        next if asp.blank? || asp.finished_on.blank?
+
+        if asp.persisted?
+          asp.update_columns(finished_on: nil)
+        else
+          asp.finished_on = nil
+        end
+      end
     end
 
     trait :for_ect do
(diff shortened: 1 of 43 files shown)


spec/factories/training_period_factory.rb
```diff
@@ -119,6 +119,21 @@
...
+      # An ongoing training period is only valid when its parent at-school period
+      # is also ongoing (otherwise the TP extends past the parent). If the parent
+      # was implicitly created with a finished_on date, make it ongoing so the
+      # envelope validation stays satisfied.
+      after(:build) do |tp|
+        asp = tp.at_school_period
+        next if asp.blank? || asp.finished_on.blank?
+
+        if asp.persisted?
+          asp.update_columns(finished_on: nil)
```
avinhurry: This feels a bit unexpected to me. Could this silently turn finished `at_school_periods` into ongoing ones?

Since this uses `update_columns`, it also bypasses validations/callbacks and may hide the invalid state this PR is trying to catch.

Should specs create an ongoing parent explicitly instead?

avinhurry (approved):

spec/factories/training_period_factory.rb
ltello: I agree with you. I don't like it much either but the alternative is to modify loads and loads of specs.

avinhurry: Maybe one to card up for the backlog

ltello: I have reverted this change and fixed the affected specs

avinhurry (approved):
Thanks for going through the pain to fix those tests, i'm sure it will help future us 💯

PR #2970, opened 2026-05-26
4005 induction tutor email get partnership rel note

StephenC-V1:
### Context
Provide release note content to support changes for induction tutor email

Add new section to release note

### Guidance to review
Visual in review app

### DIFF
--- a/app/views/api/release_notes/release_notes.yml
+++ b/app/views/api/release_notes/release_notes.yml
@@ -21,6 +21,36 @@
 #
 #     **bold** and _italic_ text
 
+- title: Updated the induction_tutor_email attribute on the GET partnerships and GET schools endpoints
+  date: 2026-05-26
+  tags:
+    - sandbox-release
+    - production-release
+  body: |
+    We’ve updated the `induction_tutor_email` attribute on the GET partnerships endpoint and GET schools endpoint so that it is populated more consistently.
+
+    ## What’s changed
+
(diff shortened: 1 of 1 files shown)


app/views/api/release_notes/release_notes.yml
```diff
@@ -21,6 +21,38 @@
...
+    - sandbox-release
+    - production-release
+  body: |
+    We’ve updated the `induction_tutor_email` attribute on the GET partnerships endpoint and GET schools endpoint so that it’s populated more consistently.
+
+    ## What’s changed
+
+    Where no induction tutor email address is available from register early career teachers (RECT) records, `induction_tutor_email` will now be populated with the school’s main email address held in GIAS.
+    
+    Previously, this field could be empty if no induction tutor was recorded.
```
easeynathan, line 36: Ross just explained to me that for GET schools responses, the field will continue to be NULL if the school has not expressed an interest of wanting to use the training provider making the call to the endpoint.

Would we benefit from a line/section saying what hasn't changed?

StephenC-V1: Yes, maybe that should be included.

To clarify, null will be returned if a provider makes a call where the school does not have an EOI for the provider making that call?  

'Note that the field will be null if a call is made to the GET schools endpoint by a provider who has not had an expression of interest from the school.'

easeynathan: @ethax-ross @andycrawshaw could one of you please confirm the GET schools response where expression_of_interest = null? Is induction_tutor_email simply null as well?

PR #2975, opened 2026-05-26
Reverse fee structure foreign keys part 2

tonyheadford:
### Context

After the database changes and reassignment of the fee structure associations [in part one](#2957) have been actioned, part 2 deals with removing the old unneeded foreign keys from `Contract` and updating the models, services and specs to support the new reversed foreign keys.

There's quite a bit here that I think needs a careful look at to ensure this hasn't broken any behaviour that wasn't represented in specs and that it works as expected.

- reversed `has_one` and `belongs_to` relationships added to `Contract`, `Contract::BandedFeeStructure` and `Contract::FlatRateFeeStructure`
- model validation changes
- Updates to specs to support the changes (mainly creation order)
- DB migration to remove old foreign keys

### Guidance to review

This is the second part branched from #2957 so perhaps start there

There will be a third and final PR to remove the now-redundant foreign keys on `Contract`.

### DIFF
--- a/app/models/contract.rb
+++ b/app/models/contract.rb
@@ -47,7 +47,7 @@ class Contract < ApplicationRecord
       errors.add(:banded_fee_structure, "Banded fee structure must be provided for ITTECF_ECTP contracts") if banded_fee_structure.blank?
     elsif ecf_contract_type?
       errors.add(:flat_rate_fee_structure, "Flat rate fee structure must be blank for ECF contracts") if flat_rate_fee_structure.present?
-      errors.add(:banded_fee_structure, "Banded fee structure must be provided for ECF contracts") if banded_fee_structute.blank?
+      errors.add(:banded_fee_structure, "Banded fee structure must be provided for ECF contracts") if banded_fee_structure.blank?
     end
   end
 end
--- a/app/models/contract.rb
+++ b/app/models/contract.rb
@@ -24,14 +24,16 @@ class Contract < ApplicationRecord
             presence: { message: "VAT rate is required" },
             numericality: { in: 0..1, message: "VAT rate must be between 0 and 1" }
 
-  validate :fee_structures_are_correct_for_contract_type
-
   with_options if: :ittecf_ectp_contract_type? do
+    validates :flat_rate_fee_structure, presence: { message: "Flat rate fee structure must be provided for ITTECF_ECTP contracts" }
+    validates :banded_fee_structure, presence: { message: "Banded fee structure must be provided for ITTECF_ECTP contracts" }
     validates :ecf_contract_version, presence: { message: "ECF contract version must be provided for ITTECF_ECTP contracts" }
     validates :ecf_mentor_contract_version, presence: { message: "ECF mentor contract version must be provided for ITTECF_ECTP contracts" }
   end
 
   with_options if: :ecf_contract_type? do
+    validates :flat_rate_fee_structure, absence: { message: "Flat rate fee structure must be blank for ECF contracts" }
+    validates :banded_fee_structure, presence: { message: "Banded fee structure must be provided for ECF contracts" }
     validates :ecf_contract_version, presence: { message: "ECF contract version must be provided for ECF contracts" }
   end
 
@@ -39,15 +41,5 @@ class Contract < ApplicationRecord
     return 0 unless lead_provider.vat_registered
 
     vat_rate
-  end
-
-  def fee_structures_are_correct_for_contract_type
-    if ittecf_ectp_contract_type?
-      errors.add(:flat_rate_fee_structure, "Flat rate fee structure must be provided for ITTECF_ECTP contracts") if flat_rate_fee_structure.blank?
-      errors.add(:banded_fee_structure, "Banded fee structure must be provided for ITTECF_ECTP contracts") if banded_fee_structure.blank?
-    elsif ecf_contract_type?
-      errors.add(:flat_rate_fee_structure, "Flat rate fee structure must be blank for ECF contracts") if flat_rate_fee_structure.present?
-      errors.add(:banded_fee_structure, "Banded fee structure must be provided for ECF contracts") if banded_fee_structure.blank?
-    end
   end
 end
--- a/app/models/contract.rb
+++ b/app/models/contract.rb
@@ -11,8 +11,8 @@ class Contract < ApplicationRecord
   belongs_to :active_lead_provider
   has_one :contract_period, through: :active_lead_provider
   has_one :lead_provider, through: :active_lead_provider
-  has_one :flat_rate_fee_structure, class_name: "Contract::FlatRateFeeStructure", inverse_of: :contract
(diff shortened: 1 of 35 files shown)


app/models/contract/banded_fee_structure.rb
```diff
@@ -2,14 +2,16 @@ class Contract::BandedFeeStructure < ApplicationRecord
...
+  belongs_to :contract
   has_many :bands,
            -> { order(min_declarations: :asc) },
            class_name: "Contract::BandedFeeStructure::Band",
            inverse_of: :banded_fee_structure,
            dependent: :destroy
-  has_one :contract, inverse_of: :banded_fee_structure
 
   # Validations
+  validates :contract_id, uniqueness: { message: "Contract with the same banded fee structure already exist" }, allow_nil: true
```
ethax-ross: I don't think we want to `allow_nil` on the `contract` associations? Sorry I missed that it was added on to the earlier PR: https://github.com/DFE-Digital/register-early-career-teachers-public/pull/2957/changes/cae1ffbac1e2156140937bee437e3e3635ae1a92

app/models/contract.rb
```diff
@@ -24,23 +24,14 @@ class Contract < ApplicationRecord
             presence: { message: "VAT rate is required" },
             numericality: { in: 0..1, message: "VAT rate must be between 0 and 1" }
 
+  validate :fee_structures_are_correct_for_contract_type
+
   with_options if: :ittecf_ectp_contract_type? do
-    validates :flat_rate_fee_structure,
```
ethax-ross, line 28: Do these need to change? it might just be my preference but I find them easier to read within the `with_options` 🤔

spec/models/contract/flat_rate_fee_structure_spec.rb
```diff
@@ -10,5 +10,20 @@
     it { is_expected.to validate_presence_of(:fee_per_declaration).with_message("Fee per declaration is required") }
     it { is_expected.to validate_numericality_of(:recruitment_target).only_integer.is_greater_than(0).with_message("Value must be greater than 0") }
     it { is_expected.to validate_numericality_of(:fee_per_declaration).is_greater_than(0).with_message("Amount must be greater than 0") }
+
+    context "uniqueness" do
```
ethax-ross: I think we can use shoulda here?

```
it { is_expected.to validate_uniqueness_of(:contract_id).with_message("...") }
```

spec/models/contract/banded_fee_structure_spec.rb
```diff
@@ -1,7 +1,7 @@
 RSpec.describe Contract::BandedFeeStructure, type: :model do
   describe "associations" do
     it { is_expected.to have_many(:bands).order(min_declarations: :asc).class_name("Contract::BandedFeeStructure::Band").inverse_of(:banded_fee_structure).dependent(:destroy) }
-    it { is_expected.to have_one(:contract).inverse_of(:banded_fee_structure) }
+    it { is_expected.to belong_to(:contract).optional }
```
ethax-ross: Can we add the new uniqueness validation here as well?

app/services/active_lead_providers/seed_from_previous.rb
```diff
@@ -77,12 +77,16 @@ def create_new_contract
     # would run those validations before the siblings exist.
     previous_contract = previous_latest_contract
 
+    # fee structures now belong_to a contract and can be added first
+    banded_fee_structure = create_banded_fee_structure(previous_contract.banded_fee_structure)
```
ethax-ross: I think we could switch these to a `build` instead of `create` and the outer `create!` on the contract will cascade

spec/factories/contract_factory.rb
```diff
@@ -2,13 +2,10 @@
   factory(:contract) do
     association :active_lead_provider
 
-    for_ittecf_ectp
```
ethax-ross, line 5: Again just my preference, but base factories should create something valid IMO - I don't mind if we want to be explicit with the traits when calling, but I think it should default to one or the other contract types

spec/models/contract/banded_fee_structure/band_spec.rb
```diff
@@ -151,32 +151,32 @@
       let(:active_lead_provider) { FactoryBot.create(:active_lead_provider) }
 
       it "is valid when creating the first band for the active lead provider" do
-        contract = FactoryBot.create(:contract, :for_ecf, active_lead_provider:)
-        banded_fee_structure = FactoryBot.create(:contract_banded_fee_structure, contract:)
+        banded_fee_structure = FactoryBot.create(:contract_banded_fee_structure)
+        FactoryBot.create(:contract, :for_ecf, active_lead_provider:, banded_fee_structure:)
```
ethax-ross: I think I see why you made `contract_id` optional :) instead of creating the `banded_fee_structure` first with a `nil` `contract_id` we could build it and the create on `Contract` will persist both together, satisfying the `Contract` validation and `contract_id` foreign key, something like:

```
banded_fee_structure = build(:contract_banded_fee_structure)
contract = create(:contract, :for_ecf, active_lead_provider:, banded_fee_structure:)
```

ethax-ross (commented):
Thanks Tony, left a few thoughts

spec/factories/contract_factory.rb
tonyheadford: I'll have another look but I was tripping up in one spec where it was the ecf type and trying to add the fee structure to the created object, it all feels a little bit the wrong way around.

spec/models/contract/banded_fee_structure/band_spec.rb
tonyheadford: Yes, you can't create a contract without a fee structure and you need the contract id in the fee structure to be present if you want to enforce the validation

app/services/active_lead_providers/seed_from_previous.rb
tonyheadford: Thanks will have a look at that

app/models/contract/banded_fee_structure.rb
tonyheadford: That might've been added after you had a look to be honest as I was hitting issues in the order of creation.

app/models/contract.rb
peterdavidhamilton: I'd concur; also likely more maintainable.

spec/factories/contract_factory.rb
peterdavidhamilton: I'm happy if the base factory is bare; because traits cater for both flavours the specs end up being more explicit which I like.

peterdavidhamilton (approved):
[Does this need including in the PR?](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/services/active_lead_providers/cascade_delete.rb#L56-L67)

Otherwise LGTM, and overall feels a little simpler but `band_consistency_across_active_lead_provider` still scares me

PR #2976, opened 2026-05-26
[3923] Create Rake task to populate events from ECF1 data

thenapking:
### Context

In the admin console, for each declaration we show a state history which relies on events.  Unfortunately all the historical ECF1 declarations did not have events created for them, so we need to backfill the data.

My initial idea was to use the built in methods that record events.  But since each of these triggers a background job that just runs one create statement this could easily overload the job processing processes and be difficult to clear down.
Therefore this task runs in batches, locking each statement and its declaration and creating the events directly.

### Guidance to review
Product Review testing for 2024:
image

For 2025
image

In order to review this, I have run the task on all the statements.  We should expect only the Aug/Oct statements from 2024 and 2025 to have declarations processed.  There may be too many declarations to check individually, but we can spot check.  The task itself outputs the total number of declarations of each type per statement processed which should tally against the statement view.

### DIFF
--- a/lib/backfill/record_statement_payment_events.rb
+++ b/lib/backfill/record_statement_payment_events.rb
@@ -15,25 +15,30 @@ module Backfill
     end
 
     def process
-      counts = DECLARATION_TYPES.index_with { 0 }
+      processed = DECLARATION_TYPES.index_with { 0 }
+      updated = DECLARATION_TYPES.index_with { 0 }
 
       ActiveRecord::Base.transaction do
         paid_declarations.find_each do |declaration|
+          processed[declaration_type(declaration)] += 1
+
           next unless declaration_event_created?(declaration, statement, :paid)
 
-          counts[declaration_type(declaration)] += 1
+          updated[declaration_type(declaration)] += 1
         end
 
         clawed_back_declarations.find_each do |declaration|
+          processed[:clawed_back] += 1
+
           next unless declaration_event_created?(declaration, declaration.clawback_statement, :clawed_back)
 
-          counts[:clawed_back] += 1
+          updated[:clawed_back] += 1
         end
 
-        record_statement_authorized_for_payment_event!
+        record_statement_authorised_for_payment_event!
       end
 
-      counts
+      format_output(processed, updated)
     end
 
   private
@@ -53,8 +58,8 @@ module Backfill
       ).process
     end
 
-    def record_statement_authorized_for_payment_event!
-      if statement_authorized_event_exists?
+    def record_statement_authorised_for_payment_event!
+      if statement_authorised_event_exists?
         puts("Statement: #{statement.id} Authorised for payment event already exists")
         return
       end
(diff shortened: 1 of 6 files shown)


lib/backfill/record_statement_payment_events.rb
```diff
@@ -0,0 +1,97 @@
...
+        clawed_back_declarations.find_each do |declaration|
+          next unless declaration_event_created?(declaration, declaration.clawback_statement, :clawed_back)
+
+          counts[:clawed_back] += 1
+        end
+
+        record_statement_authorized_for_payment_event!
+      end
+
+      counts
```
ericaporter: Minor thing: since the count only increments when an event is created, would a rerun show lower/zero counts for declarations whose events already exist?

That seems right for avoiding duplicate events, but might be confusing if we expect the output to tally with the statement view. Maybe they could be labelled separately as created / skipped or similar

lib/backfill/record_statement_payment_events.rb
```diff
@@ -0,0 +1,97 @@
...
+          contract_period_year: active_lead_provider.contract_period_year
+        },
+        author_name: "System",
+        author_type: "system"
+      )
+
+      puts("Statement: #{statement.id} Authorised for payment event created")
+    end
+
+    def statement_authorized_event_exists?
```
ericaporter: Really minor - should we use the UK spelling for authorized or the US? We have the UK spelling for the event_type below

spec/lib/backfill/record_declaration_event_spec.rb
```diff
@@ -0,0 +1,108 @@
...
+      it "does not create the event" do
+        expect { process }
+          .not_to change(Event, :count)
+      end
+
+      it { is_expected.to be_falsy }
+    end
+
+    context "when the declaration is a clawback declaration" do
+      let(:status) { :clawed_back }
```
ericaporter, line 84: You could set up a clawback declaration here just to be super clear, but not a blocker

spec/lib/backfill/record_statement_payment_events_spec.rb
```diff
@@ -0,0 +1,159 @@
...
+          event_type: "statement_authorised_for_payment",
+          statement:
+        )
+
+        expect { subject }
+          .not_to change(
+            Event.where(event_type: "statement_authorised_for_payment"),
+            :count
+          )
+      end
```
ericaporter, line 210: All the important scenarios are covered - you could add one for where a declaration event already exists to prove it is skipped and not included in the returned counts - but not essential

ericaporter (approved):
Looks good!! 👍

lib/backfill/record_statement_payment_events.rb
thenapking: Oh dear.  My spelling appears to be really terrible.  I hadn't noticed this and will switch it to the UK spelling!

PR #2977, opened 2026-05-26
[3546] Fix ECT email field screen reader label

avinhurry:
## Fix ECT email field accessibility label

Separate the page heading and guidance copy from the email field label in the ECT registration flow.

Previously, the email field reused the full page question as its label and rendered guidance copy inside the field block. This meant screen readers could announce duplicated and unnecessary content when focusing on the input.

**This change:**
- adds an explicit `<h1>` for the page heading
- moves the guidance content outside the field block
- changes the input label to `Email address`
- hides the label visually while keeping it available to screen readers

This matches the mentor registration flow and improves accessibility for screen reader users.


thenapking (approved):
LGTM

peteryates (approved):
Looks good.

I probably wouldn't hide the label though.

JSJohal (approved):

PR #2981, opened 2026-05-27
Add registration window closed banner to schools area

ericaporter:
### Context
There will be no open contract period between 1st-14th June 2026 (inclusive dates). This PR is to add a banner to the service between these dates to inform users.

- Adds a "Registration not currently open" banner to _notification_banners visible to all authenticated users between 1–14 June 2026, reusing Schools::RegistrationWindow.closed?

### Guidance to review
Use the Time Travel functonality to travel to a date within 1–14 June to verify the banner appears within schools area only, and to 15 June to verify it disappears.


app/views/layouts/shared/_notification_banners.html.erb
```diff
@@ -11,3 +11,10 @@
     This service will be unavailable from midday on 27 April until 9am on 28 April.
   <% end %>
 <% end %>
+
+<% if Schools::RegistrationWindow.closed? %>
+  <%= govuk_notification_banner(title_text: "Important") do %>
+    <h3 class="govuk-notification-banner__heading">Registration not currently open</h3>
```
peteryates: The banner component lets you set headings directly:

```suggestion
  <%= govuk_notification_banner(title_text: "Important") do |notification_banner| %>
    <% notification_banner.with_heading("Registration not currently open", tag: "h3") %>
```

You can do the similar with `#with_body` too.

ericaporter: It worked for `with_heading` but got ActionView::Template::Error: undefined method error for 'with_body' 😢

spec/requests/schools/registration_window_banner_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    it "displays on schools pages" do
+      get "/school/home/ects"
+      expect(response.body).to include("Registration not currently open")
+    end
+
+    it "does not display on non-schools pages" do
+      get "/support"
+      expect(response.body).not_to include("Registration not currently open")
+    end
```
cpjmcquillan: I'm not sure this is really testing it doesn't display on non-schools pages, more that it isn't displayed on the support page specifically.

As far as I can tell, this would be displayed on any page using the "full" layout —which is quite a few (e.g Admin pages etc).

If we're happy with that, this looks good to me!

cpjmcquillan (approved):

app/views/layouts/shared/_notification_banners.html.erb
peteryates: Ah my bad, any content in the block automatically ends up in the body, `#with_heading` just saves you from having to add the header element manually.

spec/requests/schools/registration_window_banner_spec.rb
peteryates: How we use layouts is probably worth a wider discussion.

Now there's more variation, it might make sense for us to have separate layouts for `appropriate_bodies`, `schools` and `admin` - rather than `application` (2 columns in 1:2) and `full` (1 column).

Originally I thought we'd be using the left column for secondary nav so I made that the default.

ericaporter: Good point about the full layout - I will scope this to only render when the current user is a school user - Emily confirmed it should be school facing only

app/views/layouts/shared/_schools_registration_window_banner.html.erb
```diff
@@ -0,0 +1,9 @@
+<% if current_user&.school_user? && Schools::RegistrationWindow.closed? %>
+  <%= govuk_notification_banner(title_text: "Important") do |notification_banner| %>
+    <% notification_banner.with_heading(text: "Registration not currently open", tag: "h3") %>
+
+    <p class="govuk-body">
+      You cannot register ECTs or mentors for the 2025 or 2026 academic years until we open registration on 15 June.
```
thenapking, line 6: This is very minor but I notice some of the other error messages use `reopens_on` to provide this date rather than hardcoding it.

thenapking (approved):
LGTM.  I've tested this locally, without time travelling, by changing the registration window and this all works as intended.

app/views/layouts/shared/_schools_registration_window_banner.html.erb
ericaporter: Emily had already reviewed it by the time I saw this and wants to get in it before Monday - if I get another chance I'll update it!

PR #2984, opened 2026-05-28
Add stylesheet to aid accessibility testing

peteryates:
This puts a dashed outline around each header and a label that tells us the heading level. It's intended to make ensuring page heading hierarchies flow properly easier.

It can be enabled by setting the environment variable `ENABLE_DEBUG_CSS=true` but is disabled by default.

Screenshot From 2026-05-28 12-25-46


cpjmcquillan (approved):
Love it! Is it worth going to down to `<h5>`, and `<h6>` even if we're not using them?

avinhurry (approved):
Love this! It makes it almost impossible not to spot issues.

Image

I have a feeling it might be @JSJohal's new favourite toy...

PR #2987, opened 2026-05-29
Make fee structures non-nullable

ethax-ross:
Builds upon https://github.com/DFE-Digital/register-early-career-teachers-public/pull/2975

### Context

We want fee structures to always be associated with a `Contract`, however the `Contract` must know about it's fee structures at the time of creation to satisfy the `contract_type`/fee structure validation. This means we need to persist contracts with their fee structures.

Remove redundant `reload` from band consistency check.

Switch contract association strategy to `build` so we can save together and avoid validation issues.

Ensure bands are created consistently in factories.

Drop migration to remove old foreign keys (will re-add in follow up PR to avoid errors due to inconsistent app/DB state on deploy).

Switch validation rules back to `with_options`.

### DIFF
--- a/app/models/contract.rb
+++ b/app/models/contract.rb
@@ -24,14 +24,16 @@ class Contract < ApplicationRecord
             presence: { message: "VAT rate is required" },
             numericality: { in: 0..1, message: "VAT rate must be between 0 and 1" }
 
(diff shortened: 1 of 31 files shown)


peteryates:
I like this, `not null` feels much safer.

PR #2989, opened 2026-05-29
[3397] button to authorise statement

thenapking:
### Context

Now we are live finance users need to be able to authorise statements.  The layout etc is largely unchanged.

Update the background service to transition associated declarations to paid/clawed_back.  Refactor so that the state machine will not raise unexpected errors.
The ticket notes that it is not worth using a background job at this stage and so that is the route I've followed.

### Guidance to review
This can be tested on Aug/Oct 24 and 25 statements.

### DIFF
--- a/app/models/statement.rb
+++ b/app/models/statement.rb
@@ -74,7 +74,11 @@ class Statement < ApplicationRecord
   end
 
   def can_authorise_payment?
-    output_fee? && payable? && !marked_as_paid_at? && has_outstanding_declarations? && within_deadline?
+    output_fee? &&
+      payable? &&
+      !marked_as_paid_at? &&
+      has_outstanding_declarations? &&
+      deadline_date.before?(Date.current)
   end
 
   def referenced_by_declarations?
@@ -99,13 +103,9 @@ private
 
   def deadline_date_in_the_past
     return unless payable? || paid?
-    return if within_deadline?
+    return if deadline_date.before?(Date.current)
 
     errors.add(:deadline_date, "Deadline date must be in the past")
-  end
-
-  def within_deadline?
-    deadline_date < Date.current
   end
 
   def has_outstanding_declarations?
--- a/app/models/statement.rb
+++ b/app/models/statement.rb
@@ -74,8 +74,7 @@ class Statement < ApplicationRecord
   end
 
   def can_authorise_payment?
-    # TODO: will also need to include: `participant_declarations.any?`
-    output_fee? && payable? && !marked_as_paid_at? && deadline_date < Date.current
+    output_fee? && payable? && !marked_as_paid_at? && has_outstanding_declarations? && within_deadline?
   end
 
   def referenced_by_declarations?
@@ -100,8 +99,16 @@ private
 
   def deadline_date_in_the_past
     return unless payable? || paid?
-    return if deadline_date < Date.current
+    return if within_deadline?
 
     errors.add(:deadline_date, "Deadline date must be in the past")
+  end
+
+  def within_deadline?
+    deadline_date < Date.current
+  end
+
+  def has_outstanding_declarations?
+    payment_declarations.payment_status_payable.exists? || clawback_declarations.clawback_status_awaiting_clawback.exists?
   end
 end
--- a/app/services/statements/authorise_payment.rb
+++ b/app/services/statements/authorise_payment.rb
@@ -4,30 +4,67 @@ module Statements
 
     attr_reader :statement, :author
 
-    def initialize(statement, author:)
+    def initialize(statement:, author:)
       @statement = statement
       @author = author
     end
 
     def authorise!
+      raise NotAuthorisable unless statement.can_authorise_payment?
+
       ActiveRecord::Base.transaction do
-        raise NotAuthorisable unless statement.can_authorise_payment?
-
-        if statement.marked_as_paid_at.blank?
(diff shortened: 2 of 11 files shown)


app/models/statement.rb
```diff
@@ -96,8 +95,16 @@ def unique_lead_provider_month_year
 
   def deadline_date_in_the_past
     return unless payable? || paid?
-    return if deadline_date < Date.current
+    return if within_deadline?
```
ericaporter, line 107: Small thing - within_deadline? is defined as deadline_date < Date.current which is more like the outside the deadline. The logic is correct though. I suppose it could be something like `deadline_passed?` - depends on what fits in with the language used around statements, not a blocker

Also wondering - will we ever want to authorise payments on the deadline day where deadline_date <= Date.current?

thenapking: I did struggle with this name.  The trouble with `deadline_passed?` is then we have a negative condition when we use since we need to make sure it hasn't passed, ie `!deadline_passed?`.  I'll try to make this better, and I will check about the equality condition.

app/services/statements/authorise_payment.rb
```diff
@@ -13,21 +13,58 @@ def authorise!
...
 
-          statement.update!(marked_as_paid_at: paid_at)
-          statement.mark_as_paid!
+  private
+
+    def settle_statement!
+      statement.mark_as_paid!
+      statement.update!(marked_as_paid_at: Time.zone.now)
+      record_statement_authorised_for_payment_event!
+    end
```
ericaporter, line 28: You could set a paid_at time stamp e.g. paid_at = Time.zone.now and pass it into `statement.update!(marked_as_paid_at: paid_at)` and `record_statement_authorised_for_payment_event!(happened_at: paid_at)` and keep `happened_at:` but also not a blocker

app/services/statements/authorise_payment.rb
```diff
@@ -13,21 +13,58 @@ def authorise!
...
+        record_declaration_clawed_back_event!(declaration)
       end
     end
+
+    def declarations_payable
+      statement.payment_declarations.payment_status_payable
+    end
+
+    def declarations_awaiting_clawback
+      statement.payment_declarations.clawback_status_awaiting_clawback
```
ericaporter: Can you check if this should be `statement.clawback_declarations.clawback_status_awaiting_clawback`?

app/models/statement.rb
cpjmcquillan: `within_deadline?` definitely doesn't make sense to me.

I'd be tempted to leave it inline

```ruby
deadline_date.before?(Date.current)
```

spec/services/statements/authorise_payment_spec.rb
```diff
@@ -1,89 +1,75 @@
...
 
-      it "raises and rolls back the payment change" do
-        expect {
+      context "when there are declarations awaiting clawback" do
+        let!(:declarations_awaiting_clawback) do
+          FactoryBot.create_list(:declaration, 2, :with_ect,
+                                 declaration_type: "started",
+                                 clawback_status: "awaiting_clawback",
+                                 school_partnership:,
+                                 payment_statement: statement)
```
ericaporter: This possibly should be `clawback_statement: statement` if `declarations_awaiting_clawback` uses `clawback_declarations`

spec/views/admin/finance/authorisations/new.html.erb_spec.rb
```diff
@@ -1,8 +1,37 @@
...
+                      school_partnership:,
+                      payment_statement:)
+  end
+
+  let!(:clawed_back_declaration) do
+    FactoryBot.create(:declaration, :with_ect,
+                      declaration_type: "started",
+                      clawback_status: "awaiting_clawback",
+                      school_partnership:,
+                      payment_statement:)
```
ericaporter: Possibly this one too

app/services/statements/authorise_payment.rb
thenapking: I've double checked this and I removed that because `happened_at` is defaulted to `Time.zone.now`.  There will I suppose be a small disparity but I'm not sure that really matters.

ericaporter (approved):
LGTM! 👍

app/services/statements/authorise_payment.rb
```diff
@@ -13,21 +13,58 @@ def authorise!
       ActiveRecord::Base.transaction do
         raise NotAuthorisable unless statement.can_authorise_payment?
 
-        if statement.marked_as_paid_at.blank?
-          paid_at = Time.zone.now
+        settle_declarations!
+        refund_declarations!
```
cpjmcquillan, line 17: I wonder whether it's worth being a little more explicit here

```ruby
settle_payable_declarations!
refund_awaiting_clawback_declarations!
```

What do you think?

cpjmcquillan (approved):

app/services/statements/authorise_payment.rb
thenapking: I've thought about this more.  I chose those names for brevity.  I didn't want call this `clawback_awaiting_clawback_declarations!`, and we can't call it `clawback_declarations!`. Declarations which are `awaiting_clawback` could have been called `refundable` declarations.  Similarly you only settle what is payable.  So by using this synonym we can keep these method names short, and I, hope, clear.

PR #2991, opened 2026-05-29
Simplify start date validators

cpjmcquillan:
On my quest to eliminate the magic date (31st May 2025) from the test suite, I took the opportunity to simplify these validators.

The `current_date` argument is never used outside the tests, so we may aswell remove it.

The tests have been simplified to use the `have_error` matcher, and context blocks reorganised to reduce unnecessary duplication.

The `ECTStartDateValidator` spec no longer relies on a magic date, and instead uses preceding contract periods to determine the earliest invalid date.


peteryates (approved):
This definitely feels easier to follow.

PR #2995, opened 2026-06-01
Fix failing tests on main

cpjmcquillan:
### Context

[Failures!](https://github.com/DFE-Digital/register-early-career-teachers-public/actions/runs/26745684067/job/78820449517)

#### Stub `Schools::RegistrationWindow.closed?`
Most of our tests expect the registration window to be open!

This stubs the method before each spec to reflect that.

When we want to test the behaviour of `Schools::RegistrationWindow` we can
remove the stub.

#### Fix date-dependent tests
A bunch of our tests failed as soon as today's date moved from the 2025
contract period to the 2026 contract period.

This updates a range of tests to be more explicit about the contract period they
depend on, and the dates periods start on.

#### Fix failing specs
The API seed spec is failing as we are creating test data for 2024 and 2025,
but the 2026 contract period also exists, so we need to be more specific with
our assertions.

The `Create` declaration spec fails as the statement factory was updated to set
a `deadline_date` relative to the contract period, which for this test is 2021.
We can ensure a future `deadline_date` by setting explicitly, resulting in the
service being valid/the declaration being accepted.

### DIFF
--- a/spec/requests/schools/mentors/change_lead_provider_wizard_spec.rb
+++ b/spec/requests/schools/mentors/change_lead_provider_wizard_spec.rb
@@ -165,7 +165,7 @@ describe "Schools::Mentors::ChangeLeadProviderWizard Requests" do
               # different schedule to `started_on` without being explicit.
               # This travels toward the end of the contract period to ensure
               # that happens.
-              travel_to contract_period.finished_on.advance(months: -1)
+              travel_to contract_period.finished_on.prev_month
               # We've travelled so far we need to sign in again to ensure
               # the user is still authenticated.
               sign_in_as(:school_user, school:)
--- a/spec/requests/schools/mentors/change_lead_provider_wizard_spec.rb
+++ b/spec/requests/schools/mentors/change_lead_provider_wizard_spec.rb
@@ -9,8 +9,7 @@ describe "Schools::Mentors::ChangeLeadProviderWizard Requests" do
       :ongoing,
       teacher:,
       school:,
-      started_on:,
-      email: "mentor@example.com"
+      started_on:
     )
   end
 
@@ -94,7 +93,7 @@ describe "Schools::Mentors::ChangeLeadProviderWizard Requests" do
     end
 
     context "when signed in as a School user" do
-      let!(:user) { sign_in_as(:school_user, school:) }
+      before { sign_in_as(:school_user, school:) }
 
       context "when the current_step is invalid" do
         it "returns not found" do
@@ -159,6 +158,19 @@ describe "Schools::Mentors::ChangeLeadProviderWizard Requests" do
                                 expression_of_interest: active_lead_provider)
             end
 
+            before do
+              # To test that a different schedule is assigned, we need to make
+              # sure we are changing lead provider at a point in time that would
+              # change the schedule. We can't rely on "today" determining a
+              # different schedule to `started_on` without being explicit.
+              # This travels toward the end of the contract period to ensure
+              # that happens.
+              travel_to contract_period.finished_on.advance(months: -1)
+              # We've travelled so far we need to sign in again to ensure
+              # the user is still authenticated.
+              sign_in_as(:school_user, school:)
+            end
+
             it "assigns a new schedule" do
               subject
 
@@ -192,7 +204,6 @@ describe "Schools::Mentors::ChangeLeadProviderWizard Requests" do
 
           post(path_for_step("check-answers"))
 
-          expect(training_period.reload.finished_on).to eq(Date.yesterday)
           new_training_period = mentor_at_school_period.training_periods.ongoing.first
           expect(new_training_period.expression_of_interest.lead_provider).to eq(lead_provider)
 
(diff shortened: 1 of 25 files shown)


spec/support/features/changes_before_reported_leaving_date_helpers.rb
```diff
@@ -2,12 +2,30 @@ module ChangesBeforeReportedLeavingDateHelpers
 private
 
   # Setup
+  def given_there_is_a_contract_period
```
peteryates: I wonder if this should be `given_we_are_in_the_middle_of_a_contract_period`, which could lead to us having `given_we_are_in_the_last_week_of_a_contract_period` etc.

I like that this makes `Date.current` and its distance from contract period boundaries much less significant.

spec/requests/schools/mentors/change_lead_provider_wizard_spec.rb
```diff
@@ -159,6 +158,19 @@
...
             end
 
+            before do
+              # To test that a different schedule is assigned, we need to make
+              # sure we are changing lead provider at a point in time that would
+              # change the schedule. We can't rely on "today" determining a
+              # different schedule to `started_on` without being explicit.
+              # This travels toward the end of the contract period to ensure
+              # that happens.
+              travel_to contract_period.finished_on.advance(months: -1)
```
ethax-ross: Advancing `-1` months reads weird, what about `travel_to contract_period.finished_on.prev_month`?

spec/services/contract_periods/seed_from_previous_spec.rb
```diff
@@ -17,8 +17,12 @@
 
   describe "#schedule!" do
     let(:contract_period) do
-      FactoryBot.create(:contract_period,
-                        year: current_year)
+      FactoryBot.create(
+        :contract_period,
+        :current,
+        started_on: 1.week.from_now,
```
ethax-ross: Why the specific start/end dates here? 🤔

spec/wizards/schools/register_ect_wizard/registration_store/queries_spec.rb
```diff
@@ -141,7 +141,7 @@ def build_previous_training_period_double(provider_led:, contract_period:, lead_
     end
 
     context "when the start_date is present" do
-      let!(:contract_period) { create_contract_period(year: 2025) }
```
ethax-ross, line 144: Why do we change this to 2026 from 2025? will it work when `Date.current` is `2027`? 🤔

ethax-ross (commented):
Looks good 👏 just a few questions/minor suggestions

spec/requests/schools/mentors/change_lead_provider_wizard_spec.rb
peteryates: is there no `#retreat`? :sweat_smile:

spec/services/contract_periods/seed_from_previous_spec.rb
cpjmcquillan: Good question, otherwise we run into a validation error

https://github.com/DFE-Digital/register-early-career-teachers-public/blob/7fc3619377206270430bb31091cc8688677bd7b1/app/services/contract_periods/seed_from_previous.rb#L23

Before, this file used the current year to build the contract periods. And until today, we were always **before** that 2026 contract period had started!

ethax-ross: Ah so I think we just need to make:

```
let(:contract_period) { FactoryBot.create(:contract_period, :next) }
```

And:

```
let(:previous_contract_period) do
  FactoryBot.create(:contract_period, :current)
end
```

The naming is a bit misleading, but the `SeedFromPrevious` service can only seed a contract period that has not yet started and relies on a 'previous' contract period (to the one being seeded) being available.

spec/wizards/schools/register_ect_wizard/registration_store/queries_spec.rb
cpjmcquillan: I changed these to `Date.current` because it was a simpler change than rewriting the tests to use `FactoryBot.create(:contract_period, :current)`.

As far as I can tell, the year of the contract period isn't significant (except for reassigning from frozen contract periods). This change reflects that.

I ran all the tests in this file with a `before { travel_to 1.year.from_now }` block and they all still passed.

@ericaporter do you have any more context?

spec/requests/schools/mentors/change_lead_provider_wizard_spec.rb
cpjmcquillan: Agreed it's a bit janky. I'll update this 👌🏼

spec/support/features/changes_before_reported_leaving_date_helpers.rb
cpjmcquillan: Yeah, I think that's a bit clearer.

Specifically for these feature specs, we need to be **in** the contract period, so we can easily setup the relevant periods and say the ECT is joining/leaving without running into validation errors.

peteryates (approved):
This looks good, definitely more robust and easier to follow than before.

tonyheadford (approved):
Looks good, nice one

PR #2999, opened 2026-06-01
updating_masp_end_dates

boonali:
### Context

The migration would usually not set a finished_on date for mentor_at_school_period unless the mentor had been recorded as having left the school. However, we have to use a CSV to overwrite some ECF1 induction records to ensure they migrated at all, and sometimes this involved setting an end date (for training) which has also been used for the at_school_period incorrectly.

Nullifying 3.8k MASPs who have not recorded as having left the school.


tonyheadford (approved):
All ran fine in testing

PR #3003, opened 2026-06-02
Add `changed_lead_provider` to `TrainingPeriod.withdrawal_reason`

tonyheadford:
### Context

Adds a new withdrawal reason to the `TrainingPeriod.withdrawal_reason` enum and the underlying database 

The enum does not appear to be surfaced anywhere other than via `.values` (so no mapping except for `dasherize`) or surfaced as a statement that training has been withdrawn by the provider without specifying the reason.

- Adds `changed_lead_provider` to DB enum type `withdrawal_reasons` and in the `TrainingPeriod.withdrawal_reason` enum.
- Adds the new value to the spec.


ethax-ross (approved):
Thanks Tony, looks good just the release note needed before it can go out I think

PR #3004, opened 2026-06-02
[3940] schedules and [3941] milestones

peterdavidhamilton:
### Context

Closes [schedule](https://github.com/DFE-Digital/register-ects-project-board/issues/3940) and closes [milestone](https://github.com/DFE-Digital/register-ects-project-board/issues/3941) 

Add PoC views, services, controllers, event logging etc for the end of the contract period journey in the admin finance interface.

Out of scope:

- [milestone date validations](https://github.com/DFE-Digital/register-ects-project-board/issues/4026) follow-on ticket
- [fixes and tweaks to earlier routes](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3026) extracted from this PR following comment

### Guidance to review

cpd-ec2-review-3004-web test teacherservices cloud_admin_finance_contract-periods_2026_schedules_143

cpd-ec2-review-3004-web test teacherservices cloud_admin_finance_contract-periods_2027_schedules_155

cpd-ec2-review-3004-web test teacherservices cloud_admin_finance_contract-periods_2027_schedules_155_milestones

cpd-ec2-review-3004-web test teacherservices cloud_admin_finance_contract-periods_2027_schedules_155 (1)


app/controllers/admin/finance/schedules_controller.rb
```diff
@@ -10,10 +15,74 @@ def index
...
+    end
+
   private
 
     def set_contract_period
       @contract_period = ContractPeriod.find(params[:contract_period_id])
     end
+
+    def set_schedule
+      @schedule = Schedule.find(params[:id])
```
craigmdavidson: We should scope the `@schedule` _find_ through **ContractPeriod**.

app/controllers/admin/finance/schedules_controller.rb
```diff
@@ -10,10 +15,74 @@ def index
...
+    def destroy
+      @service = Schedules::Destroy.new(
+        author: current_user,
+        schedule: @schedule
+      )
+
+      if @service.destroy!
+        redirect_to admin_contract_period_schedules_path(@contract_period),
+                    alert: "#{@schedule.name} schedule removed"
+      else
```
craigmdavidson: 1) So the service calls things like `destroy!` which will raise, but in the controller we act as if the controller will return something truthy. We should instead be calling the service, and catching anything that was raised appropriately.

2) Also, nitpick, personally I find `@service.destroy!` confusing? Are we destroying the service?

Might be clearer along the lines of:

```
def destroy
  Schedules::Destroy.new(author: current_user, schedule: @schedule).call
  redirect_to admin_contract_period_schedules_path(@contract_period),
                    alert: "#{@schedule.name} schedule removed"
 rescue 
      redirect_to admin_contract_period_schedule_path(@contract_period, @schedule),
                    alert: "#{@schedule.name} schedule could not be removed"
end
```

3) Also, we are flashing an _alert_ in both a succces and failure scenario?

app/controllers/admin/finance/schedules_controller.rb
```diff
@@ -10,10 +15,74 @@ def index
...
+    end
+
+    def create
+      @service = Schedules::Create.new(
+        author: current_user,
+        contract_period_year: @contract_period.year,
+        identifier: schedule_params[:identifier]
+      )
+
+      if @service.create!
```
craigmdavidson: As per the comment for destroy? @service.create! reads confusingly? We are creating the service?

The @service internally calls `save!` so will never return false, so we want to be catching?

app/controllers/admin/finance/schedules_controller.rb
```diff
@@ -10,10 +15,74 @@ def index
...
+        redirect_to admin_contract_period_schedules_path(@contract_period),
+                    alert: "#{@service.schedule.name} schedule added"
```
craigmdavidson: Why do we `alert` for a successful addition?

app/models/milestone.rb
```diff
@@ -1,23 +1,30 @@
...
-  validates :milestone_date, comparison: { greater_than: :start_date, message: "Milestone date must be after the start date" }, allow_nil: true
+
+  # TODO: Add milestone start date validations (sensible values within the contract period)
+  validates :start_date,
+            presence: { message: "Enter a start date" }
+
+  # TODO: Add "if start_date.present?" to milestone validation
+  validates :milestone_date,
+            comparison: { greater_than: :start_date, message: "Milestone date must be after the start date" },
+            allow_nil: true
```
craigmdavidson, line 26: Why the TODO? Isn't this just `if: -> { start_date.present? }`?

app/views/admin/finance/schedules/show.html.erb
```diff
@@ -0,0 +1,49 @@
...
+<% end %>
+
+<div class="govuk-button-group">
+  <%=
+    govuk_button_to "Delete schedule", admin_contract_period_schedule_path(@contract_period, @schedule),
+      visually_hidden_suffix: @schedule.name,
+      method: :delete,
+      warning: true,
+      disabled: @schedule.milestones.any?
+  %>
```
craigmdavidson, line 41: Does this not mean we show the button as enabled if the contract_period has started?

If so the controller will reject it and raise an error.

app/controllers/admin/finance/milestones_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+    end
+
+  private
+
+    def set_contract_period
+      @contract_period = ContractPeriod.find(params[:contract_period_id])
+    end
+
+    def set_schedule
+      @schedule = Schedule.find(params[:schedule_id])
```
craigmdavidson: This should be scoped to the contract_period?
i.e. `@contract_period.schedules.find(params[:schedule_id])`

app/controllers/admin/finance/milestones_controller.rb
```diff
@@ -0,0 +1,63 @@
+module Admin::Finance
+  class MilestonesController < Admin::Finance::BaseController
+    before_action :set_contract_period
+    before_action :set_schedule
+    before_action :redirect_if_contract_period_started
+
+    def new
+      @milestone = Milestone.new
```
craigmdavidson: This should be scoped to the schedule? I.e. `@schedule.milestones.build`

app/services/schedules/create.rb
```diff
@@ -0,0 +1,28 @@
...
+    attr_reader :author,
+                :schedule
+
+    def initialize(author:, contract_period_year:, identifier:)
+      @author = author
+      @schedule = Schedule.new(contract_period_year:, identifier:)
+    end
+
+    def create!
+      return false unless schedule.valid?
```
craigmdavidson: This isn't ideal - it means we are both boolean returning and explosive.  

We will possibly still fail in `schedule.save!` so we will need to both rescue and interrogate the boolean response. Since we have `schedule.save!` and `record_event!` as explosive failures, it is better to treat the service as explosive, avoid the boolean return, and remove this line (as it will be covered by `schedule.save!`.

app/views/admin/finance/contract_periods/new.html.erb
```diff
@@ -14,6 +14,8 @@
       <%= f.govuk_submit "Save" %>
     <% end %>
 
-    <%= govuk_link_to "Cancel and return to Contract periods", admin_contract_periods_path %>
+    <p class="govuk-body">
+      <%= govuk_link_to "Cancel and return to Contract periods", admin_contract_periods_path %>
+    </p>
```
craigmdavidson: This is out of scope of this PR?

app/views/admin/finance/contract_periods/show.html.erb
```diff
@@ -1,4 +1,9 @@
-<% page_data(title: "Contract period #{@contract_period.year}", breadcrumbs: @breadcrumbs) %>
+<%
+  page_data(
+    title: "#{@contract_period.year} contract period",
+    breadcrumbs: @breadcrumbs
+  )
+%>
```
craigmdavidson: This is out of scope of this PR?

app/views/admin/finance/contract_periods/index.html.erb
```diff
@@ -19,6 +19,6 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-full">
-    <%= govuk_button_to "Add Contract period", new_admin_contract_period_path, method: :get %>
+    <%= govuk_button_link_to "Add Contract period", new_admin_contract_period_path %>
```
craigmdavidson: This is out of scope of this PR?

app/views/admin/finance/contract_periods/edit.html.erb
```diff
@@ -14,6 +14,8 @@
       <%= f.govuk_submit "Save" %>
     <% end %>
 
-    <%= govuk_link_to "Cancel and return to Contract periods", admin_contract_periods_path %>
+    <p class="govuk-body">
+      <%= govuk_link_to "Cancel and return to Contract periods", admin_contract_periods_path %>
+    </p>
```
craigmdavidson: This is out of scope of this PR?

app/controllers/admin/finance/schedules_controller.rb
```diff
@@ -10,10 +15,74 @@ def index
...
+      else
+        @schedule = @service.schedule
+        render :new, status: :unprocessable_content
+      end
+    rescue ActionController::ParameterMissing
```
craigmdavidson, line 41: Any other raise will 500 here?

app/controllers/admin/finance/schedules_controller.rb
```diff
@@ -10,10 +15,74 @@ def index
...
+        "Finance" => admin_finance_path,
+        "Contract periods" => admin_contract_periods_path,
+        @contract_period.year.to_s => admin_contract_period_path(@contract_period),
+        "Schedules" => admin_contract_period_schedules_path(@contract_period),
+        @schedule.name => nil
+      }
+    end
+
+    def new
+      @schedule = Schedule.new
```
craigmdavidson: We should be scoping here: i.e. `@contract_period.schedules.build`?

craigmdavidson (commented):
Looks good, a few points.

FWIW, this would be lots easier, if there this was two PRs, one for schedules, and one for milestones.

app/controllers/admin/finance/schedules_controller.rb
peterdavidhamilton: Parroting how the contract period controller behaves

app/models/milestone.rb
peterdavidhamilton: Same as the other TODO above, which was flagging the gap in validations for the ticket I am looking at now

app/controllers/admin/finance/schedules_controller.rb
peterdavidhamilton: 💯

app/models/milestone.rb
craigmdavidson: The problem being in this case it will 500 though?

app/controllers/admin/finance/schedules_controller.rb
peterdavidhamilton: I also prefer the proc-style call for functional services, but again we seem to have a pattern of bespoke public methods for them elsewhere in the service.

app/controllers/admin/finance/contract_periods_controller.rb
```diff
@@ -92,10 +92,11 @@ def set_contract_period_flags
       @has_schedules = @contract_period.schedules.any?
     end
 
-    def redirect_if_contract_period_not_editable
-      return if @editable
+    def redirect_if_contract_period_started
+      return if @contract_period.editable?
```
ethax-ross: We seem to be flip/flopping on the terminology; if we're going with 'started' instead of 'editable' should we amend the method to match, so `@contract_period.started?`

app/controllers/admin/finance/milestones_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+    before_action :set_contract_period
+    before_action :set_schedule
+    before_action :redirect_if_contract_period_started
+
+    def new
+      @milestone = @schedule.milestones.build
+    end
+
+    def create
+      @service = Milestones::Create.new(
```
ethax-ross: Do we need to expose `@service` in these?

app/controllers/admin/finance/schedules_controller.rb
ethax-ross: This is inline with how we do it in the API, but I'm happy if we want to move to a `call` pattern if that's the consensus. We had issues with `call` being misused in ECF so I have a bit of a knee-jerk aversion to it personally and I don't really see a down-side to using descriptive method names, but with `call` I've seen:

- Services changing over time where you're now left with a `call` method alongside other methods
- Poorly named classes using the pattern, e.g. `Processor.call`
- Services extracted to instance variables so you lose context, e.g. `service.call`

If it's done right/when appropriate I'm fine with it, I just think its more susceptible to being done wrong 🙂

spec/models/schedule_spec.rb
```diff
@@ -51,32 +51,137 @@
...
+    describe "#name" do
+      subject { schedule.name }
+
+      it { is_expected.to eq("Standard April") }
     end
 
-    it "returns false for non-reduced schedules" do
-      schedule = Schedule.new(identifier: "ecf-standard-april")
-      expect(schedule).not_to be_reduced_schedule
+    # TODO: use #name in the #description
```
ethax-ross: Worth just doing this change?

app/services/events/record.rb
```diff
@@ -1048,28 +1048,50 @@ def self.record_contract_period_added_event!(author:, contract_period:)
...
-        happened_at: Time.zone.now,
-        modifications:
-      ).record_event!
+      new(event_type:, author:, heading:, contract_period:, happened_at: Time.zone.now, modifications:).record_event!
+    end
+
+    def self.record_schedule_added_event!(author:, schedule:)
+      event_type = :schedule_added
+      contract_period = schedule.contract_period
+      heading = "Schedule '#{schedule.name}' added to #{contract_period.year}"
```
ethax-ross: Could make these a bit more concise with `schedule.description`?

`#{schedule.description} added/removed`
`Milestone #{milestone} added/removed to/from #{schedule.description}`

app/views/admin/finance/schedules/show.html.erb
```diff
@@ -0,0 +1,49 @@
...
+<% end %>
+
+<div class="govuk-button-group">
+  <%=
+    govuk_button_to "Delete schedule", admin_contract_period_schedule_path(@contract_period, @schedule),
+      visually_hidden_suffix: @schedule.name,
+      method: :delete,
+      warning: true,
+      disabled: @schedule.milestones.any?
```
ethax-ross, line 40: Is it worth adding `dependent: :destroy` to schedule -> milestones so we don't have to disable this?

spec/models/contract_period_spec.rb
```diff
@@ -413,4 +414,89 @@
       it { is_expected.not_to be_editable }
     end
   end
+
+  describe "#available_schedules" do
+    subject(:contract_period) { FactoryBot.create(:contract_period) }
+
+    context "without associated schedules" do
+      it "lists all possible schedules" do
+        expect(contract_period.available_schedules).to eq(%w[
```
ethax-ross: `Schedule.identifier.keys`?

spec/models/contract_period_spec.rb
```diff
@@ -413,4 +414,89 @@
...
+      end
+
+      before do
+        associated_schedules.each do |identifier|
+          FactoryBot.create(:schedule, identifier:, contract_period:)
+        end
+      end
+
+      it "lists any remaining schedules" do
+        expect(contract_period.available_schedules).to eq(%w[
```
ethax-ross: `Schedule.identifier.keys - associated_schedules`?

ethax-ross (commented):
Thanks PDH looking good, just a few minor suggestions. I agree this would have been easier to review as a few separate PRs though, there's a lot in here 😅

app/controllers/admin/finance/contract_periods_controller.rb
peterdavidhamilton: ~I was trying to align with comments in other PRs, I'll change to `@contract_period.started?`, I think that is a good choice because the model cares about its state and not what others can do because of that state~

No I've changed my mind for the time being. "editable" seems well embedded. I'll try flipping something else.

app/views/admin/finance/schedules/show.html.erb
peterdavidhamilton: I felt a little friction was prudent, especially until someone had provided product feedback.

app/controllers/admin/finance/milestones_controller.rb
```diff
@@ -0,0 +1,67 @@
...
+    end
+
+    def create
+      @milestone = Milestones::Create.new(
+        author: current_user,
+        schedule: @schedule,
+        params: milestone_params
+      ).create!
+
+      if @milestone.persisted?
```
craigmdavidson: Is the milestone ever going to come back as not persisted?

Inside the create we do milestone.save! which is explosive?

Would something more like this work better?:

```
def create
  @milestone = Milestones::Create.new(...).create_milestone!
  redirect_to admin_contract_period_schedule_path(@contract_period, @schedule),
                    alert: "#{@milestone.declaration_type.titleize} milestone added"
rescue => _e
  @milestone = @schedule.milestones.build
  render :new, status: :unprocessable_content
end
```

app/controllers/admin/finance/milestones_controller.rb
```diff
@@ -0,0 +1,67 @@
...
+      render :new, status: :unprocessable_content
+    end
+
+    def destroy
+      service = Milestones::Destroy.new(
+        author: current_user,
+        milestone: @schedule.milestones.find(params[:id])
+      )
+
+      if service.destroy!
```
craigmdavidson: Again Destroy is destructive - as per create? 

Nitpick: We shouldn't assign the service object to a variable?

app/controllers/admin/finance/milestones_controller.rb
```diff
@@ -0,0 +1,67 @@
...
+    rescue ActiveRecord::RecordInvalid,
+           ActiveModel::ValidationError
+      @schedule = @schedule.milestones.build
+      render :new, status: :unprocessable_content
+    end
+
+    def destroy
+      service = Milestones::Destroy.new(
+        author: current_user,
+        milestone: @schedule.milestones.find(params[:id])
```
craigmdavidson: assign to @milestone here?....

app/controllers/admin/finance/milestones_controller.rb
```diff
@@ -0,0 +1,67 @@
...
+      )
+
+      if service.destroy!
+        redirect_to admin_contract_period_schedule_path(@contract_period, @schedule),
+                    alert: "#{service.milestone.declaration_type.titleize} milestone removed"
```
craigmdavidson: so we can use the assigned @milestone here...?

app/controllers/admin/finance/milestones_controller.rb
```diff
@@ -0,0 +1,67 @@
...
+      else
+        redirect_to admin_contract_period_schedule_path(@contract_period, @schedule),
+                    alert: "#{service.milestone.declaration_type.titleize} milestone could not be removed"
```
craigmdavidson: ... and here?

app/controllers/admin/finance/schedules_controller.rb
```diff
@@ -10,10 +15,75 @@ def index
...
+    end
+
+    def create
+      @schedule = Schedules::Create.new(
+        author: current_user,
+        contract_period_year: @contract_period.year,
+        identifier: schedule_params[:identifier]
+      ).create!
+
+      if @schedule.persisted?
```
craigmdavidson: Again Schedules::Create is explosive? So no conditional?

app/controllers/admin/finance/schedules_controller.rb
```diff
@@ -10,10 +15,75 @@ def index
...
+      render :new, status: :unprocessable_content
+    end
+
+    def destroy
+      service = Schedules::Destroy.new(
+        author: current_user,
+        schedule: @schedule
+      )
+
+      if service.destroy!
```
craigmdavidson: Again Schedules::Destroy is explosive, so we should not have a conditional here.

craigmdavidson (changes requested):
The PR looks great, it looks like there are a couple of stray conditional checks in the controller?

app/controllers/admin/finance/milestones_controller.rb
peterdavidhamilton: Hopefully all the service objects now look better. I have applied the same [here](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3026/changes/f86d0868c22852284162c988cbd498b75bf2e1c5) to the contracts.

I am not keen on swallowing all errors here. What do we think? @craigmdavidson @ethax-ross

ethax-ross (approved):
Looks great, thanks PDH

craigmdavidson (approved):
Looks great! 😅

PR #3006, opened 2026-06-02
Revert "Reverse fee structure foreign keys part 2"

ethax-ross:
Reverts DFE-Digital/register-early-career-teachers-public#2975

Failing to deploy to staging due to some bad shapes in the contract seed data - I'll address separately then re-raise.


tonyheadford (approved):
~~We'll also need to repopulate the `contract_id` (after this has run) but otherwise~~ I think this looks OK

PR #3007, opened 2026-06-02
Fix contract seed data

ethax-ross:
We were getting duplicate `FlatRateFeeStructure`s created due to setting the `id` to `nil` and not the association in the `Contract` factory.

We were also creating and sharing fee structures across the contracts in the declarations seeds.

### DIFF
--- a/db/seeds/declarations.rb
+++ b/db/seeds/declarations.rb
@@ -91,9 +91,6 @@ ITTECF_BOUNDARIES = [
 
 print_seed_info("Creating Statements")
 
-ecf_fee_structure = FactoryBot.create(:contract_banded_fee_structure, :with_bands, declaration_boundaries: ECF_BOUNDARIES)
-ittecf_fee_structure = FactoryBot.create(:contract_banded_fee_structure, :with_bands, declaration_boundaries: ITTECF_BOUNDARIES)
-
 ucl = LeadProvider.find_by!(name: "UCL Institute of Education")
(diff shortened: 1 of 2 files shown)


tonyheadford (approved):

db/seeds/declarations.rb
```diff
@@ -116,17 +113,22 @@ def milestones
   describe_school_partnership(partnership)
 end
 
+ecf_banded_fee_structure = FactoryBot.create(:contract_banded_fee_structure, :with_bands, declaration_boundaries: ECF_BOUNDARIES)
 ucl_contract_2024 = FactoryBot.create(:contract,
                                       :for_ecf,
                                       active_lead_provider: school_partnership_2024.active_lead_provider,
-                                      banded_fee_structure: ecf_fee_structure).tap do |contract|
+                                      banded_fee_structure: ecf_banded_fee_structure,
+                                      flat_rate_fee_structure: nil).tap do |contract|
```
thenapking, line 121: I had not realised this, but that does explain some of the duplication. Do you think we could fix this in the factory instead, as we use `:with_bands` in the test suite.  Perhaps we could have one trait for ECF contracts, one for ITTECF?

ethax-ross: Ah it is fixed in the factory as well, I've just made it explicit here 🙂 we're in the process of flipping the foreign key onto the fee structure to avoid fee structures being shared between contracts as well (that was an incorrect assumption in our original modelling)

thenapking: Excellent, thanks.  I've got some open PRs which use this trait so this is very handy!

PR #3008, opened 2026-06-02
Replant the seeds on staging reseed

ethax-ross:
At the moment it runs `db:seed` which will fail due to trying to create existing statements again. We should `replant` to wipe what's there as we run a full reseed anyway. This is consistent with how the review apps reseed.


craigmdavidson (approved):
LGTM - BWDIK

PR #3009, opened 2026-06-02
Reapply "Reverse fee structure foreign keys part 2"

ethax-ross:
This is take 2 of https://github.com/DFE-Digital/register-early-career-teachers-public/pull/2975 which we had to revert due to data issues in staging that have now been resolved.


tonyheadford (approved):
LGTM

PR #3010, opened 2026-06-02
Fix induction tutor bug

cpjmcquillan:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/4040

#### Handle no contract period in induction tutor wizard
Before, if there was no contract period users would run into errors if they
attempted to update induction tutor details.

We don't need to know the contract period to update a school's induction tutor
name or email address.

This reworks the `Schools::InductionTutor::CheckAnswersStep` to handle this
scenario.

Now, users can update their school's induction tutor name or email address
without running into any errors.

[Sentry]

[Sentry]: https://dfe-teacher-services.sentry.io/issues/7511558483

#### Prevent assigning induction tutor without contract period
If there is neither a current or upcoming contract period, we have no way of
knowing which contract period to assign as the school's
`induction_tutor_last_nominated_in`.

In https://github.com/DFE-Digital/register-early-career-teachers-public/commit/f6bea6b052f943dcd131a1e39f71f852a7cb9fe2, we made changes to the
`Schools::InductionTutorDetails` service so it would not redirect users to
add or confirm details if there was no relevant contract period.

This adds some redundancy in the form of explicit redirects on the respective
wizard controllers if that is the case.

Typically we wouldn't expect any users to reach these pages, but this adds a
guard-rail which will prevent errors if they did reach these pages and attempt
to complete the wizards.

### DIFF
--- a/app/controllers/schools/induction_tutor/new_induction_tutor_wizard_controller.rb
+++ b/app/controllers/schools/induction_tutor/new_induction_tutor_wizard_controller.rb
@@ -2,6 +2,17 @@ module Schools
   module InductionTutor
     class NewInductionTutorWizardController < SchoolsController
       include InductionTutorable
+
+      before_action :redirect_when_no_contract_period
+
+    private
+
+      def redirect_when_no_contract_period
+        return if ContractPeriod.current_or_upcoming.present?
+
+        redirect_to schools_induction_tutor_path,
+                    alert: "You cannot assign or confirm an induction tutor at this time"
+      end
     end
   end
 end
--- a/app/wizards/schools/induction_tutor/check_answers_step.rb
+++ b/app/wizards/schools/induction_tutor/check_answers_step.rb
@@ -7,14 +7,10 @@ module Schools
 
       def save!
         ActiveRecord::Base.transaction do
-          record_event!
-
-          school.update!(
-            induction_tutor_name: store.induction_tutor_name,
-            induction_tutor_email: store.induction_tutor_email,
-            induction_tutor_last_nominated_in: closest_contract_period
-          )
-          true
+          old_induction_tutor_name = school.induction_tutor_name
+          assign_induction_tutor_attributes
+          school.save!
+          record_event!(old_induction_tutor_name)
         end
 
         send_confirmation_email!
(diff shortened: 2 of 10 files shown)


app/controllers/schools/induction_tutor/new_induction_tutor_wizard_controller.rb
```diff
@@ -2,6 +2,17 @@ module Schools
   module InductionTutor
     class NewInductionTutorWizardController < SchoolsController
       include InductionTutorable
+
+      before_action :redirect_when_no_contract_period
+
+    private
+
+      def redirect_when_no_contract_period
```
thenapking, line 10: Since this method is repeated it could go in the `InductionTutorable` mixin.  That was created because these controllers were all pretty much identical.  Since the `before_action` isn't needed on the Update controller we can keep that call in the controller.

app/wizards/schools/induction_tutor/check_answers_step.rb
```diff
@@ -23,19 +19,27 @@ def save!
...
-      def record_event!
+      def record_event!(old_induction_tutor_name)
         Events::Record.record_school_induction_tutor_updated_event!(
           school:,
-          old_name: school.induction_tutor_name,
+          old_name: old_induction_tutor_name,
           new_name: store.induction_tutor_name,
           new_email: store.induction_tutor_email,
-          contract_period_year: closest_contract_period.year,
+          contract_period_year: closest_contract_period&.year,
```
thenapking, line 42: When the new contract period starts, we ask the school to update the induction tutor details.  We'll then end up with an event with no contract year.  I'm not sure how much this matters, but it could instead be:
```
closest_contract_period&.year || school.induction_tutor_last_nominated_in
```

thenapking (approved):
This looks fine.  A couple of comments, one which could just make it a little more dry, the other is more optional.

app/wizards/schools/induction_tutor/check_answers_step.rb
cpjmcquillan: I'm not sure I follow what you're saying here 🤔.

If a new contract period has started, then by definition there will be a `closest_contract_period`, so the event would have a `contract_period_year`. 

The tests cover each scenario.

If there isn't a closest contract period, it makes sense IMO that the event's `contract_period_year` should be `nil`. The school won't have nominated an induction tutor for a contract period, since they will have only updated a name or email address (there is no contract period).

PR #3012, opened 2026-06-02
Investigate how to expose a better "date a transfer was created" in the API

tonyheadford:
### Context

Currently the `TrainingPeriod.created_at` is used in the transfers API as the date a transfer was `created_at`.
Obviously this is when the training period was added so it does not feel correct.

Looking at options to improve this so if we added a new attribute `transfer_initiated_at` to `TrainingPeriod` and recorded the date a school/admin told us that a teacher was leaving their school it would provide a more accurate value for the API.

This spike assumes that we'll only record when a ongoing `training_period` is closed by the parent `school_period`  closing.  It would be even simpler if we wanted to record this whenever a `training_period` closes (i.e. provider changes).

I've stopped short of writing specs at this stage, just to make sure this is going in the correct direction, so comments welcome please.

- Adds a new `datetime` column to `training_periods` named `transfer_initiated_at`
- Adds extra param to the `TrainingPeriod::Finish` methods named `leaving_school` which defaults to `false`.  This is used to decide whether to set the `transfer_initiated_at` or not
- Use the `transfer_initiated_at` in the transfers serializer (defaulting to `created_at` if not present) for the `created_at` attribute

### DIFF
--- a/app/services/ect_at_school_periods/finish.rb
+++ b/app/services/ect_at_school_periods/finish.rb
@@ -79,7 +79,7 @@ module ECTAtSchoolPeriods
 
       return if training_period.finished_on.present? && training_period.finished_on <= finished_on
 
-      TrainingPeriods::Finish.ect_training(author:, training_period:, ect_at_school_period:, finished_on:, record_event: record_event?).finish!
+      TrainingPeriods::Finish.ect_training(author:, training_period:, ect_at_school_period:, finished_on:, record_event: record_event?, leaving_school: true).finish!
     end
 
     def destroy_unstarted_training_periods!
--- a/app/serializers/api/teachers/school_transfer_serializer.rb
+++ b/app/serializers/api/teachers/school_transfer_serializer.rb
@@ -27,7 +27,7 @@ class API::Teachers::SchoolTransferSerializer < Blueprinter::Base
       transfer.status
     end
     field(:created_at) do |(transfer, _teacher, _options)|
-      transfer.leaving_training_period.created_at
+      transfer.leaving_training_period.transfer_initiated_at.presence || transfer.leaving_training_period.created_at
     end
     association :leaving, blueprint: TrainingPeriodSerializer do |(transfer, _teacher, _options)|
       training_period = transfer.leaving_training_period
--- a/app/services/mentor_at_school_periods/finish.rb
+++ b/app/services/mentor_at_school_periods/finish.rb
@@ -76,7 +76,7 @@ private
     destroy_unstarted_training_periods!(period)
 
     period.training_periods.ongoing_on(finished_on).each do |training_period|
-      TrainingPeriods::Finish.mentor_training(training_period:, mentor_at_school_period: period, finished_on:, author:).finish!
+      TrainingPeriods::Finish.mentor_training(training_period:, mentor_at_school_period: period, finished_on:, author:, leaving_school: true).finish!
     end
   end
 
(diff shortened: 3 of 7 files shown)


app/services/ect_at_school_periods/finish.rb
```diff
@@ -79,7 +79,7 @@ def finish_training_period!
 
       return if training_period.finished_on.present? && training_period.finished_on <= finished_on
 
-      TrainingPeriods::Finish.ect_training(author:, training_period:, ect_at_school_period:, finished_on:, record_event: record_event?).finish!
+      TrainingPeriods::Finish.ect_training(author:, training_period:, ect_at_school_period:, finished_on:, record_event: record_event?, leaving_school: true).finish!
```
ethax-ross, line 82: I like the idea of having a field that only gets set for transfers as it would mean we could simplify the query/transfer history builder, but I think it might not be practical due to certain scenarios where a school period may be finished but it isn't going to be classed as a transfer:

- Are they actually transferring out, or are they just finishing their training? (the last school period for a completed participant is not classed as a transfer)
- Can they finish training at a school because they deferred, for example, then resume training at the same school at a later date? (which would negate the transfer)
- Can a school period be re-opened? (which would negate the transfer)

Transfers can be a bit fluid; if an ECT finishes at a school, for example, it will class that as a transfer. If the ECT is later marked as completed then that transfer will disappear (as we no longer expect them to join a new school).

I think we may have to model this as a generic `marked_as_finished_on` or similar field that is set for all training periods when they are finished 🤔

tonyheadford: I thought that deferring or withdrawing a participant only closes the `TrainingPeriod` and that resuming would create a new `TrainingPeriod`.   I asked Connor to confirm what the behaviour is around closing school periods.

> Yeah we should only "finish" a school period when
>   - Induction closed (i.e "complete")
>   - New school tells us an ECT is joining
>   - School tells us ECT is leaving
>   - School tells us Mentor is leaving
>   - New school tells us mentor is joining and mentoring at new school only 

He also didn't think they ever reopen school periods.  We can capture when a school period is closed because the induction is completed too I think, and perhaps pass a value into the school period `Finish` service to indicate a completion event, similar to this PR where we only record the `transfer_initiated_at` when closing a school period.

I _think_ this means that if we actually want to capture a transfer we probably could, but it will be simpler to just record the point the `finished_on` is set if we'd prefer do that instead.

peteryates: > Transfers can be a bit fluid; if an ECT finishes at a school, for example, it will class that as a transfer. If the ECT is later marked as completed then that transfer will disappear (as we no longer expect them to join a new school).

Can we prefix the field with `api_` so it's clear that it doesn't mean 'transfer' in the conventional sense? :sweat_smile:

peteryates: This has a hint of déjà vu about it too - we discussed having a `closure_reason` on training periods at one point - and this would've made sense as one of them (with deferral and withdrawal) being others)

ethax-ross: Yeah there's a ticket for a `reason_for_leaving` that's been kicking about for a while: https://github.com/DFE-Digital/register-ects-project-board/issues/3925

So maybe having both is the best of both worlds; a `closure_reason` and corresponding date it was actioned might be ideal

tonyheadford: Yes that sounds good

PR #3018, opened 2026-06-03
Change status for deferred or withdrawn participants

craigmdavidson:
### Context

Resolves https://github.com.mcas.ms/DFE-Digital/register-ects-project-board/issues/4010

Changes status of participants who are deferred or withdrawn to left or leaving.

Change participant status when a teacher has withdrawn or has been deferred in line with the requirement.

Script to identify and update the `api_updated_at` for these Teachers.

### Guidance to review

Moderately simple change, the script is the most complex bit here. There is a lot of data so we process in batches, try to limit N+1s, and do the final update_at rather than updating each.

#### Results of the script run on snapshot as at 8 June 2026:
```
Teachers affected by participant status changes:
Withdrawn ECT:    from a total of 2095 training_periods
Deferred ECT:     from a total of 178 training_periods
Withdrawn mentor: from a total of 4286 training_periods
Deferred mentor:  from a total of 871 training_periods
....
- ECT withdrawn    — left: 2095, leaving: 0
- ECT deferred     — left: 178, leaving: 0
- Mentor withdrawn — left: 4286, leaving: 0
- Mentor deferred  — left: 871, leaving: 0

Done — a total of 7424 teachers have been effected
```

### DIFF
--- a/app/services/api/training_periods/teacher_status.rb
+++ b/app/services/api/training_periods/teacher_status.rb
@@ -2,7 +2,7 @@ module API::TrainingPeriods
   class TeacherStatus
     attr_reader :training_period, :teacher
 
-    delegate :started_on, :finished_on, :for_ect?, :for_mentor?, to: :training_period, allow_nil: true
+    delegate :started_on, :finished_on, :for_ect?, :for_mentor?, :withdrawn_at, :deferred_at, to: :training_period, allow_nil: true
     delegate :mentor_became_ineligible_for_funding_on, to: :teacher, allow_nil: true
 
     def initialize(latest_training_period:, teacher:)
@@ -11,7 +11,9 @@ module API::TrainingPeriods
     end
 
     def status
-      if teacher_is_ect_and_completed_induction? || teacher_is_mentor_and_completed_training?
+      if withdrawn_at.present? || deferred_at.present?
+        finished_on.future? ? :leaving : :left
+      elsif teacher_is_ect_and_completed_induction? || teacher_is_mentor_and_completed_training?
         :active # longer term we would prefer something like :complete
       elsif finished_on.present?
         finished_on.future? ? :leaving : :left
--- a/db/scripts/affected_api_updated_at_teachers_with_deferred_or_withdrawn.rb
+++ b/db/scripts/affected_api_updated_at_teachers_with_deferred_or_withdrawn.rb
@@ -30,10 +30,31 @@ end
 latest_ect_ids    = Metadata::TeacherLeadProvider.where.not(latest_ect_training_period_id: nil).select(:latest_ect_training_period_id)
 latest_mentor_ids = Metadata::TeacherLeadProvider.where.not(latest_mentor_training_period_id: nil).select(:latest_mentor_training_period_id)
 
-withdrawn_ect_periods    = TrainingPeriod.where(id: latest_ect_ids).where.not(withdrawn_at: nil).includes(ect_at_school_period: :teacher)
-deferred_ect_periods     = TrainingPeriod.where(id: latest_ect_ids).where.not(deferred_at: nil).includes(ect_at_school_period: :teacher)
-withdrawn_mentor_periods = TrainingPeriod.where(id: latest_mentor_ids).where.not(withdrawn_at: nil).includes(mentor_at_school_period: :teacher)
-deferred_mentor_periods  = TrainingPeriod.where(id: latest_mentor_ids).where.not(deferred_at: nil).includes(mentor_at_school_period: :teacher)
+# Mirrors teacher_is_ect_and_completed_induction? — only ECTs who completed induction
+# on or before the training period ended were previously returning :active and are affected.
+completed_ect_induction = <<~SQL
+  (
+    teachers.trs_induction_completed_date IS NOT NULL
+    AND (training_periods.finished_on IS NULL OR teachers.trs_induction_completed_date <= training_periods.finished_on)
+  ) OR EXISTS (
+    SELECT 1 FROM induction_periods
+    WHERE induction_periods.teacher_id = teachers.id
+      AND induction_periods.finished_on IS NOT NULL
+      AND induction_periods.outcome IS NOT NULL
+      AND (training_periods.finished_on IS NULL OR induction_periods.finished_on <= training_periods.finished_on)
+  )
+SQL
+
+# Mirrors teacher_is_mentor_and_completed_training?
+completed_mentor_training = <<~SQL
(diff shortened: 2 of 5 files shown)


app/services/api/training_periods/teacher_status.rb
```diff
@@ -11,7 +11,9 @@ def initialize(latest_training_period:, teacher:)
     end
 
     def status
-      if teacher_is_ect_and_completed_induction? || teacher_is_mentor_and_completed_training?
+      if withdrawn_at.present? || deferred_at.present?
+        finished_on.future? ? :leaving : :left
```
ethax-ross, line 15: Worth extracting this to a `leaving_status` method to avoid duplication?

spec/services/api/training_periods/teacher_status_spec.rb
```diff
@@ -38,6 +38,76 @@
...
+            FactoryBot.create(:induction_period, :pass, teacher:,
+                                                        started_on: training_period.started_on,
+                                                        finished_on: training_period.finished_on)
+          end
+
+          it { is_expected.to eq(:left) }
+        end
+      end
+
+      context "when the teacher is acting as a mentor for this training period" do
```
ethax-ross, line 55: Worth peppering a few as `leaving` to cover that branch?

db/seeds/teacher_histories.rb
```diff
@@ -1279,3 +1279,27 @@ def find_or_create_school_partnership!(school:, delivery_partner:, lead_provider
...
+
+# Deferred + left (finished_on past) — picked up by affected_api_updated_at script
+pat_deferred = Teacher.find_by!(trn: "0000043")
+pat_deferred_at_school = FactoryBot.create(:ect_at_school_period, teacher: pat_deferred, school: abbey_grove_school, started_on: Date.new(2022, 9, 1), finished_on: Date.new(2024, 7, 31))
+FactoryBot.create(:training_period, :for_ect, :school_led, ect_at_school_period: pat_deferred_at_school, school_partnership: nil, expression_of_interest: nil, started_on: Date.new(2022, 9, 1), finished_on: Date.new(2024, 7, 31), deferred_at: Date.new(2023, 3, 15), deferral_reason: "career_break")
+FactoryBot.create(:induction_period, :pass, teacher: pat_deferred, appropriate_body_period: south_yorkshire_studio_hub, started_on: Date.new(2022, 9, 1), finished_on: Date.new(2024, 1, 31), number_of_terms: 6)
+
+# Withdrawn + leaving (finished_on future) — picked up by affected_api_updated_at script
+sal_withdrawn = Teacher.find_by!(trn: "0000044")
+sal_withdrawn_at_school = FactoryBot.create(:ect_at_school_period, teacher: sal_withdrawn, school: abbey_grove_school, started_on: Date.new(2022, 9, 1), finished_on: nil)
```
ethax-ross: Worth an explicit `finished_on` in the future alongside some implicit/`nil`?

db/scripts/affected_api_updated_at_teachers_with_deferred_or_withdrawn.rb
```diff
@@ -0,0 +1,59 @@
...
+      leaving_ids << training_period.teacher_id
+    end
+  end
+
+  [left_ids, leaving_ids]
+end
+
+# Only consider the latest training period per teacher per lead provider,
+# as that's what TeacherStatus is evaluated against in the API.
+latest_ect_ids    = Metadata::TeacherLeadProvider.where.not(latest_ect_training_period_id: nil).select(:latest_ect_training_period_id)
```
ethax-ross, line 30: I think we also want to restrict these to only the ECTs/mentors that have completed training? as these are the ones that will be going from `active` to `leaving/left`? If the participant is withdrawn/deferred and _not_ completed, they will currently be `leaving/left` and we don't want to touch their updated at.

ethax-ross (commented):
Looks good, thanks Craig - just some minor suggestions and a point on the script I think we need to take into account.

It might be worth picking up/including the release note in this PR as well as it will need to go out at the same time: https://github.com.mcas.ms/DFE-Digital/register-ects-project-board/issues/4022

@andycrawshaw can we get sign off on the copy for that release note?

db/scripts/affected_api_updated_at_teachers_with_deferred_or_withdrawn.rb
craigmdavidson: I don't think that's right? 

ECTs/mentors that haven't completed training, and have a `withdrawn_at` or a `deferred_at` but don't have a `finished_on` will currently be `active` - but will now be `leaving` / `left`. Completed training doesn't come into it?

So, agree we want to add an additional filter, but we really only want to exclude those where `finished_on` is not null?

app/services/api/training_periods/teacher_status.rb
craigmdavidson: No. Currently we can look at this method and work out what the return value is; adding a leaving status would obfuscate that. 

The key difference is we can read the current helper methods, without worrying about their internals to understand the method. Adding a `leaving_status` would change that?

Is that reasonable?

db/scripts/affected_api_updated_at_teachers_with_deferred_or_withdrawn.rb
ethax-ross: When you withdraw/defer it closes the training period (so sets `finished_on`) so I don't think there will be any that are deferred/withdrawn where the training period is open? might be worth a quick check on snapshot.

I think the ones that will be changing are those that are currently withdrawn/deferred and completed, as they would be `active` under the old logic and `leaving/left` under the new logic? unless I'm missing something 🤔

craigmdavidson: No you are completely correct. When I look back over the logic, it explicitly states that. Sorry for my misunderstanding.

ethax-ross: No worries, thanks Craig

spec/services/api/training_periods/teacher_status_spec.rb
craigmdavidson: done.

app/services/api/training_periods/teacher_status.rb
ethax-ross: Yeah, fair point 👍

ethax-ross (approved):
Looks good, thanks Craig. We just need the release note before this goes out but it can go to PO review. Would you mind updating the PR description with the new numbers?

PR #3021, opened 2026-06-03
Archive participants registered in error

ericaporter:
Ticket: [here](https://github.com/DFE-Digital/register-ects-project-board/issues/3234)

### Context
Schools occasionally register someone for training in error — for example, registering a teacher as a mentor instead of an ECT, or registering someone who should have been on ITT rather than ECTE. This ticket is the MVP implementation to handle these cases manually.

- Adds a `Teachers::UndoRegistration` service which handles two branches:
  - If the participant has billable or refundable declarations (`eligible`, `payable`, `paid`, `awaiting_clawback`, `clawed_back`), sets `finished_on` to today on the school period, training periods, and mentorship periods so the teacher shows as "left" to the LP
  - If the participant has no billable declarations, deletes those periods and optionally anonymises the teacher record (clearing `trs_first_name`, `trs_last_name`, `corrected_name`, `trn`; setting `trnless: true`) unless an induction period exists
- Adds `anonymisation_reason` (PG enum) and `anonymised_at` (datetime) columns to `teachers`
- Adds a `teacher_registration_undone` event type and `Events::Record.record_undo_registration_event!`
- Adds a `support:undo_registration[period_type,period_id]` rake task as the MVP entry point, with a teacher ID confirmation prompt to prevent accidental runs
- Adds `anonymisation_reason` and `anonymised_at` to analytics config
- API IDs are always preserved regardless of branch
- The service is named `UndoRegistration` rather than `Archive` to better reflect the intent — the "archiving" (anonymisation) only occurs in one specific branch

### Guidance to review
- The rake task takes `ECTAtSchoolPeriod` or `MentorAtSchoolPeriod` as `period_type` and the period's database ID
- The event heading uses the teacher ID rather than name to avoid leaking PII in the audit log after anonymisation
- `trnless: true` is set on anonymisation to reuse the existing TRN-less teacher pattern (AC1)
- Seed data is not added as the rake task can be run against dev/staging directly

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -60,6 +60,7 @@ class Event < ApplicationRecord
     teacher_schedule_assigned_to_training_period
     teacher_defers_training_period
     teacher_resumes_training_period
+    teacher_archived
     teacher_withdraws_training_period
     teacher_changes_schedule_training_period
     teacher_declaration_voided
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -60,7 +60,7 @@ class Event < ApplicationRecord
     teacher_schedule_assigned_to_training_period
     teacher_defers_training_period
     teacher_resumes_training_period
-    teacher_archived
+    teacher_registration_undone
     teacher_withdraws_training_period
     teacher_changes_schedule_training_period
     teacher_declaration_voided
--- a/app/models/teacher.rb
+++ b/app/models/teacher.rb
@@ -18,6 +18,9 @@ class Teacher < ApplicationRecord
     completed_during_early_roll_out: "completed_during_early_roll_out",
     started_not_completed: "started_not_completed",
   }
+  enum :archived_reason, {
+    registered_in_error: "registered_in_error",
+  }
 
   # Associations
   has_many :ect_at_school_periods, inverse_of: :teacher
@@ -105,6 +108,9 @@ class Teacher < ApplicationRecord
   validates :api_mentor_training_record_id, uniqueness: { case_sensitive: false, message: "API mentor training record id already exists for another teacher" }
   validates :ect_first_became_eligible_for_training_at, immutable_once_set: true
   validates :mentor_first_became_eligible_for_training_at, immutable_once_set: true
+
+  validates :archived_reason, presence: true, if: -> { archived_at.present? }
+  validates :archived_at, presence: true, if: -> { archived_reason.present? }
 
   # Scopes
   scope :search, ->(query_string) {
--- a/app/models/teacher.rb
+++ b/app/models/teacher.rb
@@ -18,7 +18,7 @@ class Teacher < ApplicationRecord
     completed_during_early_roll_out: "completed_during_early_roll_out",
     started_not_completed: "started_not_completed",
   }
-  enum :archived_reason, {
+  enum :anonymisation_reason, {
     registered_in_error: "registered_in_error",
   }
 
@@ -109,8 +109,8 @@ class Teacher < ApplicationRecord
   validates :ect_first_became_eligible_for_training_at, immutable_once_set: true
   validates :mentor_first_became_eligible_for_training_at, immutable_once_set: true
 
-  validates :archived_reason, presence: true, if: -> { archived_at.present? }
-  validates :archived_at, presence: true, if: -> { archived_reason.present? }
+  validates :anonymisation_reason, presence: true, if: -> { anonymised_at.present? }
+  validates :anonymised_at, presence: true, if: -> { anonymisation_reason.present? }
 
   # Scopes
   scope :search, ->(query_string) {
--- a/app/services/events/record.rb
+++ b/app/services/events/record.rb
@@ -142,10 +142,10 @@ module Events
 
     # Teacher Status Events
 
-    def self.record_teacher_archived_event!(author:, teacher:, reason:, happened_at: Time.zone.now)
-      event_type = :teacher_archived
-      heading = "Teacher was archived"
-      body = "Teacher was archived. Reason: #{reason}"
+    def self.record_undo_registration_event!(author:, teacher:, reason:, happened_at: Time.zone.now)
+      event_type = :teacher_registration_undone
+      heading = "Teacher #{teacher.id} registration was undone"
+      body = "Teacher registration was undone. Reason: #{reason}"
 
       new(event_type:, author:, teacher:, heading:, body:, happened_at:).record_event!
     end
--- a/app/services/events/record.rb
+++ b/app/services/events/record.rb
@@ -144,8 +144,9 @@ module Events
 
     def self.record_teacher_archived_event!(author:, teacher:, reason:, happened_at: Time.zone.now)
       event_type = :teacher_archived
-      heading = "#{teacher_name} was archived"
+      heading = "Teacher was archived"
       body = "Teacher was archived. Reason: #{reason}"
+
       new(event_type:, author:, teacher:, heading:, body:, happened_at:).record_event!
     end
 
--- a/app/services/events/record.rb
+++ b/app/services/events/record.rb
@@ -144,7 +144,6 @@ module Events
 
     def self.record_teacher_archived_event!(author:, teacher:, reason:, happened_at: Time.zone.now)
       event_type = :teacher_archived
-      teacher_name = Teachers::Name.new(teacher).full_name
       heading = "#{teacher_name} was archived"
       body = "Teacher was archived. Reason: #{reason}"
       new(event_type:, author:, teacher:, heading:, body:, happened_at:).record_event!
--- a/app/services/events/record.rb
+++ b/app/services/events/record.rb
@@ -142,6 +142,14 @@ module Events
 
     # Teacher Status Events
 
+    def self.record_teacher_archived_event!(author:, teacher:, reason:, happened_at: Time.zone.now)
+      event_type = :teacher_archived
+      teacher_name = Teachers::Name.new(teacher).full_name
+      heading = "#{teacher_name} was archived"
+      body = "Teacher was archived. Reason: #{reason}"
+      new(event_type:, author:, teacher:, heading:, body:, happened_at:).record_event!
+    end
+
     def self.record_teacher_passes_induction_event!(author:, appropriate_body_period:, induction_period:, ect_at_school_period:, mentorship_period:, training_period:, teacher:, body: nil, zendesk_ticket_id: nil)
       fail(NoInductionPeriod) unless induction_period
 
--- a/app/services/teachers/archive.rb
+++ b/app/services/teachers/archive.rb
@@ -13,12 +13,12 @@ module Teachers
       ActiveRecord::Base.transaction do
         if billable_declarations?
           finish_periods!
-          record_archived_event!
         else
-          record_archived_event!
           delete_periods!
           anonymise_teacher! unless induction_period_exists?
         end
+
+        record_archived_event!
       end
 
       API::Teachers::Query.new.teacher_by_id(teacher.id)
--- a/app/services/teachers/archive.rb
+++ b/app/services/teachers/archive.rb
@@ -1,0 +1,81 @@
+module Teachers
+  class Archive
+    attr_reader :author, :period, :reason, :teacher
+
+    def initialize(author:, period:, reason:)
+      @author = author
+      @period = period
+      @reason = reason
+      @teacher = period.teacher
+    end
+
+    def archive
+      ActiveRecord::Base.transaction do
+        if billable_declarations?
+          finish_periods!
+          record_archived_event!
+        else
+          record_archived_event!
+          delete_periods!
+          anonymise_teacher! unless induction_period_exists?
+        end
+      end
+
+      API::Teachers::Query.new.teacher_by_id(teacher.id)
+    end
+
+  private
+
+    def billable_declarations?
+      declarations.billable.exists? || declarations.refundable.exists?
+    end
+
+    def declarations
+      Declaration.where(training_period: training_periods)
+    end
+
+    def training_periods
+      period.training_periods
+    end
+
+    def mentorship_periods
+      period.mentorship_periods
+    end
+
+    def finish_periods!
+      mentorship_periods.find_each { |mp| mp.update!(finished_on: Time.zone.today) }
+      training_periods.find_each { |tp| tp.update!(finished_on: Time.zone.today) }
+      period.update!(finished_on: Time.zone.today)
+    end
+
+    def delete_periods!
+      mentorship_periods.destroy_all
+      training_periods.destroy_all
+      period.destroy!
+    end
+
+    def anonymise_teacher!
+      teacher.update!(
+        trs_first_name: nil,
+        trs_last_name: nil,
+        corrected_name: nil,
+        trn: nil,
+        trnless: true,
+        archived_reason: reason,
+        archived_at: Time.zone.now
+      )
+    end
+
+    def induction_period_exists?
+      teacher.induction_periods.exists?
+    end
+
+    def record_archived_event!
+      Events::Record.record_teacher_archived_event!(
+        author:,
+        teacher:,
+        reason:
+      )
+    end
+  end
+end
--- a/app/services/teachers/archive.rb
+++ b/app/services/teachers/archive.rb
@@ -2,6 +2,8 @@ module Teachers
   class Archive
     attr_reader :author, :period, :reason, :teacher
 
+    delegate :training_periods, :mentorship_periods, to: :period
+
     def initialize(author:, period:, reason:)
       @author = author
       @period = period
@@ -32,14 +34,6 @@ module Teachers
 
     def declarations
       Declaration.where(training_period: training_periods)
-    end
-
-    def training_periods
-      period.training_periods
-    end
-
-    def mentorship_periods
-      period.mentorship_periods
     end
 
     def finish_periods!
--- a/app/services/teachers/undo_registration.rb
+++ b/app/services/teachers/undo_registration.rb
@@ -13,11 +13,11 @@ module Teachers
 
     def undo!
       ActiveRecord::Base.transaction do
-        if billable_or_refundable_declarations?
+        if billable_or_refundable_declarations_exist?
           finish_periods!
         else
           delete_periods!
-          anonymise_teacher! unless induction_period_exists?
+          anonymise_teacher! if anonymise_teacher?
         end
 
         record_undo_registration_event!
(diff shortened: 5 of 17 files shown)


app/services/teachers/archive.rb
```diff
@@ -0,0 +1,81 @@
...
+      @period = period
+      @reason = reason
+      @teacher = period.teacher
+    end
+
+    def archive
+      ActiveRecord::Base.transaction do
+        if billable_declarations?
+          finish_periods!
+          record_archived_event!
```
thenapking: Since this happens in both branches it could be outside the if statement I suppose.  Is there any reason it's done in a different order?

spec/services/teachers/archive_spec.rb
```diff
@@ -0,0 +1,413 @@
...
+        end
+
+        it "sets finished_on to today on relevant mentorship periods" do
+          freeze_time do
+            archive
+            expect(mentorship_period.reload.finished_on).to eq(Time.zone.today)
+          end
+        end
+
+        it "does not delete the relevant periods" do
```
thenapking, line 46: We could make this test suite a bit more dry with a shared context, since there are basically two outcomes - the periods are finished or destroyed.  The latter could have two sub-contexts to cover the existence of an induction period.

lib/tasks/archive_teacher.rake
```diff
@@ -0,0 +1,23 @@
...
+  task :archive_teacher, %i[period_type period_id] => :environment do |_task, args|
+    period_types = {
+      "ECTAtSchoolPeriod" => ECTAtSchoolPeriod,
+      "MentorAtSchoolPeriod" => MentorAtSchoolPeriod
+    }.freeze
+
+    period_class = period_types.fetch(args[:period_type]) do
+      raise ArgumentError, "period_type must be ECTAtSchoolPeriod or MentorAtSchoolPeriod"
+    end
+
```
thenapking: I suppose we don't know how long this task might be in use for.  Is it worth adding a warning with a prompt in the case where it will permanently delete some data?

app/services/teachers/archive.rb
ericaporter: I spotted that this morning and was just pushing the update...

app/services/teachers/archive.rb
```diff
@@ -0,0 +1,81 @@
...
+
+    def billable_declarations?
+      declarations.billable.exists? || declarations.refundable.exists?
+    end
+
+    def declarations
+      Declaration.where(training_period: training_periods)
+    end
+
+    def training_periods
```
thenapking:
```suggestion
    delegate :training_periods, :mentorship_periods, to: :period
```

thenapking (approved):
This looks really useful.  Creating a service for this was a great idea because it means this could be given to administrators if it works reliably and it becomes a regular task.

lib/tasks/archive_teacher.rake
ericaporter: Yeah good point, I'll add in a prompt

lib/tasks/archive_teacher.rake
cpjmcquillan, line 1: I wonder if it's worth us doing a light piece of work to add this functionality to the admin console?

We might not want it to be available for everyone, but I feel like we should get in the habit of building things that don't need to be actioned via the console/rake tasks.

What do you think?

app/services/teachers/archive.rb
```diff
@@ -0,0 +1,81 @@
...
+
+    def mentorship_periods
+      period.mentorship_periods
+    end
+
+    def finish_periods!
+      mentorship_periods.find_each { |mp| mp.update!(finished_on: Time.zone.today) }
+      training_periods.find_each { |tp| tp.update!(finished_on: Time.zone.today) }
+      period.update!(finished_on: Time.zone.today)
+    end
```
cpjmcquillan, line 45: Should we use the `Finish` services here instead?

app/services/events/record.rb
```diff
@@ -142,6 +142,14 @@ def self.record_induction_period_deleted_event!(author:, modifications:, teacher
 
     # Teacher Status Events
 
+    def self.record_teacher_archived_event!(author:, teacher:, reason:, happened_at: Time.zone.now)
+      event_type = :teacher_archived
+      heading = "Teacher was archived"
+      body = "Teacher was archived. Reason: #{reason}"
+
+      new(event_type:, author:, teacher:, heading:, body:, happened_at:).record_event!
+    end
```
cpjmcquillan: Should we keep a record of the TRN? 💭

ericaporter: @cpjmcquillan I originally had trn here but then I thought since it's PII maybe we shouldn't

lib/tasks/archive_teacher.rake
ericaporter: @cpjmcquillan I know where you are coming from. Do you think we could either have a ticket for this? - happy to pick it up immediately after (this ticket is for MVP implementation recognising we will shift to a different implementation later, albeit written yonks ago) or can this be okayed by Emily and the ticket updated to include this? - then I'm happy to include it in this PR!

app/services/teachers/archive.rb
ericaporter: So I checked the existing finish services...

There are individual services for mentorship and training periods, but they are not all direct replacements. `MentorshipPeriods::Finish` supports `record_event: false`, and `TrainingPeriods::Finish.ect_training` does too. However `TrainingPeriods::Finish.mentor_training` does not currently support `record_event: false`.

Slightly broader is: `ECTAtSchoolPeriods::Finish` and `MentorAtSchoolPeriods::Finish` but I think these do too much and aren't probably aren't suitable.

spec/services/teachers/archive_spec.rb
ericaporter: I've dried these up - I haven't used shared contexts - I find they make tests less easy to read/update and couple some tests up unnecessarily making refactoring a pain - I know we already have some in this style in the project 🫤

app/services/events/record.rb
cpjmcquillan: Hmm. I'll defer to @peteryates.

I suppose we still have access to the underlying record.

I think I'm mostly concerned about the possibility of us archiving someone by mistake. If we don't keep a solid audit trail, we won't be able to undo it. And it feels like something we ought to be able to undo (within a time period).

app/services/teachers/archive.rb
```diff
@@ -0,0 +1,75 @@
+module Teachers
+  class Archive
+    attr_reader :author, :period, :reason, :teacher
+
+    delegate :training_periods, :mentorship_periods, to: :period
+
+    def initialize(author:, period:, reason:)
```
cpjmcquillan: Is it worth passing in the teacher?

At the minute, it seems like you could archive a teacher based on an `ECTAtSchoolPeriod` record and end up with some orphaned `MentorAtSchoolPeriod` records (and vice versa) 🤔.

If the entity we want to purge is the `Teacher`, I wonder if it's worth starting at the top of the tree and working our way down?

What do you think?

app/services/teachers/archive.rb
cpjmcquillan: Hmm ok.

I still feel like there ought to be a single way to "finish" a given type of period record.

Maybe those services have grown too much and now do too many things, though.

Feel free to park this for now. 

At a minimum, we should be able to call `.finish!` on each record which should simplify this a little.

app/services/teachers/archive.rb
```diff
@@ -0,0 +1,75 @@
...
+
+    def declarations
+      Declaration.where(training_period: training_periods)
+    end
+
+    def finish_periods!
+      mentorship_periods.find_each { |mp| mp.update!(finished_on: Time.zone.today) }
+      training_periods.find_each { |tp| tp.update!(finished_on: Time.zone.today) }
+      period.update!(finished_on: Time.zone.today)
+    end
```
cpjmcquillan, line 45:
```suggestion
    def finish_periods!
      mentorship_periods.find_each(&:finish!)
      training_periods.find_each(&:finish!)
      period.finish!
    end
```

lib/tasks/archive_teacher.rake
```diff
@@ -0,0 +1,32 @@
...
+    period = period_class.find(args[:period_id])
+    teacher = period.teacher
+
+    puts "WARNING: This task archives a teacher registration made in error."
+    puts "  - If billable/refundable declarations exist: periods will be closed with today's date"
+    puts "  - If no billable/refundable declarations: periods will be permanently deleted and the teacher record may be anonymised"
+    puts "Teacher ID: #{teacher.id} | Period: #{period.class.name} #{period.id}"
+    print "Type ARCHIVE to confirm: "
+
+    abort "Archive cancelled." unless $stdin.gets&.strip == "ARCHIVE"
```
cpjmcquillan: I think this is a good mitigation, but I still feel apprehensive about us doing this via a rake task.

Perhaps this is fine for now though and I'm just being overly cautious!

lib/tasks/archive_teacher.rake
cpjmcquillan: I'll chat with Emily, but otherwise happy for this to stay as a task so we can unblock archiving that participant.

app/services/teachers/archive.rb
cpjmcquillan: If we do stick with `period`, could we use `at_school_period` just to distinguish between training and mentorship periods.

I feel like that convention is something we're leaning into a little more now, especially with the `AtSchoolPeriod` concern.

What do you think?

app/services/teachers/archive.rb
```diff
@@ -0,0 +1,75 @@
...
+      teacher.update!(
+        trs_first_name: nil,
+        trs_last_name: nil,
+        corrected_name: nil,
+        trn: nil,
+        trnless: true,
+        archived_reason: reason,
+        archived_at: Time.zone.now
+      )
+    end
```
cpjmcquillan, line 63: If we're going to anonymise the teacher, could we delete them instead? 💭

lib/tasks/archive_teacher.rake
```diff
@@ -0,0 +1,32 @@
...
+      raise ArgumentError, "period_type must be ECTAtSchoolPeriod or MentorAtSchoolPeriod"
+    end
+
+    period = period_class.find(args[:period_id])
+    teacher = period.teacher
+
+    puts "WARNING: This task archives a teacher registration made in error."
+    puts "  - If billable/refundable declarations exist: periods will be closed with today's date"
+    puts "  - If no billable/refundable declarations: periods will be permanently deleted and the teacher record may be anonymised"
+    puts "Teacher ID: #{teacher.id} | Period: #{period.class.name} #{period.id}"
```
cpjmcquillan: Should we output the TRN or name here? Just to make it a bit more obvious if we're about to archive the wrong person 🤔

app/models/teacher.rb
```diff
@@ -106,6 +109,9 @@ class Teacher < ApplicationRecord
   validates :ect_first_became_eligible_for_training_at, immutable_once_set: true
   validates :mentor_first_became_eligible_for_training_at, immutable_once_set: true
 
+  validates :archived_reason, presence: true, if: -> { archived_at.present? }
+  validates :archived_at, presence: true, if: -> { archived_reason.present? }
```
cpjmcquillan: Should we validate the teacher is `trnless` if these are present?

spec/services/teachers/archive_spec.rb
```diff
@@ -0,0 +1,353 @@
...
+        it "does not set archived_at" do
+          archive
+          expect(ect_at_school_period.teacher.reload.archived_at).to be_nil
+        end
+
+        it "records an archive event" do
+          expect(Events::Record).to receive(:record_teacher_archived_event!)
+            .with(author:, teacher: ect_at_school_period.teacher, reason: :registered_in_error)
+          archive
+        end
```
cpjmcquillan: Should we add these examples to the context blocks for payable and paid declaration scenarios too?

cpjmcquillan (commented):
I've added a few comments, but I think this is pretty much there 🙌🏼

app/services/events/record.rb
peteryates: I think it's fine to put PII in the events.

But, I guess I'm not really clear on what archiving means.

If someone's archived will they be registrable by other schools/appropriate bodies?

Like I said in my other comment, I wonder if it's worth teasing apart the idea or archiving vs undoing a registration.

lib/tasks/archive_teacher.rake
thenapking: Or we could have the prompt be re-entering the TRN to confirm?

app/models/teacher.rb
ericaporter: Yeah that makes sense, I'll add a validation for that

app/services/teachers/archive.rb
ericaporter: So it was intentional not to pass teacher since the task is to only remove or finish the one that was identified as wrong (not any other valid ones)

Happy to update the naming to `at_school_period` though - that's def clearer

lib/tasks/archive_teacher.rake
ericaporter: Maybe if we do some renaming of this task and move away from "archive" we can choose something more thought provoking 😆

lib/tasks/archive_teacher.rake
ericaporter: I have updated so we re-enter the ID to confirm... not all teachers have TRNs

app/services/teachers/archive.rb
ericaporter: The ticket says so we would still have a record of their API ID if we ever get queries in future about the person - but appreciate the req might have changed

app/services/teachers/archive.rb
cpjmcquillan: I'm not sure I really understand.

At the minute, it seems as though this could be possible:

- "Undo" registration for teacher using `ECTAtSchoolPeriod` and ID
- There are no billable or refundable declarations, so we delete periods associated with the `ECTAtSchoolPeriod`
- There is no induction period associated to the teacher, so we anonymise the teacher
- **But** there were some `MentorAtSchoolPeriod` records associated to the teacher
- Now there are mentor records with a "blank" teacher record

This doesn't feel correct. Is there something I'm missing here? 🤔

ericaporter: I think you mean if a teacher has both an ECT and a mentor registration - undoing the erroneous one will leave an orphaned but valid record?

I think we probably need to add in another check to anonymise only if they don't have any other periods to guard against this

app/models/teacher.rb
```diff
@@ -18,6 +18,9 @@ class Teacher < ApplicationRecord
     completed_during_early_roll_out: "completed_during_early_roll_out",
     started_not_completed: "started_not_completed",
   }
+  enum :anonymisation_reason, {
+    registered_in_error: "registered_in_error",
+  }
```
cpjmcquillan, line 23: This naming is much more explicit and clearer 👌🏼

app/models/event.rb
```diff
@@ -59,6 +59,7 @@ class Event < ApplicationRecord
     teacher_schedule_assigned_to_training_period
     teacher_defers_training_period
     teacher_resumes_training_period
+    teacher_registration_undone
```
cpjmcquillan, line 63: I think it's helpful this mirrors the naming of the service object, but I wonder if `teacher_registration_reverted` and `Teachers::RevertRegistration` would be a small improvement 💭.

I appreciate this might be splitting hairs though!

spec/models/teacher_spec.rb
```diff
@@ -381,6 +381,53 @@
...
+
+      context "when anonymised_at is present but anonymisation_reason is missing" do
+        subject { FactoryBot.build(:teacher, anonymisation_reason: nil, anonymised_at: Time.zone.now) }
+
+        it { is_expected.to be_invalid }
+
+        it "has a validation error on anonymisation_reason" do
+          subject.valid?
+          expect(subject.errors.messages[:anonymisation_reason]).to be_present
+        end
```
cpjmcquillan: We could lean on the `have_error` matcher here to trim this down to a single test.

What do you think?

spec/models/teacher_spec.rb
```diff
@@ -381,6 +381,53 @@
...
+
+      context "when anonymisation_reason is present but anonymised_at is missing" do
+        subject { FactoryBot.build(:teacher, anonymisation_reason: "registered_in_error", anonymised_at: nil) }
+
+        it { is_expected.to be_invalid }
+
+        it "has a validation error on anonymised_at" do
+          subject.valid?
+          expect(subject.errors.messages[:anonymised_at]).to be_present
+        end
```
cpjmcquillan: Same here ✂️

app/services/teachers/undo_registration.rb
```diff
@@ -0,0 +1,75 @@
...
+
+  private
+
+    def billable_or_refundable_declarations?
+      declarations.billable.exists? || declarations.refundable.exists?
+    end
+
+    def declarations
+      Declaration.where(training_period: training_periods)
+    end
```
cpjmcquillan: If we wanted to use a single query, I think we could do something like this

```ruby
def billable_or_refundable_declarations
  Declaration.where(training_period: training_periods)
    .merge(Declaration.billable.or(Declaration.refundable))
    .exists?
end
```

app/services/teachers/undo_registration.rb
```diff
@@ -0,0 +1,75 @@
...
+
+    def finish_periods!
+      mentorship_periods.find_each(&:finish!)
+      training_periods.find_each(&:finish!)
+      at_school_period.finish!
+    end
+
+    def delete_periods!
+      mentorship_periods.destroy_all
+      training_periods.destroy_all
```
cpjmcquillan: It's a little bit less performant, but I wonder if we should call `:destroy!` to avoid silent failures.

```ruby
mentorship_periods.each(&:destroy!)
training_periods.each(&:destroy!)
```

What do you think?

cpjmcquillan (approved):
Couple of small comments, plus the change we discussed to restrict undo-ing when a teacher has other `at_school_period` records. 

But other than that this looks good. Nice one!

app/models/event.rb
ericaporter: We discussed leaving this one as is

app/services/teachers/undo_registration.rb
```diff
@@ -0,0 +1,75 @@
...
+      mentorship_periods.find_each(&:finish!)
+      training_periods.find_each(&:finish!)
+      at_school_period.finish!
+    end
+
+    def delete_periods!
+      mentorship_periods.each(&:destroy!)
+      training_periods.each(&:destroy!)
+      at_school_period.destroy!
+    end
```
cpjmcquillan, line 51: Do we want to use `find_each` here too for consistency? 

I don't think it will really make any difference though, I doubt we're going to be running this on teachers with more than 1000 (default batch size) periods.

PR #3024, opened 2026-06-03
[3173] Blazer queries for RECT comms to schools

rwrrll:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/3173

Builds and syncs the Blazer queries requested in the ticket.

If it's a new database (e.g. a review app) then this will be automatically triggered by `db:prepare` seeds step. Otherwise (e.g. production) the rake task included here can be used to upsert the queries into Blazer.

I also refactored the token generation a bit to better encapsulate it.

### Guidance to review

On a review app, going to `/admin/blazer` should list the 4 queries here.

The queries should produce the expected results as per the ticket requirements.


app/services/schools/reminder_email_opt_out_token.rb
```diff
@@ -1,46 +1,43 @@
...
 
     PURPOSE = "school-reminder-email-opt-out"
 
-    def self.generate_for(school_id:)   = new(school_id:).generate
-    def self.valid?(school_id:, token:) = new(school_id:).valid?(token)
+    class << self
+      def generate_for(school_id:)   = new(school_id:).generate
+      def valid?(school_id:, token:) = new(school_id:).valid?(token)
+
+      def token_sql(school_id_sql)
```
thenapking: The name `school_id_sql` is a little confusing, and I was really surprised to see that "s.id" is passed in as the argument to this function.  I think it might be better to have an argument called `sql_table_alias`, so either "s" or "schools" is passed in to the function.  Then this function simply appends ".id" to the `sql_table_alias`, and uses that.

db/seeds/blazer_queries/school_comms.rb
```diff
@@ -0,0 +1,204 @@
...
+
+      def recipient_note
+        "Falls back to the GIAS contact email and a '#{HEADTEACHER_NAME_PLACEHOLDER}' " \
+          "placeholder name when a school has no registered SIT."
+      end
+
+      def opt_out_url_sql
+        token_sql = Schools::ReminderEmailOptOutToken.token_sql("s.id")
+
+        "'#{opt_out_base_url}?school_id=' || s.id || '&token=' || #{token_sql}"
```
thenapking, line 200: In the context of a Ruby function with strings, it's easy to misunderstand the postgres concat operator `||`.  I think `"CONCAT('#{opt_out_base_url}?school_id=', s.id, '&token=', #{token_sql})"` is marginally easier to read, especially since this also uses ruby string interpolation.  In fact it might be easier to separate it up like this:
```
params = [
    "CONCAT('school_id=', s.id)",
    "CONCAT('token=', #{token_sql})"
  ].join(", '&', ")

  "CONCAT('#{opt_out_base_url}?', #{params})"
```
Or even just make each param its own method?

spec/seeds/blazer_queries/school_comms_spec.rb
```diff
@@ -0,0 +1,218 @@
...
+    it "defines the four school comms queries" do
+      expect(definitions.pluck(:name)).to contain_exactly(
+        "Comms: Registrations opening (June)",
+        "Comms: Start of term reminder (September)",
+        "Comms: Start of term reminder (January / April)",
+        "Comms: Partnership created but no ECTs or mentors follow-up"
+      )
+    end
+
+    it "bakes the opt-out host, path and signing secret into every statement" do
```
thenapking, line 42: I'm going to raise this for discussion, although I think the impact is likely low.  It's probably not a good idea to be baking in secrets to a query.  Obviously this is just for opting out of emails, and the only people who have access to these queries are admins.  Another way of doing this would be to have some sort of task that generates the tokens and puts them in a table this query can access.  I think that also has flaws as well, as the tokens are just as visible to some one with DB access.
What happens when the secret gets rotated? Do we just delete the queries and re-run this task?

db/seeds/blazer_queries/school_comms.rb
```diff
@@ -0,0 +1,204 @@
...
+              s.urn,
+              gs.name AS school_name,
+              #{recipient_name_sql},
+              #{recipient_email_sql},
+              #{opt_out_url_sql} AS reminder_email_opt_out_url
+            FROM schools s
+            INNER JOIN gias_schools gs ON gs.urn = s.urn
+            WHERE #{not_a_childrens_centre_sql}
+              AND #{not_opted_out_sql}
+              AND EXISTS (
```
thenapking: Small thing but we could make this subquery a method, with an name that explains precisely what it's doing. I think this is for finding whether the school has a partnership in the current contract period. I know there's nothing dynamic here and it's not used repeatedly but it might help with readability.

thenapking (commented):
Looks like this will be very useful. Left a couple of comments on things that are a little hard to read.  I've taken a look at the results on the review app and it does appear to link to the correct school when you logout as admin and follow the link.

db/seeds/blazer_queries/school_comms.rb
```diff
@@ -0,0 +1,204 @@
...
+
+      def not_opted_out_sql
+        "(s.opted_out_of_reminder_emails_until IS NULL " \
+          "OR s.opted_out_of_reminder_emails_until < CURRENT_DATE)"
+      end
+
+      def recipient_name_sql
+        "CASE WHEN s.induction_tutor_email IS NOT NULL " \
+          "THEN s.induction_tutor_name " \
+          "ELSE #{quote(HEADTEACHER_NAME_PLACEHOLDER)} END AS recipient_name"
```
ericaporter, line 184: Small thing (& might be on purpose) - the fallback uses the placeholder "colleague" with the GIAS primary/secondary contact email. The ticket says to use the GIAS headteacher first name, last name and email when there is no registered SIT. Is the placeholder intentional because we do not have reliable headteacher name fields, or should the query expose those fields separately and then also have a fallback to the placeholder colleague?

db/seeds/blazer_queries/school_comms.rb
```diff
@@ -0,0 +1,204 @@
...
+                INNER JOIN lead_provider_delivery_partnerships lpdp
+                  ON lpdp.id = sp.lead_provider_delivery_partnership_id
+                INNER JOIN active_lead_providers alp
+                  ON alp.id = lpdp.active_lead_provider_id
+                INNER JOIN contract_periods cp ON cp.year = alp.contract_period_year
+                WHERE sp.school_id = s.id
+                  AND cp.range @> CURRENT_DATE
+              )
+              AND NOT EXISTS (
+                SELECT 1 FROM ect_at_school_periods e WHERE e.school_id = s.id
```
ericaporter: Is this excluding all ect_at_school_periods (and below the mentor_at_school_periods) rather than just those with no ects / mentors in the current contract? I'm wondering if they need to be scoped

app/services/schools/reminder_email_opt_out_token.rb
rwrrll: Yeah you’re right, it’s a bit opaque at the call site, `token_sql("s.id")` doesn't obviously read as "the school id".

I'm a little hesitant on sql_table_alias though, it assumes the argument is always a bare table alias, and that the id column is always named id. Right now it takes an arbitrary SQL expression that evaluates to the id, the spec leans on this by passing a literal '42'. With an alias arg that test would have to become 42.id, which isn't valid SQL. Also the `(#{school_id_sql})::text` wrapping in the body only makes sense because we're expecting an expression rather than an identifier.

As a middle ground, I made it a keyword arg so the call site documents itself. This is actually in line with the other class methods that take a `school_id` kwarg too. I think helps address the surprise without losing the flexibility!

db/seeds/blazer_queries/school_comms.rb
rwrrll: I do agree with your point that `||` might read oddly inside an interpolated Ruby string. I had a look at the alternatives though, and think I'd lean towards keeping it:

- No nesting, you read the URL template left-to-right, which gives the clearest possible view of "what shape is this URL".
- `||` and `CONCAT` handle NULL differently: if any operand is NULL, `||` causes the whole result to be NULL but `CONCAT` silently treats NULLs as empty strings. So if someone borked the school ID parameter, `||` would produce a NULL URL, rather than `CONCAT` rendering a URL that might look legit but not work.
- I think the join thing is mentally harder to parse versus what `||` does in PG
- Method per param spreads out what’s happening

spec/seeds/blazer_queries/school_comms_spec.rb
rwrrll: Definitely a valid concern, but has been considered. The purpose of this secret is isolated to opting schools out of emails, so if you have access to this secret, that’s all you can do with it. There’s no way of giving users a Blazer query that includes opt out links without giving them the ability to opt schools out of emails, so exposing the signing secret doesn’t introduce any additional risk over exposing the opt-out links themselves.

Storing a per-school token in the database would still expose it in this Blazer query, with the added downside of having an extra column / token to manage, whereas this is just one ENV secret. If it ever does need to be rotated, just re-run the rake task, no need to delete the queries first.

db/seeds/blazer_queries/school_comms.rb
rwrrll: Yep agreed, done!

db/seeds/blazer_queries/school_comms.rb
rwrrll: Yes exactly right, there’s no reliable headteacher name field and the school contact email isn’t necessarily the headteacher. I forgot to write this up in the ticket comments which you’ve reminded me I need to do!

db/seeds/blazer_queries/school_comms.rb
rwrrll: Verrrrry good spot! I confirmed with Emily and you're right, it should be "has at least one partnership this period but no participants this period". Adjusted, thanks very much!

db/seeds/blazer_queries/school_comms.rb
thenapking: I take your point about `CONCAT` handling `NULL` differently. But something like this is still easier to read imo:
```
def opt_out_url_sql
  "'#{opt_out_base_url}?' || #{url_params}"
end

def token_sql
  Schools::ReminderEmailOptOutToken.token_sql(school_id_sql: "s.id")
end

def url_params
  query_params = [
    "'school_id=' || s.id",
    "'token=' || #{token_sql}"
  ]

  query_params.reduce { |sql, param| "#{sql}, '&', #{param}" }
end
```

It just really makes it clear that we're joining a url and params, which I didn't clock on first reading this because I was confused by the syntax.  It does spread it out a bit but only in order to add clarity.  If you find `.join(", '&', ")` unclear because of its confusion in this context there's always reduce, but maybe that's too much.  
It's not a blocker though.

app/services/schools/reminder_email_opt_out_token.rb
thenapking: OK. I can see your concerns here and casting this to text does make it a bit clearer.  Is still think `school_id_sql` is not super clear. `school_id_sql_expression` might be clearer, or just a comment to explain where this is used and what it is?

thenapking (approved):
Looks good now.

PR #3026, opened 2026-06-04
[3940] related fixes and housekeeping

peterdavidhamilton:
### Context

Extracted out-of-scope edits from `3940-schedules-crud`

This commit aligns with the style and pattern of the schedules and milestones
#3004


ethax-ross (approved):

ethax-ross (approved):
Thanks for making these all consistent PDH. I feel like we're trending towards using exceptions for control flow though 🤔 maybe something to revisit when we look at refactoring the journey in general.

PR #3028, opened 2026-06-04
Care less about 31st May in tests

cpjmcquillan:
Before, lots of our tests were focused on 31st May (and 2025 in particular).

In #2995 we fixed a bunch of tests that failed on `main` once we moved from the 2025 to the 2026 contract period.

This reworks a handful of tests that didn't break, but had a similar focus on the 2025 contract period and/or the 31st May.

Both of those are arbitrary, so this removes some of that focus.

Where tests are intended to focus on the end of a contract period, then they do that explicitly. They don't assume that 31st May is always the end of the contract period.

Also, tests focus less on the 2025 contract period and more on the "current" contract period.

This should make our test suite more robust and ensure they stay relevant over time.

### DIFF
--- a/spec/wizards/schools/register_ect_wizard/wizard_spec.rb
+++ b/spec/wizards/schools/register_ect_wizard/wizard_spec.rb
@@ -1,5 +1,5 @@
 RSpec.describe Schools::RegisterECTWizard::Wizard do
-  let(:school) { FactoryBot.create(:school, :state_funded) }
+  let(:school) { FactoryBot.create(:school, :state_funded, create_contract_period: false) }
   let(:user) { FactoryBot.create(:user) }
   let(:store) { SessionRepository.new(session: {}, form_key: :register_ect_wizard) }
   let(:wizard) { described_class.new(current_step: :find_ect, author: user, step_params: {}, store:, school:) }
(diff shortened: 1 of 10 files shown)


spec/wizards/schools/register_ect_wizard/wizard_spec.rb
```diff
@@ -628,16 +610,14 @@
...
-          started_on: Date.new(2025, 6, 1),
-          finished_on: Date.new(2026, 5, 31),
-          enabled: false
-        )
+      # We want to ensure the start date is in the future and within the
+      # current contract period.
+      # This travels to some point in the contract period so that always holds
+      # no matter when "today" actually is
+      let(:date_to_travel_to) { current_contract_period.started_on.next_month }
+      let(:start_date) { 1.week.from_now }
```
peteryates: Is the `#from_now` here relative to the date we've travelled to?

peteryates (commented):
Looks much better. Left a question (admittedly didn't look too deep) but if the dates are relative to the travelled to day, perhaps we could make that more obvious 🤔

spec/wizards/schools/register_ect_wizard/wizard_spec.rb
cpjmcquillan: Yes it is.

I think the nested contexts and complex setup make each test a bit confusing to be honest! 

I've tweaked those examples slightly which I think is a bit clearer.

```ruby
let(:date_to_travel_to) { current_contract_period.started_on.next_month }
let(:start_date) { date_to_travel_to + 1.week }
```

It's not perfect, but I hope that plus the comment helps.

peteryates (approved):
Nice!

PR #3030, opened 2026-06-05
backdating_ASPs_TPs_initial_attempt

boonali:
### Context
At the point of migration, approximately 2% of period combinations were migrated via the ECONOMY route, just using the current or most-recent of a participant-provider-school-cohort set of IRs.

This will mean it is an incomplete record of the participant's training history. It will result in incorrect dates associated with the participant in GET transfer responses, will result in some mentorships not successfully migrating and will be more difficult for users to interpret records when queries are made about the user for grant funding purposes.

Backdating ASPs and TPs to the later of
- previous RECT ASP finished_on + 1 day
- earliest ECF1 IR for the PP / URN combo

Also some leftover nullifying of end dates from #4043 (first 54 rows).

### Guidance to review

No errors in migration and correct changes have been made.


tonyheadford (approved):
Tested in migration with no errors

PR #3037, opened 2026-06-08
Fix metadata nightly alert false positives

ethax-ross:
### Context

We run a nightly metadata refresh; in an ideal world this would serve as an integrity check and not flag any changes, as our metadata is updated via callbacks when relevant objects are changed.

We've found this isn't the case for all metadata attributes, however. The `ect_assigned_mentor_latest_school_period_id` of the `Metadata::TeacherLeadProvider` is a moving target and depends on `MentorshipPeriod`s that are ongoing/future dated at any given point in time.

- Clean up handler updates/change tracking

We had rolled our own change/dirty tracking originally as we were using `upsert_all` for performance reasons. We since changed to call `update!` on individual objects as we need callbacks for the declarative touch operations.

Simplify change tracking and update; `update!` will only update changed attributes and we can use `ActiveRecord`'s built in dirty tracking for logging the changes to Sentry.

- Support excluding metadata changes from alerting

Our nightly refresh (which runs daily at 1am) is currently updating this field for mentorship periods that satisfy the criteria in the coming day. Instead of running another nightly job _before_ the current one to update these we are opting to silence the alert for this attribute and allow it to be updated with the existing job.

- Improve track_changes tests

Ensure we are sending the correct context to Sentry when tracking changes.

### Guidance to review

In v4 we will be moving away from metadata, so we don't want to spend a lot of time trying to fix this and the risk of not alerting on this reasonably low.

### DIFF
--- a/app/services/metadata/handlers/base.rb
+++ b/app/services/metadata/handlers/base.rb
@@ -28,6 +28,10 @@ module Metadata::Handlers
 
   protected
 
+    def alert_exclusions = []
+
+  private
+
     def lead_provider_ids
       @lead_provider_ids ||= LeadProvider.pluck(:id)
     end
@@ -43,7 +47,9 @@ module Metadata::Handlers
     # Allows individual handlers to exclude attributes that we can't
     # reliably update at the moment they change. Prevents them from
     # being included in the alerts.
-    def alertable_changes(saved_changes) = saved_changes
+    def alertable_changes(saved_changes)
+      saved_changes.excluding(alert_exclusions)
+    end
 
     def alert_on_changes(metadata)
       return unless @alert_on_changes
--- a/app/services/metadata/handlers/base.rb
+++ b/app/services/metadata/handlers/base.rb
@@ -32,39 +32,21 @@ module Metadata::Handlers
       @lead_provider_ids ||= LeadProvider.pluck(:id)
     end
 
-    def changes?(metadata, changes)
-      changes.stringify_keys!
-      metadata.attributes.slice(*changes.keys) != changes
-    end
-
-    def commit_changes!(metadata, changes)
-      return unless metadata.new_record? || changes?(metadata, changes)
-
+    def update_metadata!(metadata, latest_attributes)
       # As we touch other models when metadata is updated via ActiveRecord callbacks,
       # we need to ensure this remains a method that will trigger those callbacks
       # (rather than using something like upsert_all which would otherwise be quicker).
-      metadata.update!(changes)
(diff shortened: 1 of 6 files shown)


cpjmcquillan (approved):

app/services/metadata/handlers/base.rb
```diff
@@ -32,39 +32,30 @@ def lead_provider_ids
...
+    # being included in the alerts.
+    def alertable_changes(saved_changes) = saved_changes
 
-      changed_attributes_with_history = changes.each_with_object({}) do |(key, new_value), hash|
-        old_value = metadata.attributes[key]
+    def alert_on_changes(metadata)
+      return unless @alert_on_changes
 
-        next if old_value == new_value
+      alertable_changes = alertable_changes(metadata.saved_changes)
```
craigmdavidson: Do we want to name the local variable in a way that shadows the `alertable_changes` method?

app/services/metadata/handlers/teacher.rb
```diff
@@ -16,6 +16,15 @@ def destroy_all_metadata!
...
     end
 
+  protected
+
+    def alertable_changes(saved_changes)
+      # Depends on mentorship periods that are ongoing/future dated, which
+      # is a moving target, so the nightly refresh will often update this.
+      # We don't want to alert on this change when that happens.
+      saved_changes.excluding("ect_assigned_mentor_latest_school_period_id")
+    end
```
craigmdavidson, line 26: 🤔 Instead of manipulating the `saved_changes` directly, the _template method_ could just be a list of exclusions?

i.e. in base

```
  def alertable_changes(saved_changes)
    saved_changes.excluding(alert_exclusions)
  end

  def alert_exclusions = [] # template method
```

then in `teacher.rb`

```
def alert_exclusions
  # Depends on mentorship periods that are ongoing/future dated, which
  # is a moving target, so the nightly refresh will often update this.
  # We don't want to alert on this change when that happens.
  ["ect_assigned_mentor_latest_school_period_id"]
end
```

This would make for simpler tests in the classes that implement methods?

```
expect(teacher.alert_exclusions).
  to contain_exactly("ect_assigned_mentor_latest_school_period_id")
```

craigmdavidson (approved):
LGTM - a couple of nitpicks for your consideration.

app/services/metadata/handlers/teacher.rb
ethax-ross: I've refactored to use a template method but left the spec as it is to avoid making `alert_exclusions` public

PR #3038, opened 2026-06-08
Increase Band B for EDT in 2025

tonyheadford:
### Context

We need to increase the `max_declarations` level from `3102` to `4000` for Band B in 2025 for EDT to ensure they can submit the correct data.

Add a script to update the level

### Guidance to review

This will need to be run in migration and checked by an analyst before applying to production.

### DIFF
--- a/db/scripts/increase_band_b_for_edt.rb
+++ b/db/scripts/increase_band_b_for_edt.rb
@@ -3,20 +3,18 @@
 #
 # $ kubectl exec -it <pod-name> -- bin/rails runner db/scripts/increase_band_b_for_edt.rb
 
-begin
-  name = "Education Development Trust"
-  contract_period_year = 2025
+name = "Education Development Trust"
+contract_period_year = 2025
 
-  lead_provider = LeadProvider.find_by!(name:)
-  active_lead_provider = ActiveLeadProvider.find_by!(lead_provider:, contract_period_year:)
+lead_provider = LeadProvider.find_by!(name:)
+active_lead_provider = ActiveLeadProvider.find_by!(lead_provider:, contract_period_year:)
 
-  contract = Contract.find_by!(active_lead_provider:)
+contract = Contract.find_by!(active_lead_provider:)
 
-  band = contract.banded_fee_structure.bands.order(min_declarations: :desc).second
+band = contract.banded_fee_structure.bands.order(min_declarations: :desc).second
 
-  raise "Max declarations #{band.max_declarations} is not 3102!" if band.max_declarations != 3102
+raise "Max declarations #{band.max_declarations} is not 3102!" if band.max_declarations != 3102
 
-  band.update!(max_declarations: 4000)
+band.update!(max_declarations: 4000)
 
-  puts "Band B updated: #{band.min_declarations} - #{band.max_declarations}"
-end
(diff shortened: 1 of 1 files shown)


db/scripts/increase_band_b_for_edt.rb
```diff
@@ -0,0 +1,22 @@
...
+begin
+  name = "Education Development Trust"
+  contract_period_year = 2025
+
+  lead_provider = LeadProvider.find_by!(name:)
+  active_lead_provider = ActiveLeadProvider.find_by!(lead_provider:, contract_period_year:)
+
+  contract = Contract.find_by!(active_lead_provider:)
+
+  band = contract.banded_fee_structure.bands.order(min_declarations: :desc).second
```
peteryates: Do we have a particular record we want to update in mind here? Ordering and getting the second match feels a little 'imprecise'

db/scripts/increase_band_b_for_edt.rb
```diff
@@ -0,0 +1,20 @@
...
+#
+# $ kubectl exec -it <pod-name> -- bin/rails runner db/scripts/increase_band_b_for_edt.rb
+
+name = "Education Development Trust"
+contract_period_year = 2025
+
+lead_provider = LeadProvider.find_by!(name:)
+active_lead_provider = ActiveLeadProvider.find_by!(lead_provider:, contract_period_year:)
+
+contract = Contract.find_by!(active_lead_provider:)
```
ethax-ross: There could be multiple contracts for an active lead provider; it might be safer to do something like this which will raise if there are more than one:

```
contract = Contract.where(active_lead_provider:).sole
```

db/scripts/increase_band_b_for_edt.rb
```diff
@@ -0,0 +1,20 @@
...
+
+band = contract.banded_fee_structure.bands.order(min_declarations: :desc).second
```
ethax-ross: I think this is falling foul of our default ordering on `bands` (which is `min_declarations: :asc` and why you have to choose `second` here, confusingly!).

I would just do `contract.banded_fee_structure.bands.last` - we only ever increase the `max_declarations` of the last band.

ethax-ross (commented):
Thanks Tony just a couple of minor suggestions. 

@andycrawshaw I seem to recall there being two lead providers that needed their bands increase, but maybe I'm mis-remembering 🤔 I can't find the original thread to check

db/scripts/increase_band_b_for_edt.rb
tonyheadford: TIL `sole` 🥇

db/scripts/increase_band_b_for_edt.rb
tonyheadford: Good spot and I didn't know that either, thanks.

ethax-ross (approved):
Looks good thanks Tony. @andycrawshaw I think they wanted to do the same for Teach First looking back at the original thread, not 100% sure though could you maybe confirm with them?

db/scripts/increase_band_b_for_edt.rb
peteryates: :exploding_head:

PR #3040, opened 2026-06-08
Add aks-download-tmp-file to Makefile

rwrrll:
### Context

No ticket for this feature. I needed to produce a CSV from production data and download it.

Add a task to `Makefile` to download files from production. Copied and adapted from https://github.com/DFE-Digital/npq-registration/blob/main/Makefile#L233-L240


avinhurry (approved):
Nice! I'm sure this will come in handy more times than we expect.

PR #3042, opened 2026-06-08
[4027] Update cookies page to reflect current cookie usage

avinhurry:
### Summary

Updates the cookies page so it reflects the cookies the service currently uses, including the configured session cookie name, `id_token` for DfE Sign-in journeys, and clearer expiry wording.

Also fixes session handling so `id_token` is only stored when provided by DSI, while stale `id_token` cookies are still cleared for OTP sign-ins.

| Before | After |
|--------|-------|
|image|image|


cpjmcquillan (approved):
Commit breakdown really helpful. LGTM! 🚀

peteryates (approved):
⭐

PR #3047, opened 2026-06-09
[3173] Fix for duplicate scheme in comms out-opt host

rwrrll:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/3173 but after deployment it turns out that production has "https" baked into the action mailer host configuration.

Rather than always prepend the scheme to the URL, only add it if it's not already present

### DIFF
--- a/db/seeds/blazer_queries/school_comms.rb
+++ b/db/seeds/blazer_queries/school_comms.rb
@@ -205,11 +205,17 @@ module BlazerQueries
         host = options[:host].presence
         raise "action_mailer default_url_options host is not configured" if host.blank?
 
-        scheme = host == "localhost" ? "http" : "https"
-        authority = [host, options[:port]].compact.join(":")
+        # In production the configured host can already include the scheme
+        # (e.g. "https://example.com"); only add one when it's missing.
+        unless host.include?("://")
+          scheme = host == "localhost" ? "http" : "https"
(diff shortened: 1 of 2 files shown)


db/seeds/blazer_queries/school_comms.rb
```diff
@@ -205,11 +205,17 @@ def opt_out_base_url
         host = options[:host].presence
         raise "action_mailer default_url_options host is not configured" if host.blank?
 
-        scheme = host == "localhost" ? "http" : "https"
-        authority = [host, options[:port]].compact.join(":")
+        # In production the configured host can already include the scheme
+        # (e.g. "https://example.com"); only add one when it's missing.
+        unless host.include?("://")
+          scheme = host == "localhost" ? "http" : "https"
```
peteryates: This line's not the easiest to parse mentally 😅

avinhurry (approved):

peteryates (approved):
Makes sense.

db/seeds/blazer_queries/school_comms.rb
peteryates:
```suggestion
          scheme = (host == "localhost") ? "http" : "https"
```

cpjmcquillan (approved):

PR #3051, opened 2026-06-10
Touch `SchoolPartnership#api_updated_at` on `participants_currently_training` change

ethax-ross:
### Context

We surface the `participants_currently_training` in the `API::SchoolPartnershipSerializer`, however when a participant starts/finishes their training we don't touch the `api_updated_at` of the `SchoolPartnership`.

This has confused lead providers; when something changes in the serialized response the `updated_at` should reflect the change.

- Add job to touch SchoolPartnership#api_updated_at

As the `participants_currently_training` is a moving target, we have opted to add a nightly job that will touch relevant partnerships for participants that start training on that day, or finished training the previous day.

I opted to keep the `participant` terminology for consistency with the serializedfield name and used `updated_all` for efficiency as callbacks can safely be skipped.

- Add RefreshAllMetadataJob to review/staging

We only ran this in production/sandbox previously as it was only intended to be an integrity check. That changed recently as we now rely on the nightly refresh to update the `ect_assigned_mentor_latest_school_period_id` on
`Metadata::TeacherLeadProvider`, so we should run it on review/staging environments in order to avoid consistency issues in the metadata.

### Guidance to review

Create a participant due to start training tomorrow or finish training today, wait a day and check the `updated_at` in the partnerships response correctly reflects the change.

### DIFF
--- a/app/jobs/touch_school_partnership_for_participants_currently_training_job.rb
+++ b/app/jobs/touch_school_partnership_for_participants_currently_training_job.rb
@@ -1,0 +1,14 @@
+# Ensure changes to `participants_currently_training` in the
+# API::SchoolPartnershipSerializer are reflected in the `updated_at`.
+#
+# Job is scheduled to run daily at 12:01am.
+class TouchSchoolPartnershipForParticipantsCurrentlyTrainingJob < ApplicationJob
+  def perform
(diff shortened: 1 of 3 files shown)


peterdavidhamilton (approved):
LGTM

PR #3053, opened 2026-06-10
Create a copy of ECF 1 guidance documentation for reference

tonyheadford:
### Context

Convert and move the ECF1 guidance docs to the RECT so we have a copy available (especially as some were generated from templated code and had to be "scraped" from a running ECF1 instance) should we need them for reference.

Adds ECF1 docs in markdown format

### Guidance to review

Probably best to use the "View file" functionality on Github to see the rendered view and be able to navigate links etc.
Note that not all links go anywhere, I guess over time things have changed and this hasn't been updated as they don't go anywhere on the ECF1 original either (the `<schemaobject>/items` links for instance)

### DIFF
--- a/documentation/ecf1_guidance/index.md
+++ b/documentation/ecf1_guidance/index.md
@@ -1,5 +1,6 @@
-[!IMPORTANT]
-This is a copy of the legacy ECF1 API Guidance
+> [!IMPORTANT]
+> This is a copy of the legacy ECF1 API Guidance
+
 
 # About the API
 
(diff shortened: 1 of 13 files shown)


documentation/ecf1_guidance/index.md
```diff
@@ -0,0 +1,17 @@
+# About the API
```
ethax-ross, line 5: Is it worth a big warning at the top of here to make it clear this is an import of the legacy ECF API documentation to avoid any potential confusion? 🤔

ethax-ross (approved):
Thanks Tony, some broken links etc but as mentioned at stand up we're fine with that 👍

documentation/ecf1_guidance/index.md
tonyheadford: Yes, good idea

PR #3054, opened 2026-06-10
[3403 & 3404] Add training period relationship queries (Foundations) 

avinhurry:
### Summary

Add `TrainingPeriods::PeriodRelationships`, a small read only query object for reasoning about related training periods.

The service builds on existing domain concepts such as `TrainingPeriod#siblings`, `ongoing_today`, and `started_after` to expose:

- `current_active_period`
- `future_periods`

### Why?

Both the upcoming [admin contract period change](https://github.com/DFE-Digital/register-ects-project-board/issues/3404) and [schedule change journeys](https://github.com/DFE-Digital/register-ects-project-board/issues/3403) need to reason about:

- the current active training period
- future training periods created due to school transfers
- related training periods for the same participant role

This service provides a shared place for those relationship queries without introducing any ticket specific eligibility or mutation logic.

This will allow both tickets to be worked on in parallel without duplicating logic.

### DIFF
--- a/app/services/training_periods/period_relationships.rb
+++ b/app/services/training_periods/period_relationships.rb
@@ -7,10 +7,7 @@ module TrainingPeriods
     end
 
     def current_active_period
-      related_periods
-        .ongoing_today
-        .order(started_on: :desc, created_at: :desc)
-        .first
+      related_periods.ongoing_today.first
     end
 
     def future_periods
--- a/app/services/training_periods/period_relationships.rb
+++ b/app/services/training_periods/period_relationships.rb
@@ -1,0 +1,32 @@
+module TrainingPeriods
+  class PeriodRelationships
+    attr_reader :training_period
+
+    def initialize(training_period:)
+      @training_period = training_period
+    end
+
+    def current_active_period
+      related_periods
+        .ongoing_today
+        .order(started_on: :desc, created_at: :desc)
+        .first
+    end
+
+    def future_periods
+      @future_periods ||= related_periods
+        .started_after(Time.zone.today)
+        .earliest_first
+    end
+
+  private
+
+    def related_periods
+      TrainingPeriod.where(id: related_period_ids)
+    end
+
+    def related_period_ids
+      @related_period_ids ||= [training_period.id, *training_period.siblings.ids]
+    end
+  end
+end
--- a/app/services/training_periods/related_periods.rb
+++ b/app/services/training_periods/related_periods.rb
@@ -11,7 +11,7 @@ module TrainingPeriods
     end
 
     def future_periods
-      @future_periods ||= related_periods
+      related_periods
         .started_after(Time.zone.today)
         .earliest_first
     end
--- a/app/services/training_periods/related_periods.rb
+++ b/app/services/training_periods/related_periods.rb
@@ -1,5 +1,5 @@
 module TrainingPeriods
-  class PeriodRelationships
+  class RelatedPeriods
     attr_reader :training_period
 
     def initialize(training_period:)
--- a/spec/services/training_periods/related_periods_spec.rb
+++ b/spec/services/training_periods/related_periods_spec.rb
@@ -1,4 +1,4 @@
-RSpec.describe TrainingPeriods::PeriodRelationships do
+RSpec.describe TrainingPeriods::RelatedPeriods do
   subject(:relationships) { described_class.new(training_period:) }
 
   let(:today) { Date.new(2026, 2, 1) }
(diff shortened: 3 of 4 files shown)


app/services/training_periods/period_relationships.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    def initialize(training_period:)
+      @training_period = training_period
+    end
+
+    def current_active_period
+      related_periods
+        .ongoing_today
+        .order(started_on: :desc, created_at: :desc)
+        .first
```
peteryates: There can only be one training period at a time, so wouldn't `related_periods.ongoing_today.first` work here?

avinhurry: Yeah, I was being overly defensive here, could cause future devs a bit of confusion. Updated to rely on the existing training period overlap validation.

Thanks!

app/services/training_periods/related_periods.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    def current_active_period
+      related_periods.ongoing_today.first
+    end
+
+    def future_periods
+      @future_periods ||= related_periods
+        .started_after(Time.zone.today)
+        .earliest_first
+    end
```
cpjmcquillan, line 17: Take this with a grain of salt, but `TrainingPeriods::PeriodRelationships` doesn't feel particularly revealing to me as a name. I'm not sure what alternatives there are though.

I played around with something like this.

```ruby
module TrainingPeriods
  class Related
    def self.to(training_period) = new(training_period)

    def initialize(training_period)
      @training_period = training_period
    end
    
    def current = relatives.ongoing_today.first
    def future = relatives.started_after(Time.zone.today).earliest_first

    private

    attr_reader :training_period

    def relatives
      TrainingPeriod.where(id: [training_period.id, *training_period.siblings.ids])
    end
  end
end
```

and you could use it like this

```ruby
TrainingPeriods::Related.to(training_period).future
```

What do you think?

avinhurry: I spent some time thinking about the current name here.

I think `Related` is a bit too vague on its own though. When I see: 

```ruby
TrainingPeriods::Related.to(training_period)
```

I wonder what the training period is related to. Periods, participants, schedules etc.

I also think your proposed refactor is a bit more implicit then we tend to be elsewhere in the codebase. The current version is more verbose but very explicit:

```ruby
relationships = TrainingPeriods::PeriodRelationships.new(training_period:)

relationships.current_active_period
relationships.future_periods
```

Maybe `TrainingPeriods::RelatedPeriods` is a better compromise? WDYT

cpjmcquillan: I think I prefer `TrainingPeriods::RelatedPeriods` to `TrainingPeriods::PeriodRelationships` 👍🏼

What about this?

```ruby
relatives = TrainingPeriods::Relatives.new(training_period:)
relatives.current_training_period
relatives.future_training_periods
```

cpjmcquillan, line 17: It feels a bit strange that we're memoizing one of the methods but not the other.

FWIW, I think ActiveRecord's query cache would come in to play anyway in the context of a request or job, so I'm not sure we need the memoization here.

spec/services/training_periods/related_periods_spec.rb
cpjmcquillan, line 1: I think I'd find this test a bit easier to read if the dates were inlined. I don't think the repetition is that problematic. It might be easier to understand if the periods were grouped by date too (i.e historical ect and training period) rather than by type.

cpjmcquillan (approved):

app/services/training_periods/related_periods.rb
ericaporter: My tuppence 😄 ... 

I’d slightly prefer `TrainingPeriods::RelatedPeriods` over `TrainingPeriods::Related` / `TrainingPeriods::Relatives`, because it is clearer and easier to remember / search for

avinhurry: I've gone with `TrainingPeriods::RelatedPeriods` as that seems to be the preference for now.

Thanks all!

avinhurry: Good shout, i've removed the memoization from the other one.

spec/services/training_periods/related_periods_spec.rb
avinhurry: I've grouped the setup by timeline but kept the named dates because they describe the scenario and keep the related periods aligned.

PR #3055, opened 2026-06-10
Add schedule lookup across contract periods (Admin change contract period journey)

avinhurry:
### Summary

Add `TrainingPeriods::MatchingSchedule`, a query object for finding schedules with the same identifier across contract periods.

### Why?

This logic will be needed by the [upcoming contract period change journey](https://github.com/DFE-Digital/register-ects-project-board/issues/3404) and is extracted here to keep the lookup in a single place.

### DIFF
--- a/app/services/schedules/equivalent_schedule.rb
+++ b/app/services/schedules/equivalent_schedule.rb
@@ -1,0 +1,17 @@
+module Schedules
+  class EquivalentSchedule
+    attr_reader :source_schedule, :target_contract_period
+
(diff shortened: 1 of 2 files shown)


cpjmcquillan (commented):
I might be missing something here, but I don't really understand what the problem this service object solves.

What's wrong with this on its own? 🤔

```ruby
Schedule.find_by(
  schedule: source_schedule.identifier, 
  contract_period: target_contract_period
)
```

PR #3056, opened 2026-06-10
Statement fee type restriction

craigmdavidson:
### Context

resolves https://github.com.mcas.ms/DFE-Digital/register-ects-project-board/issues/4018

* Remove the ability to select a **Statement**'s `fee_type` instead assign it based the the **Statement** `month` value.

### Guidance to review

- This is exactly the kind of change best suited to a `before_validation` callback on the **Statement** model, but this approach is off the table. Because this is used in multiple places (both Create and Update), we extract a small service object to handle with that. Unfortunately though we then get ServiceObjects calling ServiceObjects.

- Because we have removed the ability for a user to assign the `fee_type` there is no need to show an error message.

### DIFF
--- a/app/services/statements/create.rb
+++ b/app/services/statements/create.rb
@@ -5,6 +5,7 @@ module Statements
     def initialize(author:, params:)
       @author = author
       @statement = Statement.new(params)
+      @statement.fee_type = FeeTypeForMonth.new(month: @statement.month).call
     end
 
     def call
--- a/app/services/statements/create.rb
+++ b/app/services/statements/create.rb
@@ -5,10 +5,11 @@ module Statements
     def initialize(author:, params:)
       @author = author
       @statement = Statement.new(params)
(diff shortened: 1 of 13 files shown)


app/services/statements/create.rb
```diff
@@ -5,6 +5,7 @@ class Create
     def initialize(author:, params:)
       @author = author
       @statement = Statement.new(params)
+      @statement.fee_type = FeeTypeForMonth.new(month: @statement.month).call
```
craigmdavidson: This would be cleaner as a `:before_validation` callback on Statement, but if we want to use service objects, this is the way to go - since we need to assign/reassign fee type in both **Statements::Create** and **Statements::Update**.

ethax-ross (approved):
Looks good 👍  I do wonder if its worth just dropping the `fee_type` column all together though and having a method on `Statement` along the lines of:

```
def fee_type
  Statements::FeeTypeForMonth.new(month:).call
end
```

That would remove the need to have it in the update/create services.

PR #3059, opened 2026-06-11
[4038] Services to open, close and replace GIAS::Schools

thenapking:
### Context
As per the linked ticket there are 5 scenarios where schools can change following updates in GIAS.  Openings, closures, 1-1 replacements, splits and merges.  This PR introduces services to perform, but *not automate*, the first 3.  The latter two are more complex and require input from the school so we know which teachers go where.  They would also need more complex service built.  

This does not automate opening/closing/replacing schools for several reasons.  Firstly there is a backlog of changes which have happened since we launched the new service, and there is a [ticket](https://github.com/DFE-Digital/register-ects-project-board/issues/4030) to deal with that.  The intention is that these services can be used to clear the backlog, and that process should help refine them before they are automated.

Secondly, althought the importer script has functionality to create new schools, I think a slightly different approach might be needed in order to automate replacements, openings, and closures.  The intention behind these service classes is that once the importer has been run fully, each changed `GIAS::School` would be run through a separate service to reconcile its associated records.  Something along these lines:
```
case
when gias_school.replaceable?
  GIAS::Schools::Replace.new(gias_school).replace!

when gias_school.closeable?
  GIAS::Schools::Close.new(gias_school).close!

when gias_school.openable?
  GIAS::Schools::Open.new(gias_school).open!
end
```
To know that a school which is marked `closed` is not being replaced or split requires import of all the `GIAS::SchoolLink` records.  Therefore it is safest if reconcilliation is run after a full import.  Also we wouldn't want a job to run this on all schools in the database because there are schools which closed before this service started.  Running a reconcilliation after records are imported seems the safest path to automation.

Three services are introduced.  The close service is the most complicated, as it needs to modify periods.  Periods which have not started are destroyed, and others are finished on the closure date.  It is worth noting that destroying unstarted ect or mentor periods does not necessarily destroy all unstarted mentorship_periods.  There may be mentorship periods in the future where one or more of the teachers has already started at the school.  These need to be removed first before removing any unstarted ect_at_school_periods.

The classes uses the date supplied by the GIAS record to update any associated periods and create events. This can be in the past to help reconcile existing records, and to deal with timing issues arising from the GIAS import.

The services are designed to be run manually from the console at the moment.  

Because the GIAS data can be unreliable. a cautious approach is taken.  Whilst successor schools are often created with predecessor records, this is not always the case so those are not used in testing replacements.  Also there is a link_type for closure.  However this has only been used historically in one or two of cases in the past, and all new closures appear to come through without this.

Also there is some delicacy around the dates, and the `proposed_to_close` and `proposed_to_open` statuses.  This PR works on the assumption that until a school is open and its `opened_at` date is current, it is not openable.  Equally for closed schools.

Shared Rspec contexts have also been created which hopefully will be useful when testing services for splitting or merging schools.

### Guidance to review

There are several schools added to the seeds in order to test running this from the console.  Before the updates only Monsters College, Frank McCay Technical College and Monsters Primary School should be open and visible from the admin interface.   

Once the below are run from the console, Monsters Junior should be open, the James P. Sullivan Preparatory School should have replaced Monsters Primary and moved the teachers across.  Nothing should have happened to Monsters College as it is being split into the Abigail Hardscrabble High School for Girls and the Michael Wazowski School of Excellence.  Neither of these schools should exist in the service yet. The Frank McCay Technical College should be closed.

To test the services run the following from the console:
```
monsters_junior = GIAS::School.find_by(urn:  9_123_506)
GIAS::Schools::Open.new(monsters_junior).open!
frank_mccay_technical = GIAS::School.find_by(urn:  9_123_505)
GIAS::Schools::Close.new(frank_mccay_technical ).close!
monsters_primary = GIAS::School.find_by(urn:  9123503)
GIAS::Schools::Replace.new(monsters_primary).replace!
monsters_college = GIAS::School.find_by(urn:  9123500)
GIAS::Schools::Replace.new(monsters_college).replace!
```

The first three should succeed.  The last should fail in the console, as it is splitting Monsters College into the Michael Wazowski School of Excellence and the Hardscrabble High School for Girls.


app/services/gias/schools/close.rb
```diff
@@ -0,0 +1,110 @@
...
+      MentorshipPeriod.started_after(gias_school.closed_on).joins(:mentee).where(ect_at_school_periods: { school_id: school.id })
+    end
+
+    def unstarted_mentorship_periods_for_mentor_at_school_periods
+      MentorshipPeriod.started_after(gias_school.closed_on).joins(:mentor).where(mentor_at_school_periods: { school_id: school.id })
+    end
+
+    def finish_ongoing_periods!
+      ongoing_ect_at_school_periods.each do |ect_at_school_period|
+        ECTAtSchoolPeriods::Finish.new(
```
peteryates, line 31: Currently when we call `ECTAtSchoolPeriods::Finish#finish!` it records a `teacher_left_school_as_ect` event.

I wonder if we could add a line to the event's body or something to make it clear the period ended because the school closed.

Perhaps this is more of a nice to have though.

app/models/gias/school.rb
```diff
@@ -17,8 +17,14 @@ class GIAS::School < ApplicationRecord
...
   has_one :school, foreign_key: :urn, primary_key: :urn, inverse_of: :gias_school
-  has_many :gias_school_links, class_name: "GIAS::SchoolLink", foreign_key: :urn, dependent: :destroy, inverse_of: :from_gias_school
   has_many :contract_period_metadata, class_name: "Metadata::SchoolContractPeriod", through: :school
+  has_many :gias_school_links, class_name: "GIAS::SchoolLink", foreign_key: :urn, dependent: :destroy, inverse_of: :from_gias_school
+
+  has_many :successor_links,   -> { where(link_type: GIAS::SchoolLink::SUCCESSOR_LINK_TYPES) },   class_name: "GIAS::SchoolLink", foreign_key: :urn, primary_key: :urn
+  has_many :predecessor_links, -> { where(link_type: GIAS::SchoolLink::PREDECESSOR_LINK_TYPES) }, class_name: "GIAS::SchoolLink", foreign_key: :urn, primary_key: :urn
+
+  has_many :successors,   class_name: "GIAS::School", through: :successor_links, source: :to_gias_school
+  has_many :predecessors, class_name: "GIAS::School", through: :predecessor_links, source: :from_gias_school
```
cpjmcquillan, line 28: Some weird formatting going on here

app/services/gias/schools/close.rb
```diff
@@ -0,0 +1,89 @@
...
+      ect_at_school_periods.starting_tomorrow_or_after
+    end
+
+    def ongoing_mentor_at_school_periods
+      mentor_at_school_periods.ongoing_on(Date.current)
+    end
+
+    def ongoing_ect_at_school_periods
+      ect_at_school_periods.ongoing_on(Date.current)
+    end
```
cpjmcquillan: I think these could use `ongoing_today`

cpjmcquillan: FWIW, I'm not sure how much value these methods add. Is the indirection worth it?

`mentor_at_school_periods.starting_tomorrow_or_after` is pretty easy to understand IMO.

What do you think?

app/models/gias/school.rb
```diff
@@ -63,4 +69,48 @@ def closed?
...
+    closed_status? && successors.empty? && !school_closure_recorded?
+  end
+
+  def openable?
+    open_status? && school_not_yet_opened? && predecessors.empty? && successors.empty?
+  end
+
+  def replaceable?
+    closed_status? && successor&.open_status? && successor.school_not_yet_opened? && school_is_being_replaced?
+  end
```
cpjmcquillan: I wonder if naming these like `can_be_closed?` would be a bit more explicit and easier to read than `closeable?`

spec/support/shared_examples/gias_school_updates.rb
```diff
@@ -0,0 +1,38 @@
...
+end
+
+RSpec.shared_examples "does not change the schools URN or record an event" do
+  it "does not update the school's URN" do
+    expect { subject }.not_to(change { gias_school.school.reload.urn })
+  end
+
+  it "does not record a school changed event" do
+    expect { subject }.not_to change(Event, :count)
+  end
```
cpjmcquillan: Does this expectation ever fail? I don't think we perform jobs in tests by default, so I don't think the Event count would change. It might be better to check the enqueued `RecordEventJob` jobs 💭

spec/services/gias/schools/close_spec.rb
```diff
@@ -0,0 +1,180 @@
...
+    it_behaves_like "records a school closed event"
+
+    context "when the school is open" do
+      let(:gias_school) { FactoryBot.create(:gias_school, :open, :with_school) }
+
+      it_behaves_like "does not destroy or finish any periods"
+
+      it "does not record an event" do
+        expect { service }.not_to(change(Event, :count))
+      end
```
cpjmcquillan: Here too, I'm not convinced these tests would typically ever fail

app/services/gias/schools/open.rb
```diff
@@ -8,16 +8,22 @@ def initialize(gias_school)
...
+      end
+
+    private
+
+      def open_school!
         ActiveRecord::Base.transaction do
           gias_school.create_school!
           record_school_opened_event!
         end
       end
```
cpjmcquillan, line 25: I think these were clearer without the `open_school!` private methods

db/seeds/gias_data.rb
cpjmcquillan, line 1: Are these seeds we want to add for this PR, or seeds we want to add forever?

cpjmcquillan (approved):
A few comments of things I think we ought to address. Not sure if we want to do this here, or in a follow up PR. 

What do you think? @rwrrll

app/services/gias/schools/close.rb
thenapking: Once the data is cleaned up, I agree.  However we need to be able to run this on schools which have already closed but we haven't updated the data for yet.

app/services/gias/schools/open.rb
thenapking: Separating this out, allows a dev to run the private method from the console, which will be useful until we have the data cleaned up.  After that point, I agree we can move this back into the main method.

db/seeds/gias_data.rb
thenapking: These are seeds we want to add semi-permanently, as they will be useful for testing scenarios where we are cleaning up the GIAS data, writing the other service classes and so on.  But once that's done, they could probably be removed.

app/services/gias/schools/close.rb
cpjmcquillan: `ongoing_today` is defined as `ongoing_on(Time.zone.today)`. 

I'm not sure what you mean about the data needing to be cleaned up?

thenapking: We have changes from April and May that we need to process with this class.  Unless we enter the date the change happened in GIAS they will all happen when we process them and the data will be wrong.

thenapking: I've now tried to make the class more brief.

app/services/gias/schools/close.rb
thenapking: We discussed this when product reviewing yesterday.  Our feeling was that, it wasn't that odd seeing this message, especially when it's on the same date the school closes.

PR #3061, opened 2026-06-11
[4038] Service to destroy mentor periods

thenapking:
### Context
In order to close a school when its status in GIAS changes to closed, we need to finish all open periods on the closure date.  However if any of these periods, or their associated training or mentorship periods start in the future that will cause an error.  So these periods will need to be deleted.  

Currently there is only a service to destroy ect_at_school_periods.  This abstracts this to a module, which can then be used to build the mentor service.
Additionally, because we have schools in the past, we will need to be able to run this service for a particular date in the past.

### Guidance to review
This adds a service class which could be reviewed from the console.  However we could try running some of the wizard journeys which destroy ECT periods in the future.

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -41,6 +41,7 @@ class Event < ApplicationRecord
     teacher_registered_as_ect
     teacher_left_school_as_ect
     teacher_ect_at_school_period_deleted
+    teacher_mentor_at_school_period_deleted
     teacher_registered_as_mentor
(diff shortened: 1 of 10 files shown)


avinhurry (approved):
Setting a default `actioned_at` makes me slightly nervous, since it would be easy to forget and accidentally run this against today. I think we can live with that though, leaving the ball in your court.

Other than that, this looks good to me.

PR #3062, opened 2026-06-11
Mentors are only eligible for their first provider training if they have never had mentor training

cpjmcquillan:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/4060

It's best to review commit by commit, but the significant change to behaviour is outlined below.

#### Mentors are only eligible if they have never had mentor training
Before, mentors were eligible for their first provider-led training if:

- they were mentoring an ECT doing provider-led training, and
- they were not ineligible for mentor funding, and
- they had no current of future mentor training periods at any school

This was incorrect, though.

In fact, a mentor should **not** be eligible for their first provider-led
training if they have previously finished mentor training.

We realised this was an issue when investigating [this bug][bug]. We were
attempting to create training periods for mentors when we ought not to.

As a follow up to this, we should identify mentors with multiple mentor training
periods and rectify them. That is covered by [this ticket][ticket].

This broadens the check for mentor training periods in the
`MentorAtSchoolPeriod::Eligibility` service to include **all** mentor training
periods (i.e current, future, **and** finished).

Now, mentors are only eligible for their first provider-led training if:

- they are mentoring an ECT doing provider-led training, and
- they are not ineligible for mentor funding, and
- they have no mentor training periods at any school

One distinction that isn't immediately clear looking at this service in
isolation is that it only applies when a school is assigning a mentor (either
after registering an ECT or as part of the change mentor journey).

Its primary purpose (although this isn't explicit in the code) is to determine
whether or not we ought to create a training period for the mentor when we
assign them. This is explored in the [mentor assignment mega-graph][excalidraw].

[bug]: https://github.com/DFE-Digital/register-ects-project-board/issues/4052
[ticket]: https://github.com/DFE-Digital/register-ects-project-board/issues/4088
[excalidraw]: https://excalidraw.com/#json=gf2x4hXzmTgK0p1JDpEF1,4-HYy0nLM70zmAKIKVkUBQ


app/services/mentor_at_school_periods/eligibility.rb
```diff
@@ -5,7 +5,7 @@ def self.for_first_provider_led_training?(ect_at_school_period:, mentor_at_schoo
 
       ect_at_school_period.provider_led_training_programme? &&
         Teachers::MentorFundingEligibility.new(trn: mentor_at_school_period.teacher.trn).eligible? &&
-        mentor_at_school_period.teacher.mentor_training_periods.current_or_future.none?
+        mentor_at_school_period.teacher.mentor_training_periods.none?
```
cpjmcquillan: This is the key change ⤴️

app/services/mentor_at_school_periods/assignment/eligibility.rb
```diff
@@ -0,0 +1,21 @@
+module MentorAtSchoolPeriods
+  module Assignment
+    class Eligibility
+      class << self
+        def for_first_provider_led_training?(ect_at_school_period:, mentor_at_school_period:)
+          return false unless mentor_at_school_period && ect_at_school_period
+          return false unless ect_at_school_period.provider_led_training_programme?
+          return false unless teacher_eligible_for_mentor_funding?(mentor_at_school_period.teacher)
+
+          mentor_at_school_period.teacher.mentor_training_periods.none?
```
ericaporter, line 10: Looking at the AC - do we want to only check for previous provider-led TPs here instead of school-led too?

The naming of the method makes it sound like we are checking whether this is the mentor’s first provider-led training, but it looks like we are just checking if the mentor has any mentor training periods

app/wizards/schools/ects/change_mentor_wizard/edit_step.rb
```diff
@@ -21,34 +21,27 @@ def save!
...
         def mentors_for_select
-          current_mentorship_period = current_mentor_at_school_period
-
-          if current_mentorship_period.present?
-            eligible_mentors.excluding(current_mentorship_period)
-          else
-            eligible_mentors
-          end
+          eligible_mentors.excluding(current_mentor_at_school_period)
         end
```
ericaporter, line 25: Can `current_mentor_at_school_period` ever be nil here? the previous version handled nil explicity so wondering if we need to keep the guard in for clarity?

app/wizards/schools/ects/change_mentor_wizard/edit_step.rb
```diff
@@ -21,34 +21,27 @@ def save!
...
 
       private
 
-        def mentor_id_is_selectable
-          return errors.add(:mentor_at_school_period_id, "Select a mentor from the list provided") if mentor_at_school_period_id.blank?
-
-          errors.add(:mentor_at_school_period_id, "Select a mentor from the list provided") unless mentors_for_select.exists?(id: mentor_at_school_period_id)
+        def mentor_exists
```
ericaporter: Maybe `mentor_is_selectable` is more descriptive here rather than mentor_exists

spec/services/ect_at_school_periods/switch_mentor_spec.rb
```diff
@@ -145,6 +145,33 @@ module ECTAtSchoolPeriods
           end
         end
 
+        context "when the mentor has a finished provider-led training period" do
```
ericaporter, line 148: Could we also add a context for a mentor with only previous school-led mentor training? That case is specified in the AC, and it would clarify whether school-led-only history should or should not block creating a provider-led mentor TP

spec/services/mentor_at_school_periods/assignment/eligibility_spec.rb
```diff
@@ -49,6 +49,35 @@
       end
     end
 
+    context "when mentor has a finished training period at the same school" do
```
ericaporter, line 52: Should we be specifying provider-led in the context descriptions?

& Maybe add in one explicit school-led context?

app/services/mentor_at_school_periods/assignment/eligibility.rb
cpjmcquillan: [Mentors can only have provider-led training](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/b62a32ce746d9dc44abed566abab5366d4b4348d/app/models/training_period.rb#L81), so it seemed superfluous (I had it but then I removed it).

Happy to add it if we want to be more explicit, but tried to keep things simple.

What do you think?

app/wizards/schools/ects/change_mentor_wizard/edit_step.rb
cpjmcquillan: It could, but `excluding(nil)` is a `noop`. I don't think the guard adds anything as far as I can tell!

spec/services/ect_at_school_periods/switch_mentor_spec.rb
cpjmcquillan: Mentor training periods can only be provider-led, so I don't think we need it (we couldn't create the training period)

spec/services/mentor_at_school_periods/assignment/eligibility_spec.rb
cpjmcquillan: Same here

app/services/mentor_at_school_periods/assignment/eligibility.rb
ericaporter: If mentors can only have provider-led training periods then it would be superfluous. 

I wonder whether the domain assumption is obvious enough for readers of the code base... it probably is...

ericaporter (approved):
LGTM! 👍

PR #3065, opened 2026-06-11
[4026] add milestone date validations

peterdavidhamilton:
### Context

Closes https://github.com/DFE-Digital/register-ects-project-board/issues/4026

At present there is no validation on a milestone's start date resulting in any date being accepted. What we need is for a milestone to not predate June 1st of the contract period year.

Whilst the validation rule is trivial the challenge with the change was getting the test suite to be reliable. 

1. Date fixtures in specs now reference the contract in question.
2. We factory contract periods in sequence which meant this change introduced errors in those specs.
3. The way factories were chained meant we could be left with an orphaned record that would cause specs to fail or be flaky.
4. Some factories persisted records during the build phase when they weren't meant to.

Add milestone start date validation that ensures it is bounded by the contract period of the schedule.

- only initialise when building a factory so chained associations are not persisted
- pin the active lead provider factory to the current period where necessary
- update spec fixture dates to assert the contract period year not hard-coded values
- pin declaration create specs to 2024 otherwise they are flaky
- ensure no ghost contract periods remain if they were overwritten

### Guidance to review

cpd-ec2-review-3065-web test teacherservices cloud_admin_finance_contract-periods_2027_schedules_323_milestones

### DIFF
--- a/app/models/milestone.rb
+++ b/app/models/milestone.rb
@@ -39,14 +39,14 @@ class Milestone < ApplicationRecord
 
 private
 
-  # Contract periods start on 1st June
+  # NB: Contract periods typically start on 15th June, 2 weeks after the milestone start date.
   def start_date_within_contract_period
     return unless schedule&.contract_period
 
-    contract_start = schedule.contract_period.started_on
+    contract_start = Date.new(schedule.contract_period_year, 6, 1)
 
     return if start_date >= contract_start
 
-    errors.add(:start_date, "The start date must be on or after the contract start date (#{contract_start.to_fs(:govuk)})")
+    errors.add(:start_date, "The start date must be on or after the #{contract_start.to_fs(:govuk)}")
   end
 end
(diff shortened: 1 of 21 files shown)


craigmdavidson (approved):
LGTM

ethax-ross (commented):
Thanks PDH. Just checking against snapshot it appears some existing milestones fall foul of these validation rules; can we dig into why that is and if its the rule or the milestone that needs adjusting?

```
invalid_milestones = Milestone.all.select(&:invalid?)

invalid_milestones.count
=> 14

invalid_milestones.map(&:errors).map(&:full_messages).uniq
=> 
[["Start date The start date must be on or after the contract start date (5 June 2023)"],
 ["Start date The start date must be on or after the contract start date (16 June 2025)"],
 ["Start date The start date must be on or after the contract start date (18 June 2024)"],
 ["Start date The start date must be on or after the contract start date (15 June 2026)"]]
```

PR #3066, opened 2026-06-12
Reseed API seeds in a job on staging env

ethax-ross:
We made this change to the review app deploy to speed it up; this brings it across to the reseed action for staging.

[Example run](https://github.com/DFE-Digital/register-early-career-teachers-public/actions/runs/27404110139/job/80988841249)

Changes:

- Reseed API seeds in a job on staging env
- Reduce amount of API seeds
- Make API seed verbose logging optional
- Increase staging worker RAM

I've ran the job a whole bunch of times to reseed staging now and it seems stable.


cpjmcquillan (approved):

app/services/api_seed_data/school_scenarios.rb
```diff
@@ -11,46 +11,46 @@ def plant
 
       # 5x schools that have minimum 1 participant registered with the lead provider as training them,
```
tonyheadford: Super minor but all the comments now have the wrong number in them

tonyheadford (approved):
Looks great, one minor comment regarding the code comments

app/services/api_seed_data/school_scenarios.rb
ethax-ross: Good spot! I've updated

PR #3067, opened 2026-06-12
Fix submit declaration error message

craigmdavidson:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/3618

Changes error message.

### Guidance to review

Checks if we the Teacher is known to the lead provider.

### DIFF
--- a/app/services/api/declarations/create.rb
+++ b/app/services/api/declarations/create.rb
@@ -151,7 +151,19 @@ module API::Declarations
       return if errors[:declaration_date].any?
       return if training_period
 
-      errors.add(:teacher_type, "The entered '#/teacher_type' is not recognised for the given participant. Check details and try again.")
+      if teacher_registered_with_lead_provider?
+        errors.add(:teacher_type, "The entered '#/teacher_type' is not recognised for the given participant. Check details and try again.")
+      else
+        errors.add(:teacher_api_id, "Your update cannot be made as the '#/teacher_api_id' is not recognised. Check participant details and try again.")
+      end
+    end
+
+    def teacher_registered_with_lead_provider?
+      return false unless teacher
+
+      [teacher.ect_training_periods, teacher.mentor_training_periods].any? do |periods|
+        periods.includes(:lead_provider).where(active_lead_providers: { lead_provider_id: }).exists?
+      end
     end
 
     def validate_milestone_exists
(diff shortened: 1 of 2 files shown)


app/services/api/declarations/create.rb
```diff
@@ -151,7 +151,19 @@ def teacher_type_exists
       return if errors[:declaration_date].any?
       return if training_period
 
-      errors.add(:teacher_type, "The entered '#/teacher_type' is not recognised for the given participant. Check details and try again.")
+      if teacher_registered_with_lead_provider?
+        errors.add(:teacher_type, "The entered '#/teacher_type' is not recognised for the given participant. Check details and try again.")
+      else
+        errors.add(:teacher_api_id, "Your update cannot be made as the '#/teacher_api_id' is not recognised. Check participant details and try again.")
```
ethax-ross: Would it be worth pulling this out into a separate validation rule, as its not really to do with teacher type on this branch of the conditional? 🤔 something like `validate teacher_registered_with_lead_provider`? then we can ensure they get the correct validation message with the ordering of the validation rules

craigmdavidson: done

app/services/api/declarations/create.rb
```diff
@@ -150,17 +151,34 @@ def teacher_type_exists
...
+      return if errors[:teacher_api_id].any? || errors[:lead_provider_id].any?
+      return if teacher_registered_with_lead_provider?
+
+      errors.add(:teacher_api_id, "Your update cannot be made as the '#/teacher_api_id' is not recognised. Check participant details and try again.")
+    end
+
+    def teacher_registered_with_lead_provider?
+      return false unless teacher
+
+      [teacher.ect_training_periods, teacher.mentor_training_periods].any? do |periods|
```
ethax-ross, line 169: We could use the metadata here, something like:

```
teacher
  .lead_provider_metadata
  .where.not(
    latest_ect_training_period: nil,
    latest_mentor_training_period: nil
  )
  .exists?(lead_provider_id:)
```

ethax-ross (approved):
Looks good, just a take it or leave it suggestion; I think its worth putting this one through PO review

app/services/api/declarations/create.rb
craigmdavidson: Surely if we don't _need_ to use metadata we shouldn't - that makes it harder to remove later?

PR #3068, opened 2026-06-12
Fix school funding eligibility GIAS foreign key

rwrrll:
### Context

Correcting FK as per https://github.com/DFE-Digital/register-ects-project-board/issues/3784#issuecomment-4432490471 prior to https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3059

Change SchoolFundingEligibility URN foreign key from `School` to `GIAS::School`

> [!IMPORTANT]
> This is 1 of 2, see #3082 for part 2


cpjmcquillan (commented):
I think this makes sense.

Is it worth renaming the column on `school_funding_eligibilities` to reflect we're referencing the `gias_schools` table? (i.e `gias_school_urn` rather than `school_urn`? 🤔

Also, do we need to think about breaking the migration into several stages, or is this safe to do in one?

cpjmcquillan (approved):

ethax-ross (commented):
Looks good past Connor's comments. As the migrations are ran before the web apps update there will be a small period of time they are out of sync, so +1 for splitting this up to avoid errors during deploy.

avinhurry (approved):

cpjmcquillan (approved):

PR #3069, opened 2026-06-12
patching_one_mentorship_period

boonali:
### Context

As part of ticket #3988, a number of missing mentorship periods were added to RECT via the son of patch csv. Emily has spotted a case where a participant has 2 mentorship periods which overlap

Moving the most-recent MP to after the initial one, and ending before the finished_on of the ECT ASP.


tonyheadford (approved):
Tested fine in migration env

PR #3070, opened 2026-06-12
Update DfE Sign-in access denied pages

ericaporter:
Ticket:[here](https://github.com/DFE-Digital/register-ects-project-board/issues/4076)

**Context**

This updates the two DfE Sign-in access denied pages shown when a user cannot access the service.

There are two separate states:

1. the user does not have a school or organisation attached to their DfE Sign-in account
2. the user has selected a school or organisation, but does not have approved access to the Register early career teachers service for that organisation

**Changes proposed in this pull request**

- Updated the access denied page titles/headings for both DfE Sign-in access states.
- Replaced the old generic access denied content with the new ticket/prototype copy.
- Moved the DfE Sign-in link from the shared wrapper into the relevant page partials so each state has the correct request-access CTA:
   - request access to a school or organisation
   - request access to the service
- Updated the access denial feature spec to cover the new content for both page variants

**Guidance to review**
- Check the copy matches the ticket/prototype for both access denied states.
- Check the correct partial is shown depending on whether an organisation name is present.
- Check the DfE Sign-in request-access link text is correct for each state.

### DIFF
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -1,14 +1,21 @@
 class PagesController < ApplicationController
   after_action :allow_search_engine_indexing, only: :home
 
+  DFE_SIGN_IN_REQUEST_ORGANISATION_URL =
+    "https://services.signin.education.gov.uk/request-organisation/search"
+
+  DFE_SIGN_IN_MY_SERVICES_URL =
+    "https://services.signin.education.gov.uk/my-services"
+
   def home
     redirect_to(post_login_redirect_path) and return if authenticated?
   end
 
   # Unrecognised DfE Sign In user (org/role)
   def access_denied
-    @dfe_sign_in_profile_url = Rails.application.config.dfe_sign_in_profile
     @organisation_name = session[:invalid_user_organisation_name]
+    @dfe_sign_in_request_organisation_url = DFE_SIGN_IN_REQUEST_ORGANISATION_URL
+    @dfe_sign_in_my_services_url = DFE_SIGN_IN_MY_SERVICES_URL
   end
 
   def support
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -1,12 +1,6 @@
 class PagesController < ApplicationController
   after_action :allow_search_engine_indexing, only: :home
 
-  DFE_SIGN_IN_REQUEST_ORGANISATION_URL =
-    "https://services.signin.education.gov.uk/request-organisation/search"
-
-  DFE_SIGN_IN_MY_SERVICES_URL =
-    "https://services.signin.education.gov.uk/my-services"
-
   def home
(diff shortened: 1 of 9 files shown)


app/views/pages/access_denied/_without_organisation.md.erb
```diff
@@ -1,4 +1,18 @@
...
 
-If you think you should have access, you can request it from your organisation's approver.
-You request access directly from DfE Sign-in.
+You cannot use this service until you have access.
+
+## What you need to do
+
+In your DfE Sign-in account, request access to the school or organisation.
+
+Request access to the 'Register early career teachers'.
```
cpjmcquillan: Are we missing a word here?

```suggestion
Request access to the 'Register early career teachers' service.
```

cpjmcquillan (approved):

app/controllers/pages_controller.rb
```diff
@@ -1,14 +1,21 @@
 class PagesController < ApplicationController
   after_action :allow_search_engine_indexing, only: :home
 
+  DFE_SIGN_IN_REQUEST_ORGANISATION_URL =
+    "https://services.signin.education.gov.uk/request-organisation/search"
+
+  DFE_SIGN_IN_MY_SERVICES_URL =
+    "https://services.signin.education.gov.uk/my-services"
```
peteryates: I think these should be env-specific, can we add them as environment variables?

https://github.com/DFE-Digital/register-early-career-teachers-public/blob/23582cbdd8b80d5eae4b6b120a16bd1c30f09e7e/config/terraform/application/config/production.yml#L13-L19

Happy for that to happen as a follow-up if this is urgent.

peteryates (approved):
Looks good, left a comment on making this work across all the envs.

app/controllers/pages_controller.rb
ericaporter: Definitely. Now updated 👍

config/application.rb
```diff
@@ -77,6 +77,14 @@ class Application < Rails::Application
...
 
     config.dfe_sign_in_profile = ENV.fetch("DFE_SIGN_IN_PROFILE", "https://test-profile.signin.education.gov.uk")
+    config.dfe_sign_in_request_organisation_url = ENV.fetch(
+      "DFE_SIGN_IN_REQUEST_ORGANISATION_URL",
+      "https://services.signin.education.gov.uk/request-organisation/search"
+    )
+    config.dfe_sign_in_my_services_url = ENV.fetch(
+      "DFE_SIGN_IN_MY_SERVICES_URL",
+      "https://services.signin.education.gov.uk/my-services"
+    )
```
cpjmcquillan, line 87: The other `DFE_SIGN_IN` configurations default to the `test` environment. Should we bring these inline with those?

cpjmcquillan (approved):
One small comment, but otherwise LGTM

PR #3071, opened 2026-06-12
Relax validation to enable schedule to be changed to reduced on completed ECTs

tonyheadford:
### Context

Currently, providers cannot change the schedule for an ECT once they have completed training. In some cases, ECTs on a reduced schedule complete before the provider is aware/able to move them to a reduced schedule.

This change modifies the validation to allow a provider to change the schedule for an ECT that has completed their training. This is restricted to only permitting a move to a reduced schedule.

In these instances we would also not want to add a new training period or reopen the last/completed training period, but only modify the last training period's schedule.

**NOTE**: I left a comment in the `Teachers::ChangeSchedule` regarding whether it should include any of the validation around the changing of the schedule for completed ECTs as wasn't sure if there should be something in there too regarding that or whether it was designed to as a one-size-fits-all kind of thing and those validations are left for the API (or UI) services.

* modify the validation to allow providers to submit a change to a reduced schedule for a completed ECT
* change logic in the `Teachers::ChangeSchedule` service to update the existing rather than add a new training_period in these cases.

### Guidance to review

When checking for a completed ECT, they need to have a completed induction period (i.e. an induction period with an outcome).

### DIFF
--- a/app/services/api/teachers/change_schedule.rb
+++ b/app/services/api/teachers/change_schedule.rb
@@ -158,7 +158,7 @@ module API::Teachers
       return if errors[:teacher_api_id].any?
       return unless training_period&.teacher_completed_training?
       # allow an ECT that has completed training to move to a reduced schedule
-      return if schedule&.reduced_schedule?
+      return if schedule&.reduced_schedule? && training_period&.for_ect?
 
       errors.add(:teacher_api_id, "You cannot change this participant's schedule as they have completed their training or induction.")
     end
--- a/app/services/api/teachers/change_schedule.rb
+++ b/app/services/api/teachers/change_schedule.rb
@@ -157,6 +157,8 @@ module API::Teachers
     def trainee_not_completed
       return if errors[:teacher_api_id].any?
       return unless training_period&.teacher_completed_training?
+      # allow an ECT that has completed training to move to a reduced schedule
+      return if schedule&.reduced_schedule?
 
       errors.add(:teacher_api_id, "You cannot change this participant's schedule as they have completed their training or induction.")
     end
--- a/spec/services/api/teachers/change_schedule_spec.rb
+++ b/spec/services/api/teachers/change_schedule_spec.rb
@@ -151,6 +151,16 @@ RSpec.describe API::Teachers::ChangeSchedule, type: :model do
               expect(subject).to have_one_error_per_attribute
               expect(subject).to have_error(:teacher_api_id, "You cannot change this participant's schedule as they have completed their training or induction.")
             end
+
+            context "when schedule is reduced" do
+              let(:schedule_identifier) { "ecf-reduced-april" }
+
+              if trainee_type == :ect
+                it { is_expected.to be_valid }
+              else
+                it { is_expected.to have_error(:schedule_identifier, "Mentors cannot be placed on a reduced schedule. Assign them to a different schedule.") }
+              end
+            end
           end
 
           context "guarded error messages" do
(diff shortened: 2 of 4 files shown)


app/services/api/teachers/change_schedule.rb
```diff
@@ -157,6 +157,8 @@ def can_move_to_frozen_contract_period
     def trainee_not_completed
       return if errors[:teacher_api_id].any?
       return unless training_period&.teacher_completed_training?
+      # allow an ECT that has completed training to move to a reduced schedule
```
ethax-ross, line 160: Can we change this to `return if schedule&.reduced_schedule && training_period&.for_ect?` otherwise I think it will result in mentors getting the 'mentor can't be on a reduced schedule' error instead of the 'you cannot change as they have completed' error?

spec/services/api/teachers/change_schedule_spec.rb
```diff
@@ -151,6 +151,16 @@
...
               expect(subject).to have_error(:teacher_api_id, "You cannot change this participant's schedule as they have completed their training or induction.")
             end
+
+            context "when schedule is reduced" do
+              let(:schedule_identifier) { "ecf-reduced-april" }
+
+              if trainee_type == :ect
+                it { is_expected.to be_valid }
+              else
+                it { is_expected.to have_error(:schedule_identifier, "Mentors cannot be placed on a reduced schedule. Assign them to a different schedule.") }
```
ethax-ross, line 162: Oddly in the ticket it says ECTs and mentors should be able to go onto a reduced schedule, but we block mentors from reduced schedules explicitly 🤔 one to double check with Andy/Nathtan

spec/services/teachers/change_schedule_spec.rb
```diff
@@ -89,6 +89,57 @@
...
+                expect(training_period.schedule).to eq new_schedule
+              end
+
+              it "doesn't alter the finished_on date of the training_period" do
+                service.change_schedule
+                expect(training_period.finished_on).to eq training_period_finished_on
+              end
+            end
+
+            # NOTE: this case is caught in the API::Teachers::ChangeSchedule but not in this service
```
ethax-ross: I think this is fine; we intentionally try and keep the underlying services quite flexible and the outer services enforce the specific use cases we want (so from the API we only allow completed ECTs to move onto a reduced schedule, but if this service was called from the admin UI we might allow them to change a completed ECT to another schedule type, for example)

ethax-ross (commented):
Looks good thanks Tony. Just a minor suggestion, but can we also double check that the ticket is incorrectly asking for mentors to be allowed onto reduced schedules? I think this is wrong:

> we need to allow Lead Providers to change the schedule to reduced for both mentors and ECTs

app/services/api/teachers/change_schedule.rb
tonyheadford: Changed. This now results in having both errors.  The reduced schedule mentor check is much earlier in the validations, is this OK or should be change the order/add guards?

spec/services/api/teachers/change_schedule_spec.rb
tonyheadford: I've asked Nathan to confirm

tonyheadford: tbh I thought this was only for ECTs too.

app/services/api/teachers/change_schedule.rb
ethax-ross: thanks, lets add a guard in that case; in general we try and only surface one error at a time

spec/services/api/teachers/change_schedule_spec.rb
tonyheadford: Nathan has confirmed that it is definitely ECTs only

ethax-ross (approved):
Looks good, if we can guard against multiple errors that would be ideal but otherwise good for PO review

PR #3079, opened 2026-06-15
LeadProviderDeliveryPartnerships CRUD

craigmdavidson:
### Context

Resolves https://github.com.mcas.ms/DFE-Digital/register-ects-project-board/issues/4045

CRUD screens for an AcitveLeadProvider's LeadDeliveryPartnership:

#### From the existing ActiveLeadProviders screen, we can now link to LeadProviderDeliveryPartnerships...

Screenshot 2026-06-16 at 11 49 24

#### Where we see them (enabled if the ContractPeriod has not year started)...

Screenshot 2026-06-16 at 11 49 32

#### ..and disabled if it has

Screenshot 2026-06-16 at 11 53 37

#### When we want to add a DeliveryPartner we see the add form...

Screenshot 2026-06-16 at 11 50 11

#### With a typeahead...
Screenshot 2026-06-16 at 11 50 19

#### When we want to remove a delivery partner we see the confirmation screen...
Screenshot 2026-06-16 at 11 49 43

#### And can click through to see the DeliveryPartner itself (the existing screen...)
Screenshot 2026-06-16 at 11 50 34

### Guidance to review

* Standard pattern for our Rails controller. 
* No Edit/Update required. The LeadProviderDeliveryPartnership table is just a many-to-many mapping table from an ActiveLeadProvider to a DeliveryPartner.

### DIFF
--- a/app/models/lead_provider_delivery_partnership.rb
+++ b/app/models/lead_provider_delivery_partnership.rb
@@ -4,9 +4,11 @@ class LeadProviderDeliveryPartnership < ApplicationRecord
   belongs_to :active_lead_provider
   belongs_to :delivery_partner
   has_many :school_partnerships
-  has_many :events
+  has_many :events, dependent: :nullify
   has_one :lead_provider, through: :active_lead_provider
   has_one :contract_period, through: :active_lead_provider
+
+  delegate :name, to: :delivery_partner, prefix: true
 
   touch -> { delivery_partner }, on_event: %i[create destroy], timestamp_attribute: :api_updated_at
 
--- a/app/services/lead_provider_delivery_partnerships/create.rb
+++ b/app/services/lead_provider_delivery_partnerships/create.rb
@@ -1,0 +1,27 @@
+module LeadProviderDeliveryPartnerships
+  class Create
+    attr_reader :lead_provider_delivery_partnership, :author
+
+    def initialize(author:, active_lead_provider:, params:)
+      @author = author
+      @lead_provider_delivery_partnership = LeadProviderDeliveryPartnership.new(
+        params.merge(active_lead_provider:)
+      )
+    end
+
+    def call
+      ActiveRecord::Base.transaction do
+        lead_provider_delivery_partnership.save!
+        Events::Record.record_lead_provider_delivery_partnership_added_event!(
+          author:,
+          delivery_partner: lead_provider_delivery_partnership.delivery_partner,
+          lead_provider: lead_provider_delivery_partnership.lead_provider,
+          contract_period: lead_provider_delivery_partnership.contract_period,
+          lead_provider_delivery_partnership:
+        )
+      end
+
+      lead_provider_delivery_partnership
+    end
+  end
+end
--- a/app/services/lead_provider_delivery_partnerships/destroy.rb
+++ b/app/services/lead_provider_delivery_partnerships/destroy.rb
@@ -1,0 +1,33 @@
+module LeadProviderDeliveryPartnerships
+  class Destroy
+    class DeletionError < StandardError; end
+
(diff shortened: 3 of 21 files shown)


app/models/lead_provider_delivery_partnership.rb
```diff
@@ -4,10 +4,12 @@ class LeadProviderDeliveryPartnership < ApplicationRecord
   belongs_to :active_lead_provider
   belongs_to :delivery_partner
   has_many :school_partnerships
-  has_many :events
+  has_many :events, dependent: :nullify
```
ethax-ross, line 7: Can we update the specs for these changes

app/services/lead_provider_delivery_partnerships/destroy.rb
```diff
@@ -0,0 +1,33 @@
+module LeadProviderDeliveryPartnerships
+  class Destroy
```
ethax-ross, line 2: Can we add specs for the create/destroy services

app/services/lead_provider_delivery_partnerships/create.rb
```diff
@@ -0,0 +1,27 @@
+module LeadProviderDeliveryPartnerships
+  class Create
```
ethax-ross, line 2: It looks like we have a service that can create/destroy delivery partnerships [here](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/86b84c45845dff1050a36d0c2414be7e7dd47ff5/app/services/admin/delivery_partners/add_lead_providers.rb) which is tailored to the current admin DPs views, but I wonder if we should call out to that 😕 I didn't realise we namespaced some services under `admin`

ethax-ross (commented):
This looks/works great, just a couple of minor questions/suggestions 👍

app/services/lead_provider_delivery_partnerships/create.rb
peteryates: If anything I think that one in `Admin` should call here. It looks like it does some extra validation to avoid duplicates (as the admin section we add multiple `ActiveLeadProvider` records to a `DeliveryPartner` to get some new `LeadProviderDeliveryPartnersip` records.

Perhaps this class should be invoked from here?

https://github.com/DFE-Digital/register-early-career-teachers-public/blob/35689fea7f824cd8801f77a6e11aed6641b01083/app/services/delivery_partners/add_lead_provider_pairings.rb#L42-L45

Then at least there's only one way to build `LeadProviderDeliveryPartnership` models and the event creation is with it.

app/models/lead_provider_delivery_partnership.rb
craigmdavidson: Done

app/services/lead_provider_delivery_partnerships/create.rb
craigmdavidson: Done.

app/services/lead_provider_delivery_partnerships/destroy.rb
craigmdavidson: Done.

app/services/lead_provider_delivery_partnerships/create.rb
craigmdavidson: I'm leaving this as a separate commit. As the changes are not specific to this piece of work.

craigmdavidson: a025124909d525314ffa2ecba4a431cf603f1ac6

tonyheadford (approved):
Looks good, appreciate the screen shots in the PR description too

ethax-ross (approved):
Looks great 👍

PR #3080, opened 2026-06-15
[4031] Part 1 - Add ActiveLeadProvider::Band model and backfill

peterdavidhamilton:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/4031

[Band modelling](https://github.com/DFE-Digital/register-ects-project-board/issues/4031)

-  part 1 (this PR)
- [part 2](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3115)
- [part 3](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3131)
- [part 4](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3148)

- Introduce a new model for tiers (`ActiveLeadProvider::Band`) of a band (`Contract::BandedFeeStructure::Band` to be renamed `BandTerms`)
- Backfill data to the new table
- Correct 2 errant `max` values in the process
- Connect new model to the fee structure band and active provider

**The pivot to `ActiveLeadProvider::Band` with `#allocation_order` and `#capacity`**

- The data model is changing to simplify how band tiers are bucketed.
- The volume of declarations in a band is recorded by #capacity.
- The position of the band in the tiers is recorded by #allocation_order.

image

### Guidance to review

Migration run against snapshot:
https://www.register-early-career-teachers.education.gov.uk/admin/blazer/queries/377-pdh-bands

Correctly migrate the max values identified: https://github.com/DFE-Digital/register-ects-project-board/issues/4053

- `#4` Education Development Trust: `2000 + 2000 + 3125 + 1975 = 9,100`
- `#5` Ambition Institute: `2000 + 2000 + 6000 + 5600 = 15,600`
- `111` in total
- 
Started to introduce new Blazer configs for this data remodel which will aid in reviewing the data on staging (or review).

Screenshot 2026-06-24 at 12 00 11

```
ActiveLeadProvider::Band
  associations
    is expected to belong to active_lead_provider required: false
    is expected to have many terms class_name => Contract::BandedFeeStructure::Band inverse_of => band
  validations
    is expected to validate that :capacity cannot be empty/falsy, producing a custom validation error on failure
    is expected to validate that :active_lead_provider cannot be empty/falsy, producing a custom validation error on failure
    is expected to validate that :allocation_order looks like an integer greater than 0, producing a custom validation error on failure
    is expected to validate that :capacity looks like an integer greater than 0, producing a custom validation error on failure
  immutability
    with multiple bands
      allows deleting the last band
      prevents changing the capacity of a band that is not the last
      prevents deleting a band that is not the last
      prevents changing the allocation order of a band that is not the last
    with only one band
      prevents changing the allocation order of the last band
      allows changing the capacity of the last band
  #min_declarations
    with a subsequent band
      sums previous capacities plus one
    without allocation_order
      is expected to be nil
    with the first band
      is expected to eq 1
  #max_declarations
    with the first band
      equals the capacity
    with a subsequent band
      sums all capacities
    when unallocated
      is expected to be nil
  #allocation_order
    auto-assigns the next position
    when the allocation order of a persisted band is edited
      raises an error
```


app/models/contract/banded_fee_structure/band.rb
```diff
@@ -1,16 +1,23 @@
 class Contract::BandedFeeStructure::Band < ApplicationRecord
   self.table_name = "contract_banded_fee_structure_bands"
 
+  # Associations
   belongs_to :banded_fee_structure,
              class_name: "Contract::BandedFeeStructure"
 
+  belongs_to :contract_band_capacity,
+             class_name: "Contract::BandCapacity"
```
ethax-ross: Worth having this just as `capacity` so it's a bit more concise as we're already in the context of a contract/band? i.e. `band.capacity` rather than `band.contract_band_capacity`?

db/migrate/20260612102325_create_contract_band_capacities.rb
```diff
@@ -0,0 +1,15 @@
+class CreateContractBandCapacities < ActiveRecord::Migration[8.1]
+  def change
+    create_table :contract_band_capacities do |t|
+      t.references :active_lead_provider, foreign_key: true
+      t.integer :min_declarations, null: false, default: 1
+      t.integer :max_declarations, null: false
+      t.timestamps
+    end
+
+    add_reference :contract_banded_fee_structure_bands,
```
ethax-ross: Can we add an index to ensure we only reference each band capacity once per fee structure? something like:

```
add_index :contract_banded_fee_structure_bands,
          [:contract_banded_fee_structure_id, :contract_band_capacity_id],
          unique: true
```

We could also add a unique index on `min_declarations` for capacity and active LP

app/models/contract/band_capacity.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  # Associations
+  belongs_to :active_lead_provider
+
+  has_many :contract_banded_fee_structure_bands,
+           class_name: "Contract::BandedFeeStructure::Band",
+           inverse_of: :contract_band_capacity
+
+  # Validations
+  validates :min_declarations,
```
ethax-ross: Can we bring across the validation to ensure the min/max declarations are in sequence and without gaps? and maybe a rule to say the first band has to have a min declarations of 1

db/migrate/20260612111906_backfill_contract_band_capacities.rb
```diff
@@ -0,0 +1,32 @@
+class BackfillContractBandCapacities < ActiveRecord::Migration[8.1]
+  def up
+    Contract::BandedFeeStructure::Band.all.find_each do |band|
```
ethax-ross: `all` is redundant here, might be worth some `includes` to avoid N+1s too

db/migrate/20260612111906_backfill_contract_band_capacities.rb
```diff
@@ -0,0 +1,32 @@
...
+      active_lead_provider = band.banded_fee_structure.contract.active_lead_provider
+      capacity = Contract::BandCapacity.where(active_lead_provider:, min_declarations: band.min_declarations).max_by(&:max_declarations)
```
ethax-ross: Not sure we need the `max_by` as there should only ever be one band capacity with a given `min_declarations` now?

ethax-ross (commented):
Thanks PDH, I've left some initial thoughts, I'll try and have a think about naming in a bit

app/models/contract/band.rb
```diff
@@ -0,0 +1,54 @@
+# Allocated buckets with varying capacities.
+# The first band (A) pays the most.
+#
+class Contract::Band < ApplicationRecord
```
craigmdavidson: This is not a **Contract::Band** as it spans multiple **Contracts**?

The **Contract** has **Contract::BandedFeeStructure::BandTerms** which is the other half of **Contract::BandedFeeStructure::Band**,

I.e. the **ActiveLeadProvider::Band** has_many **Contract::BandedFeeStructure::BandTerms**?

app/models/contract/band.rb
```diff
@@ -0,0 +1,54 @@
...
+            }
+  validates :capacity,
+            presence: { message: "Capacity is required" },
+            numericality: {
+              greater_than: 0,
+              only_integer: true,
+              message: "Capacity must be a number greater than zero"
+            }
+
+  validate :allocation_orders_are_sequential_and_contiguous_from_one
```
craigmdavidson: Does this not make `validates :allocation_order, ...`  redundant?

app/models/contract/band.rb
```diff
@@ -0,0 +1,54 @@
...
+
+    orders = active_lead_provider_bands_excluding_self.to_a.push(self).map(&:allocation_order)
+    expected = (1..orders.max).to_a
+
+    if orders.min != 1
+      errors.add(:allocation_order, "The first band's allocation order must be 1")
+    elsif orders.sort != expected
+      errors.add(:allocation_order, "Allocation orders must be sequential without gaps")
+    end
+  end
```
craigmdavidson: Since `allocation_order` is auto-assigned and immutable (or will be), they will be sequential and contiguous if we just check that:

```
# next_allocation_order will likely be a method anyway since we will want to auto-assign it
next_allocation_order = (active_lead_provider.bands.maximum(:allocation_order) || 0) + 1
if allocation_order != next_allocation_order
  errors.add(:allocation_order, "The allocation_order should be #{expected_allocation_order}")
end
```

That would mean no active-records created, just a fast little query:

```
SELECT maximum(allocation_order)
FROM bands
WHERE active_lead_provider_id = ?
```

app/models/active_lead_provider.rb
```diff
@@ -8,13 +9,16 @@ class ActiveLeadProvider < ApplicationRecord
...
+  has_many :bands, -> { order(allocation_order: :asc) }, class_name: "Contract::Band"
 
+  # Validations
   validates :contract_period_year,
             presence: { message: "Choose a contract period" },
             uniqueness: { scope: :lead_provider_id, message: "Contract period and lead provider must be unique" }
   validates :lead_provider_id, presence: { message: "Choose a lead provider" }
 
-  scope :for_contract_period, ->(year) { where(contract_period_year: year) }
+  # Scopes
```
craigmdavidson, line 20: 🤔 Do we need to explicitly point these things out?

peterdavidhamilton: It was a convention when I joined so I adopted it. I am not against it.

app/models/contract/band.rb
```diff
@@ -0,0 +1,54 @@
+# Allocated buckets with varying capacities.
+# The first band (A) pays the most.
+#
+class Contract::Band < ApplicationRecord
+  self.table_name = "contract_bands"
+
+  # Associations
+  belongs_to :active_lead_provider
+
+  has_many :contract_banded_fee_structure_bands,
```
ethax-ross: Worth just calling this out as the terms to begin with to avoid renaming later? What do we think about shortening the association name,  so something like `band.terms` rather than `band.contract_banded_fee_structure_terms`?

app/models/contract/band.rb
```diff
@@ -0,0 +1,54 @@
...
+
+private
+
+  def allocation_orders_are_sequential_and_contiguous_from_one
+    return unless active_lead_provider
+    return unless allocation_order
+
+    orders = active_lead_provider_bands_excluding_self.to_a.push(self).map(&:allocation_order)
+    expected = (1..orders.max).to_a
+
```
ethax-ross: I think this will hold as we will only allow updating the last band (so its safe to `push` it on the end), but I wonder if it would be more robust as something like:

```
def allocation_orders_are_sequential_and_contiguous_from_one
  return unless active_lead_provider && allocation_order

  orders = active_lead_provider_bands_excluding_self
             .pluck(:allocation_order)
             .append(allocation_order)
             .sort

  expected = (1..orders.size).to_a

  return if orders == expected

  message = if orders.first != 1 ?
                       "The first band's allocation order must be 1" 
                     else 
                        "Allocation orders must be sequential without gaps"
                     end

  errors.add(:allocation_order, message)
end
```

We could also make it immutable after creation with `attr_readonly :allocation_order`

app/models/contract/band.rb
```diff
@@ -0,0 +1,54 @@
...
+    if orders.min != 1
+      errors.add(:allocation_order, "The first band's allocation order must be 1")
+    elsif orders.sort != expected
+      errors.add(:allocation_order, "Allocation orders must be sequential without gaps")
+    end
+  end
+
+  def active_lead_provider_bands_excluding_self
+    Contract::Band.where(active_lead_provider:).where.not(id: id || 0)
```
ethax-ross: `Contract::Band.where(active_lead_provider:).excluding(self)`?

spec/models/contract/band_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+          let(:allocation_order) { 2 }
+
+          it "is invalid" do
+            expect(new_band).to be_invalid
+            expect(new_band.errors[:allocation_order]).to include("The first band's allocation order must be 1")
+          end
+        end
+      end
+
+      context "with existing bands" do
```
ethax-ross: Worth covering if we updated an existing band to become out of order?

spec/models/contract/band_spec.rb
```diff
@@ -0,0 +1,70 @@
+RSpec.describe Contract::Band, type: :model do
+  describe "associations" do
+    it { is_expected.to belong_to(:active_lead_provider) }
+    it { is_expected.to have_many(:contract_banded_fee_structure_bands).class_name("Contract::BandedFeeStructure::Band").inverse_of(:contract_band) }
+  end
+
+  describe "validations" do
+    let(:active_lead_provider) { FactoryBot.create(:active_lead_provider) }
+
+    it { is_expected.to validate_presence_of(:allocation_order).with_message("Allocation order is required") }
```
ethax-ross: Are we missing the presence check for `active_lead_provider`?

config/blazer.yml
```diff
@@ -12,13 +12,15 @@ data_sources:
       batch_status: ['pending', 'processing', 'processed', 'completing', 'completed', 'failed']
       batch_type: ['action', 'claim']
       urn: "SELECT urn, CONCAT(name, ' (', urn, ')') FROM gias_schools ORDER BY name ASC"
+      active_lead_provider_id: "SELECT active_lead_providers.id, lead_providers.name FROM active_lead_providers JOIN lead_providers ON lead_providers.id=active_lead_providers.lead_provider_id ORDER BY lead_providers.name ASC"
```
ethax-ross, line 15: I'm not familiar with Blazer, what does this stuff do?

db/migrate/20260612111906_backfill_contract_bands.rb
```diff
@@ -0,0 +1,41 @@
...
+        # Bucket position within its fee structure (like #letter):
+        #
+        allocation_order = original_band.banded_fee_structure.bands.sort_by(&:min_declarations).index(original_band) + 1
+
+        # Bucket volume:
+        #   81 - 1 + 1 = 81
+        #   162 - 82 + 1 = 81
+        #   243 - 163 + 1 = 81
+        #
+        capacity = original_band.max_declarations - original_band.min_declarations + 1
```
ethax-ross: We have a `capacity` method on `Band` so I think you could just do `original_band.capacity`

ethax-ross (commented):
Looking good just a few thoughts/suggestions

config/blazer.yml
peterdavidhamilton: Dynamic names and links basically. It can make checking data a bit easier.

app/models/contract/band.rb
craigmdavidson: Thinking about this....
In addition to never being able to modify the allocation_order, nor explicitly set it in the first place. 

We also want to: 
- make the record itself immutable unless it is the last one. I.e. you can increase the capacity (only) of the last record, but not the second last one. 
- You should not even be able to delete it unless it is the last one.

If we add these constraints then the `allocation_order` is the increment of the band count. 

And if it the record, and in particular the `allocation_order` _is immutable_ then we should we not leverage that?

concretely `self.allocation_order = active_lead_provider.bands.count`

We can then use a standard counter_cache to avoid any subsequent lookups, 
and means we technically do not need this validation at all.

But if we want it - it can be simply
```
def allocation_order_must_match_expected
    return if active_lead_provider.nil?

    expected_order = active_lead_provider.bands.count + 1
    if allocation_order != expected_order
      errors.add(:allocation_order, "must equal #{expected_order}")
    end
  end
```

Surely asserting the sequence of all Bands is outwith any given Band's responsibility. 
If anything, it should only only care that's `allocation_order` makes sense given what's already there?

spec/models/contract/band_spec.rb
craigmdavidson: It should not be possible to update an existing band to become out of order, so if anything we should test that?

spec/models/contract/band_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+          it "is invalid" do
+            expect(new_band).to be_invalid
+            expect(new_band.errors[:allocation_order]).to include("The first band's allocation order must be 1")
+          end
+        end
+      end
+
+      context "with existing bands" do
+        before do
+          FactoryBot.create(:contract_band, active_lead_provider:, allocation_order: 1)
```
craigmdavidson: `FactoryBot.create(:contract_band, active_lead_provider:, allocation_order: 1)` 
Does this bypass the idea that we shouldn't specify `allocation_order`?

app/models/contract/band.rb
peterdavidhamilton: I like the clarity of the message. Were you thinking we override the setter?

spec/models/contract/band_spec.rb
peterdavidhamilton: I see your point. Something for the tidy up PR or something to nail down now?

craigmdavidson: Because the creation/mutation/destruction of these bands are fundamental to their behaviour I would be tempted to lock it down now? 

Because we use service objects rather than callbacks, the overhead is significantly higher but we want our factory to be `let(:band_a) { FactoryBot.create :active_lead_provider_band, active_lead_provider: } ` so we would need something like:

```
FactoryBot.define do
  factory :active_lead_provider_band, class: "ActiveLeadProvider::Band" do
    to_create do |_instance, evaluator|
      ActiveLeadProvider::Bands::Create.call(capacity: evaluator.capacity)
    end
  end
end
```

This also lays the foundation for:
> It should not be possible to update an existing band to become out of order, so if anything we should test that?

craigmdavidson: To be honest, if we are going to play with tiny service objects for creation, then we _should_ be doing this for every object factory @ethax-ross  . 

And we are definitely _not_ doing that, 😬 so maybe this is an unfair ask from me.

app/models/contract/band.rb
craigmdavidson: It's tricky, in a standard Rails app I would use callbacks judiciously to this:
 `before_validation :set_allocation_order, if: :new_record?` 

Then make it immutable:
```
validate :prevent_update, on: :update, unless: :last?
before_destroy :abort_destruction, unless: :last?
attr_readonly ::allocation_order
```

Avoiding the callbacks makes it pretty tricky, so because of thatg we need to be strict about the service object use! :rage:  😆 .

app/models/contract/band.rb
tonyheadford: Yes good shout

app/models/contract/band.rb
tonyheadford: Just a thought... do we actually need the `allocation_order` to be anything other than sequential and if so gaps don't matter?  In that case, it sounds possible to make it an auto increment value at the DB and a read only attr in the model - does that make it simpler or muddier?

craigmdavidson: 🤔 Auto-incrementing at the DB level would mean that we would have an arbitrary increasing sequence of numbers (since PostgreSQL doesn't support scoping) e.g. 234, 252, 321, 322, 327. 

This would be pretty confusing compared to having band allocation_order of 1,2,3,4,5, and the simple sequence maps nicely to display_names: 
`def display_name = "Band #{allocation_order.to_s(36)}"`

And as shown above we can do this easily in the app at effectively zero cost: 
```
def set_allocation_order
  self.allocation_order = active_lead_provider.bands.count
end
```

So that means the only down side is because we don't use callbacks - orchestration becomes more convoluted, so we need to be strict in our service object usage.

But wait....
https://github.com/DFE-Digital/register-early-career-teachers-public/blame/main/app/models/metadata/base.rb#L7
`before_update :ensure_metadata_namespace` !!! 

@ethax-ross j'accuse! 😆

db/migrate/20260612111906_backfill_contract_bands.rb
tonyheadford, line 1: Should this be in a migration or under `db/scripts`?
As we're referencing models that will be renamed as you note this will break migrations into a new database in the future (you'd need to load the schema instead so not the end of the world, but worth considering?)

app/models/contract/band.rb
tonyheadford: I was forgetting that they're surfaced in the UI, good point

app/models/contract/band.rb
peterdavidhamilton: Done

app/models/contract/band.rb
craigmdavidson: `active_lead_provider.bands.excluding(self)` 

;-P

app/models/contract/band.rb
peterdavidhamilton: I have had a stab at it, similar but different.

app/models/contract/band.rb
peterdavidhamilton: What I'm picturing might not be accurate but I think it could be cleaner left to Rails. My attempt is deploying now.

db/migrate/20260612111906_backfill_contract_bands.rb
peterdavidhamilton: I think I am ok with it because of how `db:prepare` switched to loading the schema in `7.1` and that's what we use in `bin/setup`.

app/models/active_lead_provider/band.rb
```diff
@@ -0,0 +1,55 @@
...
+              greater_than: 0,
+              only_integer: true,
+              message: "Capacity must be a number greater than zero"
+            }
+
+  validate :allocation_orders_are_sequential_and_contiguous_from_one
+
+  before_validation :assign_allocation_order,
+                    on: :create,
+                    if: -> { allocation_order.blank? }
```
craigmdavidson: Do the assign even if `allocation_order` is not blank?

```
# Given there are three bands already
assert_equal 3, active_lead_provider.bands

# Let's say somebody breaks all the rules and tries to assign a validation order
band_d = active_lead_provider.create! capacity: 42, allocation_order: 42
assert_equal 4, band_d.allocation_order
```

app/models/active_lead_provider/band.rb
```diff
@@ -0,0 +1,55 @@
...
+
+private
+
+  def assign_allocation_order
+    self.allocation_order = self.class.where(active_lead_provider:).maximum(:allocation_order).to_i + 1
+  end
```
craigmdavidson, line 58: `self.allocation_order = active_lead_provider.bands.count + 1`

app/models/active_lead_provider/band.rb
```diff
@@ -0,0 +1,55 @@
...
+  def allocation_orders_are_sequential_and_contiguous_from_one
+    return unless active_lead_provider && allocation_order
+
+    siblings = self.class.where(active_lead_provider:).excluding(self)
+    orders = siblings.pluck(:allocation_order).append(allocation_order).sort
+    expected = (1..orders.size).to_a
+
+    return if orders == expected
+
+    errors.add(:allocation_order, "The first band's allocation order must be 1") if orders.first != 1
```
craigmdavidson: This will currently pass the validation if the sequence is 1,2,3,5,8?
If wont quick return at line #50 - since the sequence doesn't match, but will not be caught by an errors.add?

The more fundamental problem is this validation validates the state of another object, not this object. If this happens the horse has bolted? We have a DI issue.

The only thing this object should be validating is it's own state. 
(i.e. ^)

If we make it impossible to be invalid - then we don't need to validate it?

app/models/contract/band.rb
ethax-ross: > But wait.... https://github.com/DFE-Digital/register-early-career-teachers-public/blame/main/app/models/metadata/base.rb#L7 `before_update :ensure_metadata_namespace` !!!
> 
> @ethax-ross j'accuse! 😆

Metadata got an exception to the rule because it just wasn't viable to do it reliably at the service level 😅

spec/models/contract/band_spec.rb
ethax-ross: That's what I mean - I'll work on my phrasing 🙂

spec/models/contract/band_spec.rb
ethax-ross: I think lets lock it down in validation in this PR but worry about creation later so this isn't growing any more arms/legs here

app/models/active_lead_provider/band.rb
peterdavidhamilton: The reason I triggered if it was missing was for the test coverage, and I think that if the interface omits the field in the form then we lean into this being the default but can still assert values and validations. I'd like to retain tests for those because it is good documentation.

app/models/active_lead_provider/band.rb
peterdavidhamilton: I was thinking about race conditions but yeah simpler is better for now 👍🏼

app/models/active_lead_provider/band.rb
```diff
@@ -0,0 +1,88 @@
...
+
+  # @return [Boolean]
+  def editable?
+    active_lead_provider.bands.last == self
+  end
+
+  # @return [Boolean]
+  def allocated?
+    allocation_order.present? && persisted?
+  end
```
craigmdavidson, line 43: Nitpick: `allocated?` is confusing here? 

**Declarations** are _allocated_ to bands, the order of this is handled by `allocation_order`. This method isn't checking about allocations, it's about the presence of `allocation_order` - so should the method not be named something more like `allocation_ordered?`

This also implies that not having an `allocation_order` is a realistic possibility, this should not be the case?

app/models/active_lead_provider/band.rb
```diff
@@ -0,0 +1,88 @@
...
+
+  # @return [Boolean]
+  def first_band?
+    return false unless allocated?
+
+    allocation_order == 1
```
craigmdavidson: Nitpick: Is the guard not irrelevant here?  a `nil` value for `allocation order` != 1, so is false?

app/models/active_lead_provider/band.rb
```diff
@@ -0,0 +1,88 @@
...
+              only_integer: true,
+              message: "Capacity must be a number greater than zero"
+            }
+
+  validate :first_band_allocation_order_is_one, on: :create
+
+  # Callbacks
+  before_validation :assign_allocation_order,
+                    on: :create,
+                    if: -> { active_lead_provider.present? && allocation_order.blank? }
```
craigmdavidson: Nitpick: So this means a bad `allocation_order` may be passed in, and not addressed?

If we don't do that, no further `allocation_order` validation is required, 
however if we really do want to be able to try to force a bad value in, we need a validation to prevent that. i.e. `validate :allocation_order_is_correct`?

app/models/active_lead_provider/band.rb
```diff
@@ -0,0 +1,88 @@
...
+            }
+  validates :capacity,
+            presence: { message: "Capacity is required" },
+            numericality: {
+              greater_than: 0,
+              only_integer: true,
+              message: "Capacity must be a number greater than zero"
+            }
+
+  validate :first_band_allocation_order_is_one, on: :create
```
craigmdavidson: Nitpick: I'm not sure that validating this buys us? If we are going to do this, why not just `validate :allocation_order_is_correct`?

app/models/contract/banded_fee_structure/band.rb
```diff
@@ -94,7 +108,7 @@ def band_consistency_across_active_lead_provider
         .to_a
         .push(self)
 
-    attributes_to_ignore = %w[id banded_fee_structure_id created_at updated_at].freeze
+    attributes_to_ignore = %w[id banded_fee_structure_id created_at updated_at contract_band_id].freeze
```
craigmdavidson: In the schema it looks like the relationship is to `band_id`?
https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3080/changes#diff-cbf8b25d6d853acf58fa9057841f407d29ffe90649c75cf9e3f51b2d6e3aa1d3R160

app/models/active_lead_provider/band.rb
```diff
@@ -0,0 +1,88 @@
...
+  end
+
+  # @return [Boolean]
+  def first_band?
+    return false unless allocated?
+
+    allocation_order == 1
+  end
+
+  def previous_band
```
craigmdavidson: Previous band could make sense if we go linked-list style? And is useful for things like `editable?` or `first?` 

But in this case, we are doing something quite different - using recursion rather than leveraging a data structure, which is going get us into a bit of trouble from an N+1 perspective, since it is being used to derive min_declarations? I.e.

```
self.min_declarations => 
  self.previous_band => SELECT ....
  previous.max_declarations => 
    previous.min_declarations => 
      previous.previous_band => SELECT ....
      previous_previous.max_declarations =>
          previous_previous.min_declarations =>
          previous_previous.previous_band => SELECT ....
    ... and so on?
  ```

File: app/models/active_lead_provider/band.rb [hunk 27]
```diff
@@ -0,0 +1,88 @@
...
+    return nil if first_band?
+
+    active_lead_provider.bands.find_by(allocation_order: allocation_order - 1)
+  end
+
+  def min_declarations
+    return nil unless allocated?
+    return 1 if first_band?
+
+    previous_band.max_declarations + 1
```
craigmdavidson: This is just the sum of the capacities so far?

```
previous_capacities = active_lead_provider.bands.where("allocation_order < ?", allocation_order).sum(:capacity)
```

I.e. 10 (first capacity), 10 (second capacity), 10 (third capacity) -  we are band 4:
`min_declarations = previous_capacities + 1 = 10 + 10 + 10 + 1 => 31`

spec/models/active_lead_provider/band_spec.rb
```diff
@@ -0,0 +1,204 @@
...
+
+      it { is_expected.not_to be_editable }
+    end
+  end
+
+  describe "#allocated?" do
+    context "when the band is not persisted" do
+      subject(:band) { FactoryBot.build(:active_lead_provider_band, active_lead_provider:) }
+
+      it { is_expected.not_to be_allocated }
```
craigmdavidson: Is this realistic enough for a test? I would have thought that this is not desired behaviour?

We know that this is at most transient behaviour, we also know that before any band is persisted, it will have an (internally assigned) allocation order. We also know that the band will only exist in a non-persistent state for one instruction on the call stack?

app/models/active_lead_provider/band.rb
```diff
@@ -0,0 +1,88 @@
...
+    active_lead_provider.bands.last == self
+  end
+
+  # @return [Boolean]
+  def allocated?
+    allocation_order.present? && persisted?
+  end
+
+  # @return [Boolean]
+  def first_band?
```
craigmdavidson: Nitpick: Should this method not be private? Is it not an implementation detail? 

(i.e. we should be able to inline it with no consequence)?

app/models/active_lead_provider/band.rb
craigmdavidson: Should this not be private? 

Again this is an implementation detail?

spec/models/active_lead_provider/band_spec.rb
```diff
@@ -0,0 +1,204 @@
...
+    end
+
+    context "when the band is persisted" do
+      subject(:band) { FactoryBot.create(:active_lead_provider_band, active_lead_provider:) }
+
+      it { is_expected.to be_allocated }
+    end
+  end
+
+  describe "#first_band?" do
```
craigmdavidson: This is an implementation detail rather than some specified / required behaviour? 

This only exists to optimise min_declarations and max_declarations and has no extrinsic benefit other than that? We should be able to inline `first_band?` with no descernable difference in behaviour, but this locks it's presence?

spec/models/active_lead_provider/band_spec.rb
```diff
@@ -0,0 +1,204 @@
...
+
+      before do
+        FactoryBot.create(:active_lead_provider_band, active_lead_provider:)
+      end
+
+      it { is_expected.to be false }
+    end
+  end
+
+  describe "#previous_band" do
```
craigmdavidson: As above, this is an implementation detail, rather than required behaviour?

spec/models/active_lead_provider/band_spec.rb
```diff
@@ -0,0 +1,204 @@
...
+      let(:band) { FactoryBot.create(:active_lead_provider_band, active_lead_provider:) }
+
+      it { is_expected.to have_attributes(active_lead_provider:, allocation_order: 2, capacity: 123) }
+    end
+  end
+
+  describe "#min_declarations" do
+    subject { band.min_declarations }
+
+    context "when unallocated" do
```
craigmdavidson: Is this a realistic situation, if so maybe we should be assigning `after_initialize`?

craigmdavidson (commented):
Sorry for the Friday evening onslaught!

Mostly Looks good. The key questions are the excluded attribute name and the N+1 issue.

app/models/active_lead_provider/band.rb
craigmdavidson: I.e. using `previous_capacities = active_lead_provider.bands.where("allocation_order < ?", allocation_order).sum(:capacity)` is a single fast query, with no active record created.

craigmdavidson: ... and we want it to be fast because we are going to be displaying all the **Bands** for a given **ActiveLeadProvider** at once. Even then we will still have an N+1 (unless we cache) but for say 12 bands that isn't such a big deal.

craigmdavidson: ... which begs the question - is `min_capacity` and `max_capacity` really something for a view component rather than the domain model - when we will have all the bands as loaded entities already? But that's a question for another PR.

app/models/active_lead_provider/band.rb
```diff
@@ -0,0 +1,88 @@
+class ActiveLeadProvider::Band < ApplicationRecord
+  self.table_name = "active_lead_provider_bands"
+
+  attr_readonly :allocation_order
+
+  # Associations
+  belongs_to :active_lead_provider, optional: false
```
ethax-ross: We can omit `optional: false` here as that's the default behaviour

app/models/active_lead_provider/band.rb
ethax-ross: I found this method naming misleading as well; I'll throw in the suggestion of `has_allocation_order?` as `allocation_ordered?` implies some kind of state transition to me (is the allocation on the way? 😅)

app/models/active_lead_provider/band.rb
```diff
@@ -0,0 +1,88 @@
...
+  end
+
+  def previous_band
+    return nil unless allocated?
+    return nil if first_band?
+
+    active_lead_provider.bands.find_by(allocation_order: allocation_order - 1)
+  end
+
+  def min_declarations
```
ethax-ross: I would have `min/max_declarations` public but the rest private until we have a need for them outside this class?

spec/models/active_lead_provider/band_spec.rb
```diff
@@ -0,0 +1,204 @@
...
+    end
+  end
+
+  describe "#first_band?" do
+    subject { band.first_band? }
+
+    context "when the band is allocated the first position" do
+      let(:band) { FactoryBot.create(:active_lead_provider_band, active_lead_provider:) }
+
+      it { is_expected.to be true }
```
ethax-ross: `it { is_expected.to be_first_band }` and `it { is_expected.not_to be_first_band }`

spec/factories/active_lead_provider/active_lead_provider_band_factory.rb
```diff
@@ -0,0 +1,7 @@
+FactoryBot.define do
+  factory :active_lead_provider_band, class: "ActiveLeadProvider::Band" do
+    association :active_lead_provider
+    allocation_order { nil }
```
ethax-ross: Worth just dropping `allocation_order` altogether here?

ethax-ross (commented):
Looks good, I didn't really spot anything beyond Craig's comments

spec/models/active_lead_provider/band_spec.rb
peterdavidhamilton: Removed

ethax-ross (approved):
LGTM, thanks PDH!

ethax-ross (approved):

app/models/active_lead_provider/band.rb
```diff
@@ -0,0 +1,93 @@
...
+    allocation_order == 1
+  end
+
+  # @return [Boolean]
+  def last?
+    active_lead_provider.bands.last == self
+  end
+
+  # @return [ActiveLeadProvider::Band]
+  def previous_band
```
craigmdavidson: This is still going to trigger an N+1, and in a nasty way (calculating min_declarations will be O(n^2) in fancy computer science speak)? 

The only place we use this is to calculate the `min_declarations` we could do this instead:

```
def min_declarations
    return nil unless has_allocation_order?

    active_lead_provider.bands.where("allocation_order < ?", allocation_order).sum(:capacity) + 1
end
```

craigmdavidson (approved):
LGTM

PR #3081, opened 2026-06-15
4029-rel-note-api-updated-at

StephenC-V1:
### Context
Add reelease note for updates to api_updated_at attribute

New release note entry in release_notes.yml

### Guidance to review
Fact and sanity check in review app.

### DIFF
--- a/app/views/api/release_notes/release_notes.yml
+++ b/app/views/api/release_notes/release_notes.yml
@@ -37,7 +37,7 @@
 
     ## Why we’ve made this change
 
-    The number of participants currently training is an important part of the partnerships response. Not updating `api_updated_at` when this value changed meant API consumers could miss meaningful updates unless they re-polled all data.
+    The number of participants currently training is an important part of the partnerships response. Not updating `api_updated_at` when this value changed meant lead providers could miss meaningful updates unless they re-polled all data.
 
     This change ensures that updates to participant counts are returned consistently, making it easier for providers to detect and respond to changes.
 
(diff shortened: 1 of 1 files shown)


app/views/api/release_notes/release_notes.yml
```diff
@@ -21,6 +21,35 @@
...
+    
+    ## What’s changed
+
+    The `api_updated_at` value on partnerships will now be updated whenever the number of participants currently training (ongoing participants) changes.
+    
+    Previously, changes to this participant count were not reflected in `api_updated_at`, as the value is derived dynamically rather than stored on a specific attribute.
+
+    ## Why we’ve made this change
+
+    The number of participants currently training is an important part of the partnerships response. Not updating `api_updated_at` when this value changed meant API consumers could miss meaningful updates unless they re-polled all data.
```
easeynathan: "API consumers" now looks a bit odd [and yes, I 100% appreciate it was the turn of phrase I used in the draft]

@StephenC-V1 what are your thoughts between "lead providers" vs "API consumers"? After all, lead providers are the **_only_** type of API consumer for this endpoint (now and forever), so this would be more consistent with the rest of our guidance.

StephenC-V1: Yes, if the audience is specific (and we've used 'lead providers' in the past) then that's what we should use here.

I've noticed that 'API consumers' is used elsewhere in the file. I'll change that and this example.

ethax-ross (approved):
Looks good to me! thanks Stephen

PR #3082, opened 2026-06-15
Drop legacy school_urn from school_funding_eligibilities

rwrrll:
Part 2 following #3068: contract stage of pointing school_funding_eligibilities at gias_schools. Deploy after the expand migration is fully deployed.

### DIFF
--- a/db/migrate/20260615120000_remove_school_urn_from_school_funding_eligibilities.rb
+++ b/db/migrate/20260615120000_remove_school_urn_from_school_funding_eligibilities.rb
@@ -8,17 +8,6 @@ class RemoveSchoolURNFromSchoolFundingEligibilities < ActiveRecord::Migration[8.
   end
 
   def down
-    add_column :school_funding_eligibilities, :school_urn, :bigint
-    add_index :school_funding_eligibilities, :school_urn
-
-    execute(<<~SQL)
-      UPDATE school_funding_eligibilities
-      SET school_urn = gias_school_urn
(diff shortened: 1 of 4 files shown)


db/migrate/20260615120000_remove_school_urn_from_school_funding_eligibilities.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  def down
+    add_column :school_funding_eligibilities, :school_urn, :bigint
+    add_index :school_funding_eligibilities, :school_urn
+
+    execute(<<~SQL)
+      UPDATE school_funding_eligibilities
+      SET school_urn = gias_school_urn
+      WHERE school_urn IS NULL
+    SQL
```
avinhurry: After #3068 is merged, we allow `SchoolFundingEligibility` to belong to a `GIAS::School` without there being a matching `School` row.

In that state, the rollback would set `school_urn = gias_school_urn` and then try to re-add the FK to `schools.urn`, which would fail.

Is the migration actually reversible? If it's not, is it worth making that clear with:

```ruby
  def down
    raise ActiveRecord::IrreversibleMigration
  end
```

avinhurry (approved):
I had one question/comment on the down migration. Other than that looks good.

Would you mind running `bundle exec rake erd:generate` and committing the domain model diagram too please, to keep that up to date.

cpjmcquillan (approved):

PR #3091, opened 2026-06-16
Prevent resuming participants that were deferred/withdrawn today

ethax-ross:
### Context

Currently, if a lead provider withdraws/defers a participant we close their `TrainingPeriod` by setting the `finished_on` to today.

If this happened by mistake, the lead provider will immediately try and resume the participant; this attempts to create a new `TrainingPeriod` that starts today, which results in an error (as training periods can't finish/start on the same day).

- Prevent resuming participants that were deferred/withdrawn today

We are going to look at fixing this in a more convenient way for lead providers (perhaps by undoing the withdrawal/deferral rather than creating a new training period), but for now to avoid an internal server error we are going to return an error message to ask them to try again tomorrow.

### Guidance to review

Defer/withdraw a participant (that has an ongoing school period) and then immediately try to resume their training.

### DIFF
--- a/app/services/api/teachers/resume.rb
+++ b/app/services/api/teachers/resume.rb
@@ -4,6 +4,7 @@ module API::Teachers
 
     validate :not_already_active
     validate :no_ongoing_today_training_period
+    validate :training_period_not_finished_today
     validate :school_period_ongoing_today
 
     def resume
@@ -18,6 +19,17 @@ module API::Teachers
     end
 
   private
+
+    def training_period_not_finished_today
+      return if errors[:teacher_api_id].any?
+      return unless training_period_finished_today?
+
+      errors.add(:teacher_api_id, "You cannot resume a participant on the same day they were withdrawn or deferred. Resume them tomorrow or later.")
+    end
+
+    def training_period_finished_today?
+      training_period&.finished_on&.today?
+    end
 
     def not_already_active
       return if errors[:teacher_api_id].any?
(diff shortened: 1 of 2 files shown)


app/services/api/teachers/resume.rb
```diff
@@ -19,6 +20,17 @@ def resume
 
   private
 
+    def training_period_not_finished_today
+      return if errors[:teacher_api_id].any?
+      return unless training_period_finished_today?
+
+      errors.add(:teacher_api_id, "You need to wait until tomorrow to resume this participant as they were deferred/withdrawn today.")
```
StephenC-V1: Consider 'You cannot resume a training period on the same day a participant was withdrawn or deferred. Resume the period tomorrow or later'.

'Tomorrow' on it's own makes sense, but just heading off any anxiety if the user is not able to complete the task 'tomorrow' for any reason. (assuming here that tomorrow or later is viable.)

app/services/api/teachers/resume.rb
```diff
@@ -19,6 +20,17 @@ def resume
 
   private
 
+    def training_period_not_finished_today
+      return if errors[:teacher_api_id].any?
+      return unless training_period_finished_today?
+
+      errors.add(:teacher_api_id, "You cannot resume a training period on the same day a participant was withdrawn or deferred. Resume the period tomorrow or later.")
```
cpjmcquillan: Is it worth us referring to a teacher's (or participant's) training rather than a training period? 🤔

cpjmcquillan (approved):

tonyheadford (approved):

app/services/api/teachers/resume.rb
ethax-ross: Good spot, updated 👍

PR #3092, opened 2026-06-16
Better trace errors through the logs

cpjmcquillan:
Before, it was difficult to investigate errors via the logs. You'd need to find the error in Sentry, and try and guesstimate which logs corresponded to that error.

We already add the `current_user_id` to the logs payload, but this only applies for users with a `User` record in the database (i.e internal users).

In order to better trace a user's journey through the logs, this adds their `dfe_sign_in_user_id` to the log payload, and adds it to the user's context in Sentry.

Now, you can look at an error in Sentry and use the user ID to trace the error back through the logs.

This will help us better piece together the journey a user went on before running into issues.

This is the first in a series of incremental changes to our setup that hope to make it easier to resolve support queries and fix bugs.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -24,7 +24,7 @@ class ApplicationController < ActionController::Base
     Sentry.set_user(
       email: current_user&.email,
       id: current_user.try(:id),
-      dfe_sign_in_user_id: current_user.try(:dfe_sign_in_user_id)
+      dfe_sign_in_user_fingerprint: current_user_dfe_sign_in_user_fingerprint
     )
   end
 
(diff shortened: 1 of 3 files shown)


app/controllers/application_controller.rb
```diff
@@ -21,7 +21,11 @@ def require_admin
   end
 
   def set_sentry_user
-    Sentry.set_user(email: current_user&.email)
+    Sentry.set_user(
+      email: current_user&.email,
+      id: current_user.try(:id),
+      dfe_sign_in_user_id: current_user.try(:dfe_sign_in_user_id)
```
avinhurry: I get why this helps with tracing, but isn't `dfe_sign_in_user_id` the DSI user UUID?

Feels like we should avoid putting the raw value into logs/Sentry. Could we log a hashed/fingerprinted version instead, so we can still join Sentry to logs without exposing the DSI ID?

I think the DSI team removed the UUID from URLs on their end for a similar reason.

cpjmcquillan: Yeah, that makes sense 👌🏼.

Originally I was only adding the ID to logs (I thought I could lookup the Users in DSI and infer the UUID from the URL), but then I figured we could add it in to Sentry and make it a bit more seamless. 

Ultimately I think the risk is pretty low, but there's no harm in mitigating it further!

cpjmcquillan: I'm logging off for today, but feel free to push up a commit if you fancy it!

avinhurry: Pushed a commit 👍🏽

tonyheadford (approved):
LGTM

PR #3093, opened 2026-06-16
Make API errors easier to read

tonyheadford:
### Context

Default unhandled errors are being rendered in HTML for API clients.  These should return a JSON response so that they can be easily consumed.

Add simple JSON responses for the unhandled errors.

### DIFF
--- a/spec/requests/errors_spec.rb
+++ b/spec/requests/errors_spec.rb
@@ -55,7 +55,33 @@ RSpec.describe "Errors", type: :request do
     end
   end
 
-  context "when the request format is not HTML" do
+  context "when the request format is JSON" do
+    it "renders a json 404 error for /404.json" do
+      get "/404.json"
+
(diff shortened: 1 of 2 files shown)


spec/requests/errors_spec.rb
```diff
@@ -55,7 +55,33 @@
     end
   end
 
-  context "when the request format is not HTML" do
+  context "when the request format is JSON" do
+    it "renders a json 404 error for /404.json" do
```
ethax-ross, line 59: Worth adding 429 as well?

ethax-ross (approved):

spec/requests/errors_spec.rb
tonyheadford: As discussed, these are caught by `Rack::Attack` and the response has a body with the text "Retry later" for all request content types (html/json/pdf etc).  There is already a spec for this outside of any content-type and we decided that was OK.

PR #3094, opened 2026-06-16
Add uplift target ratio to banded fee structures

ericaporter:
### Context
The provider targets component currently renders the uplift target as a hardcoded 33%.

ECF stores the source contract value as uplift_target on CallOffContract, for example 0.33 or 0.4. Some historic/source contract values may not be 33%, so RECT should not hardcode this value in the component.

- Adds uplift_target_ratio to contract_banded_fee_structures
- Adds model validation allowing nil, but ensuring any populated value is between 0 and 1
- Updates ::BandedFeeComponent to render the uplift target from `banded_fee_structure.uplift_target_ratio` instead of a hardcoded 33%
- Converts the ratio to a percentage only at display time, so 0.4 renders as 40%

### Guidance to review

N.B.
The ECF data backfill is not included in this PR. That will be done separately.


cpjmcquillan (approved):
Let's do them right away, but is it worth adding a ticket for the backfill and to change the column to non-nullable?

PR #3098, opened 2026-06-17
Update `stylelint-config-gds` to 3.0.0

peteryates:
- **Upgrade stylelint-config-gds to 3.0.0**
- **Fix warnings**

Fixes #2992

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
   "devDependencies": {
     "playwright": "^1.60.0",
     "standard": "^17.1.2",
-    "stylelint-config-gds": "^2.0.0"
+    "stylelint-config-gds": "^3.0.0"
   },
   "optionalDependencies": {
     "@esbuild/darwin-arm64": "^0.28.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -7,12 +7,13 @@
     "esbuild": "^0.28.1",
     "govuk-frontend": "6.0.0",
     "sass": "^1.101.0",
+    "stylelint": "^17.13.0",
```
avinhurry: Is this intentional? I'd expect `stylelint` to live under `devDependencies` rather than app dependencies, from what I can see it's only used by `lint:scss`.

Or have I missed something?

avinhurry (approved):

package.json
peteryates: No, I was having a bit of bother getting npm to update `stylelint-config-gds` and in forcing it through with `npm update stylelint stylelint-config-gds` didn't realise it'd added it to the dependencies list! :man_facepalming:

PR #3100, opened 2026-06-17
[4090] Fix mentor summary bug hiding subsequent MASPs at the same school

rwrrll:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/4090

Handle the case where a mentor leaves a school and then returns to the same school later, by explicitly ordering MentorAtSchoolPeriods rather than using `find_by`which gives us the first matching record by implicit order

### DIFF
--- a/app/components/schools/mentors/summary_component.rb
+++ b/app/components/schools/mentors/summary_component.rb
@@ -112,7 +112,7 @@ module Schools
       end
 
       def mentor_period_for_school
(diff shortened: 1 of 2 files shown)


tonyheadford (approved):

avinhurry (approved):

cpjmcquillan (approved):

ericaporter (approved):
Looks good! 👍

PR #3101, opened 2026-06-17
Add new attribute `most_recent_induction_period_end_date` to `API::TeacherSerializer`

tonyheadford:
### Context

To help providers understand where an ECT is with their induction we can add a new field to the serializer to surface the `finished_on` date of the teacher's most recent `InductionPeriod`.

* Adds `most_recent_induction_period_end_date` attribute to the `API::TeacherSerializer`
* surface the `most_recent_induction_period_end_date` in the `ecf_enrolment` section
* Add `most_recent_induction_period_end_date` method to `API::Teachers::InductionStatus`
* Updates specs
* Updates swagger docs

### Guidance to review

Access the participant API end point.

* for a mentor teacher, the `most_recent_induction_period_end_date` should be `null`
* for an ECT with an ongoing `InductionPeriod`, the `most_recent_induction_period_end_date` should be `null`
* for an ECT whose latest `InductionPeriod` is closed, the `most_recent_induction_period_end_date` should match the `finished_on` date of the latest `InductionPeriod`

### DIFF
--- a/app/serializers/api/teacher_serializer.rb
+++ b/app/serializers/api/teacher_serializer.rb
@@ -125,6 +125,11 @@ class API::TeacherSerializer < Blueprinter::Base
     field(:full_name) { |teacher| Teachers::Name.new(teacher).full_name }
     field(:trn, name: :teacher_reference_number)
     field(:api_updated_at, name: :updated_at)
+    field(:most_recent_induction_period_end_date) do |teacher|
+      if teacher.ect_at_school_periods.any?
+        teacher.induction_periods.started_on_or_before(Time.zone.today).latest_first.first&.finished_on
+      end
+    end
 
     association :ecf_enrolments, blueprint: TrainingPeriodSerializer do |teacher, options|
       metadata = ::API::TeacherSerializer.lead_provider_metadata(teacher:, options:)
--- a/app/serializers/api/teacher_serializer.rb
+++ b/app/serializers/api/teacher_serializer.rb
@@ -80,6 +80,11 @@ class API::TeacherSerializer < Blueprinter::Base
           API::Teachers::InductionStatus.new(teacher:).induction_end_date
         end
       end
+      field(:most_recent_induction_period_end_date) do |(training_period, teacher, _)|
+        if training_period.for_ect?
+          API::Teachers::InductionStatus.new(teacher:).most_recent_induction_period_end_date
+        end
+      end
       field(:overall_induction_start_date) do |(training_period, teacher, _)|
         if training_period.for_ect?
           API::Teachers::InductionStatus.new(teacher:).induction_start_date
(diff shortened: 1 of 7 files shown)


app/serializers/api/teacher_serializer.rb
```diff
@@ -125,6 +125,11 @@ def uplift_value(training_period, metadata, uplift_type)
     field(:full_name) { |teacher| Teachers::Name.new(teacher).full_name }
     field(:trn, name: :teacher_reference_number)
     field(:api_updated_at, name: :updated_at)
+    field(:most_recent_induction_period_end_date) do |teacher|
```
ethax-ross: One to check with @easeynathan but I think we might want this field in the `ecf_enrolments` rather than at the teacher level. I would expect to see it next to the other induction period fields, although I do get why you put it here (it is odd the other induction period fields got put under enrolments as they are teacher-specific and not training period specific, but we can thank ECF for that).

app/serializers/api/teacher_serializer.rb
```diff
@@ -125,6 +125,11 @@ def uplift_value(training_period, metadata, uplift_type)
...
+      if teacher.ect_at_school_periods.any?
+        teacher.induction_periods.started_on_or_before(Time.zone.today).latest_first.first&.finished_on
```
ethax-ross: Can this live inside `API::Teachers::InductionStatus.new(teacher:)`? we may want to think about the methods names for clarity though. I also wonder if for consistency we should maybe rename it to `most_recent_induction_end_date` - although the other two indicate overall induction start/finish, so maybe having `induction_period` here does make sense 🤔

spec/serializers/api/teacher_serializer_spec.rb
```diff
@@ -37,6 +37,29 @@
...
+      context "when the teacher is not an ECT" do
+        it { is_expected.to be_nil }
+      end
+
+      context "when the teacher is an ECT" do
+        let(:teacher) { FactoryBot.create(:teacher, :induction_in_progress) }
+        let!(:ect_at_school_period) { FactoryBot.create(:ect_at_school_period, :ongoing, teacher:) }
+        let!(:induction_period) { FactoryBot.create(:induction_period, :ongoing, teacher:) }
+
+        it { is_expected.to be_nil }
```
ethax-ross: Worth covering a few more scenarios here? so:

- no induction period
- induction period with `started_on` and `finished_on` in the future
- induction period ongoing `nil` `finished_on`
- induction period ongoing with `finished_on `

ethax-ross (commented):
Thanks Tony just left a few thoughts

app/serializers/api/teacher_serializer.rb
tonyheadford: Yes good idea, I've asked Nathan, but as you say odd that it's in the training data

tonyheadford: Yes, Nathan would like it in with the others so I'll make that change

app/serializers/api/teacher_serializer.rb
tonyheadford: Yes, I think period might make it clearer

spec/serializers/api/teacher_serializer_spec.rb
tonyheadford: Added spec for no induction period
`InductionPeriod` prevents future `started_on` or `finished_on` dates
There are specs for ongoing (nil finished_on) and closed (finished_on has a value) periods, but this has been refactored with the other changes - is this what you meant?

spec/serializers/api/teacher_serializer_spec.rb
```diff
@@ -262,6 +264,32 @@
             end
           end
 
+          describe "`most_recent_induction_period_end_date`" do
```
ethax-ross, line 267: Do we need a test under the mentor context for when an induction period is present to cover it returning `nil`?

ethax-ross (approved):
Looks good thanks Tony

PR #3083, opened 2026-06-15
[3404] Add check answers step to contract period change wizard (Slice 2)

avinhurry:
### What

Following on from #3058, this is the second slice of the contract period change journey for admins.

Add the check answers step to the contract period change wizard.

This introduces:

- check answers
- summary rows and helper logic for:
  - existing contract period
  - new contract period
  - selected partnership
- back and cancel links
- a Confirm button
- guards so the check answers step is only available after a valid contract period and partnership have been selected

## Screenshot

| `/check-answers` |
|-------------------|
|image|

### Not included

This PR does not:

- display the Change link on the Training tab
- apply the contract period change
- create or update training periods
- record events
- add success messaging

PR #3103, opened 2026-06-17
[3404] Apply current active contract period changes (Slice 3)

avinhurry:
## What

Following on from #3083, this is the third slice of the contract period change journey for admins.

This PR adds the current active period mutation slice of the contract period change journey.

This introduces:

- a service for applying contract period changes to current active training periods
- creation of a replacement training period using the selected partnership and contract period
- equivalent schedule lookup in the selected contract period
- ending the previous training period
- frozen year tracking for ECT and mentors
- contract period change event recording
- CYA submission
- redirect back to the training tab with the success message

## Not included

This PR does not:

- display the Change link on the training tab
- support future starting training period mutation
- handle same/different LP/DP future transfer behaviour
- change contract period eligibility for EOI only participants
- add the no partnership dead end page

### DIFF
--- a/app/services/admin/teachers/training_periods/change_contract_period/current_active_period.rb
+++ b/app/services/admin/teachers/training_periods/change_contract_period/current_active_period.rb
@@ -1,0 +1,135 @@
+module Admin
+  module Teachers
+    module TrainingPeriods
+      module ChangeContractPeriod
+        class CurrentActivePeriod
+          class UnsupportedTrainingPeriodError < StandardError; end
+          class ScheduleNotFoundError < StandardError; end
+
+          attr_reader :training_period, :contract_period, :school_partnership, :author
+
+          def initialize(training_period:, contract_period:, school_partnership:, author:)
+            @training_period = training_period
+            @contract_period = contract_period
+            @school_partnership = school_partnership
+            @author = author
+          end
+
+          def change_contract_period!
+            unless current_active_period?
+              raise UnsupportedTrainingPeriodError,
+                    "Contract period changes are only supported for the current active training period"
+            end
+
+            if equivalent_schedule.blank?
+              raise ScheduleNotFoundError,
+                    "No equivalent schedule found for #{training_period.schedule.identifier} in contract period #{contract_period.year}"
+            end
+
+            ActiveRecord::Base.transaction do
+              track_payments_frozen_year!
+              replacement_finished_on = training_period.finished_on
+              finish_training_period!
+              new_training_period = create_replacement_training_period!(finished_on: replacement_finished_on)
+              record_contract_period_changed_event!
+              new_training_period
+            end
+          end
+
+        private
+
+          def current_active_period?
+            relationships.current_active_period == training_period && training_period.started_on < Time.zone.today
+          end
+
+          def relationships
+            @relationships ||= ::TrainingPeriods::PeriodRelationships.new(training_period:)
+          end
+
+          def equivalent_schedule
+            @equivalent_schedule ||= Schedule.find_by(
+              identifier: training_period.schedule.identifier,
+              contract_period_year: contract_period.year
+            )
+          end
+
+          def finish_training_period!
+            if training_period.for_ect?
+              ::TrainingPeriods::Finish.ect_training(
+                author:,
+                training_period:,
+                ect_at_school_period: training_period.at_school_period,
+                finished_on: Date.yesterday
+              ).finish!
+            elsif training_period.for_mentor?
+              ::TrainingPeriods::Finish.mentor_training(
+                author:,
+                training_period:,
+                mentor_at_school_period: training_period.at_school_period,
+                finished_on: Date.yesterday
+              ).finish!
+            else
+              raise UnsupportedTrainingPeriodError, "Training period must be ECT or mentor"
+            end
+          end
+
+          def create_replacement_training_period!(finished_on:)
+            ::TrainingPeriods::Create.provider_led(
+              period: training_period.at_school_period,
+              started_on: Time.zone.today,
+              finished_on:,
+              school_partnership:,
+              expression_of_interest: nil,
+              schedule: equivalent_schedule,
+              author:
+            ).call
+          end
+
+          def record_contract_period_changed_event!
+            ::Events::Record.record_teacher_contract_period_changed_event!(
+              author:,
+              original_training_period: training_period,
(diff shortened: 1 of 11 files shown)


app/services/admin/teachers/training_periods/change_contract_period/current_active_period.rb
```diff
@@ -0,0 +1,135 @@
...
+              raise ScheduleNotFoundError,
+                    "No equivalent schedule found for #{training_period.schedule.identifier} in contract period #{contract_period.year}"
+            end
+
+            ActiveRecord::Base.transaction do
+              track_payments_frozen_year!
+              replacement_finished_on = training_period.finished_on
+              finish_training_period!
+              new_training_period = create_replacement_training_period!(finished_on: replacement_finished_on)
+              record_contract_period_changed_event!
```
ericaporter: Just a thought (not a blocker) - would it be useful to have the `record_contract_period_changed_event!` record the new training period since this is finished in the same transaction? Might make the audit trail easier to follow / debug etc.

app/services/admin/teachers/training_periods/change_contract_period/eligibility.rb
```diff
@@ -39,6 +46,17 @@ def same_lead_provider_delivery_partnership?(period, other_period)
...
           end
 
+          def base_eligible?
+            return false unless training_period.provider_led_training_programme?
+            return false if training_period.school_partnership.blank?
+
+            !finished_before_today?
+          end
+
+          def current_active_period_start_date_passed?
```
ericaporter: `current_active_period_started_before_today?` might be a little clearer - possibly splitting hairs!

app/services/admin/teachers/training_periods/change_contract_period/eligibility.rb
```diff
@@ -39,6 +46,17 @@ def same_lead_provider_delivery_partnership?(period, other_period)
               period.lead_provider_delivery_partnership == other_period.lead_provider_delivery_partnership
           end
 
+          def base_eligible?
```
ericaporter: I'd be tempted to keep this inline for readability and have:
```
return false unless training_period.provider_led_training_programme?
return false if training_period.school_partnership.blank?
return false if finished_before_today?
```

app/services/admin/teachers/training_periods/change_contract_period/current_active_period.rb
avinhurry: Good thought!

I've kept the event associated with the original training period, but passed the replacement TP into the event recorder and added it's ID to the event metadata.

This keeps the main audit trail attached to the period being ended, whilst still making it easy to trace the replacement period created by the same transaction.

app/services/admin/teachers/training_periods/change_contract_period/eligibility.rb
avinhurry: Sure, i've inlined it for now. `current_active_period_changeable?` is only temporary while this slice supports active period mutation. Once the future periods paths are implemented, the intention is to go back to a single eligibility check.

app/services/admin/teachers/training_periods/change_contract_period/eligibility.rb
avinhurry: Yeah, I agree, have updated! Thanks!

app/services/admin/teachers/training_periods/change_contract_period/eligibility.rb
```diff
@@ -13,13 +13,24 @@ def eligible?
             return false unless training_period.provider_led_training_programme?
             return false if training_period.school_partnership.blank?
             return false if finished_before_today?
+            return false if current_active_period.present? && !current_active_period_started_before_today?
```
ericaporter, line 16: I was thinking a bit more about how to make this class more readable (appreciating I just commented to bring the returns inline!)

Something like...

```
def eligible?
      return false unless ongoing_provider_led_partnership?
      return false if blocked_by_active_period?

      return is_current_active_period? if no_future_periods?
      return future_periods == [training_period] if no_current_active_period?

      future_periods? && shares_lead_provider_with_active_period?
    end

def current_active_period_changeable?
      return false unless ongoing_provider_led_partnership?
      return false unless is_current_active_period?
      return false unless current_active_period_started_before_today?

      no_future_periods?
    end

private

def ongoing_provider_led_partnership?
      training_period.provider_led_training_programme? &&
        training_period.school_partnership.present? &&
        !finished_before_today?
    end

etc.

File: spec/requests/admin/teachers/training_periods/change_contract_period_wizard_spec.rb [hunk 5]
```diff
@@ -163,15 +174,35 @@
...
+      target_schedule
+      select_contract_period_and_partnership
+
+      expect {
+        post path_for_step("check-answers"), params: { check_answers: {} }
+      }.to change { TrainingPeriod.where(ect_at_school_period:).count }.by(1)
+
+      expect(response).to redirect_to(admin_teacher_training_path(teacher))
+      expect(training_period.reload.finished_on).to eq(today.yesterday)
+      expect(replacement_training_period.contract_period).to eq(target_contract_period)
```
ericaporter, line 188: for completeness you could add in `expect(replacement_training_period.school_partnership).to eq(target_school_partnership)`

ericaporter (approved):
This is looking great - I think there is scope to make eligibility.rb a bit more readable just because there are so many rules to follow but it's non-blocking. I've left a suggestion - see what you think!

spec/requests/admin/teachers/training_periods/change_contract_period_wizard_spec.rb
avinhurry: Good shout, it's good and low cost. Have added.

Thanks.

app/services/admin/teachers/training_periods/change_contract_period/eligibility.rb
avinhurry: Thank you for taking the time to review this so thoroughly. I've pushed a version close to what you suggested, with a few naming tweaks so the LP/DP rule is explicit.

The intent is still to simplify this again once the future period mutation work comes in and we can drop the temporary `current_active_period_changeable?` path, but getting the policy readable now should make that follow up easier and reduce the risk of adding the wrong behaviour.

PR #3104, opened 2026-06-17
Fix flaky test

cpjmcquillan:
Before, this test would fail if the sampled release note title contained any punctuation.

This sanitizes the response body before we assert against it, so we don't run into [failures] comparing things like

```
Added new withdrawal reason &#39;changed_lead_provider&#39;
```

with things like

```
Added new withdrawal reason 'changed_lead_provider'
```

Now, this test should pass every time.

[failures]: https://github.com/DFE-Digital/register-early-career-teachers-public/actions/runs/27689556289/job/81896706520?pr=3099

### DIFF
--- a/spec/requests/api/release_notes_spec.rb
+++ b/spec/requests/api/release_notes_spec.rb
@@ -1,4 +1,6 @@
 RSpec.describe "Release Notes" do
+  include ActionView::Helpers::SanitizeHelper
+
   describe "GET /api/guidance/release-notes" do
     it "renders valid notes" do
       expect { get(api_guidance_release_notes_path) }.not_to raise_error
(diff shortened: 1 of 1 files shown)


spec/requests/api/release_notes_spec.rb
```diff
@@ -23,7 +25,7 @@
         get(api_guidance_release_note_path(note.slug))
 
         expect(response).to be_successful
-        expect(response.body).to include(note.title)
+        expect(sanitize(response.body)).to include(note.title)
         expect(response.body).to include(note.body)
```
avinhurry, line 29: Will we have the same escaping issue with `body`?

Could we make the spec deterministic by avoiding `sample` either by testing one fixed note or all of them?

cpjmcquillan: So [we're explicitly sanitizing the page title](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/2bb28490a5798e563d986cedb433f48e518c5993/app/views/layouts/shared/_head.html.erb#L2) actually —I'm not sure why 🤔.

The commit message doesn't reveal anything either!

peteryates: Looks like a bug. I'll sort it.

Image

PR #3105, opened 2026-06-18
Handle existing declarations with invalid declaration_date for milestone

ethax-ross:
## Context

We migrated ~4k declarations from ECF to training periods such that the schedule milestone no longer matches the declaration date.

During migration we skipped validation for this, but the validation on the model is now causing errors to throw when lead providers attempt to void these declarations.

## Changes proposed

Only apply declaration date validation if it has been changed (in doing so we can update declarations with an already-persisted, invalid declaration date).

We also want to re-validate the `declaration_date` if the training period's schedule changes, causing it to become outside the new milestone bounds.

### DIFF
--- a/spec/models/declaration_spec.rb
+++ b/spec/models/declaration_spec.rb
@@ -132,10 +132,10 @@ describe Declaration do
     end
 
     describe "declaration date relative to milestone dates" do
-      subject(:declaration) { FactoryBot.build(:declaration, declaration_type: :started, declaration_date:, training_period:) }
+      subject(:declaration) { FactoryBot.build(:declaration, :started, declaration_date:, training_period:) }
 
       let(:schedule) { FactoryBot.create(:schedule, contract_period: school_partnership.contract_period) }
-      let(:milestone) { FactoryBot.create(:milestone, declaration_type: :started, schedule:) }
+      let(:milestone) { FactoryBot.create(:milestone, :started, schedule:) }
       let(:school_partnership) { FactoryBot.create(:school_partnership) }
       let(:training_period) { FactoryBot.create(:training_period, schedule:, school_partnership:) }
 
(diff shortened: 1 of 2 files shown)


spec/models/declaration_spec.rb
```diff
@@ -145,6 +145,44 @@ def will_change_attribute(attribute_to_change:, new_value:)
...
+          expect { declaration.update!(declaration_type: "retained-1") }.not_to raise_error
+        end
+
+        it "still validates the declaration date when it is changed" do
+          expect(declaration).to be_valid
+
+          declaration.declaration_date = milestone.milestone_date.next_day
+
+          expect(declaration).to be_invalid
+          expect(declaration.errors[:declaration_date]).to include("Declaration date must be on or before the milestone date for the same declaration type.")
```
cpjmcquillan:
```suggestion
          expect(declaration).to have_error(:declaration_date, "Declaration date must be on or before the milestone date for the same declaration type.")
```

spec/models/declaration_spec.rb
```diff
@@ -145,6 +145,44 @@ def will_change_attribute(attribute_to_change:, new_value:)
...
+          different_contract_period = FactoryBot.create(:contract_period, year: school_partnership.contract_period.year - 1)
+          different_schedule = FactoryBot.create(:schedule, contract_period: different_contract_period)
+          FactoryBot.create(:milestone, :started, schedule: different_schedule)
+
+          declaration.training_period.schedule = different_schedule
+        end
+
+        it "is not valid" do
+          expect(declaration).to be_invalid
+          expect(declaration.errors[:declaration_date]).to include("Declaration date must be on or before the milestone date for the same declaration type.")
```
cpjmcquillan:
```suggestion
          expect(declaration).to have_error(:declaration_date, "Declaration date must be on or before the milestone date for the same declaration type.")
```

cpjmcquillan (approved):

tonyheadford (approved):

PR #3106, opened 2026-06-18
Allow statements to be added after a contract period has been started if not frozen

craigmdavidson:
### Context

Resolves https://github.com.mcas.ms/DFE-Digital/register-ects-project-board/issues/3332

It is possible to add **Statements** after a **ContractPeriod** has started as long as it is not frozen.

### Guidance to review

* Editing and Deleting is still not possible after a ContractPeriod has started, there is a ticket to review broad in-flight modifications.
* It is technically possible for a **ContractPeriod** to be _frozen_ before a **ContractPeriod** has started, which would prevent adding new **Statements**.


ethax-ross (approved):
It does feel a bit odd to be able to add a statement and not edit/delete it, but we'll be revisiting that in the general permissions ticket

PR #3108, opened 2026-06-18
Remove `#sanitize` from page titles

peteryates:
Calling `#sanitize` when constructing `page_title_parts` means we double-sanitize the title and end up with something like this:

```html
<title>Some Junior, Infant &amp;amp; Nursery School - Register early career teachers - GOV.UK</title>
```

When we actually want this:

```html
<title>Some Junior, Infant &amp; Nursery School - Register early career teachers - GOV.UK</title>
```

This is a bit tricky to test, and while it's a partial used everywhere and the bug happens on all pages school/organisation names are the most likely offenders.

Fixes #3107

### DIFF
--- a/spec/requests/admin/schools/overviews/show_spec.rb
+++ b/spec/requests/admin/schools/overviews/show_spec.rb
@@ -1,0 +1,27 @@
+describe "Admin::Schools::OverviewsController" do
+  let(:title_suffix) { "- Register early career teachers - GOV.UK" }
+  let(:school_name) { "Some School" }
+  let(:gias_school) { FactoryBot.create(:gias_school, name: school_name) }
+  let(:school) { FactoryBot.create(:school, gias_school:) }
+
+  include_context "sign in as DfE user"
+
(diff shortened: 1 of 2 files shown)


spec/requests/admin/schools/overviews/show_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+  describe "GET /admin/schools/:urn/overview" do
+    it "displays the school name in the title" do
+      get "/admin/schools/#{school.urn}/overview"
+
+      expect(response.body).to match("<title>#{school_name} #{title_suffix}</title>")
+    end
+
+    context "when the school name contains characters that need to be sanitized" do
+      let(:school_name) { "All Saints' Infants & Junior School" }
+      let(:sanitized_school_name) { %r{All Saints&#39; Infants &amp; Junior School} }
```
peteryates: Ignore this being a regexp, switched back to a string and will rebase/repush once the build's done

avinhurry (approved):
Nice fix! 

And the spec covers the escaping case nicely.

PR #3109, opened 2026-06-18
Drop the `govuk-template--rebranded` class

peteryates:
It's not needed [post govuk-frontend 6.0.0](https://github.com/alphagov/govuk-frontend/releases/tag/v6.0.0-rc.0).


avinhurry (approved):
🚮

PR #3110, opened 2026-06-18
Schools cannot use previous choices without an appropriate body

cpjmcquillan:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/4116

Before, schools could choose to use previous choices even if they did not have a last chosen appropriate body.

This lead to some [errors].

There are a bunch of schools without a last chosen appropriate body. This means, if they are allowed to choose to reuse their last choices, we end up trying to register an ECT with a `school_reported_appropriate_body` of `nil`.

Since we validate that ECTs at state funded schools must have a teaching school hub appropriate body, this led to errors being raised and users unable to register ECTs.

This reworks the logic in the `use_previous_ect_choices` step so that a school is only presented with that option if they have previously chosen an appropriate body.

[errors]: https://dfe-teacher-services.sentry.io/issues/7324885372/events/6941e9f94d104b7d8d2fc7aa47766591/

### DIFF
--- a/spec/features/schools/ects/register/reuse/partnership_spec.rb
+++ b/spec/features/schools/ects/register/reuse/partnership_spec.rb
@@ -36,8 +36,50 @@ RSpec.describe "Registering an ECT - reuse previous partnership" do
     and_the_new_training_period_uses_the_reused_partnership
   end
 
-  def given_i_am_logged_in_as_a_state_funded_school_user_with_previous_choices
-    context = build_school_with_reusable_provider_led_partnership
+  context "when the school has no last chosen appropriate body" do
+    scenario "does not offer reuse" do
+      given_i_am_logged_in_as_a_state_funded_school_user_with_previous_choices(
+        last_chosen_appropriate_body_present: false
+      )
+      and_i_am_on_the_schools_ects_index_page
+      and_i_start_adding_an_ect
+      and_i_click_continue
+      and_i_submit_the_find_ect_form
+      and_i_choose_that_the_details_are_correct
+      and_i_click_confirm_and_continue
+      then_i_am_on_the_email_address_page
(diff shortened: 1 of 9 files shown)


rwrrll (approved):
Code looks great!

spec/features/schools/ects/register/reuse/partnership_spec.rb
```diff
@@ -36,8 +36,50 @@
     and_the_new_training_period_uses_the_reused_partnership
   end
 
-  def given_i_am_logged_in_as_a_state_funded_school_user_with_previous_choices
-    context = build_school_with_reusable_provider_led_partnership
+  context "when the school's last chosen appropriate body is no longer valid" do
+    scenario "does not offer reuse" do
+      given_i_am_logged_in_as_a_state_funded_school_user_with_previous_choices(
+        last_chosen_appropriate_body_present: false
```
ericaporter, line 42: Very minor - would it be clearer to call the context 'when the school has no last chosen appropriate body' rather than 'no longer valid' since `last_chosen_appropriate_body_present: false` ?

spec/support/reusable_partnership_helpers.rb
```diff
@@ -10,7 +10,8 @@ module ReusablePartnershipHelpers
   )
 
   def build_school_with_reusable_provider_led_partnership(
```
ericaporter: Maybe `build_school_with_previous_provider_led_choices` would be more accurate now that last_chosen_appropriate_body_present can be false?

ericaporter (approved):
LGTM 👍

peteryates (approved):
Looks good. I think @ericaporter's context-naming suggestion is worth applying though.

PR #3114, opened 2026-06-18
[3404] Restrict partnership options & add no partnership path (Slice 4)

avinhurry:
## What

Following on from #3103, this is the 4th slice of the contract period change journey for admins.

This introduces:

- partnership options restricted to those matching the training period's current lead provider and delivery partner
- a no partnerships dead end page when the selected contract period has no matching partnership
- routing from contract period selection to the 'no partnerships' page when needed

## Screenshot

| `/no-partnerships` |
|------------|
|image|

## Not included

This PR does not:

- add future period mutation behaviour
- change contract period eligibility for EOI participants
- change the existing current active mutation behaviour

### DIFF
--- a/app/wizards/admin/teachers/training_periods/change_contract_period_wizard/no_partnerships_step.rb
+++ b/app/wizards/admin/teachers/training_periods/change_contract_period_wizard/no_partnerships_step.rb
@@ -1,0 +1,11 @@
+module Admin
+  module Teachers
+    module TrainingPeriods
+      module ChangeContractPeriodWizard
+        class NoPartnershipsStep < Step
+          def previous_step = :select_contract_period
+        end
+      end
+    end
+  end
+end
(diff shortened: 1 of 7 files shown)


app/wizards/admin/teachers/training_periods/change_contract_period_wizard/no_partnerships_step.rb
```diff
@@ -0,0 +1,12 @@
+module Admin
+  module Teachers
+    module TrainingPeriods
+      module ChangeContractPeriodWizard
+        class NoPartnershipsStep < Step
+          def previous_step = :select_contract_period
+          def next_step = :no_partnerships
```
cpjmcquillan: Should we define the next step as itself or `nil`? 🤔

cpjmcquillan (approved):

app/wizards/admin/teachers/training_periods/change_contract_period_wizard/no_partnerships_step.rb
avinhurry: Remove it, that matches the existing wizard pattern.

PR #3099, opened 2026-06-17
Schools can only choose current appropriate bodies

cpjmcquillan:
### Context

[Teams discussion](https://teams.microsoft.com/l/message/19:wrVA9ZipsSjXLCrfIhG_aeXg61D4MjtXfyz2AR2lDjw1@thread.tacv2/1781693340992?tenantId=fad277c9-c60a-4da1-b5f3-b3b8b34a82f9&groupId=5e035efe-5b2b-491b-9e3e-d832445e4ad1&parentMessageId=1781693340992&teamName=Schools%20Digital&channelName=CPD%20Register%20ECTs&createdTime=1781693340992&ngc=true&allowXTenantAccess=true)

### Changes proposed in this pull request

Before, we allowed schools to choose any appropriate body that was a teaching school hub.

In production this included a bunch of appropriate bodies that were not relevant, though.

This filters the appropriate body periods that are available to select to those that have a `dfe_sign_in_organisation_id`.

PR #3118, opened 2026-06-19
Ensure an ECTs school reported appropriate body is active

cpjmcquillan:
This follows up the work in #3099 and adds some validations to the `ECTAtSchoolPeriod` model to ensure that the `school_reported_appropriate_body` is active.

This validation applies in the context of registering an ECT.

We already validated _implicitly_ that the `school_reported_appropriate_body` was present and _explicitly_ that it was a `teaching_school_hub` AB (for state-funded schools) or a `national` or `teaching_school_hub` AB (for independent schools).

This tweaks those validations so we add an explicit error message when the `school_reported_appropriate_body` is missing. We also validate that the AB is active.

This brings the validations inline with the scope we use to filter out inactive ABs when we present options to the school when they choose an appropriate body in the register ECT journey.

### DIFF
--- a/app/models/appropriate_body_period.rb
+++ b/app/models/appropriate_body_period.rb
@@ -47,4 +47,7 @@ class AppropriateBodyPeriod < ApplicationRecord
   # TODO: consider removing once view components accept the new AB object not the ABP
   # @return [School]
   delegate :lead_school, to: :appropriate_body, allow_nil: true
+
+  # Predicates
+  def active? = dfe_sign_in_organisation_id.present?
 end
--- a/app/models/ect_at_school_period.rb
+++ b/app/models/ect_at_school_period.rb
@@ -17,13 +17,18 @@ class ECTAtSchoolPeriod < ApplicationRecord
   has_one :latest_mentorship_period, -> { latest_first }, class_name: "MentorshipPeriod"
 
   # Validations
-  validate :appropriate_body_for_independent_school,
-           if: -> { school&.independent? },
-           on: :register_ect
+  with_options on: :register_ect do
+    validates :school_reported_appropriate_body_id, presence: true
 
-  validate :appropriate_body_for_state_funded_school,
-           if: -> { school&.state_funded? },
-           on: :register_ect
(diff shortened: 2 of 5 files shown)


app/models/ect_at_school_period.rb
```diff
@@ -17,13 +17,18 @@ class ECTAtSchoolPeriod < ApplicationRecord
   has_one :latest_mentorship_period, -> { latest_first }, class_name: "MentorshipPeriod"
 
   # Validations
-  validate :appropriate_body_for_independent_school,
-           if: -> { school&.independent? },
-           on: :register_ect
+  with_options on: :register_ect do
+    validates :school_reported_appropriate_body, presence: true
```
rwrrll: The others validate onto the `:school_reported_appropriate_body_id` key (n.b. `_id`). I don't know if it matters or not, but could potentially affect serializers and things.

app/models/appropriate_body_period.rb
```diff
@@ -47,4 +47,9 @@ class AppropriateBodyPeriod < ApplicationRecord
   # TODO: consider removing once view components accept the new AB object not the ABP
   # @return [School]
   delegate :lead_school, to: :appropriate_body, allow_nil: true
+
+  # Predicates
+  def active? = dfe_sign_in_organisation_id.present?
+  def inactive? = dfe_sign_in_organisation_id.blank?
+  def legacy? = dqt_id.present?
```
rwrrll: I didn't see these used anywhere, assuming we know we'll definitely need them in the future?

rwrrll (approved):
LGTM! Couple of notes just to double check!

app/models/appropriate_body_period.rb
cpjmcquillan: Yeah that's a fair point, I'll remove them 👍🏼

app/models/ect_at_school_period.rb
cpjmcquillan: I'm not too sure. I've updated it for consistency anyway 👍🏼

PR #3119, opened 2026-06-19
Add EOI seeding functionality

peteryates:
### Context

The new seeding format allowed us to build expressions of interest but not describe them properly.

Now, by selecting:

* a school that doesn't have a `SchoolPartnership` with the `LeadProvider` in the given `ContractPeriod`
* a `LeadProvider` that is active in the given `ContractPeriod`

There's a new example in the seeds to demonstrate:

```ruby
matthew_goode = teacher(next_trn, "Matthew Goode") do
  description("Expression of interest with teach first")
 
  ect_at_school_period(mallory_towers, "2025-09-01") do
    training_period(teach_first, 2025, "2025-09-01")
  end
end
```

The output will be:

```
🌱   🧑‍🏫 Matthew Goode - 0000017
🌱     ℹ️ Expression of interest with teach first
🌱     🏫 ECT at Malory Towers since 2025-09-01
🌱       📒 expression of interest registered with Teach First on 2025-09-01
```

- **Add the ability to create EOIs in seeds**
- **Remove randomness from school partnership seeding**
- **Make the cops target Ruby 4.0**


avinhurry (approved):
This looks fine for the new EOI seed case but I think the `school_partnerships` change could have a downstream impact on the API seed scenarios, for example [`APISeedData:MentorScenario` skips when there are less than 2 school partnerships](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/services/api_seed_data/mentor_scenarios.rb#L14-L20), so this may unintentionally drop some API seed coverage, might be worth double checking.

cpjmcquillan (approved):

.rubocop.yml
```diff
@@ -18,7 +18,7 @@ inherit_mode:
     - Exclude
 
 AllCops:
-  TargetRubyVersion: 3.4
+  TargetRubyVersion: 4.0
```
cpjmcquillan, line 21: Could we bring this in from the `.ruby-version` file? 🤔

```
TargetRubyVersion: <%= File.read(".ruby-version").strip %>
```

We should probably update the Gemfile to define it like this

```
ruby file: ".ruby-version"
```

peteryates: I didn't know that was possible, but yes, totally agree!

PR #3125, opened 2026-06-22
New contact index and show

craigmdavidson:
### Context

Resolves https://github.com.mcas.ms/DFE-Digital/register-ects-project-board/issues/4049

Replace current contract `index` page with a table style view, and a new `show` details page.

#### Updated index page with buttons (placeholders)..

Screenshot 2026-06-22 at 16 06 19

#### Updated index page without buttons (when contracts are non-editable)...

Screenshot 2026-06-22 at 16 06 37

### The contract details page
Screenshot 2026-06-22 at 16 06 59

### Guidance to review

* Modifications to regular Rails controller with index / show methods.


app/controllers/admin/finance/active_lead_providers/contracts_controller.rb
```diff
@@ -3,6 +3,7 @@ class ContractsController < Admin::Finance::BaseController
     layout "full"
 
     before_action :set_active_lead_provider
+    before_action :set_contract, only: %i[show]
```
craigmdavidson, line 6: This sets us up for the other CRUD actions, rather than setting it in the method, then moving it back up here.

app/components/admin/finance/contracts/statement_range_description.html.erb
```diff
@@ -0,0 +1 @@
+<%= statement_range_description -%>
```
ethax-ross: I think if you change the `def statement_range_description` to `def call` in the component you could omit this view entirely since it's just rendering `statement_range_description`

tonyheadford (approved):
Nice

app/components/admin/statements/provider_targets_component/banded_fee_component.rb
```diff
@@ -50,7 +50,7 @@ def render? = banded_fee_structure.present?
               row.with_cell(header: true) { "Band" }
               row.with_cell(header: true, numeric: true, text: "Min")
               row.with_cell(header: true, numeric: true, text: "Max")
-              row.with_cell(header: true, numeric: true, text: "Payment amount per participant")
+              row.with_cell(header: true, numeric: true, text: "Fee per declaration")
```
ethax-ross, line 53: Thinking about this change more; we might have to have different terminology depending on if we're viewing a statement vs a contract.

In the scope of a statement a participant can only have one declaration either paid or clawed back (as they are always on different statements) so 'payment amount per participant' is accurate. I'm not sure if the same participant can have multiple declarations _of different types_ on the same statement, though 🤔 that's maybe something we can verify looking at the snapshot DB?

In the scope of a contract this is the amount paid per declaration, so 'payment amount per participant' is misleading, as we may pay for two declarations for the same participant over the course of a contract.

@craigmdavidson would you mind checking snapshot to see if a declaration appears multiple times for the same participant on the same statement?

cc @cpjmcquillan - flagging for visibility as this impacts the existing statements views

app/views/admin/finance/active_lead_providers/contracts/index.html.erb
```diff
@@ -1,33 +1,55 @@
...
+            <% end %>
+          <% end %>
+        <% end %>
       <% end %>
-    </section>
+    <% end %>
   <% end %>
 <% end %>
+
+<% if @active_lead_provider.editable? %>
```
ethax-ross, line 48: I think we also want to prevent adding a contract if there are no bands on the active lead provider yet? happy to defer this to a future ticket though as I think it will depend on PDH's changes?

ethax-ross (approved):
Looks great, thanks Craig

app/views/admin/finance/active_lead_providers/contracts/index.html.erb
craigmdavidson: Exactly - deferred to the CRUD controller for this.

app/components/admin/statements/provider_targets_component/banded_fee_component.rb
craigmdavidson: > @craigmdavidson would you mind checking snapshot to see if a declaration appears multiple times for the same participant on the same statement?

👍 

I'm not sure I understand the statement issue though, even if there are multiple declarations on multiple statements, the amount is paid per declaration rather than participant?

I.e. a participant has 2 training periods, 2 declarations are submitted, and has a fee for declaration of £100, we would pay £200 (one for the first declaration and on for the second)?

but if a participant has 2 training periods, 2 declarations are submitted, and has a "payment amount per participant", we would only pay £100 - despite there being 2 declarations?

What am I missing?

ethax-ross (approved):

app/components/admin/statements/provider_targets_component/banded_fee_component.rb
craigmdavidson: > @craigmdavidson would you mind checking snapshot to see if a declaration appears multiple times for the same participant on the same statement?

I can confirm this is the case looking on migration:

```
statement = Statement.find(745)

teacher = Teacher.find(216784)

statement.payment_declarations
    .where(id: teacher.ect_declarations)
    .or(statement.payment_declarations.where(id: teacher.mentor_declarations))
```

Yields:
```
[
  {type: "started", declaration_id: 412042, payment_statement_id: 745, training_period_id: 126022},
  {type: "retained-1", declaration_id: 730659, payment_statement_id: 745, training_period_id: 126022},
  {type: "retained-2", declaration_id: 147173, payment_statement_id: 745, training_period_id: 126022},
  {type: "retained-3", declaration_id: 864309, payment_statement_id: 745, training_period_id: 126022},
  {type: "retained-4", declaration_id: 499852, payment_statement_id: 745, training_period_id: 126022},
  {type: "completed", declaration_id: 577563, payment_statement_id: 745, training_period_id: 126022}
]

```

PR #3127, opened 2026-06-22
[4111] Create Teachers::Merge service

rwrrll:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/4111

To deal with support requests, sometimes we need to merge a teacher's records into another one. This PR creates a service to handle that work.

### Guidance to review

n.b. there's an admin UI coming down the pipe so I've left this console-driven for now.

### DIFF
--- a/app/services/teachers/merge.rb
+++ b/app/services/teachers/merge.rb
@@ -1,0 +1,70 @@
+module Teachers
+  class Merge
+    class MergeError < StandardError; end
+
+    attr_reader :author, :source, :destination, :event_note, :zendesk_ticket_id
+
+    def initialize(author:, source:, destination:, event_note: nil, zendesk_ticket_id: nil)
+      @author = author
+      @source = source
+      @destination = destination
+      @event_note = event_note
+      @zendesk_ticket_id = zendesk_ticket_id
+    end
+
+    def merge!
+      ensure_distinct_teachers
+
+      ActiveRecord::Base.transaction do
+        move_school_periods
+        move_induction_records
+        record_teacher_id_change
+        refresh_metadata
+        record_merge_events
+        anonymise_source
+      end
+
+      destination
+    end
+
+  private
+
+    def ensure_distinct_teachers
+      raise MergeError, "Cannot merge a teacher into itself" if source == destination
+    end
+
+    def move_school_periods
+      source.ect_at_school_periods.find_each { |period| period.update!(teacher: destination) }
+      source.mentor_at_school_periods.find_each { |period| period.update!(teacher: destination) }
+    end
+
+    def move_induction_records
+      source.induction_periods.find_each { |period| period.update!(teacher: destination) }
+      source.induction_extensions.find_each { |extension| extension.update!(teacher: destination) }
+    end
+
+    def record_teacher_id_change
+      TeacherIdChange.create!(
+        teacher: destination,
+        api_from_teacher_id: source.api_id,
+        api_to_teacher_id: destination.api_id
+      )
+    end
+
+    # The declarative refresh hook only re-points the destination's metadata on
+    # reassignment, so refresh the destination explicitly and tear down the
+    # stale source rows so their latest_*_training_period_id FKs don't dangle
+    def refresh_metadata
+      source.lead_provider_metadata.destroy_all
+      Metadata::Manager.new.refresh_metadata!([destination])
+    end
+
+    def record_merge_events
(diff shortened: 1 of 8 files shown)


spec/services/teachers/merge_spec.rb
```diff
@@ -0,0 +1,131 @@
...
+        allow(Teachers::Anonymise).to receive(:new).with(teacher: source, reason: :teacher_record_merged).and_return(anonymiser)
+
+        service.merge!
+
+        expect(anonymiser).to have_received(:anonymise!)
+      end
+
+      it "retains the source teacher's api_id so API consumers can still follow stored references" do
+        expect { service.merge! }.not_to(change { source.reload.api_id })
+      end
```
avinhurry: Can consumers actually follow stored references?

This verifies that `source.api_id` is retained, but not that the old participant ID can still be resolved after the merge. Should we assert that behaviour directly instead?

app/services/teachers/merge.rb
```diff
@@ -0,0 +1,70 @@
...
+      source.induction_extensions.find_each { |extension| extension.update!(teacher: destination) }
+    end
+
+    def record_teacher_id_change
+      TeacherIdChange.create!(
+        teacher: destination,
+        api_from_teacher_id: source.api_id,
+        api_to_teacher_id: destination.api_id
+      )
+    end
```
avinhurry, line 57: Should we also migrate any existing `source.teacher_id_changes` to the destination?

At the moment we create a new mapping from `source.api_id` to `destination.api_id`, but any existing `TeacherIdChange` records remain attached to the source teacher. Should those historical `api_id` mappings also be moved to the destination so they continue to resolve to the surviving teacher after the merge?

avinhurry (commented):
Left a couple of API related questions. Might be worth getting one of them to have a look too, but up to you if you think it would be helpful.

db/schema.rb
cpjmcquillan, line 1: Not sure what's going on with the ordering changes here. The schema should be alphabetical since #2684.

I wonder if someone is running `rails db:migrate` with a different version (i.e not `bin/rails db:migrate` or `bundle exec rails db:migrate`) 🤔

db/migrate/20260622120000_add_teacher_record_merged_anonymisation_reason.rb
```diff
@@ -0,0 +1,9 @@
+class AddTeacherRecordMergedAnonymisationReason < ActiveRecord::Migration[8.1]
+  def up
+    add_enum_value :anonymisation_reasons, "teacher_record_merged"
+  end
+
+  def down
+    # enum values can't be removed in PostgreSQL
```
cpjmcquillan, line 7: Is it worth raising an [error](https://api.rubyonrails.org/classes/ActiveRecord/IrreversibleMigration.html)?

cpjmcquillan (approved):

db/migrate/20260622120000_add_teacher_record_merged_anonymisation_reason.rb
rwrrll: Yep of course, not sure why I didn't (it's too hot!)

db/schema.rb
rwrrll: Yeah I thought it was me! But looks like c25f0b1a9 edited the schema.rb strings in-place to rename from "archived_reasons" to "anonymisation_reasons" rather than redumping it. I wonder if there's a rubocop or something for that!

spec/services/teachers/merge_spec.rb
rwrrll: Yeah good shout, exercises the serializer now to make sure the end behaviour is what we want here.

app/services/teachers/merge.rb
rwrrll: :+1: added this

app/services/teachers/merge.rb
```diff
@@ -0,0 +1,75 @@
...
+
+      ActiveRecord::Base.transaction do
+        move_school_periods
+        move_induction_records
+        move_teacher_id_changes
+        record_teacher_id_change
+        refresh_metadata
+        record_merge_events
+        anonymise_source
+      end
```
thenapking, line 26: Events are also linked to a teacher record so we may want to move those over too.

avinhurry (approved):
Looks good to me given that we want to preserve the old `api_id` resolving to the original teacher.

PR #3128, opened 2026-06-22
[3404] Allow contract period changes for EOI only training periods (Slice 5)

avinhurry:
## What

Following on from #3114, this is the 5th slice of the contract period change journey for admins.

This introduces:

- eligibility for current active EOI only training periods
- contract period options restricted to years where the current lead provider has an active lead provider record
- skipping the school partnership selection step for EOI only periods
- CYA content for EOI only periods showing:
  - lead provider
  - delivery partner as `No delivery partner confirmed`
 - mutation support to create the replacement training period with the equivalent active lead provider in the selected contract period

This slice keeps the controller on `current_active_period_changeable?` because we only handle current active mutations for now. Once future period changes are in, we can drop `current_active_period_changeable?` and make `eligible?` the source of truth.

## Screenshot

| `/check-answers` |
|------------|
|image|

## Not included

This PR does not:

- add future period mutation behaviour
- allow changing the lead provider for EOI only periods
- turn an EOI only period into a confirmed school partnership

### DIFF
--- a/app/services/admin/teachers/training_periods/change_contract_period/eligibility.rb
+++ b/app/services/admin/teachers/training_periods/change_contract_period/eligibility.rb
@@ -10,9 +10,15 @@ module Admin
           end
 
           def eligible?
-            return false unless provider_led_with_partnership?
+            return false unless provider_led_with_partnership_or_eoi?
             return false if finished_before_today?
             return false if blocked_by_current_active_period?
+
+            if eoi_only?
+              return current_active_period? if no_future_periods?
+
+              return false
+            end
 
             return current_active_period? if no_future_periods?
             return only_future_period? if no_current_active_period?
@@ -21,7 +27,7 @@ module Admin
           end
 
           def current_active_period_changeable?
-            return false unless provider_led_with_partnership?
+            return false unless provider_led_with_partnership_or_eoi?
             return false if finished_before_today?
             return false unless current_active_period?
             return false unless current_active_period_started_before_today?
@@ -43,9 +49,9 @@ module Admin
             @future_periods ||= relationships.future_periods.to_a
           end
 
-          def provider_led_with_partnership?
+          def provider_led_with_partnership_or_eoi?
             training_period.provider_led_training_programme? &&
-              training_period.school_partnership.present?
+              (training_period.school_partnership.present? || training_period.only_expression_of_interest?)
           end
 
           def blocked_by_current_active_period?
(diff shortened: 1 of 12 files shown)


app/services/admin/teachers/training_periods/change_contract_period/eligibility.rb
```diff
@@ -10,18 +10,24 @@ def initialize(training_period:)
...
+            return false unless provider_led_with_partnership_or_eoi?
             return false if finished_before_today?
             return false if blocked_by_current_active_period?
 
+            if eoi_only?
+              return current_active_period? if no_future_periods?
+
+              return false
+            end
+
```
ericaporter, line 22: I think this could be `return current_active_period? && no_future_periods? if eoi_only?` if you wanted to keep it to a single line

avinhurry: Yeah, I originally had that, but I found it a bit easier to read with the explicit `if` branch so I changed it. 

Happy to switch it back if you prefer.

spec/wizards/admin/teachers/training_periods/change_contract_period_wizard/wizard_spec.rb
```diff
@@ -124,6 +145,28 @@
...
+
+      before do
+        FactoryBot.create(:active_lead_provider, lead_provider: active_lead_provider.lead_provider, contract_period: target_contract_period)
+      end
+
+      it "only returns contract periods with an equivalent active lead provider" do
+        expect(wizard.contract_periods).to contain_exactly(target_contract_period)
+      end
+    end
   end
```
ericaporter, line 171: It might be worth adding a spec where other_contract_period has an ActiveLeadProvider record for a different lead provider.

That would confirm the contract period is excluded unless the EOI’s current lead provider is active for that year.

ericaporter (approved):
This is looking great!

app/services/admin/teachers/training_periods/change_contract_period/eligibility.rb
ericaporter: No strong preference from me — the explicit branch is clear, so happy to leave it as it is.

spec/wizards/admin/teachers/training_periods/change_contract_period_wizard/wizard_spec.rb
avinhurry: Good idea, that strengthens the existing coverage. I've added it to the setup for the existing spec rather than adding a separate example.

Thanks!

PR #3134, opened 2026-06-23
Handle multiple partnerships when changing schedule

ethax-ross:
### Context

In RECT we allow multiple `SchoolPartnership`s for the same `LeadProvider`, `School` and `ContractPeriod` (but must have different `DeliveryPartner`s).

We don't currently account for this in our change schedule endpoint; we just find the first `SchoolPartnership` with the correct `School`, `LeadProvider` and `ContractPeriod`.

- Handle multiple partnerships when changing schedule

### Guidance to review

We want to prioritise the `SchoolPartnership`s as follows:

- Retain the existing `SchoolPartnership` if not changing `ContractPeriod`
- Find a `SchoolPartnership` with the existing/same `DeliveryPartner` in the new `ContractPeriod`
- Choose the latest `SchoolPartnership` in the new `ContractPeriod`

### DIFF
--- a/app/services/api/teachers/change_schedule.rb
+++ b/app/services/api/teachers/change_schedule.rb
@@ -89,21 +89,45 @@ module API::Teachers
     def school_partnership_exists_if_changing_contract_period
       return if errors[:contract_period_year].any?
       return unless training_period
-      return if contract_period == training_period.contract_period
+      return unless contract_period_changing?
       return if school_partnership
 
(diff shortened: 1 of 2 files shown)


peterdavidhamilton (approved):

craigmdavidson (approved):
The change itself, looks good to me, though I do find this class generally quite difficult to understand for what it is.

PR #3136, opened 2026-06-23
[3404] Support future contract period changes (Slice 6)

avinhurry:
## What

Following on from #3128, this is the 6th slice of the contract period change journey for admins.

This introduces:

- a future period mutation service which updates eligible future training periods in place
- wizard submission routing to use the future period mutation path when the selected training period starts in the future
- access to the wizard for eligible future periods

## Not included

This PR does not:

- add future period EOI behaviour
- change the current active mutation behaviour
- allow changing LP/DP through this journey


spec/services/admin/teachers/training_periods/change_contract_period/future_period_spec.rb
```diff
@@ -0,0 +1,160 @@
...
+    end
+
+    it "raises an unsupported training period error" do
+      expect {
+        service_call
+      }.to raise_error(
+        described_class::UnsupportedTrainingPeriodError,
+        "Contract period changes are only supported for eligible future training periods"
+      )
+    end
```
ericaporter, line 145: You could add in a test here to cover an example of a TP starting today to cover `started_on > Time.zone.today` and make it explicit what happens to periods starting today.

app/wizards/admin/teachers/training_periods/change_contract_period_wizard/check_answers_step.rb
```diff
@@ -14,10 +15,12 @@ def previous_step
...
             true
-          rescue CurrentActivePeriodChange::UnsupportedTrainingPeriodError
+          rescue CurrentActivePeriodChange::UnsupportedTrainingPeriodError,
+                 FuturePeriodChange::UnsupportedTrainingPeriodError
             errors.add(:base, "Training period is not eligible for contract period change")
             false
-          rescue CurrentActivePeriodChange::ScheduleNotFoundError
+          rescue CurrentActivePeriodChange::ScheduleNotFoundError,
+                 FuturePeriodChange::ScheduleNotFoundError
             errors.add(:base, "A matching schedule could not be found for the selected contract period")
```
ericaporter, line 24: Would it be worth adding a test for the new FuturePeriodChange::UnsupportedTrainingPeriodError and FuturePeriodChange::ScheduleNotFoundError error handling? I think the future happy path is covered but couldn't see any covering the expected validation errors

ericaporter (approved):
LGTM! 👍  Suggestions for a couple of extra tests you could add if they would be helpful/ aren't covered elsewhere

spec/services/admin/teachers/training_periods/change_contract_period/future_period_spec.rb
avinhurry: It's covered in the [eligibility spec](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/spec/services/admin/teachers/training_periods/change_contract_period/eligibility_spec.rb#L159-L165), but I've added it here too to make the future period boundary clearer.

app/wizards/admin/teachers/training_periods/change_contract_period_wizard/check_answers_step.rb
avinhurry: Yeah good shout. I've added specs for the error mappings for both future periods and current active cases.

Thanks!

PR #3139, opened 2026-06-23
Reorganise the layouts to remove duplication

peteryates:
Nearly all of the pages in the app place their content either:

* in a two-thirds column with optional sidebar
* in a full width column

Both of these layouts contain _mostly_ the same structure but vary in how the contents of `<main>` are laid out - one full width column vs one two thirds with column with an optional 1 third width sidebar.

The only difference between `full` and `two_thirds` should be the width. Previously there were other changes to (related to what's printed).

This commit makes the following changes:

* make two thirds the application's default layout explicitly to two thirds
* remove any non-width differences between 'full' and 'two_thirds' layouts
* remove the `minimal` layout, we weren't using it anywhere


app/views/layouts/shared/_page.html.erb
peteryates, line 1: GitHub's detecting this as a move, but really I deleted `minimal.html.erb` because we don't appear to be using it anywhere.

app/controllers/application_controller.rb
```diff
@@ -2,11 +2,10 @@ class ApplicationController < ActionController::Base
   include Authentication
   include DfE::Analytics::Requests
   include TimeTravellable
+  include Pagy::Backend
```
peteryates: Cosmetic, just putting the `include` statements together.

peteryates: I realise now that this change made more sense in an earlier iteration where I was setting `layout` here 😅  I'll remove this change from the PR as it's unrelated to anything else now.

app/views/layouts/shared/_page.html.erb
```diff
@@ -0,0 +1,29 @@
...
+
+  <body class="govuk-template__body" data-turbo="false">
+    <div class="govuk-!-display-none-print">
+      <%= render partial: "layouts/shared/header" %>
+    </div>
+
+    <div class="govuk-width-container">
+      <div class="govuk-!-display-none-print">
+        <%= render partial: "layouts/shared/banners" %>
+        <%= render partial: "layouts/shared/schools_registration_window_banner" %>
```
avinhurry, line 13: Moving this to the shared page wrapper means it now shows on  application/two_thirds pages as well, not just full layout pages.

It's supposed to be school side only, will this still preserve that?

avinhurry (commented):
Much nicer! Just had a question about the registration banner

avinhurry (approved):

app/views/layouts/shared/_page.html.erb
peteryates: On `main`, `application.html.erb` is two thirds and most pages use it.

It looks like just 3 controllers use `full`:

```
rg layout app/controllers/schools
app/controllers/schools/access_controller.rb
3:    layout "full"

app/controllers/schools/mentors_controller.rb
3:    layout "full"

app/controllers/schools/ects_controller.rb
3:    layout "full"
```

Ultimately though, if we want to do school specific things I think we should either have different layouts or be using `content_for`.

The partial in question is only shown to school users though, ther'es a `if current_user&.school_user?` at the top.

PR #3145, opened 2026-06-24
Contract CRUD

craigmdavidson:
### Context

Resolves https://github.com/DFE-Digital/register-ects-project-board/issues/4050

Create, update, and delete **Contracts** for a given **ActiveLeadProvider**.

#### Big create / edit form:

Screenshot 2026-06-24 at 15 27 32

### Guidance to review

* Standard Rails controller, accepts nested attributes for **FlatRateFeeStructure** and **BandedFeeStructure** and **Bands** (or what will become **BandTerms** when that PR comes in).
* Currently attempts to default to any existing **Band** structure, this will go away when the **ActiveLeadProvider::Band** is implemented. We can also then check that the **ActiveLeadProvider** is `banded?` as a pre-requisite for setting up a **Contract**.

### DIFF
--- a/app/services/contracts/build.rb
+++ b/app/services/contracts/build.rb
@@ -1,0 +1,33 @@
+module Contracts
+  class Build
+    attr_reader :active_lead_provider
+
+    def initialize(active_lead_provider:)
+      @active_lead_provider = active_lead_provider
+    end
+
+    def call
+      contract = active_lead_provider.contracts.build
+      banded_fee_structure = contract.build_banded_fee_structure
+
+      # Temporarily seed bands from the previous contract's band_structure until
+      # ActiveLeadProvider::Band is implemented and can provide this data directly.
+      existing_contract = active_lead_provider.contracts
+                            .joins(banded_fee_structure: :bands)
+                            .includes(banded_fee_structure: :bands)
+                            .first
+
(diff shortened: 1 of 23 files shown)


app/services/contracts/build.rb
```diff
@@ -0,0 +1,33 @@
...
+                            .first
+
+      if existing_contract&.banded_fee_structure&.bands&.any?
+        existing_contract.banded_fee_structure.bands.each do |band|
+          banded_fee_structure.bands.build(band.attributes.slice("min_declarations", "max_declarations",
+                                                                 "fee_per_declaration", "output_fee_ratio", "service_fee_ratio"))
+        end
+      else
+        banded_fee_structure.bands.build
+      end
```
craigmdavidson, line 27: ⚠️ Depending on which lands first; the Band changes, or this; this will need to be switched out.

app/views/admin/finance/active_lead_providers/contracts/index.html.erb
```diff
@@ -47,6 +47,6 @@
 
 <% if @active_lead_provider.editable? %>
```
craigmdavidson, line 48: Still missing `<% ... && @active_lead_provider.banded? %> ` due to ordering.

PR #3146, opened 2026-06-24
[3404] Support future EOI contract period changes (Slice 7)

avinhurry:
## What

Following on from #3136, this is the 7th slice of the contract period change journey for admins.

This introduces the following for EOI only training periods:

- eligibility
- future period mutation support
- updating the EOI to the equivalent active lead provider in the selected contract period
- keeping the training period EOI only, with no school partnership selected
- check answers handling for missing equivalent active lead providers

## Not included

This PR does not:

- change the no partnerships journey
- allow changing lead provider or delivery partner through this journey
- change current active mutation behaviour

### DIFF
--- a/app/services/admin/teachers/training_periods/change_contract_period/eligibility.rb
+++ b/app/services/admin/teachers/training_periods/change_contract_period/eligibility.rb
@@ -14,14 +14,9 @@ module Admin
             return false if finished_before_today?
             return false if blocked_by_current_active_period?
 
-            if eoi_only?
-              return current_active_period? if no_future_periods?
-
-              return false
-            end
-
             return current_active_period? if no_future_periods?
             return only_future_period? if no_current_active_period?
+            return future_period? if eoi_only?
 
             future_period? && same_partnership_as_current_active_period?
           end
--- a/app/services/admin/teachers/training_periods/change_contract_period/eligibility.rb
+++ b/app/services/admin/teachers/training_periods/change_contract_period/eligibility.rb
@@ -14,11 +14,14 @@ module Admin
             return false if finished_before_today?
             return false if blocked_by_current_active_period?
 
-            return current_active_period? if no_future_periods?
-            return only_future_period? if no_current_active_period?
-            return future_period? if eoi_only?
-
-            future_period? && same_partnership_as_current_active_period?
+            case future_periods.count
+            when 0
+              current_active_period?
+            when 1
+              single_future_period_changeable?
+            else
+              false
+            end
           end
 
         private
@@ -48,20 +51,17 @@ module Admin
             current_active_period == training_period
           end
 
-          def no_future_periods?
-            future_periods.empty?
-          end
-
           def no_current_active_period?
             current_active_period.blank?
           end
 
-          def only_future_period?
+          def training_period_is_the_only_future_period?
             future_periods == [training_period]
           end
 
-          def future_period?
-            future_periods.include?(training_period)
+          def single_future_period_changeable?
+            training_period_is_the_only_future_period? &&
+              (no_current_active_period? || eoi_only? || same_partnership_as_current_active_period?)
           end
 
           def same_partnership_as_current_active_period?
--- a/app/wizards/admin/teachers/training_periods/change_contract_period_wizard/check_answers_step.rb
+++ b/app/wizards/admin/teachers/training_periods/change_contract_period_wizard/check_answers_step.rb
@@ -23,7 +23,8 @@ module Admin
                  FuturePeriodChange::ScheduleNotFoundError
             errors.add(:base, "A matching schedule could not be found for the selected contract period")
             false
-          rescue CurrentActivePeriodChange::ActiveLeadProviderNotFoundError
+          rescue CurrentActivePeriodChange::ActiveLeadProviderNotFoundError,
+                 FuturePeriodChange::ActiveLeadProviderNotFoundError
             errors.add(:base, "An active lead provider could not be found for the selected contract period")
             false
           end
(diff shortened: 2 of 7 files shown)


app/wizards/admin/teachers/training_periods/change_contract_period_wizard/check_answers_step.rb
```diff
@@ -23,7 +23,8 @@ def save!
                  FuturePeriodChange::ScheduleNotFoundError
             errors.add(:base, "A matching schedule could not be found for the selected contract period")
             false
-          rescue CurrentActivePeriodChange::ActiveLeadProviderNotFoundError
+          rescue CurrentActivePeriodChange::ActiveLeadProviderNotFoundError,
```
thenapking, line 26: I appreciate that you are following a pattern here established by previous PRs, so take this comment as a possible suggestion for refactoring in the future, rather than with this PR itself.  
In other wizards we only advance to the `check_answers` step when the previous steps are valid, and so adding errors at this step seems wrong to me.  Also we're raising and catching errors instead of using validations, which I feel would fit more with the pattern of the other wizards.
I think it might make more sense to have a validator service, which the steps and the services that update the `training_period` use, and the latter service simply returns if the validation fails.  That shouldn't happen because the steps have also used the same validation service.  Then there's no need to raise and catch errors at this step.  
Just a thought though!

spec/services/admin/teachers/training_periods/change_contract_period/future_period_spec.rb
```diff
@@ -122,6 +122,76 @@
...
+      event = Event.where(event_type: "teacher_training_period_contract_period_changed").sole
+
+      expect(training_period.reload).to have_attributes(
+        teacher_id: teacher.id,
+        ect_at_school_period_id: future_ect_at_school_period.id,
+        mentor_at_school_period_id: nil,
+        started_on: future_started_on,
+        finished_on: future_finished_on
+      )
+      expect(training_period.school_partnership).to be_nil
```
thenapking, line 172: Is there any scope for making some of these expectations into a shared context which could also be used by `current_active_period_spec`.  It might make it more readable and comparible.

app/services/admin/teachers/training_periods/change_contract_period/eligibility.rb
```diff
@@ -14,14 +14,9 @@ def eligible?
...
 
-            if eoi_only?
-              return current_active_period? if no_future_periods?
-
-              return false
-            end
-
             return current_active_period? if no_future_periods?
             return only_future_period? if no_current_active_period?
+            return future_period? if eoi_only?
```
thenapking: I think this might be in the wrong order.  I think your test only tests the scenario where there is one future period so passes because of the `only_future_period? if no_current_active_period?` condition, not because `eoi_only?` is `true`

spec/services/admin/teachers/training_periods/change_contract_period/eligibility_spec.rb
```diff
@@ -97,44 +106,42 @@
...
-          :with_only_expression_of_interest,
-          ect_at_school_period: FactoryBot.create(
-            :ect_at_school_period,
-            teacher:,
-            school:,
-            started_on: future_started_on,
-            finished_on: nil
-          ),
-          expression_of_interest: training_period.expression_of_interest,
+  context "when there are current and future EOI only periods" do
```
thenapking, line 116: The code suggests there could be more than one future period because you test `future_periods.include?(training_period)`.  The previous test on eligability asserts there's only one future training period.  I'm not entirely sure if there can be two future training periods, but if there can be then we should test it.

avinhurry: Good point. The current implementation allows any eligible future period to be changeable, including future EOI only periods. I'll clarify with product what should happen when there is more than one future period and make any changes in a follow up PR. 

It's an edge case but worth thinking about.

Thank you!

app/services/admin/teachers/training_periods/change_contract_period/eligibility.rb
avinhurry: That ordering is intentional because the only future rule applies regardless of whether the period has a partnership or an EOI. The EOI specific branch is for the case where there is both a current period and a future EOI only period, where we don't have a LP/DP partnership to compare.

[This test](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/cb2921d396a5be79da14191723dd9b43bddccf62/spec/services/admin/teachers/training_periods/change_contract_period/eligibility_spec.rb#L116-L146) passes because of this line:

```ruby
return future_period? if eoi_only?
```

spec/services/admin/teachers/training_periods/change_contract_period/future_period_spec.rb
avinhurry: I did consider it, but I think keeping these explicit is clearer here.

The two services have similar EOI expectations, but they mutate differently. The current active ends the old period and creates a replacement but future updates the same period in place. A shared context would need to abstract that away which I think makes the specs less direct and harder to understand correctly.

app/wizards/admin/teachers/training_periods/change_contract_period_wizard/check_answers_step.rb
avinhurry: Yeah, I agree a shared validator would be a good improvement.

For this PR i've kept the existing structure because we are already using validations where the wizard collects input and the CYA rescue is acting as the final guard around the actual mutation. That protects against stale wizard state or underlying data changing between the earlier step validation and final submit.

I'll have a think but will do this in a follow up PR to keep this focused on the future EOI slice.

spec/services/admin/teachers/training_periods/change_contract_period/eligibility_spec.rb
avinhurry: It's very unlikely that we'll ever have multiple future periods, but if it does happen we've decided not to show any change links. That avoids the risk of changing one future period and leaving the others on a different contract period.

app/services/admin/teachers/training_periods/change_contract_period/eligibility.rb
thenapking: The conditions are now much easier to understand, thanks!

thenapking (approved):
LGTM

PR #3151, opened 2026-06-25
Format UUIDs consistently in the admin section

### LINKED ISSUE #2317: Format UUIDs in `<code>` tags

peteryates:
### Context

This resumes @peterdavidhamilton's work from #2628 which I hijacked and crashed.

Replaces the `.app-code` and `.app-mono` CSS class use with `<code>` and `<pre>` and introduces a new helper method, `format_uuid` which will make consistently formatting all UUIDs in the application easier.

The styles should look/feel more or less how they did before.

Fixes #2317 

- **Add format_uuid helper that wraps uuids in `<code>`**
- **Use format_uuid on all the UUID values in admin**
- **Replace .app-mono and .app-code with code/pre styles**
- **Replace hardcoded content_tag with direct tag helpers**

### DIFF
--- a/app/assets/stylesheets/admin.scss
+++ b/app/assets/stylesheets/admin.scss
@@ -194,14 +194,23 @@ $_current-link-border-width: govuk-spacing(1);
   overflow-wrap: break-word;
 }
 
-.app-mono {
-  padding: 2px 5px;
-  font-family: monospace;
-  background-color: $govuk-template-background-colour;
-  white-space: pre-wrap;
-}
-
 .govuk-table__row--no-border > th,
 .govuk-table__row--no-border > td {
   border-bottom: 0;
+}
+
+code {
+  background: $govuk-template-background-colour;
+  padding-inline: .4rem;
+}
+
+pre {
+  background: $govuk-template-background-colour;
+  white-space: pre-wrap;
+  @include govuk-font($size: 16);
+
+  code {
+    background: inherit;
+    padding-inline: initial;
+  }
 }
--- a/app/assets/stylesheets/admin/teachers.scss
+++ b/app/assets/stylesheets/admin/teachers.scss
@@ -1,13 +1,5 @@
 @use "govuk-frontend/dist/govuk" as *;
 
-.app-code {
-  padding: 2px 5px;
-  font-family: monospace;
-  background-color: $govuk-template-background-colour;
-  white-space: pre-wrap;
-  font-size: 0.8em;
-}
-
 .app-teachers-filters {
   @include govuk-media-query($from: tablet) {
     align-items: flex-end;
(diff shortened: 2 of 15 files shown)


app/assets/stylesheets/admin.scss
```diff
@@ -194,14 +194,23 @@ $_current-link-border-width: govuk-spacing(1);
...
+
+pre {
+  background: $govuk-template-background-colour;
+  white-space: pre-wrap;
+  @include govuk-font($size: 16);
+
+  code {
+    background: inherit;
+    padding-inline: auto;
+  }
```
peteryates, line 215: This stops the `code` rule above interfering with `pre > code` in case the two diverge in colour, and gets rid of the inline padding.

cpjmcquillan (approved):
`Nice`

avinhurry (commented):
Do these need updating too?

- https://github.com/DFE-Digital/register-early-career-teachers-public/blob/1c1994256200b60c6f0521d2fd703f5eb0381f6f/app/views/admin/teachers/declarations/_declaration.html.erb

- https://github.com/DFE-Digital/register-early-career-teachers-public/blob/1c1994256200b60c6f0521d2fd703f5eb0381f6f/app/views/admin/finance/voids/new.html.erb

app/assets/stylesheets/admin.scss
avinhurry: is `padding-inline: auto;` valid CSS? I thought padding [only accepts lengths/percentage values](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/padding-inline), so i'd expect the browser to ignore this and keep the global `code` padding applied inside `pre > code` but I haven't checked.

peteryates: Oh you're right - I meant `initial` here.

avinhurry: Nice one, that makes sense.

avinhurry (approved):
Looks great, nice one

app/assets/stylesheets/admin.scss
peteryates: it's `margin-inline` that has auto! :man_facepalming:  I knew I hadn't made it up _entirely_.

https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/margin-inline#syntax

PR #3153, opened 2026-06-25
Validate induction period dates the same as other intervals

cpjmcquillan:
Before c4c88645de47f8446dfe272f28a69f3f92c83215, we needed to override the date order validation for induction periods. This was because the range was still defined with an exclusive end.

Now that the range uses inclusives ends like the rest of our `Period` models, we can remove this validation override.


thenapking (approved):
LGTM

avinhurry (approved):

PR #3154, opened 2026-06-25
Fix specs failing on main

avinhurry:
## Summary

[The `Statement` model now validates that `payment_date` is later than `deadline_date`](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/3c5530f7f974e7bfaeaef9b4d45ce8343eb164a2/app/models/statement.rb#L37-L41).

These specs were overriding `deadline_date` to a past date and relying on the factory default `payment_date`, which could make the statement invalid.

Set `payment_date` explicitly in the affected examples so the setup remains valid and the specs continue to test `MarkAsPayable` behaviour.


peteryates (approved):
This looks sensible to me, but would it be worth adjusting the statements factory to set the payment date to something valid given the generated deadline date?

PR #3155, opened 2026-06-25
[3404] Include original frozen contract period option (Slice 8)

avinhurry:
## What

Following on from #3146, this is the 8th slice of the contract period change journey for admins.

The `2021` and `2022` contract periods are disabled in production. The wizard was starting from enabled contract periods, which meant the participant's original frozen contract period was filtered out before the frozen year exception could apply.

This PR updates the wizard so that:

- enabled contract periods are still shown
- the participants original frozen 2021/2022 contract period is shown, even though it is disabled
- the other frozen contract period remains hidden
- unrelated disabled contract periods remain hidden

## Not included

This PR does not:

- change the mutation behaviour
- change the no partnerships journey

### DIFF
--- a/app/wizards/admin/teachers/training_periods/change_contract_period_wizard/wizard.rb
+++ b/app/wizards/admin/teachers/training_periods/change_contract_period_wizard/wizard.rb
@@ -3,8 +3,6 @@ module Admin
     module TrainingPeriods
       module ChangeContractPeriodWizard
         class Wizard < ApplicationWizard
-          class UnexpectedTrainingPeriodTypeError < StandardError; end
-
           PartnershipOption = Data.define(:id, :name)
 
           attr_accessor :store, :teacher_id, :training_period_id, :author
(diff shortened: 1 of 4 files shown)


app/wizards/admin/teachers/training_periods/change_contract_period_wizard/wizard.rb
```diff
@@ -52,7 +52,13 @@ def teacher_name
           delegate :school, to: :training_period
 
           def contract_periods
-            scope = ContractPeriod.enabled.most_recent_first
+            scope = if original_frozen_contract_period_year.in?([2021, 2022])
+                      ContractPeriod.enabled.or(ContractPeriod.where(year: original_frozen_contract_period_year))
+                    else
+                      ContractPeriod.enabled
+                    end
```
peteryates: This feels a bit business-logicy to live in the depths of the wizard.

peteryates (approved):
Change looks good but I think the logic should perhaps be elsewhere. Could we have a scope that takes params, or a service object maybe?

app/wizards/admin/teachers/training_periods/change_contract_period_wizard/wizard.rb
avinhurry: Yeah, agreed. I don’t think a `ContractPeriod` scope is quite right here because the rules need training period, frozen years and EOI context.

I've extracted a small service object for the available contract periods instead, which keeps the wizard focused on the flow.

Thank you 🙇🏽

PR #3156, opened 2026-06-25
Replace primary navigation component with a service

peteryates:
The primary navigation component doesn't really have any component functionality in it, it just contains logic that determines which arguments are through passed to [govuk_service_navigation](https://govuk-components.x-govuk.org/components/service-navigation/). Many of the tests repeat/overlap the tests from the components library too, which isn't ideal.

As a result, it can be converted to a service object with minimal changes, and we can pass the arguments directly into `#govuk_service_navigation`.

This is hugely beneficial because one of the pieces of feedback from the accessibility report points out that we have several elements on the page aren't contained within [ARIA landmarks](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/landmark_role) (see DFE-Digital/register-ects-project-board#4082). The [govuk_header](https://govuk-components.x-govuk.org/components/header/#header-with-service-navigation-and-phase-banner) component has slots for the service navigation and the phase banner, which will render them inside the `<header>`, so this PR paves the way for that work.

## Changes

1. move the component to `app/services/navigation`
2. rewrite the tests so they test what the class does not what `#govuk_header` renders
3. replace the calls to the primary nav component with calls to `#govuk_header` using the arguments from `Navigation::Primary#govuk_header_arguments`


app/views/layouts/api_docs.html.erb
```diff
@@ -30,10 +30,14 @@
...
-      current_user: current_user,
-    ) %>
+    <%=
+      govuk_service_navigation(
+        **Navigation::Primary.new(
+          current_path: request.fullpath,
+          current_user: current_user,
+        ).govuk_header_arguments
+      )
+    %>
```
avinhurry, line 40: The nav is still being rendered outside of the header.

Image

Do we want this instead?


```suggestion
    <% header.with_service_navigation(
      **Navigation::Primary.new(
        current_path: request.fullpath,
        current_user: current_user
      ).govuk_header_arguments
    ) %>
  <% end %>
```

https://govuk-components.x-govuk.org/components/header/#header-with-service-navigation-and-phase-banner

If i've missed something that is intended for follow up work feel free to ignore me.

avinhurry (approved):
Had a question on whether we want to use `header.with_service_navigation` instead of `govuk_service_navigation` but the refactor itself looks great.

app/views/layouts/api_docs.html.erb
peteryates: Yeah, I was going to move both the secondary nav and the phase banner inside in a separate PR. This was just groundwork.

PR #3161, opened 2026-06-29
4148-update-lp-docs-site

StephenC-V1:
Update to show new changes

### Context
We've updated the reg opening email since this was last shown on the site. Update with new changes

Update to new version of email
### Guidance to review
Sanity check against Notify version

### DIFF
--- a/documentation/communication-emails-to-schools/reg-open-to-sit.md
+++ b/documentation/communication-emails-to-schools/reg-open-to-sit.md
@@ -26,27 +26,46 @@ Hello ((sit_name)),
 
 We’re sending you this email as our records show that you’re the induction tutor for ((school_name)).
 
-You can now register early career teachers (ECTs) and mentors for the 2026-2027 academic year with [register early career teachers](https://www.register-early-career-teachers.education.gov.uk/).
+You need to register early career teachers (ECTs) and mentors for training for the 2026 to 2027 academic year. Do this as soon as possible so that we can arrange funding and access to training materials.
+
+## Signing into the registration service is different
+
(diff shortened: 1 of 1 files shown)


documentation/communication-emails-to-schools/reg-open-to-sit.md
```diff
@@ -26,27 +26,46 @@ Hello ((sit_name)),
...
+1. Schools need to appoint a DfE Sign-in approver. This person verifies the identity of users before giving access to services. The approver is often the headteacher or administrator.   
+2. Users who want access to DfE services need to create a DfE Sign-in account. They use their account to select their school and request access to services.
+
+You can get help with [DfE Sign-in](https://help.signin.education.gov.uk/contact-us).
+
+## If you’re not expecting to register ECTs or mentors this year
+
+You do not need to do anything if your school is not expecting any ECTs or mentors for academic year 2026-27.
+
+If your school does not currently have ECTs or mentors and you do not expect new ECTs or mentors this term, you can [stop reminder emails from the DfE](#).
```
avinhurry, line 55: It's a bit confusing that this link goes nowhere. Might be worth removing the link from this version?

avinhurry (approved):
Left a comment about the link that goes nowhere, other than that, this looks good to me.

documentation/communication-emails-to-schools/reg-open-to-sit.md
StephenC-V1: The URL is generated for each school prior to sending from Notify, so I'm not sure how we'd get it to work here. However, it's helpful to show lead providers that school users have the option so I'd be reluctant to remove. 
 
As this page is for LP's to see what we send schools, I'm happy (reasonably) that it doesn't  have the 'full functionality'.

PR #3162, opened 2026-06-29
Remove redundant guard clause in `ReviewMentorEligibilityStep`

cpjmcquillan:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/4097

#### Remove redundant early return
It's hard to discern **why** this guard clause exists. And I don't want to tear
something down just because I don't understand it.

But after doing some extensive mapping of the mentor assignment journeys in
[Excalidraw] it looks like this branch is redundant.

The only entrypoint to the `Schools::AssignExistingMentorWizard` journey is by
submitting the mentorships form (i.e `Schools::MentorshipsController`).

The wizard is kicked off if the chosen mentor is eligible for their first
provider led training. By definition, that can only be true if the ECT in
question is undertaking provider-led training.

If an ECT is doing provider-led training, then they will have either a school
partnership or an expression of interest. So they will have a lead provider too,
whether that's through the school partnership or the expression of interest.

I can't see a scenario in which the ECT would not have a lead provider, so the
guard clause seems like it has no bearing on anything.

This removes the guard clause, with the intention of simplifying this journey
slightly.

[Excalidraw]: https://excalidraw.com/#json=iIK7sogwBa6ZtXmI9B5-Z,DwPh88qnr0Xwu4Tv_ySGqQ

#### Refactor for readability
This refactors the `ReviewMentorEligibilityStep` in the
`Schools::AssignExistingMentorWizard` to make things a little easier to read.

For now, the method definitions needed for the `TrainingPeriodSources` concern
are grouped together with a comment.

Down the line, I think things would be clearer and easier to unpick if we
rethink that abstraction. Especially since we are using it in places alongside
the `Schedules::Reuse` concern.

We've run into a number of bugs with `ContractPeriod` assignment since going
live, and we've made a bunch of changes in this area.

### Guidance to review

- [ ] Verify mentor assignment (and training eligibility/training period creation) behaviour is as expected

### DIFF
--- a/app/wizards/schools/assign_existing_mentor_wizard/review_mentor_eligibility_step.rb
+++ b/app/wizards/schools/assign_existing_mentor_wizard/review_mentor_eligibility_step.rb
@@ -19,8 +19,6 @@ module Schools
       end
 
       def create_mentor_training_period!
-        return unless lead_provider
-
         ActiveRecord::Base.transaction do
           training_period = TrainingPeriods::Create.provider_led(
             period: mentor_at_school_period,
--- a/app/wizards/schools/assign_existing_mentor_wizard/review_mentor_eligibility_step.rb
+++ b/app/wizards/schools/assign_existing_mentor_wizard/review_mentor_eligibility_step.rb
@@ -9,24 +9,27 @@ module Schools
     private
 
       def persist
-        AssignMentor.new(
-          ect: wizard.context.ect_at_school_period,
-          mentor: wizard.context.mentor_at_school_period,
-          author: wizard.author
-        ).assign!
-
+        assign_mentor!
         create_mentor_training_period!
+      end
+
+      def assign_mentor!
+        AssignMentor.new(
+          ect: ect_at_school_period,
+          mentor: mentor_at_school_period,
+          author:
+        ).assign!
       end
 
       def create_mentor_training_period!
         ActiveRecord::Base.transaction do
           training_period = TrainingPeriods::Create.provider_led(
             period: mentor_at_school_period,
-            started_on:,
+            started_on: mentor_at_school_period.started_on,
             school_partnership: earliest_matching_school_partnership,
             expression_of_interest:,
-            mentee: wizard.context.ect_at_school_period,
-            author: wizard.author
+            mentee: ect_at_school_period,
+            author:
           ).call
 
           record_training_period_event!(training_period)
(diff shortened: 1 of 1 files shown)


app/wizards/schools/assign_existing_mentor_wizard/review_mentor_eligibility_step.rb
```diff
@@ -9,24 +9,27 @@ def next_step = :confirmation
...
-
-        create_mentor_training_period!
       end
 
       def create_mentor_training_period!
         ActiveRecord::Base.transaction do
           training_period = TrainingPeriods::Create.provider_led(
             period: mentor_at_school_period,
-            started_on:,
+            started_on: mentor_at_school_period.started_on,
```
cpjmcquillan, line 28: The behaviour here hasn't changed, but it's now a lot clearer.

Is this intentional?

I'd expect the training period to start on the day the mentor is assigned, rather than the day the mentor started at the school.

cc @mason-emily @Aaminap 

For context, this is the bit in the graph where a School chooses to assign an existing mentor for an ECT, the mentor is eligible for their first provider-led training, and the school confirms that the ECT's lead provider will provide training to the mentor. This is marked out in pink ⤵️

Image

app/wizards/schools/assign_existing_mentor_wizard/review_mentor_eligibility_step.rb
```diff
@@ -37,34 +38,31 @@ def create_mentor_training_period!
 
       def record_training_period_event!(training_period)
         Events::Record.record_teacher_starts_training_period_event!(
-          author: wizard.author,
+          author:,
           teacher: mentor_at_school_period.teacher,
           school: mentor_at_school_period.school,
```
thenapking, line 43: `school` is defined below so we can omit it.
```suggestion
          school:,
```

app/wizards/schools/assign_existing_mentor_wizard/review_mentor_eligibility_step.rb
```diff
@@ -9,26 +9,27 @@ def next_step = :confirmation
...
+          ect: ect_at_school_period,
+          mentor: mentor_at_school_period,
+          author:
         ).assign!
-
-        create_mentor_training_period!
       end
 
       def create_mentor_training_period!
-        return unless lead_provider
```
thenapking, line 22: I'm not completely convinced this is unnecessary, only because there is no test scenario which covers the absence of a `lead_provider`.  Obviously creating the training period will fail and rollback the transaction, but won't it also raise an unhandled error?  I'm not really sure it would be better to add error handling here in place of a guard clause which is easy to understand.

cpjmcquillan: [There aren't any provider-led training periods without a lead provider through the school partnership or the expression of interest](https://www.register-early-career-teachers.education.gov.uk/admin/blazer/queries/390-provider-led-training-periods-without-a-lead-provider), **and** [we validate that all provider-led training periods have a school partnership or an expression of interest](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/26888de9daeeaf4fcd169cb3502466788e1f9aa1/app/models/training_period.rb#L75) when we create them.

...which is why I think this is unnecessarily defensive and redundant.

thenapking: OK that does make sense from the flow diagram, but it is far from obvious in the code in isolation.  If we could document this or make it obvious with some name changes then that might be sufficient.  My worry is that, without this specific diagram attached, a future dev could make changes, where the guard becomes necessary but misses that.  It could be as simple as a comment about the `lead_provider_for_ect_training` method, or a rename of that method to make it clear the ECT will always be in training.  Or a comment in the specs saying that a lead_provider will always be present so we don't need to test the scenario where it is absent. Or even rename the wizard to something like `AssignExistingMentorToECTInTraining`.  It's a bit verbose, but then its totally clear.

avinhurry: I think this is ok to remove. We've checked the data , and the model validation means provider-led training periods should always have a school partnership or EOI, so the `nil` provider case doesn't look reachable.

If we want to make the assumption clearer, i'd prefer a small spec for the EOI only path rather than a comment that we need to keep up to date but I don't think keeping the guard is necessary.

avinhurry (approved):

PR #3168, opened 2026-06-29
first_pass_declaration_fixes

boonali:
### Context

At the point of migration, x170 declarations failed to migrate successfully to RECT.

x89 of these were voided or submitted
x6 of these have since been created
The remaining x55 are all paid or clawed_back declarations from the 2021 or 2022 cohorts.

In some cases, the migrator didn't know which partnership to use or set up.

In some cases, there was no room left to set up the necessary periods for the correct cohort associated with the declarations.

In some cases, both will have been true.

The patch changes in this PR aim to patch relevant ASPs,TPs & create the declarations for these participants.


tonyheadford (approved):
Tested fine on migration

PR #3174, opened 2026-06-30
Fix ECT contract period reuse when switching to provider-led

ericaporter:
### Context
This PR fixes a bug where changing an ECT from school-led back to provider-led did not reuse provider-led training history from an earlier ECT-at-school period.

The existing reuse logic only looked at training periods on the **current** ECT-at-school period. In this journey, the relevant previous provider-led period can belong to an earlier ECT-at-school period for the same teacher, so the lookup missed it and fell back to the transition date’s contract period.

- Overrides the provider-led lookup in ECTAtSchoolPeriods::SwitchTraining
- Searches the teacher’s previous ECT provider-led training periods when switching from school-led back to provider-led
- Restricts the lookup to previous provider-led periods that started before the transition date
- Adds regression coverage for reusing provider-led contract-period history from an earlier ECT-at-school period

### Guidance to review
- The main thing to review is whether the override in ECTAtSchoolPeriods::SwitchTraining is the right place for this logic

N.B. I kept the change local to this service because Schedules::Reuse is shared by other flows, and this bug is specific to the ECT switch-training journey where the relevant provider-led history may belong to an earlier ECT-at-school period for the same teacher.

### DIFF
--- a/app/services/ect_at_school_periods/switch_training.rb
+++ b/app/services/ect_at_school_periods/switch_training.rb
@@ -66,6 +66,7 @@ module ECTAtSchoolPeriods
           .teacher
           .ect_training_periods
           .provider_led_training_programme
+          .where(ect_at_school_periods: { school_id: @ect_at_school_period.school_id })
           .where("training_periods.started_on < ?", date_of_transition)
           .latest_first
           .first
--- a/app/services/ect_at_school_periods/switch_training.rb
+++ b/app/services/ect_at_school_periods/switch_training.rb
@@ -56,7 +56,19 @@ module ECTAtSchoolPeriods
     attr_reader :school, :lead_provider
 
     def training_period_for_switch
-      @ect_at_school_period.current_or_next_training_period || @ect_at_school_period.latest_training_period
+      @ect_at_school_period.current_or_next_training_period ||
+        @ect_at_school_period.latest_training_period
+    end
+
+    def most_recent_provider_led_period
+      @most_recent_provider_led_period ||=
+        @ect_at_school_period
+          .teacher
+          .ect_training_periods
+          .provider_led_training_programme
+          .where("training_periods.started_on < ?", date_of_transition)
+          .latest_first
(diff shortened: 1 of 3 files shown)


spec/features/schools/ects/change/training_programme_spec.rb
```diff
@@ -75,6 +75,31 @@
...
+
+    when_i_change_the_training_programme
+    and_i_choose_the_lead_provider
+    and_i_continue
+    then_i_am_asked_to_check_and_confirm_the_change
+
+    and_i_confirm_the_change
+    then_i_see_the_provider_led_confirmation_message
+    and_the_new_provider_led_period_uses_the_previous_contract_period
+  end
```
avinhurry: This feels quite heavy for what it's proving. I think we can cover this more effectively at the unit level, and given how slow the suite is getting i'd avoid adding feature coverage unless it's buying us something meaningfully different.

The new service level test already looks sufficient to me, so i'd be inclined to drop this feature spec.

avinhurry (approved):
I left a suggestion on the test coverage/performance side, but the fix itself looks spot on to me.

app/services/ect_at_school_periods/switch_training.rb
```diff
@@ -56,7 +56,19 @@ def to_provider_led
...
-      @ect_at_school_period.current_or_next_training_period || @ect_at_school_period.latest_training_period
+      @ect_at_school_period.current_or_next_training_period ||
+        @ect_at_school_period.latest_training_period
+    end
+
+    def most_recent_provider_led_period
+      @most_recent_provider_led_period ||=
+        @ect_at_school_period
+          .teacher
+          .ect_training_periods
```
rwrrll, line 67: I think this will span previous schools, just checking whether that's what we want?

ericaporter: Good point, I'll add a filter on the current school

PR #3176, opened 2026-06-30
Relax permissions in the opening registration screens

ethax-ross:
### Context

Initially we opted to block operations for statements, lead provider delivery partnerships and contracts in a contract period that has started. 

After reviewing this, we would like to relax the permissions so that we can create/update/delete these when in a
contract period as long as it is not payments frozen (and subject to other conditions for statements).

- Relax permissions around statements

When creating or updating statements we should only be able to do so when the `deadline_date` has not yet passed.

When deleting a statement we should block if there are any associated declarations (which is the existing behaviour).

Update `Statement` model with additional validation rules; note I've used a `context` here for only allowing creation
of statements with a `deadline_date` in the past to avoid a large refactor, but we may want to try and tackle this
separately.

- Relax permissions around lead provider delivery partnerships

Currently we restrict the creation/deletion of lead provider delivery partnerships to only contract periods that have not yet started.

We want to relax this restriction to instead allow it for contract periods until they are payments frozen.

- Relax permissions around contracts

Allow adding/deleting contracts for contract periods as long as they are not payments frozen.

The behaviour change was already introduced with the change to `active_lead_provider.editable?`, this just updates the tests.

- Rework `Statement#deadline_date` validation rule

Moving to opt-out rather than opt-in.

### Guidance to review

We can pick up the remaining permission checks around contracts/bands as part of the tickets to add contracts/bands that are currently in-progress.

What do we think about the conditional validation rule around `deadline_date`? I also tried having it in the `Create` service but it felt awkward and disjointed.

### DIFF
--- a/app/models/active_lead_provider.rb
+++ b/app/models/active_lead_provider.rb
@@ -35,5 +35,6 @@ class ActiveLeadProvider < ApplicationRecord
   }
 
   delegate :name, to: :lead_provider, prefix: true
-  delegate :editable?, to: :contract_period
+
+  def editable? = !contract_period.payments_frozen?
 end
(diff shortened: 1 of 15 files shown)


peterdavidhamilton (approved):
The context and will_save_change_to seems wise especially if it avoids touching more.

PR #3177, opened 2026-06-30
Disable `actions` column to stop API IDs from wrapping

peteryates:
The actions column, unless suppressed, limits the width the 'value' column can take.

This causes long strings, like UUIDs, to wrap, which makes them even harder to read than they already are.

| Before | After |
| ------ | ----- |
| Screenshot From 2026-06-30 14-26-26 |Screenshot From 2026-06-30 14-26-37 |


avinhurry (approved):
Thanks Pete 🙇🏽

PR #3148, opened 2026-06-24
[4031] Part 4 - Retire old database fields

peterdavidhamilton:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/4031

[Band modelling](https://github.com/DFE-Digital/register-ects-project-board/issues/4031)

- [part 1](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3080)
- [part 2](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3115)
- [part 3](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3131)
- part 4 (this PR)
- [part 5](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3178)

### Changes proposed in this pull request

The data has been migrated in part 1, reconnected in part 3 and can now be removed.

⚠️ **NB: This PR is destructive.**

### Guidance to review

Calculators should be functioning and reports/statements displaying correct values using the ALP bands min/max values.

PR #3131, opened 2026-06-23
[4031] Part 3 - Switch calculators over

peterdavidhamilton:
### Context

[Band modelling](https://github.com/DFE-Digital/register-ects-project-board/issues/4031)

- [part 1](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3080)
- [part 2](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3115)
- part 3 (this PR)
- [part 4](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3148)
- [part 5](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3178)

### Changes proposed in this pull request

Migrate calculators to use `ActiveLeadProvider::Band#capacity`

- Enforce an ALP band per term and relax restriction on max values
- Remove traits that build declaration boundaries
- Update specs to create ALP bands
- Update methods to fetch capacity from ALP band
- Retire min/max methods and validations from the term

### Guidance to review

Screenshot 2026-07-03 at 10 41 56

Screenshot 2026-07-03 at 10 41 27

PR #3115, opened 2026-06-18
[4031] Part 2 - Rename band terms

peterdavidhamilton:
### Context

[Band modelling](https://github.com/DFE-Digital/register-ects-project-board/issues/4031)

- [part 1](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3080)
- part 2 (this PR)
- [part 3](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3131)
- [part 4](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3148)

### Changes proposed in this pull request

**No functional change.**

- rename old table appending `_terms`
- rename model `Contract::BandedFeeStructure::Band` to `Contract::BandedFeeStructure::BandTerm`
- rename association and methods to `band_terms` to expose the surface
- update tests and inline comments

image

PR #3178, opened 2026-06-30
[4031] Part 5 - optimisation and refactor

peterdavidhamilton:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/4031

[Band modelling](https://github.com/DFE-Digital/register-ects-project-board/issues/4031)

- [part 1](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3080)
- [part 2](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3115)
- [part 3](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3131)
- [part 4](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3148)
- part 5 (this PR)

No functional change. 

- Improve type checking of calculators by adding boolean predicates.
- Allow direct access to band values from the band_term.
- Change to using a factoried contract with associated band/term objects.

### DIFF
--- a/app/components/admin/statements/clawbacks_component.rb
+++ b/app/components/admin/statements/clawbacks_component.rb
@@ -61,7 +61,7 @@ module Admin
 
         return declaration_type_label if calculator.flat_rate?
 
-        "#{declaration_type_label} (Band #{declaration_type_output.band_term.band.letter})"
+        "#{declaration_type_label} (Band #{declaration_type_output.band_term.letter})"
       end
 
       delegate :number_to_pounds, to: :helpers
--- a/app/components/admin/statements/clawbacks_component.rb
+++ b/app/components/admin/statements/clawbacks_component.rb
@@ -51,24 +51,17 @@ module Admin
 
       def caption_text(calculator)
         return "Clawbacks" if contract.ecf_contract_type?
+        return "ECT clawbacks" if calculator.banded?
 
-        case calculator
-        when PaymentCalculator::Banded
-          "ECT clawbacks"
-        when PaymentCalculator::FlatRate
-          "Mentor clawbacks"
-        end
+        "Mentor clawbacks"
       end
 
       def payment_type(calculator, declaration_type_output)
         declaration_type_label = declaration_type_output.declaration_type.humanize
 
-        case calculator
-        when PaymentCalculator::Banded
-          "#{declaration_type_label} (Band #{declaration_type_output.band_term.band.letter})"
-        when PaymentCalculator::FlatRate
-          declaration_type_label
-        end
+        return declaration_type_label if calculator.flat_rate?
+
+        "#{declaration_type_label} (Band #{declaration_type_output.band_term.band.letter})"
       end
 
       delegate :number_to_pounds, to: :helpers
@@ -78,7 +71,7 @@ module Admin
         @calculators ||= PaymentCalculator::Resolver
           .new(statement:, contract:)
           .calculators
-          .sort_by { it.is_a?(PaymentCalculator::Banded) ? 0 : 1 }
+          .sort_by { it.banded? ? 0 : 1 }
       end
     end
   end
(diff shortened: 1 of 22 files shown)


app/components/admin/statements/clawbacks_component.rb
```diff
@@ -51,24 +51,17 @@ def uplift_row(calculator)
 
       def caption_text(calculator)
         return "Clawbacks" if contract.ecf_contract_type?
+        return "ECT clawbacks" if calculator.banded?
```
ethax-ross, line 54: I wonder if we want to be more explicit and raise here when an unexpected calculator is passed in (potentially in the future)? 🤔

app/components/admin/statements/output_payments_component.rb
```diff
@@ -17,21 +17,18 @@ def render?
       delegate :contract, to: :statement, private: true
 
       def calculators
-        @calculators ||= PaymentCalculator::Resolver.new(statement:, contract:).calculators
-          .sort_by { |c| c.is_a?(PaymentCalculator::Banded) ? 0 : 1 }
-          .map { presenter_for it }
+        @calculators ||= PaymentCalculator::Resolver.new(statement:, contract:)
+          .calculators
+          .sort_by { it.banded? ? 0 : 1 }
```
ethax-ross, line 22: Not sure if its any clearer, but you could `.sort_by { !it.banded? }`

app/services/payment_calculator/banded.rb
```diff
@@ -7,6 +7,9 @@ class Banded
     attribute :banded_fee_structure
     attribute :declaration_selector
 
+    def banded? = true
```
ethax-ross, line 10: It can be one for later, but I wonder if a concern would be more extensible (then the including class just overrides the method(s) that are `true`)

app/services/statements/declaration_selection.rb
```diff
@@ -29,10 +29,8 @@ def calculators
 
     def selected_ids_for(calculator)
       case calculator
-      when PaymentCalculator::Banded
-        selected_ids_for_banded(calculator)
-      when PaymentCalculator::FlatRate
-        selected_ids_for_flat_rate(calculator)
+      when PaymentCalculator::Banded   then selected_ids_for_banded(calculator)
```
ethax-ross, line 32: Do we want to use `banded?` and `flat_rate?` instance methods here for consistency?

ethax-ross (approved):
Looks good just minor thoughts/suggestions

app/components/admin/statements/output_payments_component.rb
peterdavidhamilton: It would be nice but `sort_by` doesn't work over booleans. Something to finesse though I agree.

app/services/statements/declaration_selection.rb
peterdavidhamilton: I did originally but then dropped the idea because of how the existing spec was written

app/components/admin/statements/clawbacks_component.rb
peterdavidhamilton: This speaks to the extensiblity with concerns and type checking... will there be more calculators ever?

ethax-ross: Yeah potentially, not this year but in the future if the CMs agree different terms with lead providers we could have other fee structures/calculators

app/components/admin/statements/output_payments_component.rb
ethax-ross: Ah that's annoying, I thought you could sort on booleans 😞

app/components/admin/statements/clawbacks_component.rb
peterdavidhamilton: Ok I'll keep that in mind. 👍🏼

app/components/admin/statements/output_payments_component.rb
peterdavidhamilton: I think I am making this go away whilst I update the CRUD PR. Fingers crossed.

PR #3179, opened 2026-07-01
Only allow user accounts with DfE email addresses

peteryates:
### Context

We currently have two groups of users who get a user account (a record in the `users` table):

* DfE users
* School users who we give ad hoc access to in non-prod envs for UR purposes

We want to limit all of our production users to being DfE staff for security purposes, and we want their access to stop once they've lost access to their DfE email address.

This PR adds a validation rule that ensures all user accounts have an email address that ends in `@education.gov.uk`.

### Notes

This will break the school signing-in via OTP functionality.

I think we need to either:

* formalise it and make it a fully-fledged `role` and feature
* remove it

Alternatively, if we were to add a new role we could adjust the validation so it only ensures DfE staff users need `@education.gov.uk` accounts.

> [!IMPORTANT]
> @avinhurry and I discussed this and decided that for now we'll go ahead and lock down accounts to `@education.gov.uk` ones, and I'll [raise a new issue](https://github.com/DFE-Digital/register-ects-project-board/issues/4164) to remove the functionality that supports schools logging in with an OTP

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1,4 +1,6 @@
 class User < ApplicationRecord
+  VALID_EMAIL_SUFFIX = "@education.gov.uk"
+
   ROLES = {
     admin: "Admin",
     user_manager: "User manager",
@@ -16,6 +18,8 @@ class User < ApplicationRecord
             presence: { message: "Enter an email address" },
             uniqueness: { message: "Email address already used, enter another" },
             notify_email: true
+  validate :ensure_email_belongs_to_dfe, if: -> { email.present? }
+
   validates :otp_school_urn,
             reference_number_format: {
               allow_blank: true,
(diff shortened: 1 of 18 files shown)


app/models/user.rb
```diff
@@ -1,4 +1,6 @@
 class User < ApplicationRecord
+  VALID_EMAIL_SUFFIX = "@education.gov.uk"
```
avinhurry, line 2: Are we dropping digital for good now? 😢

avinhurry (approved):

app/models/user.rb
peteryates: I was the only person who had a digital one so I changed it yesterday :( The initial version of this branch did account for both, but this simplifies the code, tests and error messages.

PR #3180, opened 2026-07-01
Only allow reuse of AB if active

ericaporter:
Ticket [here](https://github.com/DFE-Digital/register-ects-project-board/issues/4135)
### Context

Currently, if the school’s previously chosen appropriate body is no longer active, the school is still directed to the previous choices step. 

This PR updates the register ECT previous choices eligibility check so that previous choices can only be reused when the school’s last chosen appropriate body is active.

- Replaces the presence check with an active appropriate body check for both provider-led and school-led reuse paths, using `school.last_chosen_appropriate_body.active?`

### Guidance to review

The main behaviour to check is that UsePreviousECTChoicesStep is only allowed when school.last_chosen_appropriate_body is active.

### DIFF
--- a/app/wizards/schools/register_ect_wizard/use_previous_ect_choices_step.rb
+++ b/app/wizards/schools/register_ect_wizard/use_previous_ect_choices_step.rb
@@ -34,17 +34,21 @@ module Schools
         provider_led_reusable? || school_led_reusable?
       end
 
+      def previous_appropriate_body_active?
+        school.last_chosen_appropriate_body&.active? == true
+      end
+
       def provider_led_reusable?
         return false unless school.provider_led_training_programme_chosen?
         return false if school.last_chosen_lead_provider.blank?
-        return false if school.last_chosen_appropriate_body_id.blank?
+        return false unless previous_appropriate_body_active?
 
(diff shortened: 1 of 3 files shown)


app/wizards/schools/register_ect_wizard/use_previous_ect_choices_step.rb
```diff
@@ -34,17 +34,21 @@ def reusable_available?
         provider_led_reusable? || school_led_reusable?
       end
 
+      def previous_appropriate_body_active?
+        school.last_chosen_appropriate_body&.active? == true
+      end
```
cpjmcquillan: Comparing `something == true` feels totally unnecessary. 

I understand if we want to avoid chaining too many methods, but that isn't something we follow with any uniformity across the project.

So I'm not convinced `school.last_chosen_appropriate_body&.active?` is any less readable than `previous_appropriate_body_active?`.

Do we need it? 🤔

ericaporter: I was going for readability - but happy to revert!

cpjmcquillan (approved):

PR #3184, opened 2026-07-01
Align admin teacher pagination with display name sorting

avinhurry:
## Summary

Order the admin teacher search scope using the same display name logic used when rendering teacher rows.

Previously, teachers were ordered by their trs names in SQL, but the UI displayed corrected names (in Ruby). Because the ordering and rendering used different name logic, the visible ordering and pagination could become inconsistent making the alphabetical ordering appear incorrect on the UI.

The search scope now mirrors `Teachers::Name` in SQL by preferring `corrected_name`, falling back to the TRS full name, and using "Unknown" when neither name is present.

### DIFF
--- a/app/services/admin/teachers/search.rb
+++ b/app/services/admin/teachers/search.rb
@@ -211,12 +211,12 @@ module Admin
         <<~SQL.squish
           LOWER(
             COALESCE(
-              NULLIF(teachers.corrected_name, ''),
+              NULLIF(TRIM(teachers.corrected_name), ''),
               NULLIF(
                 CONCAT_WS(
                   ' ',
-                  NULLIF(teachers.trs_first_name, '.'),
-                  NULLIF(teachers.trs_last_name, '.')
+                  NULLIF(NULLIF(TRIM(teachers.trs_first_name), ''), '.'),
+                  NULLIF(NULLIF(TRIM(teachers.trs_last_name), ''), '.')
                 ),
                 ''
               ),
--- a/app/services/admin/teachers/search.rb
+++ b/app/services/admin/teachers/search.rb
@@ -13,7 +13,7 @@ module Admin
 
       def teacher_scope
         preload_associations(filtered_teacher_scope)
-          .reorder(:trs_first_name, :trs_last_name, :id)
+          .reorder(Arel.sql(display_name_order_sql), :id)
       end
 
     private
@@ -203,6 +203,25 @@ module Admin
         {
           latest_training_period: :schedule
         }
+      end
+
+      def display_name_order_sql
+        <<~SQL.squish
+          LOWER(
+            COALESCE(
+              NULLIF(teachers.corrected_name, ''),
+              NULLIF(
+                CONCAT_WS(
+                  ' ',
+                  NULLIF(teachers.trs_first_name, '.'),
+                  NULLIF(teachers.trs_last_name, '.')
+                ),
+                ''
+              ),
+              'Unknown'
+            )
+          )
+        SQL
       end
     end
   end
--- a/app/services/admin/teachers/search.rb
+++ b/app/services/admin/teachers/search.rb
@@ -206,6 +206,8 @@ module Admin
       end
 
       def display_name_order_sql
+        # Mirrors Teachers::Name in SQL so ordering and pagination use the same
+        # display name logic used by the UI.
         <<~SQL.squish
           LOWER(
             COALESCE(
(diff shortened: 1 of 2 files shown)


app/services/admin/teachers/search.rb
```diff
@@ -204,6 +204,25 @@ def training_period_preload
           latest_training_period: :schedule
         }
       end
+
+      def display_name_order_sql
```
rwrrll, line 208: Could be worth adding a tiny comment above this method mentioning that it's intended to mirror the Ruby method

app/services/admin/teachers/search.rb
```diff
@@ -204,6 +204,25 @@ def training_period_preload
...
+              NULLIF(
+                CONCAT_WS(
+                  ' ',
+                  NULLIF(teachers.trs_first_name, '.'),
+                  NULLIF(teachers.trs_last_name, '.')
+                ),
+                ''
+              ),
+              'Unknown'
+            )
```
rwrrll, line 224: `.reject { |n| n.blank? || n == "." }` drops empty string TRS name parts before joining, but `CONCAT_WS` only skips NULL arguments (not empty strings), so e.g.:

- `trs_first_name= ""; trs_last_name = "Smith"`: Ruby full_name = "Smith" but SQL order key = " smith" (leading space)

- `corrected_name = nil` & both TRS names "": Ruby = "Unknown" but SQL order key = " " (a lone space)

Haven't thought it through thoroughly but maybe a quick `TRIM` would solve this! If you think it's even worth addressing 🤷

spec/services/admin/teachers/search_spec.rb
```diff
@@ -15,6 +15,22 @@
...
+          corrected_name: "Aaron Aardvark"
+        )
+      end
+      let!(:other_teacher) { FactoryBot.create(:teacher, trs_first_name: "Bob", trs_last_name: "Builder") }
+
+      it "orders teachers by the displayed full name" do
+        expect(teacher_scope).to eq([teacher_with_corrected_name, other_teacher])
+      end
+    end
+
```
rwrrll, line 33: We don't currently exercise the literal "." case, empty string vs nil for either name source, or the "Unknown" fallback (i.e. the cases mentioned above)... worth adding? IDK, maybe it's a bit overkill, up to you!

rwrrll (approved):
Some bits to maybe consider but beats what we have at the moment so ✅ !

app/services/admin/teachers/search.rb
avinhurry: I kinda figured the commit message would do that job but happy to add, it's not obvious. I've added.

app/services/admin/teachers/search.rb
avinhurry: Very good catch! Thanks!

I've updated the SQL to `TRIM` and normalise the name values before constructing the display name so it more closely mirrors `Teachers::Name`.

spec/services/admin/teachers/search_spec.rb
avinhurry: Yeah, we should. I initially kept it minimal because I'd expect people to be searching here most of the time, but now that you've mentioned it I've added a regression spec to cover those edge cases.

app/services/admin/teachers/search.rb
```diff
@@ -13,7 +13,7 @@ def initialize(query_string: nil, role: nil, contract_period: nil)
 
       def teacher_scope
         preload_associations(filtered_teacher_scope)
-          .reorder(:trs_first_name, :trs_last_name, :id)
+          .reorder(Arel.sql(display_name_order_sql), :id)
```
peteryates, line 16: Sorry I know this horse has already bolted but this will make the query apply the functions below to every record in `teachers` every time the page is loaded without using any indexes.

It's fine for now but I wonder if this should/could be a generated column. Also if the column was a `citext` one it would mean no on-the-fly `lower` would be needed and it would use the table's index.

On the negative side, the name is _kinda_ business logicy, even though it's kind of common sense, and putting it in SQL on the table would make it a bit harder to adjust.

🤔

avinhurry: Yeah, that's fair. This fixes the pagination/display name mismatch, but the unfiltered page will still need to compute the expression to sort the teacher rows unless we give postgres a matching index.

We could add a concurrent expression index, but before we get too clever with SQL I'm going to check with Emily whether alphabetical sorting is actually useful here. The teachers table is ~269k rows in production, so i'm not sure we gain much from alphabetical ordering.

This page is probably used mainly via search, I think it may be more valuable to use a simple stable order like `created_at DESC, id DESC`. That would remove this custom display name sort completely and avoid adding more complexity.

avinhurry: PR to fix is here https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3342

PR #3187, opened 2026-07-01
[4142] Create school partnerships for GIAS Merge service

thenapking:
### Context

In order to merge schools following GIAS changes we will need to move or recreate `at_school_periods` in the new school.  If those `at_school_periods` have `training_periods` with confirmed school partnerships at the old school, then those partnerships need to be recreated at the new school.  If the a partnership already exists with the same `active_lead_provider` and `delivery_partner` then we should reuse that.  

This service finds or creates the required partnerships to move or recreate `at_school_periods` in the new school. 

At the moment this class is fairly simple. I expect that it may grow slightly more complex as we encounter some edge cases.  I looked at using similar classes such as `SchoolPartnerships::CreateFromPrevious` service.   But since that finds the closest match service I felt that might actually not be what is required in this situation.

### Guidance to review
At the moment there is nothing to product review.


app/services/school_partnerships/recreate.rb
```diff
@@ -0,0 +1,37 @@
+module SchoolPartnerships
+  class Recreate
```
cpjmcquillan: I think we should try and be a bit more explicit with the naming here so we can reveal more of the intent.

It might be worth adding a comment at the top of the class too outlining the purpose of this class.

Unless I'm missing something, I think this is specifically intended to create a school partnership at a school when the partnership already exists at another school (and the schools are being merged or amalgamated). Is that correct?

```ruby
module Schools
  module SchoolPartnerships
    class Create
      def self.for_gias_merge(...) = new(...).for_gias_merge

      def initialize(
        existing_school_partnership:, 
        school_without_partnership:, 
        author: Events::SystemAuthor.new
      )
        # assign them
      end

      def for_gias_merge
        # the code
      end
    end
  end
end
```

What do you think about something like this?

It's a little more verbose, but I think it communicates the intent more clearly and removes ambiguity about when this class should be used.

thenapking: This is absolutely the purpose and so that refactoring looks really sensible.  One other thing I was considering was making these explicit support services in the `GIAS::Schools` module, rather than lumping them in with other services.

app/services/gias/schools/merge/school_partnerships.rb
```diff
@@ -0,0 +1,41 @@
+module GIAS::Schools
+  module Merge
+    class SchoolPartnerships
```
cpjmcquillan, line 3: Just another small comment on naming here.

I think we have precedent across the application for modules that group together functionality for the underlying `ActiveRecord` model.

For example, we do `TrainingPeriods::Create.provider_led(...)`.

I think we should follow a similar convention, so `GIAS::Schools::SchoolPartnerships::Merge.resolve` instead.

What do you think?

cpjmcquillan: If we wanted to reflect the fact we're merging `Schools`, not `GIAS::Schools` (as per Teams discussion), what about something like this?

```ruby
module Schools::GIASMerge
  class SchoolPartnerships
    # the code
  end
end
```

PR #3188, opened 2026-07-02
deletions_first_batch

boonali:
### Context

As part of migration, a number of 1-day At School Periods and Training Periods were created with start dates prior to September 2021. This is not to be allowed within RECT.

The pre-September 2021 periods need to either be removed if they don't link to any declarations, or if they do, their subsequent periods need to be moved forward to allow the erroneous period to fit in.

This PR will delete all those ASPs and TPs with no declarations attached.

### Guidance to review

The relevant periods will not appear in RECT.


tonyheadford (approved):
Tested successfully in migration

PR #3191, opened 2026-07-02
updating_induction_periods

boonali:
### Context
We have a handful of induction period records where the trs_induction_completion_date doesn't line up with the latest date on our induction period.

Somehow, perhaps due to a short-lived bug, we have:
- 2 induction records per teacher
- the earlier one has an outcome on it (which is invalid, only the latest one should)
- in most cases the finished_on of the later one matches the trs_induction_completed_date (the last one is 1 day out)

For teachers 17932, 18556, 25957, 26697, remove the pass from the earlier induction period and set it on the later one

For teacher 72453, remove the pass from the earlier induction period and set it on the later one, and change the finished_on of the later one's end date to the next day, so it matches the trs_induction_completed_date

### Guidance to review
Run query below, should be zero results and the aforementioned changes should be in place.

select t.id, t.trs_induction_completed_date, ip.finished_on
from teachers t
inner join induction_periods ip on t.id = ip.teacher_id
where t.trs_induction_completed_date is not null
and outcome is not null
and t.trs_induction_completed_date <> ip.finished_on;


peteryates (approved):
💅  I've eyeballed it and it looks spot on. Still worth doing a dry run though.

tonyheadford (approved):
Tested with real data successfully

PR #3193, opened 2026-07-03
Contract CRUD with ALP band integration fix

peterdavidhamilton:
### Context

This branch (`pdh/contract-crud-with-alp-bands-integration`) applies @craigmdavidson's [Contract CRUD](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3145) atop the full ALP band work (`pdh/4031-band-modelling-part-5`) and [joins the dots](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3193/changes/525f95326e7ee5ec668f37755adcc356597a2476).

Closes #3145 

On top of the orignal work:

- CRUD controllers, services, event logging...
- Stimulus controller to automatically calculate the service fee % from the output fee %
- See original PR...

We are adding:

- ~Seed 2027 contract~
- Make use of new contract `with_bands_and_band_terms` trait
- Avoid sorting calculators
- Update specs for any changed behaviour
- JS testing
- Some snagging and extra coverage

### Guidance to review

cpd-ec2-review-3193-web test teacherservices cloud_admin_finance_contract-periods_2026_active-lead-providers_238_contracts

cpd-ec2-review-3193-web test teacherservices cloud_admin_finance_contract-periods_2026_active-lead-providers_238_contracts_1558


app/javascript/controllers/band_controller.js
```diff
@@ -0,0 +1,19 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class BandController extends Controller {
+  static targets = ["outputPercentage", "servicePercentage"]
+
+  connect() {
+    const target = this.servicePercentageTarget
+    target.readOnly = true
+    target.tabIndex = -1
+    target.style.opacity = "0.5"
```
ethax-ross: Can we move the styling to CSS instead of embedding it here?

app/services/payment_calculator/resolver.rb
```diff
@@ -9,7 +9,7 @@ class ContractMismatchError < StandardError; end
     attribute :contract
     attribute :statement
 
-    # @return [Array<PaymentCalculator::FlatRate, PaymentCalculator::Banded>]
+    # @return [Array<PaymentCalculator::Banded, PaymentCalculator::FlatRate>] in canonical display order
```
ethax-ross, line 12: Not sure about moving display logic into the service 🤔 maybe its fine as long as we have tests in the components in case someone changes it here without realising?

app/models/contract/banded_fee_structure/band_term.rb
```diff
@@ -35,6 +35,22 @@ class Contract::BandedFeeStructure::BandTerm < ApplicationRecord
 
   delegate :capacity, :min_declarations, :max_declarations, :letter, to: :band
 
+  def output_fee_percentage
```
ethax-ross, line 38: Can we add tests for these

app/models/contract.rb
```diff
@@ -15,6 +15,9 @@ class Contract < ApplicationRecord
   has_one :banded_fee_structure, class_name: "Contract::BandedFeeStructure", inverse_of: :contract, dependent: :destroy
   has_many :statements, inverse_of: :contract
 
+  # Scopes
+  scope :most_recent_first, -> { order(created_at: :desc) }
```
ethax-ross, line 19: Can we add a spec

app/services/contracts/create.rb
```diff
@@ -0,0 +1,18 @@
...
+
+    def initialize(author:, active_lead_provider:, params:)
+      @author = author
+      @active_lead_provider = active_lead_provider
+      @params = params
+    end
+
+    def call
+      contract = active_lead_provider.contracts.build(params)
+      contract.save!
```
ethax-ross: For the create/destroy/update can we wrap the save and event in a transaction

spec/services/contracts/create_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+      flat_rate_fee_structure_attributes: {
+        recruitment_target: 500,
+        fee_per_declaration: 100,
+      },
+    }
+  end
+
+  before { allow(Events::Record).to receive(:record_contract_created_event!) }
+
+  it "creates and returns a contract for the active lead provider, and records the created event" do
```
ethax-ross: Worth breaking up `params` so we can do more complete assertions? so:

```
let(:banded_fee_structure_attributes) { ... }

expect(contract.banded_fee_structure).to have_attributes(banded_fee_structure_attributes)
```

spec/services/contracts/create_spec.rb
```diff
@@ -0,0 +1,47 @@
+describe Contracts::Create do
```
ethax-ross, line 1: Can we also test when creation fails no event is emitted? ditto for `Update`

app/services/events/record.rb
```diff
@@ -1019,6 +1019,35 @@ def self.record_active_lead_provider_deleted_event!(author:, lead_provider:, con
...
+      heading = "Contract updated: #{contract.description} for #{lead_provider.name}"
+
+      new(event_type:, author:, heading:, active_lead_provider:, lead_provider:, happened_at:, modifications:).record_event!
+    end
+
+    def self.record_contract_deleted_event!(author:, active_lead_provider:, happened_at: Time.zone.now)
+      event_type = :contract_deleted
+      lead_provider = active_lead_provider.lead_provider
+      contract_period = active_lead_provider.contract_period
+      heading = "Contract deleted for #{lead_provider.name} in #{contract_period.year}"
```
ethax-ross: Worth passing in the `contract` for deletion so we can use `contract.description` for consistency? it should be fine even though we've destroyed it, it'll remain in memory

app/javascript/controllers/band_controller.js
```diff
@@ -0,0 +1,19 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class BandController extends Controller {
```
ethax-ross, line 3: Do we want to add unit tests for stimulus controllers? or cover them in feature specs?

app/javascript/controllers/band_controller.js
```diff
@@ -0,0 +1,19 @@
...
+    target.readOnly = true
+    target.tabIndex = -1
+    target.style.opacity = "0.5"
+    target.style.cursor = "not-allowed"
+    target.style.backgroundColor = "transparent"
+  }
+
+  updateServiceFee() {
+    const output = Number.parseFloat(this.outputPercentageTarget.value)
+    this.servicePercentageTarget.value = Number.isNaN(output) ? "" : (100 - output).toFixed(2)
```
ethax-ross: Can we round this so we get `50` instead of `50.00` and maybe cap the output fee to be 0-100?

ethax-ross (commented):
Looking good, just some minor thoughts/suggestions. We're missing request/view specs as well though, can we add those too?

There's also a bit of weirdness in the form; when it first loads you get banded and flat rate fields, but if you select the ITTECF ECTP contract type and submit (so you get an error) it removes the flat rate fields and you can't get them back. Is it worth another small stimulus controller to show/hide the fields?

app/services/payment_calculator/resolver.rb
peterdavidhamilton: It's not so much moving display logic as saying banded always comes first. I agree though, I'll add extra coverage

spec/models/contract_spec.rb
```diff
@@ -19,6 +19,28 @@
     it { is_expected.to have_many(:statements).inverse_of(:contract) }
   end
 
+  describe "scopes" do
+    describe ".most_recent_first" do
+      let(:active_lead_provider) { FactoryBot.create(:active_lead_provider) }
+      let(:other_active_lead_provider) { FactoryBot.create(:active_lead_provider) }
+      let!(:contract_1) { FactoryBot.create(:contract, active_lead_provider:) }
```
ethax-ross: Could this be flaky as they are evaluated in quick succession/could end up with identical timestamps? probably not, but might be worth wrapping in `travel_to() { }` blocks for explicitness?

app/models/contract_period.rb
```diff
@@ -67,6 +67,10 @@ def started_on_or_before_today?
     started_on <= Time.zone.today
   end
 
+  def finished_on_before_today?
```
ethax-ross, line 70: Can we add a spec

app/views/admin/finance/active_lead_providers/contracts/_fields.html.erb
```diff
@@ -0,0 +1,120 @@
...
+        inputmode: "decimal" %>
+
+  <%= bfs.govuk_text_field :setup_fee,
+        label: { text: "Setup fee (£)", size: "s" },
+        width: 10,
+        inputmode: "decimal" %>
+
+  <h3 class="govuk-heading-s">Bands</h3>
+
+  <table class="govuk-table">
```
ethax-ross: FWIW there is a component for tables we could use https://govuk-components.x-govuk.org/components/table/

db/seeds/contract_periods.rb
```diff
@@ -32,6 +32,11 @@ def describe_contract_period(cp)
     detailed_evidence_types_enabled: true,
     uplift_fees_enabled: false },
   { year: 2026,
+    enabled: true,
+    mentor_funding_enabled: true,
+    detailed_evidence_types_enabled: true,
+    uplift_fees_enabled: false },
+  { year: 2027,
```
ethax-ross: Can we remove this from the seeds before merging; I'm expect it might have a bearing on other teams/testing 🤔

app/views/admin/finance/active_lead_providers/contracts/_fields.html.erb
```diff
@@ -0,0 +1,120 @@
+<%= content_for(:error_summary) { form.govuk_error_summary } %>
+
+<%= form.govuk_collection_radio_buttons :contract_type,
+      Contract.contract_types.keys,
+      :itself,
+      ->(type) { type.humanize.upcase },
+      legend: { text: "Contract type", size: "s" } %>
+
+<%= form.govuk_text_field :ecf_contract_version,
+      label: { text: "ECF contract version", size: "s" },
```
ethax-ross, line 10: I think ECF contract versions can be binned off now; I'll raise a separate ticket for it but that should reduce the number of fields slightly at least 🙂

ethax-ross (commented):
Looks great, I'm still seeing the issue when switching contract type though (I checked and I can replicate it on the review app too) -- shall we jump on a call and I can show you what I'm doing?

ethax-ross (commented):
LGTM! 👏

ethax-ross (approved):

PR #3194, opened 2026-07-03
Speed up prepare-app-env CI action

rwrrll:
A couple of changes intended to speed up CI runs (no ticket)

- Don't install playwright deps already shipped with Ubuntu runner
- Cache built assets

Seems to reduce the `prepare-app-env` action's runtime by somewhere between 25-50% (10-20s)


avinhurry (approved):
We are relying on `ubuntu-latest` continuing to ship with the playwright system deps, so if that changes this stops working. Did you have any thoughts on that?

This looks good as things stand though.

cpjmcquillan (approved):

PR #3196, opened 2026-07-03
Fix flaky spec/services/api/declarations/clawback_spec.rb

rwrrll:
### Context

https://github.com/DFE-Digital/register-early-career-teachers-public/actions/runs/28666457963/job/85019283544?pr=3195

The above job failed due to a flaky (unrelated) spec, where randomness could cause the statement's `payment_date` not to fall after the `deadline_date`. 

By fixing `payment_date` to Date.tomorrow the validation can't fail.


avinhurry (approved):
Thanks @rwrrll!

PR #3197, opened 2026-07-03
Add `finished_on` to metadata refresh trigger

ethax-ross:
### Context

We are not refreshing the `teacher` metadata when the `finished_on` of an `InductionPeriod` is set. It's a long shot (because `finished_on` is likely set at the same time as `outcome`) but this may be contributing to a metadata refresh alert we're seeing that effects the transfer status of the `Teacher` (which is tied to induction completion).

Ensure we refresh metadata when the `InductionPeriod` is finished.

### Guidance to review

Relevant Sentry errors:

https://dfe-teacher-services.sentry.io/issues/7444944076/?environment=production&project=4508369974788096&query=is%3Aunresolved&referrer=issue-stream
https://dfe-teacher-services.sentry.io/issues/7590280327/?environment=production&project=4508369974788096&query=is%3Aunresolved&referrer=issue-stream

### DIFF
--- a/app/models/induction_period.rb
+++ b/app/models/induction_period.rb
@@ -21,7 +21,7 @@ class InductionPeriod < ApplicationRecord
 
   refresh_metadata -> { teacher },
                    on_event: %i[create destroy update],
-                   when_changing: %i[outcome]
+                   when_changing: %i[outcome finished_on]
 
   # Validations
   validates :appropriate_body_period_id, presence: { message: "Select an appropriate body" }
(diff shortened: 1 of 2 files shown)


peterdavidhamilton (approved):
Whilst I am not yet familiar enough with metadata the rationale behind the change is sound. 

`finished_on` should be the trigger for a refresh because `outcome` having a value is not guaranteed, it is `nil` for a release.

PR #3201, opened 2026-07-06
dp_id_updates

boonali:
### Context

When creating a number of declarations as part of ticket #4033, 16 fields for _delivery_partner_when_created_id_ were incorrectly populated. Therefore in the UI, they were showing as the wrong DP.

Amending the 16 incorrect _delivery_partner_when_created_id_ fields for the correct DP.

### Guidance to review

Check in the UI for the correct DP name for the declarations in the patch.


peteryates (approved):
Looks good - these are the ones we discussed last week right?

PR #3205, opened 2026-07-06
[4105] bands CRUD

tonyheadford:
> [!IMPORTANT]
> I'm expecting there will be some rebasing needed on top of the work that @peterdavidhamilton has been doing as this has been off to the side a bit so that I'm not treading on the toes of the larger piece of work

### Context

Adding the UI to manage fee capacity bands for the active lead providers

Adds `Admin::Finance::ActiveLeadProvider::BandsController`
Adds CRUD actions to controller (except show)
Adds `ActiveLeadProviders::Bands::(Create|Update|Destroy)` services
Adds `BandHelper` to expose some the Bands service to the views

### Guidance to review

image

image

image

#### When no contracts and contract period has not started

image


app/controllers/admin/finance/active_lead_providers/bands_controller.rb
```diff
@@ -0,0 +1,96 @@
...
+    def new
+      set_breadcrumbs
+      @band = @active_lead_provider.bands.new
+    end
+
+    def create
+      @band = @active_lead_provider.bands.new(band_params)
+
+      if @band.valid?
+        @band.save!
```
ethax-ross: Can we align this with the other controllers, so a service per action (create/update/delete with associated events)?

app/helpers/bands_helper.rb
```diff
@@ -0,0 +1,25 @@
+module BandsHelper
+  def label_for(band:)
+    "Band #{band.letter}"
+  end
+
+  def capacity_description_for(band:)
+    "#{band.min_declarations} - #{band.max_declarations}"
+  end
+
+  def editable?(band:)
```
ethax-ross: We've been putting equivalent methods in the models in other places; is it worth moving some of these and just keeping view-related helper methods in here?

spec/views/admin/finance/active_lead_providers/bands/edit.html.erb_spec.rb
```diff
@@ -0,0 +1,48 @@
+RSpec.describe "admin/finance/active_lead_providers/bands/edit.html.erb" do
```
ethax-ross, line 1: Worth a spec for `new.html.erb` as well?

app/services/admin/finance/bands.rb
```diff
@@ -0,0 +1,52 @@
+module Admin::Finance
```
ethax-ross: This is different to how we've been structuring the services for similar actions in the opening reg journey, can we make it consistent with the established pattern? So services for [create](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/services/statements/create.rb), [destroy](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/services/statements/destroy.rb) and [update](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/services/statements/update.rb)?

ethax-ross (commented):
Thanks Tony, I haven't had time to do a thorough review but I think its worth looking at some of the similar controllers/services/screens for other parts of the opening reg journey to make this more consistent with how those have been approached?

app/controllers/admin/finance/active_lead_providers/bands_controller.rb
```diff
@@ -0,0 +1,103 @@
+module Admin::Finance::ActiveLeadProviders
+  class BandsController < Admin::Finance::BaseController
+    include BandsHelper
+
+    before_action :set_active_lead_provider
+    before_action :set_band, only: %i[edit update delete]
+    before_action :redirect_if_contracted_or_inside_contract_period, only: %i[create delete]
```
ethax-ross: `redirect_unless_bands_can_be_added_or_removed`?

app/helpers/bands_helper.rb
```diff
@@ -0,0 +1,9 @@
+module BandsHelper
+  def label_for(band:)
```
ethax-ross: As view helpers all get mixed together is it worth prefixing these with `band_` to avoid potential clashes?

app/models/active_lead_provider/band.rb
```diff
@@ -28,13 +30,20 @@ class ActiveLeadProvider::Band < ApplicationRecord
...
+  validate :capacity_change_increases_value
+
   # Callbacks
   before_update :abort_update, unless: :last?
   before_destroy :abort_destruction, unless: :last?
   before_validation :assign_allocation_order,
                     on: :create,
                     if: -> { active_lead_provider.present? }
 
+  before_create :abort_if_contracted_or_inside_contract_period, unless: :allow_creation_when_contracted_or_after_contract_period_start
```
ethax-ross: Is it worth just having this as a regular `validate ... on: :create` then we don't need to handle errors in the `create` action on the controller?

Would 'contract period started' make more sense than 'inside contract period'?

app/models/active_lead_provider/band.rb
```diff
@@ -28,13 +30,20 @@ class ActiveLeadProvider::Band < ApplicationRecord
               message: "Capacity must be a number greater than zero"
             }
 
+  validate :capacity_change_increases_value
```
ethax-ross: `validate :capacity_can_only_increase`?

spec/models/active_lead_provider/band_spec.rb
```diff
@@ -1,5 +1,6 @@
 RSpec.describe ActiveLeadProvider::Band, type: :model do
```
ethax-ross, line 1: Can we add tests for `last?`, `editable?` and `deletable?`

app/models/active_lead_provider.rb
```diff
@@ -37,4 +37,8 @@ class ActiveLeadProvider < ApplicationRecord
   delegate :name, to: :lead_provider, prefix: true
 
   def editable? = !contract_period.payments_frozen?
+
+  def bands_can_be_added_and_removed?
```
ethax-ross, line 42: Can we add a spec

app/services/active_lead_providers/seed_from_previous.rb
```diff
@@ -96,7 +96,8 @@ def build_banded_fee_structure(previous_fee_structure)
     new_fee_structure = previous_fee_structure.dup
 
     previous_fee_structure.band_terms.each do |previous_term|
-      new_band = active_lead_provider.bands.create!(capacity: previous_term.capacity)
+      new_band = active_lead_provider.bands.create!(capacity: previous_term.capacity,
+                                                    allow_creation_when_contracted_or_after_contract_period_start: true)
```
ethax-ross: I don't think we should be using this bypass outside of factories/'normal' seed data. Could we not get around this by creating all the `Bands` before we create all the `Contracts` (which is the order we're imposing on users)?

app/services/events/record.rb
```diff
@@ -94,6 +96,7 @@ def initialize(
       @metadata = metadata || modifications
       @zendesk_ticket_id = zendesk_ticket_id
       @contract_period = contract_period
+      @band = band
```
ethax-ross, line 99: We can maybe pick it up in a separate ticket, but I wonder if its worth adding `Band` and `Contract` as associations on `Event`

app/services/events/record.rb
```diff
@@ -1308,6 +1311,54 @@ def self.record_school_changed_event!(author:, school:, old_gias_school:, new_gi
       new(event_type:, author:, heading:, school:, happened_at:, metadata:).record_event!
     end
 
+    ## band events
+
+    def self.record_active_lead_provider_band_added_event!(author:, band:)
```
ethax-ross, line 1344: Can we add tests for these

spec/requests/admin/finance/active_lead_providers/bands_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+      include_context "sign in as finance DfE user"
+
+      it "returns success" do
+        get path
+        expect(response.status).to eq(200)
+      end
+    end
+  end
+
+  describe "POST .../bands" do
```
ethax-ross, line 60: Can we also cover:

```
GET .../bands/:id/edit
PATCH .../bands/:id
GET .../bands/:id/delete
DELETE .../bands/:id
```

ethax-ross (commented):
Thanks Tony, this is looking good just some minor suggestions.

app/models/active_lead_provider/band.rb
tonyheadford: yes, much better

app/helpers/bands_helper.rb
tonyheadford: yes, really good point, thanks

app/controllers/admin/finance/active_lead_providers/bands_controller.rb
tonyheadford: yes, much clearer, thanks

app/services/events/record.rb
tonyheadford: I did wonder about it, yes a separate ticket would make sense

app/models/active_lead_provider/band.rb
tonyheadford: The service will still raise an error if `.save!` fails, but moved to a validation

spec/requests/admin/finance/active_lead_providers/bands_spec.rb
tonyheadford: I've dropped `GET .../bands/:id/delete` as it seemed overkill but have added specs for the the other others

ethax-ross (approved):
LGTM

PR #3206, opened 2026-07-06
Semantic logger 5.0.0 in dev

peteryates:
This change adds config for us to use semantic logger in dev as well as production, and adds colour to the dev logs to make reading them easier.

It sets the log level for `$stdout` to `info` but leaves `log/development.log` as `debug`.

Fixes #3173


rwrrll (approved):
LGTM! :shipit:

cpjmcquillan (approved):

PR #3212, opened 2026-07-07
Fix n+1 query on the personas page and refactor

peteryates:
This moves the logic from the `PersonasController` into a service class that does the same thing but now does all the querying at the start rather than n+1. And it has some specs.

I know this isn't a 'real' page and users won't actually see it, but the less noise we have the easier it will be to see real n+1 problems.


avinhurry (approved):
Nice one.

> the less noise we have the easier it will be to see real n+1 problems.

Totally agree.

PR #3214, opened 2026-07-07
Decommission the `paritycheck` environment

peteryates:
The parity check env was used to test API parity between ECF1 and ECF2 during the data migration. Now the migration is done and the service has been live for a while, it can be removed.


cpjmcquillan (approved):
🔥

ethax-ross (approved):
Thanks 👍  couple of other bits we may want to nuke:

https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/README.md?plain=1#L22
https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/.rubocop.yml#L113

PR #3215, opened 2026-07-07
[3404] Broaden partnership options for current contract period changes

avinhurry:
## Summary

Updating the contract period change wizard so current active training periods show all partnerships for the selected school and contract period.

When moving contract period, the school may not be working with the same LP/DP in the target year, or that DP may no longer be working with that LP. This gives admins the flexibility to choose the correct partnership.

Future training period changes remain restricted to the current active LP/DP when the future period is eligible.


ericaporter (approved):
Looks good! 🌞

PR #3216, opened 2026-07-08
Fix mentor assignment date overlap bug

cpjmcquillan:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/4181

This is best reviewed commit by commit. 

This fixes the bug outlined in [this issue](https://github.com/DFE-Digital/register-ects-project-board/issues/4181) by reworking the `Schools::AssignMentor` service so that it no longer tries to infer whether a newly registered mentor is mentoring only at their new school. 

Now, when a school registers a mentor that is effectively transferring (i.e only mentoring at the new school), we explicitly pass a flag that prevents the new mentor assignment (i.e `mentorship_period` from starting today). 

From a business perspective, this ensures mentor assignments can still be backdated when the mentor is moving schools. We only want that to apply in that specific scenario, though. It is never relevant when we are assigning a mentor that is already at the school, and this change prevents that.

### Guidance to review

- [x] School can change an ECT's mentor when the ECT has a current and a previous mentorship that overlap the date the mentor joined the school
- [x] School can assign a mentor
- [x] School can register a new mentor


app/services/ect_at_school_periods/switch_mentor.rb
```diff
@@ -31,9 +31,9 @@ def switch
 
     def assign_mentor!
       Schools::AssignMentor.new(
-        author:,
-        ect: ect_at_school_period,
-        mentor: mentor_at_school_period
+        ect_at_school_period:,
```
ericaporter, line 34: This is much clearer - there are a few other similarly named variables/keyword args across the project - it would be good to get those renamed too (under another ticket)

app/services/schools/assign_mentor.rb
```diff
@@ -1,101 +1,85 @@
...
 
-    def initialize(author:, ect:, mentor:)
+    def initialize(ect_at_school_period:, mentor_at_school_period:, author:, mentoring_at_new_school_only: false)
+      @ect_at_school_period = ect_at_school_period
+      @mentor_at_school_period = mentor_at_school_period
       @author = author
-      @ect = ect
-      @mentor = mentor
+      @mentoring_at_new_school_only = mentoring_at_new_school_only
     end
```
ericaporter, line 10: I wonder whether mentoring_at_new_school_only describes the wizard answer more than the behaviour AssignMentor actually cares about. Inside this service, the flag mainly controls whether Date.current is included in earliest_possible_start, so something like allow_backdated_start: might keep the service boundary a bit cleaner. I can see the reasoning behind the current name though, so I’m not strongly against it.

spec/wizards/schools/register_mentor_wizard/check_answers_step_spec.rb
```diff
@@ -120,8 +120,34 @@
         expect(subject.mentor).to have_received(:register!)
       end
 
-      it "assigns the created mentor to the ECT" do
-        expect(Schools::AssignMentor).to have_received(:new).with(ect: wizard.ect, mentor: new_mentor, author:)
+      context "when the mentor is mentoring only at the new school" do
+        before { store.mentoring_at_new_school_only = "yes" }
+
```
ericaporter, line 125: Worth checking - but I think this might not be testing the intended 'yes' branch since subject.save! runs in the parent before, so the call to AssignMentor.new has already happened by the time that value is set

spec/services/schools/assign_mentor_spec.rb
```diff
@@ -232,6 +220,84 @@
...
+            school: mentee.school
+          )
+        end
+        let!(:previous_mentorship_period) do
+          FactoryBot.create(
+            :mentorship_period,
+            started_on: mentee.started_on,
+            finished_on: previous_mentor.finished_on,
+            mentee:,
+            mentor: current_mentor
```
ericaporter: Possibly intentional but should it be previous_mentor here instead?

spec/services/schools/assign_mentor_spec.rb
```diff
@@ -232,6 +220,84 @@
       end
     end
 
+    context "when the mentor moved schools in the past" do
```
ericaporter: Could we add a regression test for assigning an ECT back to a previous mentor?

The support case I’m looking at currently- the ECT had mentor A from 1 June 2025 to 23 June 2026, then mentor B from 24 June 2026 onwards, and the school hit an error when trying to change back to mentor A.

I think the expected behaviour would be to close the current mentor B period and create a new non-overlapping mentor A period starting today, rather than trying to backdate into the old mentor A period. The support data still needs correcting manually but this would help confirm that the change mentor journey is covered. I am also happy to add this scenario retrospectively once this PR is merged!

spec/services/schools/assign_mentor_spec.rb
cpjmcquillan: Yep definitely, good spot!

spec/services/schools/assign_mentor_spec.rb
cpjmcquillan: Yes, I'll add a test here 👌🏼

app/services/schools/assign_mentor.rb
cpjmcquillan: I addressed this comment, but it surfaced heaps of tests around dates and their interaction with one another. I ended up doing some more refactoring to abstract that away.

That takes this PR off on a bit of a tangent, though, so we might want to pull those commits out into a separate PR.

Either way, this is ready for review again now!

ericaporter: This feels much clearer and keeps the wizard behaviour language separate

ericaporter (approved):
LGTM! 👍

PR #3218, opened 2026-07-08
[3726] Refactor statement components

rwrrll:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/3726

Looking for duplication and different approaches across all the statement components we built, and trying to iron it all out to end up with a neater and more coherent set of statement components.

### Guidance to review

I figured it would be good to look at and discuss real code rather than hypothetical changes, so everything here just for consideration, it's no worries to go in a different direction if we think there's a better way!

Trying to keep each refactor in its own commit for easier review.


app/components/admin/statements/payment_overview/ecf_component.rb
```diff
@@ -16,14 +16,11 @@ def rows
       private
 
         def uplifts
-          raise ArgumentError, "Expected exactly 1 calculator for ECF contract type" unless calculators.one?
```
thenapking, line 19: I think removing these errors is a very good idea.  It's not the components job to check this, the classes should always be returning the correct kind of information.

app/services/payment_calculator/collection.rb
```diff
@@ -0,0 +1,30 @@
+module PaymentCalculator
+  class Collection
+    class MissingCalculatorError < StandardError; end
```
thenapking, line 3: This is a sensible abstraction.  My only question would be whether this could be rolled into the `Resolver` class itself.  Perhaps `Resolver` wouldn't be the right name.  But I don't see why there couldn't be one class that returns the calculators, optionally ordered, or returns calculators of a specific kind, eg banded/flat-rate.

rwrrll: Yes nice idea and worth exploring! Hopefully fairly straightforward to try out and compare the two versions now this is extracted...

app/components/admin/statements/uplift_fees_component.rb
```diff
@@ -20,17 +20,10 @@ def render?
...
-        @calculators ||= PaymentCalculator::Resolver.new(contract:, statement:).calculators
-      end
+      delegate :contract, :calculators, to: :statement
 
       def uplifts
-        raise ArgumentError, "Expected exactly 1 calculator for ECF contract type" unless calculators.one?
-        raise ArgumentError, "Expected Banded calculator for ECF contract type" unless calculators.first.banded?
-
-        @uplifts ||= calculators.first.uplifts
+        @uplifts ||= calculators.banded_calculator.uplifts
```
thenapking, line 25: A query, in the ECF component uplifts are: `banded_calculator.uplifts.total_billable_amount`.  Why the discrepancy here?  I can see that this is a straight refactor of what the code originally said, so I presume this is correct, but it would be good to refactor it in a way to make it make sense!

app/components/admin/statements/calculator_presenters.rb
```diff
@@ -0,0 +1,22 @@
+module Admin
+  module Statements
+    module CalculatorPresenters
+    private
```
thenapking: Could we put this in a `concerns` subfolder?

app/components/admin/statements/clawbacks_component/base_presenter.rb
```diff
@@ -0,0 +1,34 @@
+module Admin
```
thenapking, line 1: When we were doing this originally we discussed different approaches to aggregating and displaying rows in the tables.  It's great this brings clawbacks in line with output payments component.  I took a different approach when building the declarations component, which iirc would have been cumbersome for the output payments.  So I think if we're going to use this approach we should also refactor the declarations component to use presenters rather than relying on layout.

app/components/admin/statements/payment_overview_component.rb
```diff
@@ -3,6 +3,8 @@ module Statements
     class PaymentOverviewComponent < ApplicationComponent
       delegate :number_to_pounds, to: :helpers
       delegate :contract, to: :statement
+      delegate :calculators, to: :statement, private: true
+      delegate :banded_calculator, to: :calculators, private: true
 
       attr_reader :statement
```
thenapking, line 10: Perhaps we should consider also using separate presenters for the overview component, rather than the separate components `PaymentOverview::ECFComponent` and `PaymentOverview::IttecfEctpComponent`,  that inherit from this base component?

thenapking (commented):
This looks like it is coming along well.  I've left a couple of questions and ideas, which might be things you were going to get to anyway.

app/services/payment_calculator/collection.rb
rwrrll: Had a proper look and I think they are separate responsibilities: `Resolver` constructs the calculators by dispatching on contract type, whereas `Collection` queries/iterates a _set_ of calculators. I think merging them would put "how to build from a contract" and "how to slice a set" in one class, which suits this particular application of rendering screens, but isn't necessarily a structural weld we want. That said, the underlying motivation to have one clean public door is basically already there: `Collection.for(statement)` is the single entrypoint, and `Resolver` is effectively its private builder. I think I'd lean towards keeping the split, but happy to go with a tie breaker if anyone disagrees with me!

app/components/admin/statements/uplift_fees_component.rb
rwrrll: Yep agree, this was confusing. I've renamed to `uplift_fees` so it matches its row label, and `uplifts` now consistently means "the uplifts object" everywhere (as in `UpliftFeesComponent`).

app/components/admin/statements/calculator_presenters.rb
rwrrll: Yeah makes sense!

app/components/admin/statements/clawbacks_component/base_presenter.rb
rwrrll: Yep cool beans — wanted to get a take on the presenter pattern elsewhere, but as we seem happy with that, I've changed this one over too 🚀

app/components/admin/statements/payment_overview_component.rb
rwrrll: Not sure on this one although I see the argument.

The presenter pattern in clawbacks/output-payments picks a presenter per calculator to aggregate rows. But the _overview_ varies by contract type and emits a _completely different_ fixed row set (different labels, different row count, mentor lines only for ITTECF). That's variance by type, rather than row aggregation, so presenter abstraction doesn't map as cleanly and the payoff feels smaller, subclassing a base component seems a legitimate fit here. Inclined to leave it, but again happy to take a tie-breaker into account if anyone else feels the uniformity is worth the slight awkwardness of this one!

app/services/payment_calculator/collection.rb
```diff
@@ -0,0 +1,30 @@
...
+
+    def self.for(statement)
+      new(Resolver.new(statement:).calculators)
+    end
+
+    def initialize(calculators)
+      @calculators = calculators
+    end
+
+    def each(&) = ordered.each(&)
```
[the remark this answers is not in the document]
rwrrll: Anonymous is the Ruby default and Rubocop default, I think that's more important than the fact we haven't happened to see this in the app so far.

cpjmcquillan (approved):
Nice work! 💫

PR #3220, opened 2026-07-08
Enable migration fixes (son-of-patch) to update read-only attrs

tonyheadford:
### Context

We have some migration data fixes (in a different PR) that need to update the value in `Declaration.delivery_partner_when_created_id`.  This is a `attr_readonly` attribute and as such Rails will not allow us to update the value and save it to the database.

This PR adds a workaround that temporarily drops the `attr_readonly` list on the target model so that the updates can be applied, then adds them back in again once the changes have been saved.

To enable this, the env var `UPDATE_READONLY_ATTRS` needs to be set.

### DIFF
--- a/app/services/migration_fixes/processor.rb
+++ b/app/services/migration_fixes/processor.rb
@@ -41,6 +41,7 @@ private
 
   def update!(target_object, attrs)
     return if attrs.blank?
+
     orig_readonly_attrs = []
 
     if update_readonly_attrs
--- a/app/services/migration_fixes/processor.rb
+++ b/app/services/migration_fixes/processor.rb
@@ -1,8 +1,9 @@
 class MigrationFixes::Processor
-  attr_reader :batch_refs
+  attr_reader :batch_refs, :update_readonly_attrs
 
-  def initialize
+  def initialize(update_readonly_attrs: false)
     @batch_refs = {}
+    @update_readonly_attrs = update_readonly_attrs
   end
 
   def process!(data_change: {})
@@ -40,8 +41,21 @@ private
 
   def update!(target_object, attrs)
     return if attrs.blank?
+    orig_readonly_attrs = []
+
+    if update_readonly_attrs
+      # hacky workaround as there doesn't seem to be another way to do this other
+      # than perhaps raw sql
+      orig_readonly_attrs = target_object.class._attr_readonly
+      target_object.class._attr_readonly = []
+    end
 
     target_object.update!(**attrs)
+
+    if orig_readonly_attrs.present?
+      # this reverts after the session anyway but belt and braces
+      target_object.class._attr_readonly = orig_readonly_attrs
+    end
   end
 
   def delete!(target_object)
--- a/app/services/migration_fixes/processor.rb
+++ b/app/services/migration_fixes/processor.rb
@@ -52,11 +52,9 @@ private
     end
 
     target_object.update!(**attrs)
-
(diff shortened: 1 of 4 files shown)


app/services/migration_fixes/processor.rb
```diff
@@ -41,7 +42,21 @@ def create!(object_type, attrs)
...
+      orig_readonly_attrs = target_object.class._attr_readonly
+      target_object.class._attr_readonly = []
+    end
+
     target_object.update!(**attrs)
+
+    if orig_readonly_attrs.present?
+      # this reverts after the session anyway but belt and braces
+      target_object.class._attr_readonly = orig_readonly_attrs
+    end
```
cpjmcquillan: Could we use `update_columns` here instead? 🤔

tonyheadford: Yes, I tried that but it didn't work. I tried all the work arounds I could find online but they all at best allow you to update the in memory model but not save to the database.

There are 2 different exceptions that get raised depending on how you approach it.  If you just try and `update` or assign a value to an instance you'll get an `ActiveRecord::ReadonlyAttributeError`.  You can silence that one by setting `Rails.configuration.active_record.raise_on_assign_to_attr_readonly = false` but....
You'll then hit an `ActiveRecord::ActiveRecordError`, which is also what you get when using `update_column` or `update_columns`. 

The only other option I think would be to translate the patch into raw SQL and execute it, but dropping the readonly attr list on the class (which would only persist for your session at most) seems far simpler if a little clunky

app/services/migration_fixes/processor.rb
```diff
@@ -41,7 +42,21 @@ def create!(object_type, attrs)
...
+    if update_readonly_attrs
+      # hacky workaround as there doesn't seem to be another way to do this other
+      # than perhaps raw sql
+      orig_readonly_attrs = target_object.class._attr_readonly
+      target_object.class._attr_readonly = []
+    end
+
     target_object.update!(**attrs)
+
+    if orig_readonly_attrs.present?
```
ethax-ross: The conditional looks redundant/could be dropped? in case `update!` throws an error for belt, braces and duct tape you could do:

```
...update...
ensure
    target_object.class._attr_readonly = orig_readonly_attrs
```

config/database.yml
```diff
@@ -27,7 +27,7 @@ production_default: &production_default
 
 development:
   <<: *default
-  database: ecf2_development
```
ethax-ross, line 30: Don't think we want to commit this 🙃

app/services/migration_fixes/processor.rb
ethax-ross: `update_columns` doesn't bypass readonly attributes but `update_all` does, so something like this would work:

```
if update_readonly_attrs
  target_object.class.where(id: target_object.id).update_all(**attrs)
  target_object.reload
else
  target_object.update!(**attrs)
end
```

But `update_all` also bypasses callbacks and will cause metadata to go out of sync, so I think how you have it is the best approach.

ethax-ross (approved):

config/database.yml
tonyheadford: Good spot!

app/services/migration_fixes/processor.rb
tonyheadford: I so nearly did this originally, but given you've also thought about, it make sense to add it in

PR #3222, opened 2026-07-10
Fixing n+1 queries on the school ECTs index

peteryates:
### Context

There are lots of N+1 queries on the schools index page.

The main reason is that the view components are doing too much heavy lifting, we probably want to move to a model where we pass collections of structs/objects in and the view component logic uses it directly, rather than the component issuing its own queries. Rewriting can wait though, this solves most of the problem for now.

### Final task

There's one remaining n+1 that's trickier to fix. It's the eligible mentors one, the scope excludes the current teacher - so it's run every time. This will be fixed by getting a list of eligible for the school up front, and then excluding the current teacher in Ruby rather than SQL.

- **Fix some linting errors**
- **Fix most n+1 queries on the school ECT index page**

### DIFF
--- a/app/controllers/schools/ects_controller.rb
+++ b/app/controllers/schools/ects_controller.rb
@@ -5,7 +5,20 @@ module Schools
     include Schools::InductionRedirectable
 
     def index
-      search = Teachers::Search.new(ect_at_school: school, in_progress: true, query_string: params[:q]).search
+      search = Teachers::Search
+        .new(
+          ect_at_school: school,
+          in_progress: true,
+          query_string: params[:q]
+        )
+        .search
+        .includes(
+          ect_at_school_periods: %i[latest_training_period mentorship_periods],
+          current_or_next_ect_at_school_period: [
+            { latest_training_period: %i[lead_provider delivery_partner] },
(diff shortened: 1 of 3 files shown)


app/controllers/schools/ects_controller.rb
```diff
@@ -5,7 +5,20 @@ class ECTsController < SchoolsController
...
     def index
-      search = Teachers::Search.new(ect_at_school: school, in_progress: true, query_string: params[:q]).search
+      search = Teachers::Search
+        .new(
+          ect_at_school: school,
+          in_progress: true,
+          query_string: params[:q]
+        )
+        .search
+        .includes(
```
thenapking, line 15: Is it possible we could refactor the `Teachers::Search` service itself to do this? Or do the tables/columns we'd want to have in the `includes` statement change per use context?

peteryates: Yeah it's used by multiple contexts. I did toy with the idea of passing the `includes` structure into the initializer. Perhaps if `#search` was aliased to `#scope` this usage would make more sense.

cpjmcquillan (approved):

PR #3223, opened 2026-07-10
[4182] Remove enable_finance_contract_periods flag

peterdavidhamilton:
### Context

Now live on all deployments

Remove the flag and env var

### Guidance to review

Everything under `/admin/finance/contract-periods` still works


cpjmcquillan (approved):
Should we remove it from the all the config files in `config/terraform/application/config` too?

PR #3224, opened 2026-07-10
[4142] Services to transfer ECTs and Mentors

thenapking:
### Context

In order to merge two school records, when we've received notification from GIAS that the schools are merging, we need to  move all the `ect_at_school_periods` and `mentor_at_school_periods` in the predecessor school to the successor school.  For ECTs this is easy, because a teacher can only have `ect_at_school_periods` at one school at a time, so there can be no overlaps.  `mentor_at_school_periods` are different because a teacher can mentor at two different schools simultaneosly.  

This PR introduces services which move `ect_at_school_periods` and `mentor_at_school_periods`  between two schools, on the basis that nothing overlaps.  Another PR will introduce services to deal with overlapping `mentor_at_school_periods`.  A final PR will use all these services to merge two schools.

To move an ECT, all we need to do is ensure that any related records - `training_periods` and `events` - are moved.  If those have a relation to a `school_partnership` we will need to find or create this partnership at the successor school.  Moving a mentor is exactly the same, with the provision that we should move their `mentorship_periods`.  Since `mentorship_periods` relate ECTs and mentors, and a mentor can have many ECTs, but an ECT can have only one mentor,when we come to build the service to merge schools it will process mentors first, moving any ECTs associated with them.  Finally we would move any remaining ECTs who are not associated with mentors.

#### Brief outline of expected merge schools service

```
def merge_school!
     mentor_teachers.each do |teacher|
      overlapping_mentor_at_school_periods(teacher).each do |periods|
        GIAS::Schools::MentorAtSchoolPeriods::Merge.call(periods:, predecessor_school:, successor_school:)
      end
    end

    mentors_without_overlapping_periods.each do |period|
      GIAS::Schools::MentorAtSchoolPeriods::Transfer.call(period:, predecessor_school:, successor_school:)
    end

    ects_without_mentorship_periods.each do |period|
      GIAS::Schools::ECTAtSchoolPeriods::Transfer.call(period:, predecessor_school:, successor_school:)
    end

    record_school_merged_event!
  end
end
````

### CODE REVIEW

File: app/services/gias/schools/school_partnerships/transfer.rb [hunk 1]
```diff
@@ -0,0 +1,61 @@
+module GIAS::Schools
+  module SchoolPartnerships
+    class Transfer
```
cpjmcquillan, line 3: I'm still not convinced these classes should live in the `GIAS::Schools` namespace, since we're transferring partnerships etc between `School` records, not between `GIAS::School` records.

What do you think?

app/services/concerns/periods/transferable.rb
```diff
@@ -0,0 +1,62 @@
+module Periods
+  module Transferable
```
cpjmcquillan: I'm not sure this abstraction is worth it yet. We're using it in two places, and once of those usages needs overriding already.

Is the duplication hurting us particularly?

app/services/concerns/periods/transferable.rb
```diff
@@ -0,0 +1,62 @@
...
+    end
+
+    def initialize(period:, predecessor_school:, successor_school:)
+      @period = period
+      @successor_school = successor_school
+      @predecessor_school = predecessor_school
+    end
+
+    def call
+      return unless predecessor_school == period.school
```
ericaporter: Should this guard run before the mentorship periods are transferred? I think it's only reached at the moment after super is called

app/services/concerns/periods/transferable.rb
```diff
@@ -0,0 +1,62 @@
...
+
+      ActiveRecord::Base.transaction do
+        update_events!
+        assign_training_periods
+
+        period.assign_attributes(school: successor_school)
+
+        training_periods.each(&:save!)
+
+        period.save!
```
ericaporter: Do we want to record an event that an ECT / Mentor periods has moved school here?

app/services/concerns/periods/transferable.rb
thenapking: I'm happy to remove this - it was a pretty straight forward abstraction, I don't have strong opinions either way.

app/services/gias/schools/school_partnerships/transfer.rb
thenapking: I suppose I have named this like this because they are support services for `GIAS::Merge`.  I take your point though, and did consider several other naming schemes.  I don't really want to have them in `services/schools` as they are pretty specifically for GIAS.  What about putting them in `/lib`?

cpjmcquillan: I don't think we should dump them in `/lib`.

What about something like this?

```ruby
module Schools
  module GIASAutomations
    module SchoolPartnerships
      class Transfer; end
    end
  end
end
```

or even 

```ruby
module GIASAutomations
  module Schools
    module SchoolPartnerships
      class Transfer; end
    end
  end
end
```

cpjmcquillan: @peteryates any thoughts?

thenapking: Yeah, I think that's a good idea.  We already have the `GIAS::Importer` service.  It will use these services so perhaps we could do `GIAS::Importer::Schools::Open` for instance? Or `GIAS::Importer::SchoolPartnerships::Transfer`

app/services/gias/schools/ect_at_school_periods/transfer.rb
```diff
@@ -0,0 +1,71 @@
...
+
+        def new_partnership_for(object)
+          GIAS::Schools::SchoolPartnerships::Transfer.call(object:, successor_school:)
+        end
+
+        def record_event!
+          Events::Record.record_teacher_ect_at_school_period_moved_school!(
+            teacher:,
+            ect_at_school_period:,
+            old_school_name: predecessor_gias_school.name,
```
ericaporter: You could record the old school name & URN for debugging ease

spec/services/gias/schools/ect_at_school_periods/transfer_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+  let(:gias_school) { FactoryBot.create(:gias_school, :with_school) }
+  let(:predecessor_school) { predecessor_gias_school.school }
+  let(:successor_school) { gias_school.school }
+
+  let(:ect_at_school_period) { FactoryBot.create(:ect_at_school_period, school: predecessor_school, started_on:, finished_on:) }
+
+  let(:started_on) { Date.new(2025, 1, 1) }
+  let(:finished_on) { Date.new(2025, 12, 31) }
+
+  describe ".call" do
```
ericaporter, line 13: Optional - You could consider adding a context where the ect period does not belong to the predecessor school so that nether the mentor period or linked ECT periods are transferred - covering the early return of the guard (and possibly the same in the mentor period file)

spec/factories/event_factory.rb
```diff
@@ -8,6 +8,9 @@
     author_type { :appropriate_body_user }
     event_type { Event::EVENT_TYPES.sample }
 
+    association :school
+    association :school_partnership
```
cpjmcquillan: This means every test that uses this factory will build/create a `School` record and a `SchoolPartnership` (and their associations) by default.

Are we sure we want to do that?

app/services/gias/schools/school_partnerships/transfer.rb
```diff
@@ -0,0 +1,61 @@
...
+      class InvalidObject < StandardError; end
+      class InvalidPeriodType < StandardError; end
+
+      attr_reader :object, :successor_school, :author
```
cpjmcquillan: Do these need to be public methods?

app/services/gias/schools/school_partnerships/transfer.rb
```diff
@@ -0,0 +1,61 @@
...
+
+      def initialize(object:, successor_school:, author: Events::SystemAuthor.new)
+        @object = object
+        @successor_school = successor_school
+        @author = author
+
+        raise InvalidObject unless object.respond_to?(:school_partnership)
+      end
+
+      def self.call(**args) = new(**args).call
```
cpjmcquillan: I think it's helpful if we group class methods before any instance methods, especially if this method is intended to be the entry point to the service.

What do you think about moving this before the initializer?

cpjmcquillan: Also, you can forward arguments like this


```suggestion
      def self.call(...) = new(...).call
```

app/services/gias/schools/school_partnerships/transfer.rb
```diff
@@ -0,0 +1,61 @@
+module GIAS::Schools
+  module SchoolPartnerships
+    class Transfer
+      class InvalidObject < StandardError; end
+      class InvalidPeriodType < StandardError; end
+
+      attr_reader :object, :successor_school, :author
+
+      def initialize(object:, successor_school:, author: Events::SystemAuthor.new)
```
cpjmcquillan: Everything in Ruby is an `Object`, so I think this is a little too ambiguous.

Could we be more explicit and pass in the school partnership to be recreated directly?

It simplifies things here a bit, and I think it makes things easier to follow upstream too.

What do you think?

```ruby
module GIAS::Schools
  module SchoolPartnerships
    class Transfer
      def self.call(...) = new(...).call

      def initialize(school_partnership:, succcessor_school:, author: Events::SystemAuthor.new)
        @school_partnership = school_partnership
        @successor_school = successor_school
        @author = author
      end
      
      def call
        return unless reassignment_required?

        find_or_create_school_partnership_at_successor_school
      end

      private

      attr_reader :school_partnership, :successor_school, :author

      def reassignment_required?
        return false if school_partnership.blank? || successor_school.blank?

        predecessor_school != successor_school
      end

      def find_or_create_school_partnership_at_successor_school!
          successor_school_partnership = SchoolPartnership.find_or_create_by!(
            school: successor_school,
            lead_provider_delivery_partnership:
          )

          if successor_school_partnership.previously_new_record?
            Events::Record.record_school_partnership_recreated_event!(
              old_school_partnership: school_partnership,
              new_school_partnership: successor_school_partnership,
              author:
            )
          end

          successor_school_partnership
        end
      end

      def predecessor_school = school_partnership.school
      delegate :lead_provider_delivery_partnership, to: :school_partnership, private: true
    end
  end
end
```

app/services/gias/schools/ect_at_school_periods/transfer.rb
cpjmcquillan, line 1: Could we collect records with the associations explicitly, rather than relying on `SchoolPartnerships::Transfer` returning `nil`?

```ruby
module GIAS::Schools
  module ECTAtSchoolPeriods
    class Transfer
      def self.call(...) = new(...).call

      def initialize(ect_at_school_period:, predecessor_school:, successor_school:)
        @ect_at_school_period = ect_at_school_period
        @predecessor_school = predecessor_school
        @successor_school = successor_school
      end
      
      def call
        return unless predecessor_school == ect_at_school_period.school

          update_ect_at_school_period!
          update_training_periods!
          update_events!
          record_event!
        end
      end

      private

      def update_ect_at_school_period!
        ect_at_school_period.update!(school: successor_school)
      end

      def update_training_periods!
        training_periods.where.associated(:school_partnership).each do |training_period|
          new_partnership = new_partnership_for(training_period.school_partnership)
          training_period.update!(school_partnership: new_partnership)
        end
      end

      def update_events!
        events.where.associated(:school_partnership).each do |event|
          new_partnership = new_partnership_for(event.school_partnership)
          event.update!(school_partnership: new_partnership)
        end

        events.where.associated(:school).each do |event|
          event.update!(school: successor_school)
        end
      end

      def record_event!
        Events::Record.record_teacher_ect_at_school_period_moved_school!(
            teacher:,
            ect_at_school_period:,
            old_school_name: predecessor_gias_school.name, 
            new_school: successor_school,
            happened_at: predecessor_gias_school.closed_on,
            author: Events::SystemAuthor.new
          )
      end

      attr_reader :ect_at_school_period, :predecessor_school, :successor_school

      delegate :mentorship_periods, :training_periods, :events, :teacher, to: :ect_at_school_period, private: true
    end
  end
end
```

cpjmcquillan (commented):
I think whatever we land on for naming is going to be fine. I agree it'll be helpful to have these services co-located (e.g in `gias/schools`).

I think we should lose the `object` part though and pass in the school partnerships to transfer explicitly.

Other than that, this looks good!

spec/factories/event_factory.rb
thenapking: Yeah - actually I don't think I needed this addition in the end at all so I've removed this.

app/services/gias/schools/school_partnerships/transfer.rb
thenapking: I'd originally had this like that, but had found that I needed common code outside class everytime I used it in order to determine when an update was required.  Checking the object's associations does seem to get round that, so I've updated this inline with your suggestion.

app/services/events/record.rb
```diff
@@ -381,25 +381,25 @@ def self.record_teacher_mentor_at_school_period_deleted!(author:, teacher:, scho
       new(event_type:, author:, heading:, teacher:, school:, happened_at:).record_event!
     end
 
-    def self.record_teacher_ect_at_school_period_moved_school!(author:, teacher:, ect_at_school_period:, old_school_name:, new_school:, happened_at: Time.zone.now)
+    def self.record_teacher_ect_at_school_period_moved_school!(author:, teacher:, ect_at_school_period:, old_school_name_and_urn:, new_school:, happened_at: Time.zone.now)
       event_type = :teacher_ect_at_school_period_moved_school
       teacher_name = Teachers::Name.new(teacher).full_name
-      new_school_name = Schools::Name.new(new_school).name_and_urn
+      new_school_name_and_urn = Schools::Name.new(new_school).name_and_urn
```
cpjmcquillan, line 387: Nice 👌🏼

app/services/gias/schools/mentor_at_school_periods/transfer.rb
```diff
@@ -0,0 +1,80 @@
...
+          end
+
+          events.where.associated(:school).each do |event|
+            event.update!(school: successor_school)
+          end
+        end
+
+        def update_mentorship_periods!
+          mentorship_periods.each do |mentorship_period|
+            GIAS::Schools::ECTAtSchoolPeriods::Transfer.call(ect_at_school_period: mentorship_period.mentee, predecessor_school:, successor_school:)
```
cpjmcquillan, line 54: Does this mean we end up updating some ECTs more than once? (i.e as an ECT, and as a mentee) 🤔

spec/services/gias/schools/ect_at_school_periods/transfer_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+
+      subject
+    end
+
+    context "when the ect_at_school_period has associated training periods" do
+      context "when the training_period only has an expression of interest" do
+        let!(:training_period) { FactoryBot.create(:training_period, :for_ect, :with_only_expression_of_interest, ect_at_school_period:) }
+
+        it "updates the ect_at_school_period's school to the successor school but makes no change to the training_period" do
+          expect { subject }.not_to(change { training_period })
```
cpjmcquillan: Does this definitely fail if the school partnership is set? I'd expect us to need to be more specific than asserting no change to the training period record 💭

spec/services/gias/schools/mentor_at_school_periods/transfer_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+
+      subject
+    end
+
+    context "when the mentor_at_school_period has associated training periods" do
+      context "when the training period has only an expression of interest" do
+        let!(:training_period) { FactoryBot.create(:training_period, :for_mentor, :with_only_expression_of_interest, mentor_at_school_period:) }
+
+        it "updates the mentor_at_school_period's school to the successor school but makes no change to the training_period" do
+          expect { subject }.not_to(change { training_period })
```
cpjmcquillan: Same here 💭

cpjmcquillan (approved):

app/services/gias/schools/mentor_at_school_periods/transfer.rb
thenapking: The services are called in an order to prevent this in the [final](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3244) service.  First we move each mentor with overlapping periods, and all their ects.  Then we re-query and move all the remaining mentors, and all their ects.  Then we re-query and move any ect without a mentor.

spec/services/gias/schools/ect_at_school_periods/transfer_spec.rb
thenapking: I've changed this test to very specifically test that the `school_partnership_id` does not change once reloaded.

PR #3225, opened 2026-07-10
[4142] Service to merge overlapping Mentor periods

thenapking:
### Context
In order to merge two school records, when we've received notification from GIAS that the schools are merging, we need to move all the ect_at_school_periods and mentor_at_school_periods in the predecessor school to the successor school. Because a teacher can mentor at two different schools simultaneosly, we need to identify overlapping `mentor_at_school_periods` at the two schools, and then merge them into an existing period at the successor school.  This previous [PR](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3224) introduces services to move ECTs and mentors when there are no overlaps.  

This PR introduces two services to identify and then merge the periods.  The intention is that a service to merge schools will run this process first then run the services introduced elsewhere to move mentors and ECTs with no overlaps.

I first considered leveraging some of the specific time range functionality of psql in order to identify overlapping `mentor_at_school_periods`.  After some investigation I realised there was a much more simple way to do this in Ruby.  First we order the periods at both schools by `started_on`.  Then we try to group them into overlapping sets.  So we put the first period into a group.  Then we take the next period.  If the first period is ongoing, then the next period must overlap it, and it goes in the group.  If the first period has a finished_on date, and the next periods starts after that date, then the periods don't overlap and we should start a new group.  If there's no gap the next period goes in the group.  Then we simply continue until all the periods are assigned to groups.  

It is also worth noting some features of the current validation on periods that help simplify this.  A mentor cannot have overlapping periods at the same school.  So if a mentor has two periods at the same school, the first one cannot be ongoing.  If a mentor has periods at both schools, and the earliest period is ongoing then this period will cover all the other periods.  

The tests also include scenarios where we have transitive overlaps.  Suppose we have two periods at the source school that are say 1st Jan to 31st March, then 1st May to 30th June, and a third period at the target school starting 1st March and ending 30th May.  The third period bridges the two non-overlapping periods so that we have to have a period starting 1st Jan and ending 30th June to cover it all.  If either of the later periods were ongoing then we'd need a period starting 1st Jan that is ongoing to cover all of them.

### Brief outline of expected merge schools service
```
def merge_school!
     mentor_teachers.each do |teacher|
      overlapping_mentor_at_school_periods(teacher).each do |periods|
        GIAS::Schools::MentorAtSchoolPeriods::Merge.call(periods:, predecessor_school:, successor_school:)
      end
    end

    mentors_without_overlapping_periods.each do |period|
      GIAS::Schools::MentorAtSchoolPeriods::Transfer.call(period:, predecessor_school:, successor_school:)
    end

    ects_without_mentorship_periods.each do |period|
      GIAS::Schools::ECTAtSchoolPeriods::Transfer.call(period:, predecessor_school:, successor_school:)
    end

    record_school_merged_event!
  end
end
```

### DIFF
--- a/app/services/events/record.rb
+++ b/app/services/events/record.rb
@@ -407,28 +407,30 @@ module Events
       new(event_type:, author:, heading:, teacher:, mentor_at_school_period:, school: new_school, metadata:, happened_at:).record_event!
     end
 
-    def self.record_teacher_mentor_at_school_periods_merged!(author:, teacher:, target_period:, mentor_at_school_periods:, happened_at: Time.zone.now)
+    def self.record_teacher_mentor_at_school_periods_merged!(author:, teacher:, successor_period:, mentor_at_school_periods:, happened_at: Time.zone.now)
       event_type = :teacher_mentor_at_school_periods_merged
       teacher_name = Teachers::Name.new(teacher).full_name
-      school_name = Schools::Name.new(target_period.school).name_and_urn
+      school_name = Schools::Name.new(successor_period.school).name_and_urn
 
       periods = mentor_at_school_periods.collect do |period|
         { school: period.school.name,
           started_on: period.started_on,
-          finished_on: period.finished_on }
+          finished_on: period.finished_on,
+          id: period.id,
+          urn: period.school.urn }
       end
 
-      time_period = if target_period.ongoing?
-                      "from #{target_period.started_on}"
+      time_period = if successor_period.ongoing?
+                      "from #{successor_period.started_on}"
                     else
-                      "between #{target_period.started_on} and #{target_period.finished_on}"
+                      "between #{successor_period.started_on} and #{successor_period.finished_on}"
                     end
 
       heading = "#{teacher_name}'s mentor at school periods #{time_period} were merged into a single period at #{school_name}"
 
       metadata = { periods: }
 
-      new(event_type:, author:, heading:, teacher:, mentor_at_school_period: target_period, metadata:, happened_at:).record_event!
+      new(event_type:, author:, heading:, teacher:, mentor_at_school_period: successor_period, metadata:, happened_at:).record_event!
     end
 
     def self.record_teacher_starts_training_period_event!(author:, training_period:, ect_at_school_period:, mentor_at_school_period:, teacher:, school:, happened_at:)
--- a/app/services/events/record.rb
+++ b/app/services/events/record.rb
@@ -407,6 +407,30 @@ module Events
       new(event_type:, author:, heading:, teacher:, mentor_at_school_period:, school: new_school, metadata:, happened_at:).record_event!
     end
 
+    def self.record_teacher_mentor_at_school_periods_merged!(author:, teacher:, target_period:, mentor_at_school_periods:, happened_at: Time.zone.now)
+      event_type = :teacher_mentor_at_school_periods_merged
+      teacher_name = Teachers::Name.new(teacher).full_name
+      school_name = Schools::Name.new(target_period.school).name_and_urn
+
+      periods = mentor_at_school_periods.collect do |period|
+        { school: period.school.name,
+          started_on: period.started_on,
+          finished_on: period.finished_on }
+      end
+
+      time_period = if target_period.ongoing?
+                      "from #{target_period.started_on}"
+                    else
+                      "between #{target_period.started_on} and #{target_period.finished_on}"
+                    end
+
+      heading = "#{teacher_name}'s mentor at school periods #{time_period} were merged into a single period at #{school_name}"
+
+      metadata = { periods: }
+
+      new(event_type:, author:, heading:, teacher:, mentor_at_school_period: target_period, metadata:, happened_at:).record_event!
+    end
+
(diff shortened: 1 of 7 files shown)


app/services/gias/schools/mentor_at_school_periods/merge.rb
```diff
@@ -0,0 +1,103 @@
...
+        def self.call(**args) = new(**args).call
+
+        def call
+          ActiveRecord::Base.transaction do
+            successor_period.assign_attributes(started_on:, finished_on:)
+
+            update_mentorship_periods
+            update_training_periods
+            update_events
+
```
ericaporter, line 22: Should this service call `record_teacher_mentor_at_school_periods_merged!`? I can see the recorder and event type have been added but can't see where the new merge event is created

app/services/events/record.rb
```diff
@@ -401,6 +401,30 @@ def self.record_teacher_mentor_at_school_period_moved_school!(author:, teacher:,
...
 
+    def self.record_teacher_mentor_at_school_periods_merged!(author:, teacher:, target_period:, mentor_at_school_periods:, happened_at: Time.zone.now)
+      event_type = :teacher_mentor_at_school_periods_merged
+      teacher_name = Teachers::Name.new(teacher).full_name
+      school_name = Schools::Name.new(target_period.school).name_and_urn
+
+      periods = mentor_at_school_periods.collect do |period|
+        { school: period.school.name,
+          started_on: period.started_on,
+          finished_on: period.finished_on }
```
ericaporter: Wondering if it might be useful to include the original mentor period IDs and school URNs here - only that it might be helpful once the redundant periods are deleted

app/services/gias/schools/mentor_at_school_periods/merge.rb
```diff
@@ -0,0 +1,103 @@
...
+            end
+
+            successor_period.save!
+          end
+        end
+
+      private
+
+        def successor_period
+          @successor_period ||= periods.reverse.find do |period|
```
ericaporter: It might be clearer to select the latest successor period with `max_by(&:started_on)` here instead of `reverse.find` and then no need to worry about the order of the periods

spec/services/gias/schools/mentor_at_school_periods/overlapping_spec.rb
```diff
@@ -0,0 +1,222 @@
...
+    FactoryBot.create(
+      :mentor_at_school_period,
+      teacher:,
+      school: second_school,
+      started_on: second_period_started_on,
+      finished_on: second_period_finished_on
+    )
+  end
+
+  describe "#call" do
```
ericaporter: This should be .find I think

ericaporter (commented):
A few small things to look at - overall looks solid!

app/services/events/record.rb
thenapking: Yes, that's a good idea.  I wasn't really too sure what we want to record, but that does sound sensible.

app/services/gias/schools/mentor_at_school_periods/merge.rb
thenapking: This was a rubocop thing, I couldn't find something which it didn't auto-correct back to this.  I'll try your suggestion and if not override its correction.

spec/services/events/record_spec.rb
```diff
@@ -914,6 +914,116 @@
...
+        teacher:,
+        school: second_school,
+        started_on: first_period_started_on,
+        finished_on: second_period_finished_on
+      )
+    end
+
+    let(:mentor_at_school_periods) { [first_period, second_period] }
+
+    context "when the target period is ongoing" do
```
ericaporter: You could update these two contexts to use successor period instead of target period after updating the variable names - it's still understandable though!

spec/services/gias/schools/mentor_at_school_periods/overlapping_spec.rb
```diff
@@ -0,0 +1,222 @@
...
+
+      it "returns two separate merge groups" do
+        expect(groups).to eq([
+          [first_period, second_period],
+          [third_period, fourth_period]
+        ])
+      end
+    end
+
+    context "when the earliest periods is ongoing" do
```
ericaporter: period instead of periods maybe?

spec/services/gias/schools/mentor_at_school_periods/overlapping_spec.rb
```diff
@@ -0,0 +1,222 @@
...
+
+      let(:first_period_started_on) { Date.new(2025, 1, 1) }
+      let(:first_period_finished_on) { Date.new(2025, 3, 31) }
+      let(:second_period_started_on) { Date.new(2025, 4, 1) }
+      let(:second_period_finished_on) { Date.new(2025, 6, 30) }
+
+      it { is_expected.to be_empty }
+    end
+
+    context "when there mentor periods belonging to other teachers at the schools" do
```
ericaporter: when there 'are' mentor periods...

spec/services/events/record_spec.rb
thenapking: I'd originally named them target/source but I refactored, I must have missed this when doing that!

ericaporter (approved):
Looks great! Few tiny copy updates in the specs

PR #3226, opened 2026-07-13
Spike into API versioning

ethax-ross:
> ⚠️ DO NOT MERGE

### Context

Spike into the different ways we could version various aspects of the API (controllers/serializers/queries/services/swagger docs).

- Add a v4 controller/route
- Versioned serializers (inheritance)
- Versioned serializers (views)
- Versioned query services (extract preloading)
- Versioned swagger docs
- Versioned services (inheritance)

### Guidance to review

The controllers don't need any kind of inheritance complicating things IMO; there's not much in them to repeat in different versions.

The versioned serializers using inheritance feel like they work better than using views to me; the views get a bit awkward with nested serializers as you have to define the nested field for each view (but the same, nested serializer can interpret both views): https://github.com/DFE-Digital/register-early-career-teachers-public/commit/281446fa62d9cdb7743fd0862312dedf218d4bab#diff-640b899a5b5ca8d3b1750f9718634dba7c87118bc8fdca94a9846013cd5e9d51R28-R33

For the query services I figured they are unlikely to change much between versions so went for a single query service across all versions. The difficulty then is ensuring the preloading is including the associations needed for the serializer. I shifted the `preload_associations` into the serializers, which keeps them close to the thing that causes them to change. It feels a bit off to have query logic in the serializers, though, so we may want to extract it to a separate 'preloader': https://github.com/DFE-Digital/register-early-career-teachers-public/commit/ef3e70821065e4f1ce323082892b1a5f8258db01#diff-cf117d3b2eca92ad7f8d846bdae1b15d341502ad3d5c691eb8346e7c3dee2f8cR2-R4. We also need to pass the preloading logic into the query so we run the `includes` before `find_by!` here: https://github.com/DFE-Digital/register-early-career-teachers-public/commit/64d5d5a7cc618ee0aa8833063e75b2a25528a80d#diff-42eb2092ee0c9a2ff3d4602d2de298bc7230c939daf1c6ddad7d4ce112adad9aR16-R18

The API versioning was straight forward following what we did in NPQ reg; it gives us a nice way to build on existing schemas for future versions: https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3226/changes/64d5d5a7cc618ee0aa8833063e75b2a25528a80d#diff-b833dd6ab7b49e6b179f6e1bff0da682a41011051d2410190d228fb8d2b3b631R49-R56

For the non-query services I figured we'd likely share most of the functionality between versions but some of the validation rules may change. I opted to extract a base service with commonalities and then have specific versions mirroring the controller namespaces to keep things aligned: https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3226/changes/a3b8db76f9e09b1167ed62446755c1e8e667fdab#diff-a93030f64a8d13194943a34b817bd4e5d9d0f5ad788d5834171396f632bc8062

### DIFF
--- a/app/controllers/api/v3/delivery_partners_controller.rb
+++ b/app/controllers/api/v3/delivery_partners_controller.rb
@@ -6,13 +6,18 @@ module API
           contract_period_years: extract_conditions(contract_period_years, type: :integer),
           sort:
         }
-        paginated_delivery_partners = paginate(delivery_partners_query(conditions:).delivery_partners)
+        delivery_partners = delivery_partners_query(conditions:).delivery_partners
+        paginated_delivery_partners = paginate(serializer.preload_associations(delivery_partners))
 
         render json: to_json(paginated_delivery_partners)
       end
 
       def show
-        render json: to_json(delivery_partners_query.delivery_partner_by_api_id(api_id))
+        # TODO: pass in a preloader so the query service can call it before find_by!
+        delivery_partner = delivery_partners_query.delivery_partner_by_api_id(api_id)
+        delivery_partner.strict_loading!(false)
+
+        render json: to_json(delivery_partner)
       end
 
     private
@@ -29,7 +34,7 @@ module API
 
       def serializer_options
         @serializer_options ||= {
-          lead_provider_id: current_lead_provider.id
+          lead_provider_id: current_lead_provider.id,
         }
       end
 
@@ -46,7 +51,11 @@ module API
       end
 
       def to_json(obj)
-        API::DeliveryPartnerSerializer.render(obj, root: "data", **serializer_options)
+        serializer.render(obj, root: "data", **serializer_options)
+      end
+
+      def serializer
+        API::V3::DeliveryPartnerSerializer
       end
     end
   end
--- a/app/serializers/api/delivery_partner_serializer.rb
+++ b/app/serializers/api/delivery_partner_serializer.rb
@@ -1,0 +1,35 @@
+class API::DeliveryPartnerSerializer < Blueprinter::Base
+  class AttributesSerializer < Blueprinter::Base
+    exclude :id
+
+    field :name
+    field :created_at
+    field(:api_updated_at, name: :updated_at)
+
+    view :v3 do
+      field(:cohort) do |delivery_partner, options|
+        delivery_partner
+          .active_lead_providers
+          .select { it.lead_provider_id == options[:lead_provider_id] }
+          .map { it.contract_period_year.to_s }
+      end
+    end
+
+    view :v4 do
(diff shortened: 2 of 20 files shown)


app/controllers/api/v3/delivery_partners_controller.rb
```diff
@@ -6,13 +6,18 @@ def index
           contract_period_years: extract_conditions(contract_period_years, type: :integer),
           sort:
         }
-        paginated_delivery_partners = paginate(delivery_partners_query(conditions:).delivery_partners)
+        delivery_partners = delivery_partners_query(conditions:).delivery_partners
+        paginated_delivery_partners = paginate(serializer.preload_associations(delivery_partners))
```
craigmdavidson, line 10: As discussed, consider having serializer offer it's preferred association graph, rather than have it do the loading.

app/serializers/api/v4/delivery_partner_serializer.rb
```diff
@@ -0,0 +1,21 @@
+class API::V4::DeliveryPartnerSerializer < API::V3::DeliveryPartnerSerializer
+  def self.preload_associations(results)
+    results.includes(active_lead_providers: :lead_provider)
```
craigmdavidson, line 3: As part of the consideration of having the serializers express their association loading graph consider:

```
class API::V4::DeliveryPartnerSerializer < API::V3::DeliveryPartnerSerializer
  dependencies { active_lead_providers: :lead_provider }
end
```

Where the caller can then ask the serializer for it's dependencies rather than telling it to mutate the query:

```
  delivery_partners = delivery_partners_query(conditions:).delivery_partners
  paginated_delivery_partners = paginate(delivery_partners.include(serializer.dependencies)
```

or even:

```
  paginated_delivery_partners = paginate(
    delivery_partners_query(
       conditions:, 
       dependencies: serializer.dependencies).delivery_partners))
```

app/serializers/api/delivery_partner_serializer.rb
```diff
@@ -5,18 +5,31 @@ class AttributesSerializer < Blueprinter::Base
...
+          .active_lead_providers
+          .select { it.lead_provider_id == options[:lead_provider_id] }
+          .map { it.contract_period_year.to_s }
+      end
+    end
+
+    view :v4 do
+      field :new_field do
+        "example"
+      end
```
craigmdavidson, line 21: I agree using inheritance might be better than views.

But inheritance assumes that there is no _"impedance mismatch"_ between the versions. I.e. that the entities are broadly the same shape? 

Why not just allow for explicit duplication between versions and allow them to be completely decoupled?

PR #3229, opened 2026-07-13
Remove unused ECF contract version attributes

ethax-ross:
These attributes were added to keep track of which contracts had been migrated from ECF during the RECT go-live process. We no longer maintain a contract version explicitly in RECT; instead contracts are directly linked to the statements they are applicable to.

The attributes can be safely removed from production.


cpjmcquillan (approved):

peteryates (approved):
Nice 🧹

PR #3231, opened 2026-07-13
[3404] Skip partnership selection when only one option is available

avinhurry:
### Summary

When there is only one school partnership available, it doesn't make sense to show the user a radio selection with a single option. This PR skips that page and uses the only available partnership for the check answers step.

image

### DIFF
--- a/app/wizards/admin/teachers/training_periods/change_contract_period_wizard/wizard.rb
+++ b/app/wizards/admin/teachers/training_periods/change_contract_period_wizard/wizard.rb
@@ -24,7 +24,7 @@ module Admin
 
             return steps << :check_answers if training_period_eoi_only?
             return steps << :no_partnerships unless school_partnerships.exists?
-            return steps << :check_answers if single_school_partnership
+            return steps << :check_answers if only_school_partnership
 
             steps << :select_partnership
             return steps unless selected_school_partnership_allowed?
@@ -63,7 +63,7 @@ module Admin
           end
 
           def selected_school_partnership
-            return single_school_partnership if single_school_partnership
+            return only_school_partnership if only_school_partnership
             return if store.school_partnership_id.blank?
 
             @selected_school_partnership ||= school_partnerships.find_by(id: store.school_partnership_id)
@@ -152,7 +152,7 @@ module Admin
             @current_active_period ||= ::TrainingPeriods::RelatedPeriods.new(training_period:).current_active_period
           end
 
-          def single_school_partnership
+          def only_school_partnership
             partnerships = school_partnerships.limit(2).to_a
             partnerships.first if partnerships.one?
           end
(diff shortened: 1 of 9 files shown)


app/wizards/admin/teachers/training_periods/change_contract_period_wizard/wizard.rb
```diff
@@ -117,10 +119,14 @@ def selected_lead_provider_name
             training_period.expression_of_interest_lead_provider.name
           end
 
-          def partnership_selection_required?
+          def partnership_details_required?
```
thenapking: Personally I don't feel this method name adds clarity.  I think `eoi_only?` is very clear in its relationship to `school_partnership`.  So for instance in this file you have:
```
return steps << :check_answers if eoi_only?
```
which I think is clearer than what's in `select_contract_period_step.rb`:
```
return :check_answers unless wizard.partnership_details_required?
```
That's something of a double negative, because it resolves to `unless !eoi_only?`  It's also easy to confuse with `partnership_selection_step_required?`.  Perhaps `training_period_eoi_only?` is a half way house?

app/wizards/admin/teachers/training_periods/change_contract_period_wizard/wizard.rb
```diff
@@ -150,12 +156,19 @@ def current_active_period
             @current_active_period ||= ::TrainingPeriods::RelatedPeriods.new(training_period:).current_active_period
           end
 
+          def single_school_partnership
```
thenapking: A small one but `sole_school_parntership` might be a better name?

app/wizards/admin/teachers/training_periods/change_contract_period_wizard/wizard.rb
avinhurry: Yeah, you're right. I was going to make the `eoi_only?` method public but then decided not to. I've gone with your halfway house suggestion, which is a good one. 

Thanks!

app/wizards/admin/teachers/training_periods/change_contract_period_wizard/wizard.rb
avinhurry: I've updated to `only_school_partnership` to try and improve clarity.

I avoided `sole` because it has a very specific meaning in AR which this method does not do.

spec/features/admin/teachers/change_contract_period_spec.rb
```diff
@@ -97,6 +97,12 @@ def and_there_is_a_current_active_provider_led_training_period
       lead_provider:,
       delivery_partner:
     )
+    FactoryBot.create(
```
thenapking, line 100: Just an idea, but is it worth testing that when a user selects a contract year with multiple partnerships they don't skip, and if they go back, select another year with one, they do skip straight to check answers?

thenapking (approved):
LGTM.  One small thing was that I couldn't find a teacher to test this with.  The schools with multiple partnerships in a year, I couldn't reassign teachers to that year.

spec/features/admin/teachers/change_contract_period_spec.rb
avinhurry: It's a good thought but I wouldn't cover that with a feature test. The behaviour is already covered through request and step specs.

PR #3232, opened 2026-07-13
Reuse ECT contract period across schools

ericaporter:
Relating to this [ticket](https://github.com/DFE-Digital/register-ects-project-board/issues/4131) - a further example was found

### Context
Updates ECTAtSchoolPeriods::SwitchTraining so that, when switching an ECT back to provider-led training, it looks across the teacher’s full previous provider-led training history rather than restricting the lookup to the same school

The same-school filter has been removed from `most_recent_provider_led_period`

### Guidance to review
The main scenario is an ECT who:

1. previously had provider-led training on an earlier contract period;
2. later moved to school-led training;
3. then switches back to provider-led training at a different school.

In that case, we should reuse the most recent earlier provider-led contract period rather than assigning the current one.

### DIFF
--- a/app/services/ect_at_school_periods/switch_training.rb
+++ b/app/services/ect_at_school_periods/switch_training.rb
@@ -66,7 +66,6 @@ module ECTAtSchoolPeriods
           .teacher
           .ect_training_periods
           .provider_led_training_programme
-          .where(ect_at_school_periods: { school_id: @ect_at_school_period.school_id })
           .where("training_periods.started_on < ?", date_of_transition)
           .latest_first
           .first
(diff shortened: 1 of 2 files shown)


avinhurry (approved):
Looks good!

peteryates (approved):

app/services/ect_at_school_periods/switch_training.rb
```diff
@@ -66,7 +66,6 @@ def most_recent_provider_led_period
           .teacher
           .ect_training_periods
           .provider_led_training_programme
-          .where(ect_at_school_periods: { school_id: @ect_at_school_period.school_id })
           .where("training_periods.started_on < ?", date_of_transition)
```
peteryates: Unrelated to this PR but we should be using the `#started_before` scope here I think.

ericaporter: Sure, I'll add it in!

PR #3235, opened 2026-07-14
DO NOT MERGE - RIAB api endpoint structure spike

craigmdavidson:
### Context

resolves https://github.com/DFE-Digital/register-ects-project-board/issues/4143

Details in https://github.com/DFE-Digital/register-ects-project-board/issues/4143

Spike implementation additional decisions:
* Follow `/appropriate-body/teachers/:trn/claim` pattern.
* Short-circuit appropriate body - as that is out of scope of this spike.
* Using `training_programme` rather than `induction_programme` for ease of implementation.

#### Claim
cURL:

```
 curl http://localhost:3000/api/v4/appropriate-body/teachers/4000300/claim \
    -H 'Content-Type: application/json' \
    -d '{"date_of_birth":"1990-01-01","started_on":"2025-05-02","training_programme":"school_led"}'
```

We have short-circuited authorization, so the real implementation would obviously also have an authorization token.

#### Pass

```
curl http://localhost:3000/api/v4/appropriate-body/teachers/4000300/pass \
    -H 'Content-Type: application/json' \
    -d '{"finished_on":"2026-05-02","number_of_terms":6}'
```

#### Fail

```
curl http://localhost:3000/api/v4/appropriate-body/teachers/4000300/fail \
    -H 'Content-Type: application/json' \
    -d '{"finished_on":"2026-05-02","number_of_terms":6,"fail_confirmation_sent_on":"2026-05-01"}'
```

#### Release

````
curl http://localhost:3000/api/v4/appropriate-body/teachers/4000300/release \
    -H 'Content-Type: application/json' \
    -d '{"finished_on":"2026-05-02","number_of_terms":6}'
````    

These all return a success response like:  
```
{
  "trn":"4000300",
  "full_name":"Kate Ashfield",
  "status":"ongoing",
  "started_on":"2025-05-02",
  "finished_on":null,
  "training_programme":"school_led",
  "number_of_terms":null,
  "created_at":"2026-07-14T11:28:48Z",
  "updated_at":"2026-07-14T11:28:48Z"
 }
```

And a failure responses like:

```
{"errors":[{"title":"Resource not found","detail":"Nothing could be found for the provided details"}]}
```

```
{"errors":[{"title":"Bad request","detail":"Correct json data structure required. See API docs for reference."}]}
```
`

### DIFF
--- a/app/controllers/api/v4/appropriate_body/base_controller.rb
+++ b/app/controllers/api/v4/appropriate_body/base_controller.rb
@@ -1,0 +1,32 @@
+module API
+  module V4
+    module AppropriateBody
+      # SPIKE: authentication is deferred to a separate PR. For now every request
+      # acts as the "Golden Leaf" appropriate body.
+      class BaseController < APIController
+        skip_before_action :authenticate # skip for spike
+
+        rescue_from AppropriateBodies::CloseInduction::TeacherHasNoOngoingInductionPeriod, with: :not_found_response
+
+        rescue_from ActiveModel::ValidationError do |exception|
+          render json: API::Errors::Response.from(exception.model), status: :unprocessable_content
+        end
+
+        rescue_from ActiveRecord::RecordInvalid do |exception|
+          render json: API::Errors::Response.from(exception.record), status: :unprocessable_content
+        end
+
+      private
+
+        # SPIKE: hardcoded default appropriate body
+        def appropriate_body_period
+          @appropriate_body_period ||= AppropriateBodyPeriod.find_by!(name: "Golden Leaf Teaching School Hub")
+        end
+
+        def author
+          @author ||= Events::AppropriateBodyAPIAuthor.new(appropriate_body_period:, email: "spike@example.com")
+        end
+      end
+    end
+  end
+end
--- a/app/controllers/api/v4/induction/claims_controller.rb
+++ b/app/controllers/api/v4/induction/claims_controller.rb
@@ -1,0 +1,65 @@
+module API
+  module V4
+    module Induction
+      class ClaimsController < AppropriateBody::BaseController
+        rescue_from TRS::Errors::TeacherNotFound,
+                    TRS::Errors::TeacherDeactivated,
+                    TRS::Errors::TeacherMerged,
+                    with: :not_found_response
+
+        rescue_from AppropriateBodies::ClaimAnECT::CheckECT::TeacherHasOngoingInductionPeriodWithAnotherAB,
+                    AppropriateBodies::ClaimAnECT::FindECT::TeacherHasOngoingInductionPeriodWithCurrentAB do |exception|
+          render json: { errors: API::Errors::Response.new(title: "Conflict", messages: exception.message).call }, status: :conflict
+        end
+
+        rescue_from TRS::Errors::ProhibitedFromTeaching,
+                    TRS::Errors::InductionAlreadyCompleted,
+                    TRS::Errors::QTSNotAwarded do |exception|
+          render json: { errors: API::Errors::Response.new(title: "Unprocessable", messages: exception.message).call }, status: :unprocessable_content
+        end
+
+        def create
+          return render_submission_errors unless find_ect.import_from_trs!
+
+          check_ect.begin_claim!
+
+          if register.register(**claim_params.to_h.symbolize_keys)
+            render json: API::V4::InductionPeriodSerializer.render(register.induction_period), status: :created
+          else
+            render_submission_errors
+          end
+        end
+
+      private
+
+        def submission
+          @submission ||= PendingInductionSubmission.new(
+            trn: params[:trn],
+            date_of_birth: claim_params[:date_of_birth],
+            appropriate_body_period_id: appropriate_body_period.id
+          )
+        end
+
+        def find_ect
+          AppropriateBodies::ClaimAnECT::FindECT.new(appropriate_body_period:, pending_induction_submission: submission)
+        end
+
+        def check_ect
+          AppropriateBodies::ClaimAnECT::CheckECT.new(appropriate_body_period:, pending_induction_submission: submission)
+        end
+
+        def register
+          @register ||= AppropriateBodies::ClaimAnECT::RegisterECT.new(appropriate_body_period:, pending_induction_submission: submission, author:)
+        end
+
+        def claim_params
+          params.expect(claim: %i[date_of_birth started_on training_programme])
+        end
+
+        def render_submission_errors
+          render json: API::Errors::Response.from(submission), status: :unprocessable_content
+        end
+      end
+    end
+  end
+end
--- a/app/controllers/api/v4/induction/fails_controller.rb
+++ b/app/controllers/api/v4/induction/fails_controller.rb
@@ -1,0 +1,31 @@
+module API
+  module V4
+    module Induction
+      class FailsController < AppropriateBody::BaseController
+        before_action :set_teacher
+        before_action :set_induction_period
+
+        def create
+          AppropriateBodies::RecordFail.new(teacher: @teacher, appropriate_body_period:, author:)
+            .call(**fail_params.to_h.symbolize_keys)
+
+          render json: API::V4::InductionPeriodSerializer.render(@induction_period), status: :ok
+        end
+
+      private
+
+        def set_teacher
+          @teacher = AppropriateBodies::ECTs.new(appropriate_body_period).current.find_by!(trn: params[:trn])
+        end
+
+        def set_induction_period
+          @induction_period = @teacher.ongoing_induction_period
+        end
+
+        def fail_params
+          params.expect(fail: %i[finished_on number_of_terms fail_confirmation_sent_on])
+        end
+      end
+    end
+  end
+end
--- a/app/serializers/api/v4/induction_period_serializer.rb
+++ b/app/serializers/api/v4/induction_period_serializer.rb
@@ -1,0 +1,23 @@
+# Shared response shape for the v4 claim / close endpoints — the induction
+# period in its current state.
+class API::V4::InductionPeriodSerializer < Blueprinter::Base
+  field(:trn) { |induction_period| induction_period.teacher.trn }
+  field(:full_name) { |induction_period| ::Teachers::Name.new(induction_period.teacher).full_name }
+
+  # ongoing | pass | fail | release (released = finished with no pass/fail outcome)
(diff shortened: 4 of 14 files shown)

### CODE REVIEW

File: app/controllers/api/v4/appropriate_body/base_controller.rb [hunk 1]
```diff
@@ -0,0 +1,56 @@
+module API
+  module V4
+    module AppropriateBody
+      # SPIKE: authentication is deferred to a separate PR. For now every request
+      # acts as the "Golden Leaf" appropriate body.
+      class BaseController < APIController
```
craigmdavidson, line 6: > [!NOTE]
> I am not a fan of this kind of "hiding your crap in the loft", 
but this is a spike rather than a production ready solution, 
and it is the existing pattern in the APIController.

app/controllers/api/v4/induction/fails_controller.rb
```diff
@@ -0,0 +1,31 @@
+module API
+  module V4
+    module Induction
+      class FailsController < AppropriateBody::BaseController
```
craigmdavidson, line 4: :+1: The "closing" controllers read as relatively standard Rails. 

The separation of the closing types by Controller keeps the Controller simple, and makes it harder for pratfalls in passing the wrong closing type.

app/controllers/api/v4/induction/fails_controller.rb
```diff
@@ -0,0 +1,31 @@
...
+        before_action :set_teacher
+        before_action :set_induction_period
```
craigmdavidson, line 6: > [!NOTE]
> When it comes to doing this for real, we could create a concern **TeacherScoped** to remove duplication, or even an intermediate superclass (but that smells a bit of putting your junk in the loft).

app/serializers/api/v4/induction_period_serializer.rb
```diff
@@ -0,0 +1,23 @@
+# Shared response shape for the v4 claim / close endpoints — the induction
+# period in its current state.
+class API::V4::InductionPeriodSerializer < Blueprinter::Base
```
craigmdavidson, line 3: > [!NOTE]
> Since this will be called via a 3rd party by an AppropriateBody, there is no point in returning the AppropriateBody in the response. 

**So this follows the principal minimising the data shared between systems to only what is necessary.** 

> [!NOTE]
> We purposefully do not share the id of the created InductionPeriod, to that end. We are explicitly treating the InductionPeriod as internal, and exposing concepts more appropriate to the AppropriateBody like _claim_, _release_, _passing_, _failing_ an ECT.  **This is in line with the UI treatment.**

config/routes/api.rb
```diff
@@ -30,4 +30,13 @@
     resources :schools, only: %i[index show], param: :api_id
     resources :unfunded_mentors, only: %i[index show], path: "unfunded-mentors", param: :api_id
   end
+
+  namespace :v4 do
+    scope path: "appropriate-body/teachers/:trn", module: :induction do
```
craigmdavidson, line 35: > [!NOTE]
> This spike tries to broadly follow the UI URL rules. 
> 
> If we wanted to change from `appropriate-body/teachers` to `ect-inductions`, 
that is just a matter of changing the path parameter.

app/controllers/api/v4/appropriate_body/base_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+                    TRS::Errors::QTSNotAwarded,
+                    with: :unprocessable_content_response
+
+        rescue_from ActiveModel::ValidationError, with: :model_validation_error_response
+        rescue_from ActiveRecord::RecordInvalid, with: :record_invalid_response
+
+      private
+
+        # SPIKE: hardcoded default appropriate body
+        def appropriate_body_period
```
craigmdavidson, line 22: > [!WARNING]
> This stuff is a short-circuit for Authorisation/Authentication which is out of scope of this spike. In reality we might expect to have `Current.appropriate_body` or `current_appropriate_body` or something like that.

app/services/events/appropriate_body_api_author.rb
```diff
@@ -0,0 +1,21 @@
+class Events::AppropriateBodyAPIAuthor
```
craigmdavidson, line 1: > [!WARNING]
> This is just a modified copy of the **LeadProviderAPIAuthor**. It cannot have the simplified same structure as **LeadProviderAPIAuthor** without a database migration because we store enums (`event_author_types`) in the database??! 

**So for just now we masquerade as a real person, we would need to modify the enum if doing it for real.**

app/controllers/api/v4/induction/claims_controller.rb
```diff
@@ -0,0 +1,65 @@
...
+      class ClaimsController < AppropriateBody::BaseController
+        rescue_from TRS::Errors::TeacherNotFound,
+                    TRS::Errors::TeacherDeactivated,
+                    TRS::Errors::TeacherMerged,
+                    with: :not_found_response
+
+        rescue_from AppropriateBodies::ClaimAnECT::CheckECT::TeacherHasOngoingInductionPeriodWithAnotherAB,
+                    AppropriateBodies::ClaimAnECT::FindECT::TeacherHasOngoingInductionPeriodWithCurrentAB do |exception|
+          render json: { errors: API::Errors::Response.new(title: "Conflict", messages: exception.message).call }, status: :conflict
+        end
```
craigmdavidson, line 13: These rescues are only needed in this controller, I think it makes it more obvious to have them here rather than hide them in the "loft". 

If I was doing this for real I might prefer to have use concerns for this:
- **TrsErrorRescueable**; and 
- **ClaimEctConflictRescueable**?

app/controllers/api/v4/induction/claims_controller.rb
```diff
@@ -0,0 +1,65 @@
...
+          return render_submission_errors unless find_ect.import_from_trs!
+
+          check_ect.begin_claim!
+
+          if register.register(**claim_params.to_h.symbolize_keys)
+            render json: API::V4::InductionPeriodSerializer.render(register.induction_period), status: :created
+          else
+            render_submission_errors
+          end
+        end
```
craigmdavidson, line 31: The service objects used here are not ideally suited to the API, the are structured around the UI Wizard flow (that is in and of itself a smell since they are supposed to be general purpose).

When doing this for real, it might be a good idea to restructure some of these to maybe return something useful? 

For now, to make the action clearer, I have extracted these to private helper methods, but that's masking a symptom rather than the underlying problem.

app/controllers/api/v4/induction/claims_controller.rb
```diff
@@ -0,0 +1,65 @@
...
+
+        def create
+          return render_submission_errors unless find_ect.import_from_trs!
+
+          check_ect.begin_claim!
+
+          if register.register(**claim_params.to_h.symbolize_keys)
+            render json: API::V4::InductionPeriodSerializer.render(register.induction_period), status: :created
+          else
```
craigmdavidson, line 28: > [!CAUTION]
> The problem with our service objects is they are inconsistent. They both raise errors and use return codes. This inconsistency makes them hard to work with, and adds additional conditional logic to the client code.

app/controllers/api/v4/induction/claims_controller.rb
craigmdavidson: We could use another Service Object, 
which calls other these Service Objects to "sweep the crap under the carpet" - COBOL style 😵

app/controllers/api/v4/induction/claims_controller.rb
peteryates: I agree totally on this, I'd sacrifice a bit of repetition for being able to easily see what's going on where.

app/controllers/api/v4/appropriate_body/base_controller.rb
ethax-ross: Are you referencing the concerns? what's the alternative, including them all explicitly in every controller? most are used by every controller, but there are a few that are more specific we could maybe push down 🤔

app/serializers/api/v4/induction_period_serializer.rb
ethax-ross: Will we need to backtrack if we do this and later expose teacher/induction information with a GET endpoint/webhook? or do we only ever expose their current induction state?

app/controllers/api/v4/appropriate_body/base_controller.rb
craigmdavidson: I am referencing having a superclass at all - it is only there because this is a spike, and that is out of scope of the spike itself. 

I personally think all API controllers should extend a simple ApplicationApiController that at most handles authentication which is a fundamental cross cutting concern, and that's about it. 

The "hiding your crap in the loft" covers things like rescuing from the myriad of different errors. The individual controllers express these these explicitly, this could either be directly (as per the **ClaimsController**) or via including declarative concerns, e.g. **TrsErrorRescueable**. 

This means we can read the single class and understand clearly what it does, we should never need to look in the attic. Everything in the "loft" is essentially a surprise.

If you look at what we have in the loft of our current APIController, ... https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/controllers/api_controller.rb#L2-L35 :scream: 😆

app/serializers/api/v4/induction_period_serializer.rb
craigmdavidson: > Will we need to backtrack if we do this and later expose teacher/induction information with a GET endpoint/webhook? or do we only ever expose their current induction state?

This is the key question. I spent a good bit of time looking at this specific question, going through every path I could in the UI, and as far as I can see, **we will only ever want to deal with the current induction state**. 

**So I think the answer is no.** The InductionPeriods from an API point of view are in internal implementation concern. 

See the sanity check in the Issue comments:

```
  GET ect-inductions? => Search inductions 
  GET ect-inductions/:trn => Get all inductions for a given TRN?
  GET ect-inductions/:trn/current => Get the current induction
  PATCH ect-inductions/:trn/current => Change the current induction; start or induction_programme; idempotent changes.
  POST ect-inductions/:trn/claim => Claim an ECT for induction
  POST ect-inductions/:trn/extensions => Add an extension for an ECT induction
  POST ect-inductions/:trn/close-current => Close an ECT for induction
  or if we prefer
  POST ect-inductions/:trn/pass
  POST ect-inductions/:trn/release
  POST ect-inductions/:trn/fail
```

PR #3236, opened 2026-07-14
Fix bug in `TrainingProgramme` service

ethax-ross:
### Context 

The `TrainingProgramme` service checks `TrainingPeriod` records against a given `ContractPeriod` for a `School` in order to determine the overall `induction_programme_choice` of the `School`.

It prioritises `provider-led` over `school-led`; if neither are present it will be `not-yet-known`.

Previously we were not accounting for multiple `TrainingPeriod` records with different `training_pgoramme` values against the same `AtSchoolPeriod`. For example:

```
TrainingPeriod
  .where(ect_at_school_period_id:)
  .order(training_programme: :asc)
  .pluck(:ect_at_school_period_id, :training_programme)

[[39, "provider_led"], [39, "school_led"]]
```

We would call `to_h` on the above result to get a mapping of `AtSchoolPeriod#id` -> `programme_choice`; with the way `to_h` works if there are multiple values with the same key we are left with only the last one:

```
[[39, "provider_led"], [39, "school_led"]].to_h

{39 => "school_led"}
```

In this scenario the `School` would incorrectly be identified as `school-led`. Instead, we should take into account both of the programme choices, prioritising `provider-led`.

This bug manifests as schools going from `provider-led` to `school-led` in the metadata refresh.

### Changes 

- Fix bug in `TrainingProgramme` service

### Guidance to review

After testing in the migration environment there are **139 schools effected**, with 174 incorrect induction programme types against contract periods:

```
2021 -> 14
2022 -> 47
2023 -> 48
2024 -> 42
2025 -> 18
2026 -> 5
```

### DIFF
--- a/app/services/schools/training_programme.rb
+++ b/app/services/schools/training_programme.rb
@@ -50,13 +50,18 @@ module Schools
     end
 
     def ect_training_programme(contract_period_year:)
-      ect_at_school_period_id = (
+      ect_at_school_period_ids = (
         (ect_expressions_of_interest_ids_by_contract_period_year[contract_period_year] || []) +
         (ect_at_school_period_ids_by_contract_period_year[contract_period_year] || []) +
         (school_led_ect_at_school_period_ids_by_contract_period_year[contract_period_year] || [])
       ).compact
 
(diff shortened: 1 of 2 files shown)


craigmdavidson (approved):
LGTM

PR #3237, opened 2026-07-14
Add support task for ECT withdrawal corrections

ericaporter:
Relates to this support [ticket](https://github.com/DFE-Digital/register-ects-project-board/issues/4183)

### Context

An ECT left their school in November 2023, but their original training period remained open beyond the correct leaving date.

An erroneous later training period was subsequently created and the provider withdrawal was recorded against that period. This caused the ECT’s training history to be displayed incorrectly.

The withdrawal details need to be retained on the original training period, the linked records need to use the correct leaving date, and the erroneous later period needs to be removed.

Adds a support task for correcting an ECT history where withdrawal details have been recorded against an erroneous later training period.

The task accepts:

- ECT-at-school period ID
- original training period ID
- erroneous withdrawn training period ID
- corrected end date

It displays the records and proposed changes, then requires the teacher ID to be entered before continuing.

The correction service:

- moves the withdrawal details onto the original training period
- applies the corrected end date to the ECT-at-school period
- updates the linked training and mentorship periods
- removes linked periods that start on or after the corrected end date
- suppresses leaving and period-finished events

The service validates that the supplied records belong together, that their dates support the correction, and that the relevant training periods have no declarations or events.

### Guidance to review

Run the service spec:

bundle exec rspec spec/services/teachers/ect_withdrawal_history_correction_spec.rb


lib/support/ect_withdrawal_history_correction.rb
```diff
@@ -0,0 +1,131 @@
+# Corrects an ECT history where withdrawal details were recorded against a
+# later training period instead of the relevant earlier training period.
+#
+# It moves the withdrawal details to the original period, applies the corrected
+# leaving date across linked records, and removes later periods that should no
+# longer exist. The supplied periods must have no declarations or events.
+module Teachers
+  class ECTWithdrawalHistoryCorrection
```
thenapking, line 9: This is a really very specific service, which its difficult to imagine being used for anything other than the rake task.  Wouldn't it be better if it lived in the `/lib` folder?

lib/support/ect_withdrawal_history_correction.rb
```diff
@@ -0,0 +1,131 @@
...
+    def validate_period_dates!
+      if ect_at_school_period.finished_on.present?
+        raise "ECT-at-school period is already finished"
+      end
+
+      if corrected_end_date < ect_at_school_period.started_on
+        raise "Corrected end date is before the ECT-at-school period start date"
+      end
+
+      if original_training_period.started_on > corrected_end_date
```
thenapking, line 88: Should this be >= ? Couldn't we end up with a zero length period?

lib/support/ect_withdrawal_history_correction.rb
```diff
@@ -0,0 +1,131 @@
...
+        raise "Original training period starts after the corrected end date"
+      end
+
+      if original_training_period.finished_on.present? &&
+          original_training_period.finished_on < corrected_end_date
+        raise "Original training period finishes before the corrected end date"
+      end
+
+      if erroneous_withdrawn_training_period.started_on < corrected_end_date
```
thenapking, line 99: Again maybe <= ?

app/services/teachers/ect_withdrawal_history_correction.rb
```diff
@@ -0,0 +1,131 @@
...
+      end
+    end
+
+    def validate_dependencies!
+      if original_training_period.declarations.exists?
+        raise "Original training period has declarations"
+      end
+
+      if original_training_period.events.exists?
+        raise "Original training period has events"
```
thenapking: Is it really a problem the original `training_period` has events?  Couldn't we just move the events from the erroneious period to the original? Or even delete them if the period really is erroneous?  Or is there a particular event (relating to declarations?) that we can't move/delete?

lib/tasks/ect_withdrawal_history_correction.rake
```diff
@@ -0,0 +1,131 @@
...
+    puts "  - Apply them to the original training period"
+    puts "  - Finish linked records on #{corrected_end_date}"
+    puts "  - Remove linked training and mentorship periods starting on or after that date"
+    puts "  - Do not record leaving or period-finished events"
+    puts
+
+    print "Enter the teacher ID to confirm: "
+
+    unless $stdin.gets&.strip == teacher.id.to_s
+      abort "Correction cancelled."
```
thenapking, line 100: This is a good guard!

app/services/teachers/ect_withdrawal_history_correction.rb
```diff
@@ -0,0 +1,131 @@
...
+      validate_withdrawal_details!
+      validate_dependencies!
+    end
+
+    def validate_training_periods!
+      if original_training_period == erroneous_withdrawn_training_period
+        raise "Original and erroneous training periods must be different"
+      end
+
+      unless original_training_period.ect_at_school_period ==
```
thenapking: If we derived `ect_at_school_period` from the `original_training_period` then we only need to check that `original_training_period.ect_at_school_period == erroneous_withdrawn_training_period.ect_at_school_period`.

thenapking (approved):
This looks good.  I think there's a couple of validations to double check, and we could possibly get rid of the `ect_at_school_period` passed to the service.  I also notice there are no seeds to test this on, but I am assuming we'll merge and test on the specific case on migration.

lib/support/ect_withdrawal_history_correction.rb
cpjmcquillan: Ranges with inclusive beginnings and ends mean a period that starts and finishes on the same day has a duration of 1 day, not zero

lib/support/ect_withdrawal_history_correction.rb
ericaporter: That makes sense - I’ll move it to lib/support, rename it to `Support::ECTWithdrawalHistoryCorrection`

lib/support/ect_withdrawal_history_correction.rb
ericaporter: I think `>` is right here since we have inclusive start and end dates

app/services/teachers/ect_withdrawal_history_correction.rb
ericaporter: Yeah this is completely fair - I’ve removed that check.

I’ve kept the check on the erroneous period because it is deleted, and I don’t want to move or delete its events without reviewing them first. There are no events in this case anyway.

PR #3171, opened 2026-06-29
[3980] Update school reported AB from claiming AB

rwrrll:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/3980

### Changes proposed in this pull request

Update the school-reported AB when the ECT is claimed by AB, as described in the ticket

### Guidance to review

Does it make sense to rename or have a new attribute for this? Would we want to retain the school-reported AB for any reason, if it's been claimed meaning that we know it's wrong?

PR #3242, opened 2026-07-15
Ensure the relevant AB is displayed on the ECT card

cpjmcquillan:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/4184

In #3171 we reworked the `Schools::ECTInductionDetailsComponent` so it displayed the details of the appropriate body that claimed an ECT if it was relevant.

We didn't do this work for the `Schools::ECTs::ListingCardComponent`, though.

This extracts a small service for determining the relevant appropriate body for a given ECT, then uses that service to display the relevant name in both contexts.

### Guidance to review

- [ ] Register an ECT and choose an appropriate body
- [ ] See the school reported appropriate body name on the teachers index page
- [ ] See the school reported appropriate body name on the teachers show page
- [ ] Claim an ECT at a different appropriate body
- [ ] See the claimed by appropriate body name on the teachers index page
- [ ] See the claimed by appropriate body name on the teachers show page (✨ the fix)

### DIFF
--- a/app/controllers/schools/ects_controller.rb
+++ b/app/controllers/schools/ects_controller.rb
@@ -13,6 +13,7 @@ module Schools
         )
         .search
         .includes(
+          ongoing_induction_period: :appropriate_body_period,
           ect_at_school_periods: %i[latest_training_period mentorship_periods],
           current_or_next_ect_at_school_period: [
             { latest_training_period: %i[lead_provider delivery_partner] },
--- a/app/services/ect_at_school_periods/appropriate_body.rb
+++ b/app/services/ect_at_school_periods/appropriate_body.rb
@@ -6,7 +6,7 @@ module ECTAtSchoolPeriods
 
     delegate :name, to: :appropriate_body
 
-    def from_induction? = induction_periods_reported_during_placement.any?
+    def from_induction? = claimed_by_appropriate_body.present?
 
     def from_school?
       return false if from_induction?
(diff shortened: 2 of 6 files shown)


app/services/ect_at_school_periods/appropriate_body.rb
```diff
@@ -0,0 +1,47 @@
...
+    # a previous school, are excluded, so the school keeps seeing its reported
+    # appropriate body as provisional until it's confirmed for this placement.
+    def induction_periods_reported_during_placement
+      @ect_at_school_period
+        .teacher
+        .induction_periods
+        .ongoing
+        .where(started_on: @ect_at_school_period.started_on..@ect_at_school_period.finished_on)
+        .latest_first
+    end
```
avinhurry: The method now runs a `induction_periods` query for every ECT card that uses the service. On the school ECT list we don't preload what this needs so it looks like it introduce an n+1.

Can we either preload the required induction data or avoid doing a per card query here? We just fixed most of the n+1's on this page a few days ago in https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3222, so it'd be good not to reintroduce them.

cpjmcquillan: Nice one 🙌🏼 I was going to leave a comment about this, but I hadn't gotten round to it.

I think this is a good point to start thinking about whether we ought to add some tests to some of these pages to avoid any regressions. `prosopite` does this quite well, but I'm keen to hear what people think.

Either way, we should definitely avoid re-introducing N+1s!

avinhurry: IMO, I'd be a bit wary of adding regression tests by default to these pages because we already have prosopite, which flags N+1s clearly, and I'd expect catching most of these to be part of the implementation and review process.

That said, since we removed the N+1s from this page a few days ago and have already reintroduced them, I wouldn't be opposed to adding a targeted regression test here to see how it goes. We can then decide whether it's worth expanding to other pages.

cpjmcquillan: I haven't added any regression tests yet, but I'm adding a ticket to the backlog for us to discuss/spike out an approach.

I've fixed the N+1 for now, though.

app/controllers/schools/ects_controller.rb
```diff
@@ -13,6 +13,7 @@ def index
         )
         .search
         .includes(
+          :ongoing_induction_period,
```
avinhurry: This improves things but doesn't quite finish the chain. 

It preloads `ongoing_induction_period`, but for claimed ECTs the service still calls `ongoing_induction_period.appropriate_body_period` so we should include that nested  `belongs_to` in the preload.

avinhurry (approved):
Looks good! Thank you for addressing those 🙇🏽

PR #3244, opened 2026-07-15
[4142] Service to merge GIAS::School

thenapking:
### Context

The final orchestrator service which merges two school records, when their `GIAS::School` records are merged.

This follows what has turned out to be a straightforward process.  Merge any `mentor_at_school_periods` at the two schools which overlap, and move their mentees and other associated records.  Then for the remaining `mentor_at_school_periods` move them and move their mentees and other associated records.  Finally move any remaining ect_at_school_periods.  

As with the three other services, this service only merges schools in the correct condition to be merged.  For purposes of override by a developer there is a private `merge_school!` method which can be run whilst we are testing the service.

### Guidance to review

For product review there are two sets of school merges we can look at.  Each school has a variety of mentors and ECTs which cover all scenarios.  Before merging we should observe the teachers at each school and what we expect.  The mentors are named according to the arrangement of their periods at both schools, eg "Overlapping Mentor", "Separate Mentor".  In order to product review, I will take the the reviewer through the scenarios, and then we can run the merge and observe the output to see if it worked.

```
monsters_college = GIAS::School.find(9123500)
GIAS::Schools::Merge.new(monsters_college).merge!
```

### DIFF
--- a/app/models/gias/school.rb
+++ b/app/models/gias/school.rb
@@ -92,7 +92,18 @@ class GIAS::School < ApplicationRecord
       successor.open_status? &&
       successor.opened_on_or_before_today? &&
       successor.school_not_yet_opened? &&
-      school_replaced?
+      school_being_replaced?
+  end
+
+  def can_be_merged?
+    closed_status? &&
+      closed_on_or_before_today? &&
+      !school_merger_recorded? &&
+      successors.one? &&
+      successor.open_status? &&
+      successor.opened_on_or_before_today? &&
+      successor.school.present? &&
+      school_being_merged?
   end
 
   def school_not_yet_opened?
@@ -117,7 +128,15 @@ private
     Event.where(school:, event_type: :school_closed).exists?
   end
 
-  def school_replaced?
+  def school_merger_recorded?
+    Event.where(school:, event_type: :school_merged).exists?
+  end
+
+  def school_being_replaced?
     successor_links.where(link_type: GIAS::SchoolLink::SUCESSOR).exists?
+  end
+
+  def school_being_merged?
+    successor_links.where(link_type: GIAS::SchoolLink::SUCCESSOR_MERGED).exists?
   end
 end
--- a/app/models/gias/school.rb
+++ b/app/models/gias/school.rb
@@ -98,7 +98,7 @@ class GIAS::School < ApplicationRecord
   def can_be_merged?
     closed_status? &&
       closed_on_or_before_today? &&
-      !school_merger_recorded? &&
+      no_school_merged_event_recorded? &&
       successors.one? &&
       successor.open_status? &&
       successor.opened_on_or_before_today? &&
@@ -128,8 +128,8 @@ private
     Event.where(school:, event_type: :school_closed).exists?
   end
 
-  def school_merger_recorded?
-    Event.where(school:, event_type: :school_merged).exists?
+  def no_school_merged_event_recorded?
+    Event.where(school:, event_type: :school_merged).none?
   end
 
   def school_being_replaced?
--- a/app/services/gias/schools/ect_at_school_periods/transfer.rb
+++ b/app/services/gias/schools/ect_at_school_periods/transfer.rb
@@ -1,73 +1,73 @@
-module GIAS
-  module Schools
-    module ECTAtSchoolPeriods
-      class Transfer
-        def self.call(...) = new(...).call
+module GIAS::Schools
+  module ECTAtSchoolPeriods
+    class Transfer
+      def self.call(...) = new(...).call
 
-        def initialize(ect_at_school_period:, predecessor_school:, successor_school:)
-          @ect_at_school_period = ect_at_school_period
-          @successor_school = successor_school
-          @predecessor_school = predecessor_school
-        end
-
-        def call
-          return unless predecessor_school == ect_at_school_period.school
-
-          ActiveRecord::Base.transaction do
-            update_ect_at_school_period!
-            update_training_periods!
-            update_events!
-
-            record_event!
-          end
-        end
-
-      private
-
-        attr_reader :ect_at_school_period, :predecessor_school, :successor_school
-
-        def update_ect_at_school_period!
-          ect_at_school_period.update!(school: successor_school)
-        end
-
-        def update_training_periods!
-          training_periods.where.associated(:school_partnership).each do |training_period|
-            new_partnership = successor_partnership(training_period.school_partnership)
-            training_period.update!(school_partnership: new_partnership)
-          end
-        end
-
-        def update_events!
-          events.where.associated(:school_partnership).each do |event|
-            new_partnership = successor_partnership(event.school_partnership)
-            event.update!(school_partnership: new_partnership)
-          end
-
-          events.where.associated(:school).each do |event|
-            event.update!(school: successor_school)
-          end
-        end
-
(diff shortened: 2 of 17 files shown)


db/seeds/gias_data.rb
```diff
@@ -14,14 +14,171 @@ def describe_mentorship_period(mp)
 
 def teacher(...) = TeacherHistories::TeacherBuilder.teacher(...)
 
-def populate_school(gias_school, counter = 0)
+SCENARIOS = [
```
thenapking, line 17: This is only for testing this PR.  Once this is product reviewed, I will stop these being loaded by default.  Once we've tested the 4 services in migration and found them working, I will remove this seed file entirely.

app/models/gias/school.rb
```diff
@@ -117,7 +128,15 @@ def school_closure_recorded?
     Event.where(school:, event_type: :school_closed).exists?
   end
 
-  def school_replaced?
+  def school_merger_recorded?
```
cpjmcquillan:
```suggestion
  def school_merged_event_recorded?
```

app/services/gias/schools/merge.rb
```diff
@@ -0,0 +1,71 @@
+module GIAS::Schools
+  class Merge
+    attr_reader :gias_school
```
cpjmcquillan: Do we need to expose this method, or could it be private?

spec/models/gias/school_spec.rb
cpjmcquillan, line 1: I'm not completely sure what the rationale is here for testing some of the scenarios, and not others. Or why some contexts are nested, or not 🤔

app/services/gias/schools/merge.rb
```diff
@@ -0,0 +1,71 @@
...
+        happened_at: closed_on,
+        author:
+      )
+    end
+
+    def author
+      Events::SystemAuthor.new
+    end
+
+    def schools = [school, successor_school]
```
cpjmcquillan: Is this used anywhere?

spec/services/gias/schools/merge_spec.rb
cpjmcquillan, line 1: The tests and nested contexts here feel a little complex, especially if we're unit testing each of the underlying services and methods extensively.

If we're going to stub things, maybe we could lean into it further and simplify this test file?

app/models/gias/school.rb
```diff
@@ -117,7 +128,15 @@ def school_closure_recorded?
...
+    Event.where(school:, event_type: :school_merged).exists?
```
ericaporter: The merge service records the school merged event against a successor school but is this querying the predecessor school? If so will this remain false after a successful merge?

app/services/gias/schools/merge.rb
```diff
@@ -0,0 +1,71 @@
...
+      end
+    end
+
+    def overlapping_mentor_at_school_periods(teacher)
+      GIAS::Schools::MentorAtSchoolPeriods::Overlapping.find(teacher:, schools:)
+    end
+
+    def record_school_merged_event!
+      Events::Record.record_school_merged_event!(
+        school: successor_school,
```
ericaporter: I think this differs from `school_merger_recorded?` which looks like it uses the predecessor school here

app/services/gias/schools/merge.rb
ericaporter: `GIAS::Schools::MentorAtSchoolPeriods::Overlapping.find(teacher:, schools:)` I think?

spec/models/gias/school_spec.rb
ericaporter: Maybe starting with a fully valid mergeable example then contexts to change each condition would help?

spec/services/gias/schools/merge_spec.rb
ericaporter: Could we stub the overlap, merge and transfer services here and just test that they’re called with the right records? The cannot be merged context probably doesn’t need to set up all the periods and partnerships, as it should return straight away. We could then keep one full happy-path test to check it all works together.

app/services/gias/schools/ect_at_school_periods/transfer.rb
```diff
@@ -1,73 +1,73 @@
...
+          new_school: successor_school,
+          happened_at: predecessor_gias_school.closed_on,
+          author:
+        )
       end
+
+      def predecessor_gias_school = predecessor_school.gias_school
+      def old_school_name_and_urn = ::Schools::Name.new(predecessor_school).name_and_urn
+
+      delegate :mentorship_periods, :training_periods, :events, :teacher, to: :ect_at_school_period
```
ericaporter, line 68: Are we using :mentorship_periods anywhere? We can possibly remove it

app/services/gias/schools/merge.rb
```diff
@@ -0,0 +1,68 @@
...
+    def overlapping_mentor_at_school_periods(teacher)
+      GIAS::Schools::MentorAtSchoolPeriods::Overlapping.find(
+        teacher:,
+        schools: [predecessor_school, successor_school]
+      )
+    end
+
+    def record_school_merged_event!
+      Events::Record.record_school_merged_event!(
+        school: predecessor_school,
```
ericaporter, line 52: Cool this has now been updated 👍

app/models/gias/school.rb
```diff
@@ -92,7 +92,18 @@ def can_be_replaced?
...
       successor.opened_on_or_before_today? &&
       successor.school_not_yet_opened? &&
-      school_replaced?
+      school_being_replaced?
+  end
+
+  def can_be_merged?
+    closed_status? &&
+      closed_on_or_before_today? &&
+      no_school_merged_event_recorded? &&
```
ericaporter, line 101: Is relying on an asynchronous event to tell us whether or not to merge a little bit risky?

spec/models/gias/school_spec.rb
```diff
@@ -375,6 +375,92 @@
       end
     end
 
+    describe "#can_be_merged?" do
+      subject { gias_school.can_be_merged? }
+
+      context "when the school is closed" do
+        let(:gias_school) { FactoryBot.create(:gias_school, status: :closed, closed_on:) }
```
ericaporter, line 382: I think gias_school needs a :with_school otherwise school will be nil?

ericaporter (approved):
@cpjmcquillan Left a few comments; a few tidy-ups and a qu about relying on an async event for a check but it's looking good otherwise!

app/models/gias/school.rb
cpjmcquillan: Yeah I guess it is a little.

I wonder whether we ought to have a `merged_at` timestamp on the `GIAS::School` and rely on that instead? 🤔

spec/models/gias/school_spec.rb
cpjmcquillan: Yeah, good spot.

Going to rework these unit tests so they are more robust 👍🏼

app/models/gias/school.rb
cpjmcquillan: We're doing the same thing for closing schools, so I think I'll stick with this in the interests of getting things moving. I'll add a tech debt issue to circle back and rework both approaches 👍🏼

spec/models/gias/school_spec.rb
cpjmcquillan: Was getting so bogged down with these test refactors that I've parked them for now

PR #3246, opened 2026-07-15
[4226] Use latest contract periods for change lead provider options

avinhurry:
## Summary

The mentor and ECT "change lead provider" flows were using the participant start date to decide which lead providers to show. That meant participants who started in an older contract period but were later moved to a newer one could still see the old options.

Switch both flows to use the contract period from the participant's latest training period instead, so the options shown line up with their current contract period.


cpjmcquillan (approved):
This makes sense to me.

Outside the scope of this change, but is there any reason why `Schools::ECTs::ChangeLeadProviderWizard::EditStep` shouldn't follow the same approach? I don't really understand why the `Schedules::Reuse` stuff is in there, but I'm probably missing something!

PR #3248, opened 2026-07-16
Use standard ApplicationService pattern for simpler calls

craigmdavidson:
### Context

This project makes use of Service Objects extensively.  Our Service Objects are not sure of themselves. While they only a single public method, and are verb named (e.g. Create, Find, etc), they have custom execute names. This attempt to provide additional context actually ends up obfuscating and confusing the situation since:
* The Service Object name is already a verb, so we violate DRY.
* It makes it less obvious that this thing is actually a service object rather than a "real" domain centred PORO.

In Ruby, objects that encapsulate executable logic like **Proc**, **Method**, and **Lambda** are invoked using `.call`. By naming our service method `call` we follow Ruby's native design language.

This is a widely accepted standard, that improves predictability.

#### Why not dry-rb?

I don't think dry-rb is a great fit for a Rails application. It's built around a fundamentally different philosophy, and mixing that with Rails usually means carrying two sets of patterns and conventions. 

In our case, we can get the fundamental benefits with a simple **ApplicationService** superclass and one single-line method, without introducing another abstraction or dependency.

```
  def self.call(**kwargs) = new(**kwargs).call
```

* Make service objects inherit from ApplicationService. This provides a class `call` for convenience method.
* Call service objects using their class method using the explicit dot-syntax. 

I.e.

Instead of:
```
# This is bad 
seed_from_previous = ActiveLeadProviders::SeedFromPrevious.new(active_lead_provider:)
seed_from_previous.seed!

# This is still bad
ActiveLeadProviders::SeedFromPrevious.new(active_lead_provider:).seed!

# This is still not good
ActiveLeadProviders::SeedFromPrevious.new(active_lead_provider:).call
```

Use: 
```
# This is okay
ActiveLeadProviders::SeedFromPrevious.(active_lead_provider:)

# This is best (IMO)
SeedFromPrevious.(active_lead_provider:)
```

This keeps them as close to methods in the caller as possible. The actual instance doing the work is never seen by the caller.

### Guidance to review

* Superclass ApplicationService.
* Class based invocation.
* includes namespace ~ this one might be controversial.

### DIFF
--- a/app/services/application_service.rb
+++ b/app/services/application_service.rb
@@ -1,0 +1,6 @@
+class ApplicationService
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+
+  def self.call(**kwargs) = new(**kwargs).call
+end
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -101,6 +101,9 @@ Style/HashSyntax:
 Style/Lambda:
   EnforcedStyle: literal
 
+Style/LambdaCall:
+  Enabled: false
+
 Rails/UnknownEnv:
   Environments:
     - production
--- a/app/controllers/admin/finance/active_lead_providers_controller.rb
+++ b/app/controllers/admin/finance/active_lead_providers_controller.rb
@@ -1,8 +1,11 @@
 module Admin::Finance
   class ActiveLeadProvidersController < Admin::Finance::BaseController
+    include ::ActiveLeadProviders
+
     layout "full"
 
     before_action :set_contract_period
+    before_action :set_active_lead_provider, only: %i[destroy]
     before_action :redirect_unless_contract_period_editable, only: %i[new create destroy]
 
     def index
@@ -24,11 +27,11 @@ module Admin::Finance
     end
 
     def create
-      @active_lead_provider = ::ActiveLeadProviders::Create.new(
+      @active_lead_provider = Create.(
         author: current_user,
         contract_period: @contract_period,
         lead_provider_id: active_lead_provider_params[:lead_provider_id]
-      ).call
+      )
 
       if @active_lead_provider.persisted?
         flash[:notice] = "#{@active_lead_provider.lead_provider.name} added"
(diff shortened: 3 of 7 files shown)


app/services/application_service.rb
```diff
@@ -0,0 +1,6 @@
+class ApplicationService
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+
+  def self.call(**kwargs) = new(**kwargs).call
+end
```
tonyheadford, line 6: [We had basically the same thing in the old service](https://github.com/DFE-Digital/early-careers-framework/blob/main/app/services/base_service.rb), however at some point the general consensus was that the devs didn't like the pattern.  That maybe down to how it was used by different devs but from what I can remember it was not helpful to have the code riddled with `.call` when trying to find something specific.  Perhaps there's a way to mitigate against that being an issue?

Also why do we need model behaviours in every service?

craigmdavidson: Thanks @tonyheadford 

I think the answer to that is that we should _never_ (_ever_) have the `.call` in the code.

The whole advantage of the superclass is we can eschew the `call`.
So we should always have `SeedFromPrevious.(active_lead_provider)`, i.e. use method objects like methods.

> That maybe down to how it was used by different devs but from what I can remember it was not helpful to have the code riddled with .call when trying to find something specific.

Search for the method object? What is better in my view is to have methods on the domain entities - standard Rails style? 🤷 

> Also why do we need model behaviours in every service?
If we use models we don't need they handle the constructors for us. Rails providesthisfor us, so why not take advantage of it?

craigmdavidson: > What is better in my view is to have methods on the domain entities - standard Rails style? 🤷

https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3257

tonyheadford: IMO using this form `ActiveLeadProviders::SeedFromPrevious.(active_lead_provider)` causes a "WTF, where's the receiver? There must be magic going on here" for me at least and we should avoid that kind of thing in general (causing WTF moments).
Given the choice, I'd probably go for something more akin to:
```
ActiveLeadProviderSeeder.new.seed_from_previous!(active_lead_provider:)
```

PR #3249, opened 2026-07-16
Ignore ended TRS prohibition alerts during registration

avinhurry:
## Context

Updating TRS prohibition handling so RECT only treats a teacher as prohibited from teaching when the prohibition alert is still active.

Previously, any TRS alert in the "Prohibition from teaching" category caused `trs_prohibited_from_teaching` to be `true`, even when the alert had an `endDate` in the past.

## Changes proposed

- Use the full TRS alert payload while building `TRS::Teacher` to check the alert `endDate`
- Treat prohibition alerts as active when:
  - `endDate` is blank or missing
  - `endDate` is in the future
  - `endDate` cannot be parsed
- Treat prohibition alerts as ended when `endDate` is in the past.
- Continue storing only:
  - `trs_alerts`: alert category ids
  - `trs_prohibited_from_teaching`: the calculated boolean

The end date is only used while building `TRS::Teacher` and the existing `trs_alerts` output hasn't changed and still only exposes the alert category ids.

### DIFF
--- a/lib/trs/teacher.rb
+++ b/lib/trs/teacher.rb
@@ -45,7 +45,9 @@ module TRS
       @trs_last_name = data["lastName"]
       @trs_email_address = data["emailAddress"]
       @trs_national_insurance_number = data["nationalInsuranceNumber"]
-      @trs_alerts = data.fetch("alerts", []).map { |a| a.dig(*%w[alertType alertCategory alertCategoryId]) }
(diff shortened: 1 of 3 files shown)


peterdavidhamilton (approved):
LGTM I recall @peteryates mentioning some boolean flags on teacher records could become datetimes down the line.

PR #3253, opened 2026-07-16
[4146] RIAB API authentication spike

peterdavidhamilton:
### Context

HMW generate API tokens for ABs intended for use by external parties
 
- Database migrations add ABs to existing token model, new 3rd party entity and delivery object.
- Simple PoC UI and delivery mechanism.
- Seed AB tokens using opt-in env var `AB_API` (any value)
- Smoke test `GET` route
- Add factories (no new tests yet)
- Fix existing tests
- Draft bulk script
- Update docs

```mermaid
erDiagram
    API_TOKEN {
        bigint id PK
        string token
        string description
        datetime last_used_at
        datetime created_at
        datetime updated_at
        bigint lead_provider_id FK
        bigint appropriate_body_period_id FK
        bigint api_third_party_id FK
    }
    LEAD_PROVIDER {
        bigint id PK
        string name
    }
    APPROPRIATE_BODY_PERIOD {
        bigint id PK
        bigint appropriate_body_id FK
        datetime started_at
        datetime ended_at
    }
    API_THIRD_PARTY {
        bigint id PK
        string name
        string email
    }
    API_TOKEN_DELIVERY {
        bigint id PK
        bigint api_token_id FK
        string token
        datetime expires_at
        datetime used_at
        datetime created_at
        datetime updated_at
    }

    API_TOKEN ||..|| LEAD_PROVIDER : "belongs_to"
    API_TOKEN ||..|| APPROPRIATE_BODY_PERIOD : "belongs_to"
    API_TOKEN ||..|| API_THIRD_PARTY : "belongs_to"
    API_TOKEN }|..|| API_TOKEN_DELIVERY : "has_many"
    API_TOKEN_DELIVERY ||..|| API_TOKEN : "belongs_to"
  ```

### Guidance to review

- Seed locally or on pod with or without tokens
- Generate a token using an AB persona from `/appropriate-body/tokens`
- Check logs, Notify, and DB
- Retrieve token
- Use AB's token on `/api/v3/teachers` (temp) 

riab-api-auth

---

## Questions

- **Q. Do AB tokens need to expire?**
  - Yes, post RIAB data model migration tokens will expire at the end of an AB period.

- **Q. How will AB's let us know which 3rd party they are working with?**
  - In service via a form but we plan to coordinate with 3rd parties and bulk generate all tokens in one go.

- **Q. How will we restrict access to the relevant teachers?**
  - Tokens are owned by the AB so 3rd parties authenticate as the AB.

- **Q. Do any ABs plan to use the API directly and not through 3rd parties?**
  - We must support them doing so but do not expect them to. We therefore need to surface the token once after generation.

- **Q. Is selecting a 3rd party mandatory?**
  - Yes.

- **Q. Must we be able to distinguish AB activity from 3rd party activity?**
  - No.

- **Q. Do we need roles/permissions for the above and how granular do we need them to be?**
  - A 3rd party can do anything on behalf of the AB.

- **Q. Do we anticipate tokens grant access to API versions?**
  - No. Granular access permissions are not needed for MVP.

- **Q. How will tokens be shared between ABs and 3rd parties?**
  - The service will act as a secure intermediary and pass the token directly to the 3rd party using a self-destructing link via email.

---

## Implementation

- **Keep a single token table**
  - The team has moved away from polymorphism and tech lead prefers all tokens in one place.
  - The events table pattern using multiple foreign keys sets the precedent and is extensible.

- **Each AB has a token**
  - ABs are not expected to call directly but could.
  - ABs will generate and revoke tokens in service and own the 3rd party nomination process.
  - Tokens can be displayed to the AB after creation.
  - We do not need distinguish between the AB and their 3rd party when the API is accessed.

- **No scopes on tokens**
  - A token grants access to the AB's induction periods already.
  - No additional RBAC is required.

- **UI for token manager**
  - A RESTful resource for tokens leveraging the existing token manager can be exposed to the AB.
  - Similarly, an Admin facing UI can be copied to provide a way to regenerate and distribute LP tokens.

- **Token handling**
  - ECT Manager and Mozaic will provide us an email address each and an automated email via Notify will facilitate retrieval.
  - Each token delivery will have a single recipient because the link expires after use.
  - This approach is scalable and replaces manual intervention and delays inherent in the current use of Galaxkey.

- **MVP release**
  - We can identify 3rd party choices and can therefore script token generation in bulk.
  - Ephemeral links need to remain active long enough for ECTM and Mozaic to act.

### CODE REVIEW

File: app/models/api/third_party.rb [hunk 1]
```diff
@@ -0,0 +1,8 @@
+class API::ThirdParty < ApplicationRecord
```
craigmdavidson, line 1: Why do we _care_ who the third party is? 

If an AppropriateBody says the token is for ECTManager then uses it for Mosaic do we care? What if they want to roll their own system/scripts?

Will we know, will we care?

app/controllers/appropriate_bodies/tokens_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+        delivery = API::TokenDelivery.create!(
+          api_token:,
+          expires_at: 1.hour.from_now
+        )
+
+        APITokenDeliveryMailer.with(
+          recipient_name: api_third_party.name,
+          recipient_email: api_third_party.email,
+          token_url: api_token_delivery_url(delivery.token)
+        ).api_token_email.deliver_later
```
craigmdavidson, line 28: Why do we email the token to the API Third Party rather than just show it to the Appriopriate Body?

Is there something I'm missing, or are ECTManager and Mosaic not just multi-tenanted applications?

If we just show the token to the API token to the ApproprateBody they can add it to their "profile" in the third party app?

If you wanted to wanted to connect your Stripe account to your Xero account, you would get the token from your Xero account and add it to your Stripe account (and vice versa).

craigmdavidson (commented):
Love it! 

💯 Agree on having the tokens at the AppropriateBody level rather than the third party, this way everything can be scoped to the token. It is clean, secure by design.

I do have questions about 2 things:
* Why do we even care about which third party the AppropriateBody wants to use? What if they wanted to roll their own? We are not whitelisting URLS, and would we really want to be in the business of maintaining third party lists anyway?
 
* Why would we email third parties a OTP style token? That path disempowers the Appropriate Bodies in favour of the 3rd parties? The normal pattern here would be to give the user/org (in this case the Appropriate Body) an API key and let them use the API as they please?

If we wanted more than this we could layer on a HMRC to Accounting app Government Gateway style OAuth dance?

app/controllers/appropriate_bodies/tokens_controller.rb
peterdavidhamilton: Good question. 

@mdoyle99 what does the current AB UX look like? 

My thinking was broadly that LPs have had their token delivered using Galaxkey in the past, but that has its own hurdles; so is there an option that is able to be agnostic and work for both.
 
@easeynathan asked about the prospect of getting everyone setup en masse and HMW make it really simple for lowfi smaller bodies.

app/models/api/third_party.rb
peterdavidhamilton: I think the answer is two fold, the business wants to know who has been chosen and when that changes, and the service needs to know for the email mechanism to work.

By also exposing any token in the UI we also give them the chance to roll their own.

app/controllers/appropriate_bodies/tokens_controller.rb
mdoyle99: I'm not 100% sure what the AB UX looks like on the third-party platform but the reason we thought best to email the token directly to the third party is that it removes a step for the AB and also mitigates against any potential data mishandling - e.g. if ABs are responsible for generating and sharing tokens it brings in some risks but let's chat about it in our catch-up on Weds. 

In the meantime I will have a look for screenshots of the ECTM UX as I think we had some conversations with them a while back

mdoyle99: [ECT manager screenshots](https://lucid.app/lucidspark/c5cf9b30-ccdc-4788-a1e3-e286c9d02641/edit?beaconFlowId=425382D7295B334D&page=0_0&invitationId=inv_5ba73cbb-6155-4e65-829e-0acd546f62f2#)  have pulled these screenshots of ECTM flows out of a RIAB discovery lucid - shows some of the key user journeys/homepage etc - side note that it is probably a bit out of date but gives us an idea of what things look like on their side

peterdavidhamilton: Cheers @mdoyle99 I have requested `view` access

peterdavidhamilton: First impressions are they would want a new option on their dashboard and a dedicated page rather than shoehorning something into an existing form.

mdoyle99: Yeah had similar thoughts - would be something that we need to investigate with them to assess the feasibility of adding something - would also need to see what Mosaic UX looks like.

PR #3256, opened 2026-07-17
Fix and validate swagger schema

ethax-ross:
### Context

As part of setting up the API in the Find and Use an API service we ran into errors uploading our swagger schema as it is falling foul of the OpenAPI v3 spec.

- Fix `required` notation in swagger schemas

The OpenAPI spec only wants you to specify `required` attributes as an array before defining the `properties`. Having `required: true/false` on the property itself results in a validation error.

I think we did this originally because swagger-ui wasn't putting a required asterisk next to the attribute name without it, but I checked after this change and it appears to now, so it isn't an issue any longer.

- Validate OpenAPI schema in GitHub action

Extends the existing action to also run the swagger.yaml through Redocly CLI which will validate it against the OpenAPI spec v3.

- Fix swagger schema issues

There were a number of minor issues in our swagger schema that resulted in validation errors:

- Typos
- Examples that don't match the attribute type
- Examples that don't match the enum options
- Examples with invalid UUID values
- Missing operation IDs

### Guidance to review

This prevents issues when we later upload the generated schema file to the Find and Use an API service.

### DIFF
--- a/spec/swagger_schemas/responses/particpant_transfers.rb
+++ b/spec/swagger_schemas/responses/particpant_transfers.rb
@@ -1,6 +1,6 @@
 PARTICIPANT_TRANSFERS_RESPONSE = {
   description: "Transfers for a given participant",
-  tyoe: :object,
+  type: :object,
   required: %i[data],
   properties: {
     data: { "$ref": "#/components/schemas/ParticipantTransfer" },
--- a/.github/workflows/lead_provider_openapi_check.yml
+++ b/.github/workflows/lead_provider_openapi_check.yml
@@ -59,6 +59,12 @@ jobs:
           prepare-test-database: "true"
           prepare-assets: "true"
 
+      - name: Install Redocly CLI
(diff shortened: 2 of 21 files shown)


peterdavidhamilton (approved):
If Redocly is happy, I am happy
https://github.com/DFE-Digital/register-early-career-teachers-public/actions/runs/29565741360/job/87838162916

spec/swagger_schemas/responses/particpant_transfers.rb
```diff
@@ -1,6 +1,6 @@
 PARTICIPANT_TRANSFERS_RESPONSE = {
   description: "Transfers for a given participant",
-  tyoe: :object,
+  type: :object,
```
craigmdavidson, line 3: Nice catch!

craigmdavidson (approved):
LGTM. 

Observation: We do seem to have a lot of swagger schema files to maintain. It would be good if we could somehow translate our blueprinter files into swagger schema files.

PR #3257, opened 2026-07-17
DO NOT MERGE - Make ActiveRecord classes more like models

craigmdavidson:
### Context

Okay so this is probably going to be a bridge too far, so I'll start with a blatant appeal to authority: 

> Now, the more common mistake is to give up too easily on fitting the behavior into an appropriate object, gradually slipping toward procedural programming.
> **Eric Evans, Domain Driven Design**

Sometimes we find ourselves with behaviour that doesn't naturally fit in any single class. This is often the exception rather than the rule. However, often times we find ourselves following a pattern, and have arrived a services to Create a new object. E.g. `AppropriateBodies::Create`. This has led us to a pretty anemic domain model. 

This makes it very hard to understand what any given entity can do, and hard to distinguish these pseudo-services from real complex orchestrations (when everything is a service, you can't see the forest from the trees). 

DRAFT

### Guidance to review

DRAFT


app/models/contract_period/event_recordable.rb
```diff
@@ -0,0 +1,22 @@
+class ContractPeriod
+  module EventRecordable
```
peterdavidhamilton, line 2: I was looking at the event concerns without having looked this PR first, so I hadn't expected to see multiple `EventRecordable` concerns.

My first reaction was, why not one?

Looking at `Contracts::Update` and the extra logic for modifications I can see how a dedicated concern benefits us and gives a home for any model-specific stuff.

I'd like to see us run with this a bit and see if any edgecases throw a spanner at this pattern. 

It is nice to look at

app/models/contract_period/event_recordable.rb
```diff
@@ -0,0 +1,22 @@
...
+    extend ActiveSupport::Concern
+
+    included do
+      has_many :events
+
+      after_create :record_created_event
```
peterdavidhamilton, line 8: Is it advisable to instead use `after_create_commit`, `after_update_commit` and `after_destroy_commit` to guard against rollbacks, ensuring the event is queued only once the database is changed.

craigmdavidson: I'm not sure???  

Using `after_update` keeps it in the same transaction, mirroring what's there in the service object.
https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3257/changes#diff-b85dbc5f45fa4289691f62add6db7c61f80af7366807f8ae41818cb4431df83bL23-L26
```
        contract_period.update!(params)
        record_event!(modifications)
      end
```

But using `after_create_commit` keeps it in a different transaction, mirroring what's there in the service object.
https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3257/changes#diff-1c72d5703363bde614dcd51df55ce7531166ce40e9bf791e972ba23f8ab3f7e6L14-L17
```
if active_lead_provider.save
  Events::Record.record_active_lead_provider_created_event!(author:, active_lead_provider:)
  SeedFromPrevious.(active_lead_provider:)
end
```
or do they:
https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3257/changes#diff-59dd76e7e60ffc749746dc8a951173fe6c413ae50874ffac37276f27bcc72507L17-L21
```
        contract_period.save!
        record_event!
        seed_from_previous!
      end
```


😵 Our existing service objects are not really consistent.

Which for me is a pretty decent argument for the callbacks by itself?! 🤷

PR #3261, opened 2026-07-17
Add ECT contract period correction task

ericaporter:
### Context
A small number of ECTs were moved from school-led to provider-led training using contract period 2026, even though their previous provider-led training used contract period 2024.

N.B. The application fix prevents this happening for future records

Adds a rake task which calls `Support::ECTContractPeriodCorrection` to validate and apply the changes

The task:
- validates the proposed correction
- automatically identifies the matching schedule and school partnership or - expression of interest for the replacement contract period
- displays the current and proposed values before making any changes
- requires confirmation using the teacher ID
- updates the training period within a transaction

Validates
- training period belongs to the supplied teacher
- the period is ongoing and provider-led
- the current contract period is the expected one
- the replacement schedule uses the correct contract period and schedule identifier
- the replacement partnership or expression of interest belongs to the correct lead provider and contract period

### Guidance to review
Will be run on migration env to test on realistic data

### DIFF
--- a/app/services/data_corrections/ect_contract_period_correction.rb
+++ b/app/services/data_corrections/ect_contract_period_correction.rb
@@ -1,4 +1,4 @@
-module Support
+module DataCorrections
   class ECTContractPeriodCorrection
     class InvalidCorrection < StandardError; end
 
@@ -66,6 +66,11 @@ module Support
     end
 
     def validate_training_period!
+      unless training_period.ect_at_school_period
+        raise InvalidCorrection,
+              "Training period must be for an ECT"
+      end
+
       unless teacher.id == teacher_id
         raise InvalidCorrection,
               "Training period belongs to a different teacher"
--- a/app/services/data_corrections/ect_contract_period_correction.rb
+++ b/app/services/data_corrections/ect_contract_period_correction.rb
@@ -9,7 +9,8 @@ module DataCorrections
       replacement_contract_period_year:,
       replacement_schedule: nil,
       replacement_school_partnership: nil,
-      replacement_expression_of_interest: nil
+      replacement_expression_of_interest: nil,
+      allow_finished_training_period: false
     )
       @training_period = training_period
       @teacher_id = teacher_id
@@ -21,6 +22,8 @@ module DataCorrections
         replacement_school_partnership
       @replacement_expression_of_interest =
         replacement_expression_of_interest
+      @allow_finished_training_period =
+        allow_finished_training_period
     end
 
     def preview
@@ -53,7 +56,8 @@ module DataCorrections
     attr_reader :training_period,
                 :teacher_id,
                 :current_contract_period_year,
-                :replacement_contract_period_year
+                :replacement_contract_period_year,
+                :allow_finished_training_period
 
     def validate!
       validate_training_period!
@@ -81,7 +85,8 @@ module DataCorrections
               "Training period must be provider-led"
       end
 
-      if training_period.finished_on.present?
+      if training_period.finished_on.present? &&
+          !allow_finished_training_period
         raise InvalidCorrection,
               "Training period must be ongoing"
       end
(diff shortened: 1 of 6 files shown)


app/services/data_corrections/ect_contract_period_correction.rb
```diff
@@ -0,0 +1,267 @@
...
+      validate_training_period!
+      validate_contract_period_change!
+      validate_current_partnership_or_expression_of_interest!
+      validate_replacement_schedule!
+      validate_replacement_partnership_or_expression_of_interest!
+      validate_replacement_school_partnership!
+      validate_replacement_expression_of_interest!
+    end
+
+    def validate_training_period!
```
avinhurry, line 72: Nice to see us being cautious here. Should we explicitly raise `InvalidCorrection` when the training period is not for an ECT?

The rake task accepts any training period id, and otherwise a mentor period will fail when we try to call `ect_at_school_period.teacher` on line 255.

lib/tasks/data_corrections/ect_contract_period_correction.rake
```diff
@@ -0,0 +1,97 @@
...
+         training_period_id
+         teacher_id
+         current_contract_period_year
+         replacement_contract_period_year
+         replacement_schedule_id
+         replacement_school_partnership_id
+         replacement_expression_of_interest_id
+       ] => :environment do |_task, args|
+    training_period =
+      TrainingPeriod.find(args.fetch(:training_period_id))
```
avinhurry, line 17: Is the plan to delete this rake task after we use it? When`Rake` prints the tasks it pushes the description far to the right so it makes things difficult to read.

I'm ok with it if it's temporary and it will be removed.

app/services/support/ect_contract_period_correction.rb
```diff
@@ -0,0 +1,267 @@
+module Support
+  class ECTContractPeriodCorrection
```
avinhurry: I'm wondering whether a `data_corrections` folder under `app/services` would make it clearer that this is a one off task. The rake task could then live in `lib/tasks/data_corrections/ect_contract_period_correction.rake`.

At the moment `app/services/support/ect_contract_period_correction.rb` could sound like a long lived support service to the uninitiated. Similarly, `support:ect_contract_period_correction` sits alongside `support:undo_registration`, which appears to be a long lived operational task rather than a temporary data correction.

WDYT?

avinhurry (approved):
I've left a small extra hardening suggestion and some of my thoughts on how we might start organising these temporary tasks.

Other than that, this looks very thorough, nice work ✅

app/services/support/ect_contract_period_correction.rb
ericaporter: I think data_corrections is better at reflecting the purpose and the intent - happy to update this!

We haven't been removing these tasks so far but I expect we will once functionality moves to the admin.

lib/tasks/data_corrections/ect_contract_period_correction.rake
ericaporter: We haven’t generally been deleting these tasks immediately after use but I'm sure we will remove them at some point.

There are a few options but none are ideal - the last three arguments are optional overrides -  so I could keep the four required values as rake arguments and move the optional IDs to env variables or remove them completely. I could also shorten the variable names or split it up into two tasks but I feel the current naming and inclusion makes it safer to run.

app/services/data_corrections/ect_contract_period_correction.rb
ericaporter: Good catch - I think this is worth fixing

lib/tasks/data_corrections/ect_contract_period_correction.rake
avinhurry: Yeah, I think you're right to be honest, and I don't want to derail this work.

I couldn't think of an alternative other than env variables, and I think that would hide the optional overrides. Perhaps we keep it as it is for now as a known trade off and If it causes issues for devs, we can revisit it. 

Hopefully we can remove the task by then.

avinhurry (approved):
Nice work! 

Thank you for making those changes.

PR #3263, opened 2026-07-20
Fix paragraph rendering on mentor confirmation pages

cpjmcquillan:
We're using the markdown renderer for these pages, but for some reason it doesn't render the paragraph immediately after the panel in a `<p>` tag.

This adds an extra newline to ensure it does.

<details>
<summary>Before</summary>
image

</details>

<details>
<summary>After</summary>
image

</details>


avinhurry (approved):
This fix works and happy to approve but why are we using markdown files here? Do we get any benefit from it or can we switch these all to `html.erb` files. I have a feeling this will get reintroduced in some shape or form in the future otherwise.

peteryates (approved):

mason-emily (approved):

PR #3265, opened 2026-07-20
complex_pre_092021_patches

boonali:
### Context
As part of migration, a number of 1-day At School Periods and Training Periods were created with start dates prior to September 2021. This is not to be allowed within RECT.

The pre-September 2021 periods need to either be removed if they don't link to any declarations, or if they do, their subsequent periods need to be moved forward to allow the erroneous period to fit in.

This pull request is specifically for 20 participants with more complex shapes, eg. multiple pre-09/2021 periods or a one-day period on 01/09/2021, which results in a more complex fix.

Amending the patch file for the amendments to the Training and At School periods for these 20 participants.

### Guidance to review
Ensure there are no periods prior to 1st September 2021.


tonyheadford (approved):
Tested with snapshot data and no errors

PR #3268, opened 2026-07-20
Fix n+1 query on delivery partners index

peteryates:
This PR fixes an n+1 query on the delivery partners show page by querying more efficiently.

It now uses two queries, one to retrieve the contract periods and the other to retrieve the list of appropriate bodies partnered per contract period. This is necessary so 'empty' contract periods are included in the list.

The page is really low traffic and not really a problem, but the fewer pages that trigger warnings the easier the real problem pages are to spot.

### Context

- **Add service object that returns partnership data**
- **Make delivery partner show page use new query service**

### DIFF
--- a/app/services/delivery_partners/partnership_data.rb
+++ b/app/services/delivery_partners/partnership_data.rb
@@ -1,0 +1,25 @@
+module DeliveryPartners
+  class PartnershipData
+    attr_reader :delivery_partner, :all_contract_periods
+
+    def initialize(delivery_partner)
+      @delivery_partner = delivery_partner
+      @all_contract_periods = ContractPeriod.all.latest_first.each_with_object({}) do |cp, h|
+        h[cp.year] = []
+      end
+    end
+
+    # returns a hash of contract_period years and lead provider names
+    # { 2021 => ["Lead provider 1", "Lead provider 2"] }
+    def partners_by_contract_period
+      present = ContractPeriod
+        .includes(active_lead_providers: [:lead_provider, { lead_provider_delivery_partnerships: :delivery_partner }])
+        .where(active_lead_providers: { lead_provider_delivery_partnerships: { delivery_partner: } })
+        .each_with_object({}) do |cp, h|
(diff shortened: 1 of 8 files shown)


app/services/delivery_partners/partnership_data.rb
```diff
@@ -0,0 +1,25 @@
...
+    # { 2021 => ["Lead provider 1", "Lead provider 2"] }
+    def partners_by_contract_period
+      present = ContractPeriod
+        .includes(active_lead_providers: [:lead_provider, { lead_provider_delivery_partnerships: :delivery_partner }])
+        .where(active_lead_providers: { lead_provider_delivery_partnerships: { delivery_partner: } })
+        .each_with_object({}) do |cp, h|
+          h[cp.year] = cp.active_lead_providers.map { |alp| alp.lead_provider.name }.sort
+        end
+
+      { **all_contract_periods, **present }
```
peteryates, line 22: When two hashes are merged, if the second one has keys that clash with the first, the value is updated in place maintaining the original order.

app/services/delivery_partners/partnership_data.rb
```diff
@@ -0,0 +1,25 @@
...
+    end
+
+    # returns a hash of contract_period years and lead provider names
+    # { 2021 => ["Lead provider 1", "Lead provider 2"] }
+    def partners_by_contract_period
+      present = ContractPeriod
+        .includes(active_lead_providers: [:lead_provider, { lead_provider_delivery_partnerships: :delivery_partner }])
+        .where(active_lead_providers: { lead_provider_delivery_partnerships: { delivery_partner: } })
+        .each_with_object({}) do |cp, h|
+          h[cp.year] = cp.active_lead_providers.map { |alp| alp.lead_provider.name }.sort
```
peteryates, line 19: One change I made here is that now there's an explicit alphabetic ordering of the lead providers.

avinhurry (approved):
Looks good to me. Nice to see us nipping these N+1s in the bud, even on low traffic pages. 

The less noise we have, the easier it is to spot real issues (like you said) 👍🏽

PR #3271, opened 2026-07-21
Move service nav and phase banner inside the `<header>`

peteryates:
This PR moves all the service navigation and phase banner inside the `<header>`.

Refs DFE-Digital/register-ects-project-board#4082

There should be no visible changes, essentially we're going:

| From | To |
| ---- | ----- |
| Screenshot From 2026-07-21 17-33-24 | Screenshot From 2026-07-21 17-32-55 |

It makes use of `govuk_header`'s ability to [render the phase banner and service nav as slots](https://govuk-components.x-govuk.org/components/header/#header-with-service-navigation-and-phase-banner) so they're inside the `<header>` element.

## Refactoring the API guidance pages

This was a bit more work, the header, phase banner and service nav was replicated in several places, and logic around the inversion of colours on the landing page (in the masthead) made doing it without disruption a bit painful. I took the opportunity to refactor the tops of those pages so that:

* we now use the same `_header` partial everywhere
* the landing page with the blue masthead has its own layout (`api_home`)
* duplicate headers, phase banners and service navigations have been consolidated

| | Home | Swagger | Release notes | Guidance |
| ---- | ------ | ------- | ------ | ------- |
| Before | Screenshot From 2026-07-21 17-29-26 |Screenshot From 2026-07-21 17-29-21 | Screenshot From 2026-07-21 17-29-29 |Screenshot From 2026-07-21 17-29-32 |
| After |  Screenshot From 2026-07-21 17-43-43 | Screenshot From 2026-07-21 17-29-45 | Screenshot From 2026-07-21 17-29-49 | Screenshot From 2026-07-21 17-29-52 |

## Review tips

The ~~last~~ "Simplify API changes" commit is the big one that did loads of moving around, I'd suggest going commit-by-commit. Happy to discuss/walk through the changes.

## Changes

- **Move service nav inside header element**
- **Split environment_specific_phase_banner method up**
- **Move phase banner inside the <header> element**
- **Simplify API pages by adding new home layout**

### DIFF
--- a/app/helpers/environment_helper.rb
+++ b/app/helpers/environment_helper.rb
@@ -7,10 +7,6 @@ module EnvironmentHelper
     "app-header--#{ENVIRONMENT_COLOUR}"
   end
 
-  def environment_specific_phase_banner(html_attributes: {}, tag_html_attributes: {})
-    govuk_phase_banner(**environment_specific_phase_banner_arguments(html_attributes:, tag_html_attributes:))
-  end
-
   def environment_specific_phase_banner_arguments(inverse: false)
     tag_text = ENVIRONMENT_PHASE_BANNER_TAG || "Beta"
     banner_text = ENVIRONMENT_PHASE_BANNER_CONTENT || environment_phase_banner_default_content
--- a/app/helpers/environment_helper.rb
+++ b/app/helpers/environment_helper.rb
@@ -8,10 +8,18 @@ module EnvironmentHelper
   end
 
   def environment_specific_phase_banner(html_attributes: {}, tag_html_attributes: {})
+    govuk_phase_banner(**environment_specific_phase_banner_arguments(html_attributes:, tag_html_attributes:))
+  end
+
+  def environment_specific_phase_banner_arguments(html_attributes: {}, tag_html_attributes: {})
     tag_text = ENVIRONMENT_PHASE_BANNER_TAG || "Beta"
     banner_text = ENVIRONMENT_PHASE_BANNER_CONTENT || environment_phase_banner_default_content
 
-    govuk_phase_banner(text: banner_text, tag: { text: tag_text, colour: ENVIRONMENT_COLOUR, html_attributes: tag_html_attributes }.compact, html_attributes:)
+    {
+      text: banner_text,
+      tag: { text: tag_text, colour: ENVIRONMENT_COLOUR, html_attributes: tag_html_attributes }.compact,
+      html_attributes:,
+    }
   end
 
 private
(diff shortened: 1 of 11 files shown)


app/helpers/environment_helper.rb
```diff
@@ -8,10 +8,30 @@ def environment_specific_header_colour_class
...
+                       %w[govuk-width-container x-govuk-phase-banner--inverse]
+                     else
+                       %w[govuk-width-container]
+                     end
+
+    tag_classes = if inverse
+                    %w[x-govuk-tag--inverse]
+                  else
+                    []
+                  end
```
peteryates: Possibly a nicer way to do this with `class_names`, looking at it.

app/helpers/environment_helper.rb
```diff
@@ -8,10 +8,30 @@ def environment_specific_header_colour_class
   end
 
   def environment_specific_phase_banner(html_attributes: {}, tag_html_attributes: {})
+    govuk_phase_banner(**environment_specific_phase_banner_arguments(html_attributes:, tag_html_attributes:))
```
avinhurry: Should we drop `environment_specific_phase_banner` now that `environment_specific_phase_banner_arguments` is used directly? 

The wrapper still forwards `html_attributes` and `tag_html_attributes`, which the new method no longer accepts. I couldn't find any remaining API doc page callers.

avinhurry (approved):
Looks good to me. Super clear description and nice commit story!

app/helpers/environment_helper.rb
peteryates: Ah yeah good call, I made a mental note to do that and forgot!

peteryates: > The wrapper still forwards html_attributes and tag_html_attributes, which the new method no longer accepts. I couldn't find any remaining API doc page callers.

Yeah I thought as we only use them in one place (the blue masthead) to invert the colours, it'd be cleaner to just have an `inverse` argument that does it.

app/helpers/environment_helper.rb
```diff
@@ -11,17 +11,8 @@ def environment_specific_phase_banner_arguments(inverse: false)
...
-                     else
-                       %w[govuk-width-container]
-                     end
-
-    tag_classes = if inverse
-                    %w[x-govuk-tag--inverse]
-                  else
-                    []
-                  end
+    banner_classes = class_names("govuk-width-container", "x-govuk-phase-banner--inverse" => inverse)
```
peterdavidhamilton, line 14: 👍🏼

peterdavidhamilton (approved):

PR #3278, opened 2026-07-22
[4192] Include contract period in CSV filename

avinhurry:
## Summary

To make assurance macros easier to work with we are removing hyphens from the lead provider names and including the contract period year in the declarations CSV filename.

Before: `ambition-institute-january-2021-declarations.csv`

After: `ambitioninstitute-2021-contract-january-2021-declarations.csv`


peteryates (approved):
Looks good

PR #3279, opened 2026-07-22
Add `govuk_body_element` helper

peteryates:
We render the body element in various layouts and it [should be followed immediately by a skip link](https://design-system.service.gov.uk/components/skip-link/#when-to-use-this-component).

The `govuk_body_element` helper automatically renders the govuk_skip_link component at the very top of `<body>`


avinhurry (approved):
Looks good - Do we need to update the docs site too?

PR #3281, opened 2026-07-23
Fix 4239 backdated mentorship overlap

ericaporter:
Ticket [here](https://github.com/DFE-Digital/register-ects-project-board/issues/4239)

### Context
When registering a transferring mentor for an ECT with a previous finished mentorship, the new mentorship could be backdated to the mentor’s school start date. If that date overlapped the previous mentorship, the assignment failed validation.

- keeps the previous mentorship unchanged
- creates a new mentorship from the date of the change
- prevents the new period from overlapping a previous mentorship

### DIFF
--- a/app/services/schools/mentor_assignment/mentorship_periods/dates_resolver.rb
+++ b/app/services/schools/mentor_assignment/mentorship_periods/dates_resolver.rb
@@ -5,24 +5,31 @@ module Schools
         def initialize(
           ect_at_school_period:,
           mentor_at_school_period:,
-          mentorship_can_start_today:
+          mentor_is_transferring_schools:
         )
           @ect_at_school_period = ect_at_school_period
           @mentor_at_school_period = mentor_at_school_period
-          @mentorship_can_start_today = mentorship_can_start_today
+          @mentor_is_transferring_schools = mentor_is_transferring_schools
         end
 
         def earliest_possible_start
           [
             @ect_at_school_period.started_on,
             @mentor_at_school_period.started_on,
-            (Date.current if @mentorship_can_start_today)
+            (Date.current unless backdating_allowed?)
           ].compact.max
         end
 
         def latest_possible_finish
           [@ect_at_school_period.finished_on, @mentor_at_school_period.finished_on]
             .compact.min
+        end
+
+      private
+
+        def backdating_allowed?
+          @mentor_is_transferring_schools &&
+            @ect_at_school_period.mentorship_periods.none?
         end
       end
     end
--- a/app/wizards/schools/register_mentor_wizard/check_answers_step.rb
+++ b/app/wizards/schools/register_mentor_wizard/check_answers_step.rb
@@ -50,17 +50,12 @@ module Schools
             ect_at_school_period: ect,
             mentor_at_school_period: mentor.register!(author:),
             author:,
-            mentorship_can_start_today: use_today_as_earliest_start?
+            mentor_is_transferring_schools: mentoring_at_new_school_only?
           ).assign!
         end
       rescue StandardError => e
         mentor.registered = false
         raise e
-      end
-
-      def use_today_as_earliest_start?
-        ect.mentorship_periods.exists? ||
-          !mentoring_at_new_school_only?
       end
     end
   end
--- a/app/wizards/schools/register_mentor_wizard/check_answers_step.rb
+++ b/app/wizards/schools/register_mentor_wizard/check_answers_step.rb
@@ -50,12 +50,17 @@ module Schools
             ect_at_school_period: ect,
             mentor_at_school_period: mentor.register!(author:),
             author:,
-            mentorship_can_start_today: !mentoring_at_new_school_only?
+            mentorship_can_start_today: use_today_as_earliest_start?
           ).assign!
         end
       rescue StandardError => e
         mentor.registered = false
         raise e
+      end
+
+      def use_today_as_earliest_start?
+        ect.mentorship_periods.exists? ||
+          !mentoring_at_new_school_only?
       end
     end
   end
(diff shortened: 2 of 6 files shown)


spec/services/schools/assign_mentor_spec.rb
cpjmcquillan, line 1: I'm not sure this test belongs here.

We haven't changed anything in `Schools::AssignMentor` class, and the implementation doesn't care about previous mentorships either.

The `when the mentee is currently being mentored` context is only relevant for the tests that describe `closing current mentorship`.

What do you think?

app/wizards/schools/register_mentor_wizard/check_answers_step.rb
```diff
@@ -50,13 +50,18 @@ def persist
...
+            mentorship_can_start_today: use_today_as_earliest_start?
           ).assign!
         end
       rescue StandardError => e
         mentor.registered = false
         raise e
       end
+
+      def use_today_as_earliest_start?
+        ect.mentorship_periods.exists? ||
```
cpjmcquillan: This feels like an oversimplification.

If a new mentor is transferring schools (and the date would clash with a previous mentorship period for the ECT), should we actually be using whichever is latest from `[ect.started_on, mentor.started_on, latest_mentorship_period.finished_on.next_day]`? 🤔

spec/services/schools/assign_mentor_spec.rb
ericaporter: Ah came in from other branch! Will remove!

cpjmcquillan (commented):
I think we should try and keep the boundaries between the `RegisterMentorWizard`, `Schools::AssignMentor` and `Schools::MentorAssignment::MentorshipPeriods::DateResolver` as strict as we can.

I wonder if it would be clearer and more robust if we did something like this instead?

```ruby
# app/wizards/schools/register_mentor_wizard/check_answers_step.rb
def persist
    AssignMentor.new(
      ect_at_school_period: ect,
      mentor_at_school_period: mentor.register!(author:),
      author:,
      mentor_is_transferring_schools:
    ).assign!
  end
rescue StandardError => e
  mentor.registered = false
  raise e
end

def mentor_is_transferring_schools? = !!mentoring_at_new_school_only?
```

and pass that through so

```ruby
module Schools
  module MentorAssignment
    module MentorshipPeriods
      class DatesResolver
        def initialize(
          ect_at_school_period:,
          mentor_at_school_period:,
          mentor_is_transferring_schools:
        )
          @ect_at_school_period = ect_at_school_period
          @mentor_at_school_period = mentor_at_school_period
          @mentor_is_transferring_schools = mentor_is_transferring_schools
        end

        def earliest_possible_start
          [
            @ect_at_school_period.started_on,
            @mentor_at_school_period.started_on,
            (latest_finished_mentorship_period.finished_on.next_day if latest_finished_mentorship_period),
            (Date.current unless @mentor_is_transferring_schools)
          ].compact.max
        end

        def latest_possible_finish
          [@ect_at_school_period.finished_on, @mentor_at_school_period.finished_on]
            .compact.min
        end

      private

        def latest_finished_mentorship_period
          @ect_at_school_period.mentorship_periods.finished.latest_first.take
        end
      end
    end
  end
end
```

Then we can add unit tests to the `DateResolver` to cover this scenario.

Also, do we need to consider the impact on the `latest_possible_finish` too? 🤔

What do you think?

app/wizards/schools/register_mentor_wizard/check_answers_step.rb
ericaporter: @cpjmcquillan that was my initial thought too, but product confirmed that when the ECT has had a previous mentor, the new mentorship should start from the date the change is made, rather than the day after the previous mentorship ended. Aamina left a note confirming this on the ticket.

My understanding is that the existing backdating behaviour should only apply when the ECT has no mentorship history and the mentor is transferring schools. Otherwise, the new mentorship should start on the latest of the ECT’s start date, the mentor’s start date and today.

Have I understood your point correctly?

cpjmcquillan: Ok, cool!

Let's capture that context in the commit message 🙌🏼.

I still think this should be the responsibility of the `DateResolver` class, though. What do you think?

app/services/schools/mentor_assignment/mentorship_periods/dates_resolver.rb
```diff
@@ -5,25 +5,32 @@ class DatesResolver
...
           [@ect_at_school_period.finished_on, @mentor_at_school_period.finished_on]
             .compact.min
         end
+
+      private
+
+        def backdating_allowed?
+          @mentor_is_transferring_schools &&
+            !@ect_at_school_period.mentorship_periods.exists?
+        end
```
cpjmcquillan, line 33: How do you feel about using [`none?`](https://api.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-none-3F) and flipping the negative instead? 🤔

```suggestion
        def backdating_allowed?
          @mentor_is_transferring_schools &&
            @ect_at_school_period.mentorship_periods.none?
        end
```

spec/services/schools/mentor_assignment/mentorship_periods/dates_resolver_spec.rb
```diff
@@ -26,12 +26,13 @@
...
   let(:ect_finished_on) { nil }
   let(:mentor_finished_on) { nil }
+  let(:mentor_is_transferring_schools) { false }
 
   describe "#earliest_possible_start" do
     subject(:earliest_possible_start) { service.earliest_possible_start }
 
-    context "when the mentorship can start today" do
-      let(:mentorship_can_start_today) { true }
+    context "when the mentor is not transferring schools" do
```
cpjmcquillan, line 34: Should we add contexts for the ECT having no previous mentorships/having a previous mentorship here too?

cpjmcquillan (approved):

PR #3287, opened 2026-07-27
Display an ECT's future mentorship details

cpjmcquillan:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/4265

Before, we only showed schools the `current_or_next` mentorship details for a
given ECT.

However this means when a school assigns a new mentor in the future, there is no
indication that change was successful.

This led to several schools running in to errors, and contacting support when
they failed to re-assign that same mentor again and again. Sometimes, up to 10+
times!

This reworks the ECT listing card and teacher profile components to render
upcoming mentorship details too.

The existing behaviour has been extended to display any upcoming mentorship
periods for the ECT in some hint text.

### Guidance to review

image
image

### DIFF
--- a/app/components/schools/ects/mentorship_component.html.erb
+++ b/app/components/schools/ects/mentorship_component.html.erb
@@ -1,0 +1,8 @@
+<% if mentorship.current_mentor.blank? %>
+  <%= govuk_warning_text(text: assign_or_create_mentor_link) %>
+<% else %>
+  <%= mentorship.current_mentor_name %>
+  <% if upcoming_mentorships.any? %>
+    <%= govuk_list(upcoming_mentorship_descriptions, classes: "govuk-hint") %>
+  <% end %>
+<% end %>
--- a/app/components/schools/ects/mentorship_component.rb
+++ b/app/components/schools/ects/mentorship_component.rb
@@ -1,0 +1,40 @@
+module Schools::ECTs
+  class MentorshipComponent < ApplicationComponent
+    include ECTHelper
+
+    def initialize(ect_at_school_period)
+      @ect_at_school_period = ect_at_school_period
+    end
+
+  private
+
+    def assign_or_create_mentor_link
+      govuk_link_to(
+        "Assign a mentor for this ECT",
+        assign_or_create_mentor_path(ect_at_school_period),
+        no_visited_state: true
+      )
+    end
+
+    attr_reader :ect_at_school_period
+
+    def mentorship
+      @mentorship ||= ECTAtSchoolPeriods::Mentorship.new(ect_at_school_period)
+    end
+
+    def upcoming_mentorships
+      mentorship.upcoming_mentorship_periods.reject do
+        it == mentorship.current_or_next_mentorship_period
+      end
+    end
+
+    def upcoming_mentorship_descriptions
+      upcoming_mentorships.map { upcoming_mentorship_description(it) }
+    end
+
+    def upcoming_mentorship_description(mentorship_period)
+      mentor_name = Teachers::Name.new(mentorship_period.mentor.teacher).full_name
+      "#{mentor_name} (from #{mentorship_period.started_on.to_fs(:govuk)})"
+    end
+  end
+end
--- a/app/controllers/schools/ects_controller.rb
+++ b/app/controllers/schools/ects_controller.rb
@@ -14,7 +14,10 @@ module Schools
     end
 
     def show
-      @ect_at_school_period = @school.ect_at_school_periods.find(params[:id])
+      @ect_at_school_period = @school
+        .ect_at_school_periods
+        .includes(upcoming_mentorship_periods: { mentor: :teacher })
+        .find(params[:id])
       @training_period = @ect_at_school_period.current_or_next_or_latest_training_period
       @teacher = @ect_at_school_period.teacher
     end
--- a/app/services/teachers/search.rb
+++ b/app/services/teachers/search.rb
@@ -30,7 +30,14 @@ module Teachers
                           .when(MentorshipPeriod.arel_table[:id].eq(nil)).then(0)
                           .else(1)
                           .asc,
-          { ect_at_school_periods: { started_on: :desc } }
+          {
+            ect_at_school_periods: { started_on: :desc },
+            # Calling .order overrides the ordering on the scoped associations
+            # (e.g upcoming_mentorship_periods).
+            # We need to explicitly order mentorship periods here to avoid
+            # has_many associations being loaded in the wrong order
+            mentorship_periods: { started_on: :asc },
+          }
         ]
       else
         %i[trs_last_name trs_first_name id]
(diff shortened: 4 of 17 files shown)


app/controllers/schools/ects_controller.rb
cpjmcquillan, line 1: These changes prevent N+1s

app/services/teachers/search.rb
```diff
@@ -30,7 +30,14 @@ def order
                           .when(MentorshipPeriod.arel_table[:id].eq(nil)).then(0)
                           .else(1)
                           .asc,
-          { ect_at_school_periods: { started_on: :desc } }
+          {
+            ect_at_school_periods: { started_on: :desc },
+            # Calling .order overrides the ordering on the scoped associations
+            # (e.g upcoming_mentorship_periods).
+            # We need to explicit order mentorship periods here to avoid
+            # has_one associations being loaded in the wrong order
```
cpjmcquillan:
```suggestion
            # has_many associations being loaded in the wrong order
```

spec/support/matchers/have_summary_list_row.rb
```diff
@@ -15,7 +15,7 @@ def matches?(page)
         @matching_row && @matching_row.text == @key
       else
         @sibling = @matching_row&.sibling("dd.govuk-summary-list__value", visible: @visible)
-        @sibling && @sibling.text == @value
+        @sibling && @sibling.text.strip.include?(@value)
```
avinhurry: Is this safe? We're changing the matching behaviour for every assertion. It could give us false confidence...

Did you consider keeping exact matching as the default and using a custom matcher for this special case?

I think this could cause specs to pass incorrectly for example if the name appears in the hint.

app/components/schools/ects/mentorship_component.rb
```diff
@@ -0,0 +1,40 @@
...
+
+    def upcoming_mentorships = mentorship.upcoming_mentorship_periods
+
+    def upcoming_mentorship_descriptions
+      safe_join(upcoming_mentorships.map { upcoming_mentorship_description(it) }, tag.br)
+    end
+
+    def upcoming_mentorship_description(mentorship_period)
+      mentor_name = Teachers::Name.new(mentorship_period.mentor.teacher).full_name
+      mentee_name = Teachers::Name.new(mentorship_period.mentee.teacher).full_name
```
avinhurry: This creates new n+1s because we preload the upcoming mentorship periods mentor relationships but not the mentees. We can avoid those by using what the component already has.

```suggestion
      mentee_name = Teachers::Name.new(ect_at_school_period.teacher).full_name

File: spec/views/schools/ects/show.html.erb_spec.rb [hunk 4]
```diff
@@ -68,23 +71,188 @@
...
+            started_on: current_mentorship_period.finished_on.next_day,
+            mentee: current_ect_period,
+            mentor: future_mentor_at_school_period
+          )
+        end
+
+        it "has current mentor's full name" do
+          expect(subject).to have_summary_list_row(
+            "Mentor",
+            value: "John Smith",
```
avinhurry: False positive?

The current mentors name is `Moby Dick`

Is this passing because `:include?` is  picking up `John Smith` from the hint text?

avinhurry (approved):
Thanks for making those changes. I think there's a false positive in the tests. Other than that this looks good.

spec/views/schools/ects/show.html.erb_spec.rb
cpjmcquillan: Yeah copy-pasted exactly the thing you were talking about 😆.

I'm not completely sold in using `:eql?` and `:include?`.

At the minute I'm playing around with this. It has a reasonable default, and you can extend it to some useful cases —but it feels clunky.

What do you think?

```ruby
module HaveSummaryListRow
  class Matcher
    def initialize(key, value: "", visible: :visible, matcher: [:has_text?, { exact: true, normalize_ws: true }])
      @key = key
      @value = value
      @visible = visible
      @matcher = matcher
    end

    def matches?(page)
      @page = page
      @rows = @page.find_all("dl.govuk-summary-list dt.govuk-summary-list__key", visible: @visible)
      @matching_row = @rows.find { |it| it.text == @key }

      if @value.blank?
        @matching_row && @matching_row.text == @key
      else
        @sibling = @matching_row&.sibling("dd.govuk-summary-list__value", visible: @visible)
        @sibling && @sibling.public_send(@matcher.first, @value, **@matcher.second)
      end
    end

    # ... the rest
  end
end
```

It means you can write expectations like this if you want

```ruby
it "displays upcoming mentorship details" do
  upcoming_mentorship_detail = <<~TXT.squish
    John Smith will start mentoring Baz White
    on #{future_mentorship_period.started_on.to_fs(:govuk)}
  TXT
  expect(subject).to have_summary_list_row(
    "Mentor",
    value: ".govuk-hint",
    matcher: [:has_css?, { text: upcoming_mentorship_detail }]
  )
end
```

avinhurry: Nice, yeah I like where it's going. I guess it's that trade off of flexibility vs abstraction.

app/components/schools/ects/mentorship_component.rb
```diff
@@ -0,0 +1,44 @@
...
+    def upcoming_mentorship_descriptions
+      safe_join(upcoming_mentorships.map { upcoming_mentorship_description(it) }, tag.br)
+    end
+
+    def upcoming_mentorship_description(mentorship_period)
+      mentor_name = Teachers::Name.new(mentorship_period.mentor.teacher).full_name
+      mentee_name = Teachers::Name.new(ect_at_school_period.teacher).full_name
+      <<~TXT.squish
+        #{mentor_name} will start mentoring #{mentee_name} on
+        #{mentorship_period.started_on.to_fs(:govuk)}
```
peteryates: I think the wordiness of this makes it harder to read vs this:

Image

app/components/schools/ects/mentorship_component.html.erb
```diff
@@ -0,0 +1,11 @@
+<% if mentorship.current_mentor.blank? %>
+  <%= govuk_warning_text(text: assign_or_create_mentor_link) %>
+<% else %>
+  <%= mentorship.current_mentor_name %>
+  <% if upcoming_mentorships.any? %>
+    <br>
+    <div class="govuk-hint">
+      <%= upcoming_mentorship_descriptions %>
+    </div>
```
peteryates: I think when there are `upcoming_mentorships`, this should be a non-bulleted list really.

app/components/schools/ects/mentorship_component.rb
cpjmcquillan: Agreed! @StephenC-V1 what do you think?

app/components/schools/ects/mentorship_component.html.erb
cpjmcquillan: Yeah, I did think about that. It's unlikely, but technically possible, so I'll update it 👌🏼

app/components/schools/ects/mentorship_component.rb
StephenC-V1: Yes. I'd not seen the words for this.

peteryates (approved):
Looks good, a definite improvement.

PR #3296, opened 2026-07-28
Increase Playwright navigation timeout in feature specs

avinhurry:
## Summary

CI runners occasionally take longer than the existing 3 second timeout to complete page navigation. This causes otherwise valid feature specs to fail.

Use a 10 second timeout for navigation while keeping the 3 second default for other Playwright actions. Successful navigations continue immediately, so this doesn't slow down passing specs.

Failed examples:

- https://github.com/DFE-Digital/register-early-career-teachers-public/actions/runs/30345529769/job/90230723346
- https://github.com/DFE-Digital/register-early-career-teachers-public/actions/runs/30335249431/job/90198631492
- https://github.com/DFE-Digital/register-early-career-teachers-public/actions/runs/30280383787/job/90024962044

### DIFF
--- a/spec/support/rspec_playwright.rb
+++ b/spec/support/rspec_playwright.rb
@@ -5,6 +5,7 @@ module RSpecPlaywright
   class PlaywrightMajorVersionMismatch < StandardError; end
 
   DEFAULT_TIMEOUT = 3_000
+  DEFAULT_NAVIGATION_TIMEOUT = 10_000
   PLAYWRIGHT_CLI_EXECUTABLE_PATH = "./node_modules/.bin/playwright"
 
   def self.start_browser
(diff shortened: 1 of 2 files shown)


cpjmcquillan (approved):
I think this has been tripping me up [in this commit](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3287/changes/a556f5afc5bd7270a2327de17ec0507ee4a8becb).

Preloading to prevent the N+1 makes the query a little slower, and GitHub actions are notoriously slow (compared to most local machines), and I think the blast radius of that on the ECTs page made lots of tests fail.

This seems pragmatic, and I like that it is targeted on navigation timeouts only 👌🏼.

Is there a risk that we could miss anything that we'd otherwise catch with this change? (e.g a slow query) 🤔

PR #3297, opened 2026-07-28
[4263] Interrupt significant ECT/Mentor name change

rwrrll:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/4263

Tries to detect significant name changes as per the algorithm described in the ticket, and if detected, interrupts the "change name" flow for both ECTs and Mentors with a special panel as designed in the prototype.

### Guidance to review

I was on the fence about a feature test for this since it is kind of an edge case, however there is a distinct user journey described here that's integrated across several steps, so I settled on adding one.

This PR includes a slight refactor to #4262 in [1d8d5aa](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3297/commits/1d8d5aaf1b188759224030c388ffa0c8c193a327)

Also sorry for all the Copilot noise! This seems to have been turned on ✨magically✨ for me, thanks GitHub! It's off again now 🤓

### DIFF
--- a/app/assets/stylesheets/interruption-panel.scss
+++ b/app/assets/stylesheets/interruption-panel.scss
@@ -1,0 +1,27 @@
+@use "govuk-frontend/dist/govuk" as *;
+
+.govuk-panel--interruption {
+  background: govuk-colour("blue");
+  color: govuk-colour("white");
+  text-align: left;
+
+  .govuk-body,
+  .govuk-list,
+  .govuk-warning-text__text {
+    color: govuk-colour("white");
+  }
+
+  .govuk-warning-text__icon {
+    border-color: govuk-colour("black");
+    background: govuk-colour("black");
+    color: govuk-colour("white");
+  }
+
+  .govuk-panel__body > :last-child {
+    margin-bottom: 0;
+  }
+
+  .govuk-panel__actions {
+    margin-top: govuk-spacing(6);
+  }
+}
--- a/app/assets/stylesheets/interruption-panel.scss
+++ b/app/assets/stylesheets/interruption-panel.scss
@@ -1,27 +1,6 @@
 @use "govuk-frontend/dist/govuk" as *;
 
-.govuk-panel--interruption {
-  background: govuk-colour("blue");
-  color: govuk-colour("white");
-  text-align: left;
-
-  .govuk-body,
-  .govuk-list,
-  .govuk-warning-text__text {
-    color: govuk-colour("white");
-  }
-
-  .govuk-warning-text__icon {
-    border-color: govuk-colour("black");
-    background: govuk-colour("black");
-    color: govuk-colour("white");
-  }
-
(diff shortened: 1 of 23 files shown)


app/views/schools/confirm_name_change.html.erb
```diff
@@ -0,0 +1,44 @@
+<% page_data(
+     title: "Check this name change is for the same person",
+     header: false,
+     backlink_href: @wizard.previous_step_path
+   ) %>
+
+<% teacher_full_name = @wizard.teacher_full_name %>
+
+<%= form_with(model: @wizard.current_step, url: @wizard.current_step_path) do |f| %>
+  <div class="govuk-panel govuk-panel--interruption">
```
cpjmcquillan: FYI, we should be able to use the [`govuk_panel` component](https://govuk-components.x-govuk.org/components/panel/#interruption-panels) here, and lose the custom markup/stylesheet

rwrrll: Aha thanks I didn't know about the interruption toggle! 🙌

avinhurry: I think we may be the first public repo to use it ✨

spec/services/teachers/name_change_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    it "is true when both the first and last name change" do
+      expect(significant?("John Smith", "Mary Jones")).to be(true)
+    end
+
+    it "is false when a middle name is added" do
+      expect(significant?("John Smith", "John Paul Smith")).to be(false)
+    end
+
+    it "is false for multi-part names when the first and last words are unchanged" do
```
avinhurry:
```suggestion
    it "is false when only the last part of a multi-part name changes" do
```

spec/features/schools/ects/change/name_spec.rb
```diff
@@ -27,32 +27,52 @@
...
-    and_i_should_see_the_name("Sister Mildred")
+    and_i_should_see_the_name("Miriam Mildred")
     and_i_should_see_the_name("Miriam Margolyes")
     and_i_click_the_confirmation_button
     then_i_should_be_taken_to_the_confirmation_page
-    then_the_teacher_name_should_be_corrected
+    then_the_teacher_name_should_be("Miriam Mildred")
+  end
+
+  scenario "significant change (both first and last name) shows the interruption page" do
```
avinhurry, line 41: Agree these are worthwhile 👍🏽

avinhurry (approved):
I was wondering why copilot kept picking on you... 😆

This looks great!

app/assets/stylesheets/interruption-panel.scss
cpjmcquillan, line 1: Is this a bug in `govuk-frontend`? 🤔

cpjmcquillan (approved):
🚢

PR #3299, opened 2026-07-29
Rename ongoing_today and unfinished

craigmdavidson:
### Context

Resolves https://github.com/DFE-Digital/register-ects-project-board/issues/3765

The issue is confusing naming where:
```
ongoing returns if finished_on is nil
ongoing_today returns if the start/finish cover the current date (with nil treated as future)
```
In practice, this means:
```
when started_on: 1.week.ago, finished_on: nil - 'ongoing' and 'ongoing today'
when started_on: 1.week.from_now, finished_on: 2.weeks.from_now - neither 'ongoing' nor 'ongoing today'
when started_on: 1.week.ago, finished_on: 1.week.from_now - not 'ongoing', but is 'ongoing today' ⚠️
when started_on: 1.week.from_now, finished_on: nil - 'ongoing' but is not 'ongoing today' ⚠️
```

* Renames `ongoing_today`=> `contains_today` and renames  `ongoing`=> `unfinished`.

### Guidance to review

> [!Note]
> This is a straight simple rename refactoring, but unfortunately with quite a large blast radius.

> [!Warning]
> `contains_today?` currently and continues to ignore the `start_date`. So any Interval with a `start_date` in the future would still have `contains_today?` returning true. 

Changing the implementation of `contains_today?` would no longer be a _behaviour preserving change_, with the PR already being unacceptably large. I have left this as out of scope - to be covered in a subsequent PR.  

Concretely In **Interval** `in app/models/concerns/interval.rb`, contains_today? is implemented as.
```
  def contains_today? = finished_on.nil? || finished_on.future?
```

### DIFF
--- a/app/models/concerns/interval.rb
+++ b/app/models/concerns/interval.rb
@@ -7,7 +7,7 @@ module Interval
     validate :period_dates_validation
 
     # Scopes
-    scope :ongoing, -> { where(finished_on: nil) }
+    scope :unfinished, -> { where(finished_on: nil) }
     scope :finished, -> { where.not(finished_on: nil) }
     scope :earliest_first, -> { order(started_on: "asc") }
     scope :latest_first, -> { order(started_on: "desc") }
@@ -19,7 +19,7 @@ module Interval
     scope :finished_on_or_after, ->(date) { where(finished_on: date..) }
     scope :overlapping_with, ->(period) { where(*overlapping_with_range(period.started_on, period.finished_on)) }
     scope :containing_period, ->(period) { where(*containing_range(period.started_on, period.finished_on)) }
-    scope :ongoing_on, ->(date) { where(*date_in_range(date)) }
+    scope :contains_date, ->(date) { where(*date_in_range(date)) }
     scope :closest_to, ->(date) {
       # this scope orders records by their proximity to the given date.
       # if a record covers the date, it will be prioritized.
@@ -43,9 +43,9 @@ module Interval
     }
 
     # Date relative scopes
-    scope :ongoing_today, -> { ongoing_on(Time.zone.today) }
+    scope :contains_today, -> { contains_date(Time.zone.today) }
     scope :starting_tomorrow_or_after, -> { started_on_or_after(Time.zone.tomorrow) }
-    scope :current_or_future, -> { ongoing_today.or(starting_tomorrow_or_after) }
+    scope :current_or_future, -> { contains_today.or(starting_tomorrow_or_after) }
   end
 
   # Validations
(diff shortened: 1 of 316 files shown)


app/models/concerns/interval.rb
```diff
@@ -95,9 +95,9 @@ def leaving_today_or_in_future?
 
   def invalid_date_order? = !valid_date_order?
 
-  def ongoing? = finished_on.nil?
+  def unfinished? = finished_on.nil?
 
-  def ongoing_today? = finished_on.nil? || finished_on.future?
+  def contains_today? = finished_on.nil? || finished_on.future?
```
craigmdavidson, line 100: This should include a reference to `start_date` - but in order to keep this PR a behaviour preserving change this is out of scope of this PR.

cpjmcquillan: Is it essential this predicate mirrors the `.contains_today` scope? 🤔

Would `unfinished_today?` better reflect the behaviour here?

craigmdavidson: > Would unfinished_today? better reflect the behaviour here?

:sob:! @cpjmcquillan you are killing me! 😱 😆 

Essentially this PR is to improve clarity, so nothing here is essential, 
so the answer to the _'is it essential'_ first question is no, it is not essential AFAIK. 

It would not be unfair to say that if the goal was to leave the implementation as `finished_on.nil? || finished_on.future?` then `unfinished_today` is more accurate - but I'm not sure that that is the end goal? 

Worth noting that we have a corresponding scope named `contains_today`: 
```
scope :contains_today, -> { contains_date(Time.zone.today) }
```

TBH though, I'm not super sure what these methods are actually trying to convey? (which is perhaps damning in and of itself - I'm acting as a code monkey on this ticket). I'm interpreting these as meaning: 

* `def unfinished? = finished_on.nil?` - I interpret as `no_finish_specified?`?
* `def contains_today? = finished_on.nil? || finished_on.future?` => I interpret as `current?`

Which aligns with the scope: `scope :current_or_future, -> { contains_today.or(starting_tomorrow_or_after) }`, whose name implies `contains_today` means `current`.

That said how intention revealing the names are to a large extent informed by the callers context, which I am unfamiliar with.

So on the second question, I can't really answer that (I don't know?!), I'll need to leave that one for @ethax-ross and @peteryates !

cpjmcquillan (approved):
LGTM!

I wonder if we should rename the `ongoing_` associations and methods too once this has shipped? 🤔

app/models/concerns/interval.rb
tonyheadford: The more I'm looking at it, the more it feels like `includes_today` would be a better choice and more Ruby-ish, and chain with `unfinished.includes_today` as required

tonyheadford (approved):
An improvement I think. Well done for persevering!

PR #3301, opened 2026-07-29
Set autocomplete attributes on OTP signin pages

peteryates:
These pages are used by DfE staff to log in.

There have been reports that autocompletion isn't working in Microsoft Edge on Windows (DfE-provisioned devices).

This commit sets the autocomplete attributes for the [email](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/autocomplete#email) field in an attempt to make it work.

### DIFF
--- a/app/views/otp_sessions/request_code.html.erb
+++ b/app/views/otp_sessions/request_code.html.erb
@@ -7,7 +7,7 @@
     We’ve sent you an email with a sign in code.
   </p>
 
-  <%= f.govuk_text_field :code, width: 'one-quarter', label: { text: "Sign in code", size: "s" }, inputmode: "numeric", autocomplete:"off" %>
+  <%= f.govuk_text_field :code, width: 'one-quarter', label: { text: "Sign in code", size: "s" }, inputmode: "numeric" %>
 
   <%= f.govuk_submit "Sign in" %>
 <% end %>
--- a/app/views/otp_sessions/new.html.erb
+++ b/app/views/otp_sessions/new.html.erb
@@ -7,7 +7,7 @@
     You need to request an email with a code to sign in.
   </p>
 
-  <%= f.govuk_text_field :email, type: :email, width: 'two-thirds', label: { text: "Email address", size: "s" } %>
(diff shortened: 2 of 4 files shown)


app/views/otp_sessions/request_code.html.erb
```diff
@@ -7,7 +7,7 @@
     We’ve sent you an email with a sign in code.
   </p>
 
-  <%= f.govuk_text_field :code, width: 'one-quarter', label: { text: "Sign in code", size: "s" }, inputmode: "numeric", autocomplete:"off" %>
+  <%= f.govuk_text_field :code, width: 'one-quarter', label: { text: "Sign in code", size: "s" }, inputmode: "numeric", autocomplete:"one-time-code" %>
```
cpjmcquillan: Should there be a ` ` here?
```suggestion
  <%= f.govuk_text_field :code, width: 'one-quarter', label: { text: "Sign in code", size: "s" }, inputmode: "numeric", autocomplete: "one-time-code" %>
```

cpjmcquillan (approved):

app/views/otp_sessions/request_code.html.erb
avinhurry: What benefit are we expecting this to provide? 

The OTP arrives via an email and this attribute won't let Edge extract and autofill an emailed code 🤔

peteryates: Hmm [the MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/autocomplete#one-time-code) made it sound like it'd work:

> A one-time password (OTP) for verifying user identity that is used as an additional factor in a sign-in flow. Most commonly this is a code received via some out-of-channel mechanism, such as SMS, **email**, or authenticator application.

I've removed it and reworded.

peteryates: There should've! I removed this change as per the comment below.

avinhurry: Shouldn't we keep it off, otherwise won't it suggest values that were previously typed into that field, i.e old OTP's?

PR #3302, opened 2026-07-29
[4191] Trim previous period on same school re-registration

avinhurry:
## Summary

When a school re-registers an ECT at the same school before or on their recorded leaving date, the inclusive periods would overlap and cause a 500.

This change allows the re-registration and trims the previous ECT period to the day before the new start date, matching the existing behaviour for school transfers. Registrations are still blocked when the ECT has an ongoing period with no recorded leaving date at the same school.

### DIFF
--- a/spec/services/schools/register_ect_spec.rb
+++ b/spec/services/schools/register_ect_spec.rb
@@ -553,7 +553,7 @@ RSpec.describe Schools::RegisterECT do
       let(:training_programme) { "school_led" }
       let(:lead_provider) { nil }
 
-      before { FactoryBot.create(:teacher, trn:) }
+      let!(:teacher) { FactoryBot.create(:teacher, trn:) }
 
       it "creates a TrainingPeriod" do
         expect { service.register! }.to change(TrainingPeriod, :count).by(1)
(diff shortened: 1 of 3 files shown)


spec/services/schools/register_ect_spec.rb
```diff
@@ -580,6 +580,46 @@
...
+            :ect_at_school_period,
+            teacher:,
+            school:,
+            started_on: Date.new(2026, 7, 1),
+            finished_on: recorded_leaving_date
+          )
+        end
+
+        shared_examples "successful same-school re-registration" do
+          it "finishes the existing period and allows a new registration at the same school" do
```
thenapking, line 600: What if the ect has a training or mentorship period which ends on the recorded_leaving_date (or any date between started_on.yesterday and the recorded_leaving_date ? Won't that cause a validation error?

avinhurry: This is handled by `ECTAtSchoolPeriods::Finish`. It trims any associated training period or mentorship period spanning the new cutoff to `started_on.yesterday` before updating the parent ECT period.

[I've added a spec to demonstrate](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3302/changes/7c58f691e5c2c4bf1498725e23a66eae7dc2ef49).

thenapking (approved):
LGTM

PR #3307, opened 2026-07-30
Add initial `Statement::AuditNote`s for contract managers

tonyheadford:
### Context

Add the initial statement audit notes to the database.  These are intended to be view only and attached to a Statement where there is something noteworthy for contract managers to see.

Also to preserve paragraph spacing of the audit notes, I've added HTML `<br/>` elements to the notes and made the content `.html_safe` in the view component.

Script to add the audit notes
Render note body as HTML-safe

### Guidance to review

Example of note displayed on Statement view:
image

### DIFF
--- a/app/components/admin/statements/audit_notes_component.html.erb
+++ b/app/components/admin/statements/audit_notes_component.html.erb
@@ -3,6 +3,6 @@
 
   <% audit_notes.each do |audit_note| %>
     <p class="govuk-hint govuk-!-margin-bottom-1"><%= timestamp(audit_note) %></p>
-    <p class="govuk-body"><%= audit_note.body %></p>
+    <p class="govuk-body"><%= audit_note.body.html_safe %></p>
   <% end %>
 </div>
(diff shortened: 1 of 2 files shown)


app/components/admin/statements/audit_notes_component.html.erb
```diff
@@ -3,6 +3,6 @@
 
   <% audit_notes.each do |audit_note| %>
     <p class="govuk-hint govuk-!-margin-bottom-1"><%= timestamp(audit_note) %></p>
-    <p class="govuk-body"><%= audit_note.body %></p>
+    <p class="govuk-body"><%= audit_note.body.html_safe %></p>
```
cpjmcquillan: Could we use [`sanitize`](https://api.rubyonrails.org/classes/ActionView/Helpers/SanitizeHelper.html#method-i-sanitize) here instead?

tonyheadford: Yes, good idea, one day these might be editable

cpjmcquillan (approved):

PR #3308, opened 2026-07-30
[4270] Fix incorrect "Not Available" Lead Provider message

rwrrll:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/4270

`Teachers::Search#where_ect_at` calls `eager_load(current_or_next_ect_at_school_period: …)`, which makes the whole relation eager loading. This meant that the `.includes` the controller chained was folded it into the same `LEFT OUTER JOIN`.

Because `latest_training_period.lead_provider` was resolved through the current/next period's join chain, if an ECT had no current or next period, the joined chain would be all NULL, so the latest period's lead provider comes back NULL, which triggers the "Not available" message. The same collapse would affect `delivery_partner` and `expression_of_interest_lead_provider`.

By changing from `includes` to `preload`, the query doesn't get rolled up, and we get the desired outcome.

### Guidance to review

Check the same record mentioned in the ticket on staging; it should no longer show "Not available".

### DIFF
--- a/app/controllers/schools/ects_controller.rb
+++ b/app/controllers/schools/ects_controller.rb
@@ -12,7 +12,7 @@ module Schools
           query_string: params[:q]
         )
         .search
-        .includes(
+        .preload( # .includes folds into search query and breaks when no current/next TP
           ongoing_induction_period: :appropriate_body_period,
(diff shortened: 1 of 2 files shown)


thenapking (approved):
LGTM

avinhurry (approved):

PR #3310, opened 2026-07-30
[4030] Remove opened_on conditions

thenapking:
### Context
During manual testing of the GIAS services, we discovered that a large number of open schools have no opening date. Since this affects both historical and new schools it seems best to rely on the school's status only. I've spoken to @tonyheadford about this and he thought this best. I considered adding a method to the `GIAS::Schools::Open` service to set the `opened_on` date when it was missing. But we'd also have to backfill old dates, which we might not have appropriate data to do.

Remove the `opened_on_or_before_today?` condition on `can_be_opened?`, `can_be_replaced?`, and `can_be_merged?`.  The main changes are to the events.  The `GIAS::Schools::Replace` service now uses the date the school being replaced `closed_on`, because that is still a condition on a school being replaced.  From looking at the data I could not find instances of closed schools with `closed_on = nil`

### DIFF
--- a/app/models/gias/school.rb
+++ b/app/models/gias/school.rb
@@ -82,7 +82,7 @@ class GIAS::School < ApplicationRecord
   end
 
   def can_be_opened?
-    open_status? && opened_on_or_before_today? && school_not_yet_opened? && predecessors.empty? && successors.empty?
+    open_status? && school_not_yet_opened? && predecessors.empty? && successors.empty?
   end
 
   def can_be_replaced?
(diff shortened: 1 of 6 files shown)


app/models/gias/school.rb
```diff
@@ -116,12 +114,6 @@ def closed_on_or_before_today?
     closed_on <= Date.current
   end
 
-  def opened_on_or_before_today?
-    return false if opened_on.blank?
-
-    opened_on <= Date.current
```
avinhurry, line 122: Reckon it's worth adding a spec for `when the school has a future opening date`?

avinhurry (approved):

app/models/gias/school.rb
thenapking: I thought about that, but we are now just making this depend on status, and I think in the future that test would seem a bit redundant.

cpjmcquillan (approved):

PR #3312, opened 2026-07-30
[4030] Treat amalgamations as mergers

thenapking:
### Context

From manual testing it would appear that amalgamations can be treated exactly like mergers.  We have a small number of amalgamations which have come through the wrong way round - successors identified as predecessors and vice versa.  But outside of this it would appear that amalgamations can be treated as mergers.  There are a number of proposed mergers where the successor school has not yet opened, and both schools are in `proposed_to_close` status.  The current guards will be able to handle that context.

Simply widen the criteria we use to identify the `GIAS::SchoolLink` to include `Successor - amalgamation`.

### Guidance to review

Updated the test data so that the following code from the console can test an amalgamation.

```
gias_school = GIAS::School.find(9123502)
GIAS::Schools::Merge.new(gias_school).merge!
```

### DIFF
--- a/app/models/gias/school.rb
+++ b/app/models/gias/school.rb
@@ -101,7 +101,7 @@ class GIAS::School < ApplicationRecord
       successors.one? &&
       successor.open_status? &&
       successor.school.present? &&
-      school_being_merged?
+      school_being_merged_or_amalgamated?
   end
 
   def school_not_yet_opened?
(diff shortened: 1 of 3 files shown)


avinhurry (approved):

cpjmcquillan (commented):
You mention that _sometimes_ the links come in the wrong way round?

Is there anything we need to do to mitigate this? If we automate this, could we easily undo it if they turn out to be inverted? 🤔

PR #3315, opened 2026-07-31
Fix Staging failure on invalid session handling

craigmdavidson:
### Context

After a data refresh, a signed-in user's session points at data that no longer exists. This results in a 500 error instead of signing the user.

When a user's session points at a deleted user, school, or appropriate body, Sessions::Manager#current_user now:
- logs the issue to Sentry at info level; 
- ends the session; 
- returns nil 

so the user is signed out rather than receiving a 500 error.

### Guidance to review

We introduce the concept of an InvalidSession and use that in place of StandardError. This is actually the bulk of the modifications in this PR.

> [!Note]
> Key question, sending a Sentry notification follows our current approach - but do we want to send it to Sentry at all?

If you want to check behaviour:

1. go to the review app
2. log in as an Appropriate Body leave the page open; 
3. modify the README file, push and wait for the rebuild of the review app;
4. then go back to the page and refresh, where previously this would have been a 500, it not logs out the user.

### DIFF
--- a/app/services/sessions/manager.rb
+++ b/app/services/sessions/manager.rb
@@ -30,8 +30,7 @@ module Sessions
     end
 
     # @see Sessions::User
-    # @raise [Sessions::User::UnrecognisedType]
-    # @return [Sessions::User]
+    # @return [Sessions::User, nil]
     def current_user
       @current_user ||= load_from_session
     end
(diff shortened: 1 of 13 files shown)


app/services/sessions/manager.rb
```diff
@@ -85,9 +84,10 @@ def load_from_session
 
         record_new_activity(session_user)
       end
-    rescue ArgumentError => e
-      Sentry.capture_exception(e)
+    rescue Sessions::User::InvalidSession, ArgumentError => e
+      Sentry.capture_exception(e, level: :info)
```
ethax-ross: I think we should keep this as a Sentry error; as you mentioned at stand up if we disable it then it would inevitably become useful in the future/shoot us in the foot. That being said, if we expect these errors to happen in staging I would opt to skip over the `capture_exception` for that specific environment and add a note as to why?

ethax-ross (approved):
LGTM

cpjmcquillan (approved):
LGTM :shipit:

PR #3316, opened 2026-07-31
Introduce `Schools::ECTAtSchoolPeriods::Query` object

cpjmcquillan:
This adds another layer of abstraction between the `Schools::ECTsController` and
the complex query that filters teachers.

The intention is this will make things easier to test, and allow us to
explicitly preload associations that are relevant for this use case.

This query object also leans on [`preload`][preload] rather than
[`includes`][includes]. This means we use a separate query for each association,
rather than rolling them into a single, complex query with multiple joins.

The benefits of this are two-fold:

1. It means the latest training period associations are loaded even when there
is no current or next training period (as outlined in https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3308)
2. It is more performant in this use-case (as discussed in https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3287)

[preload]: https://api.rubyonrails.org/classes/ActiveRecord/QueryMethods.html#method-i-preload
[includes]: https://api.rubyonrails.org/classes/ActiveRecord/QueryMethods.html#method-i-includes


app/services/schools/ect_at_school_periods/query.rb
```diff
@@ -0,0 +1,46 @@
...
+  class Query
+    def initialize(school:, query:)
+      @school = school
+      @query = query
+    end
+
+    def teachers
+      teachers_search
+        .search
+        .strict_loading
```
cpjmcquillan, line 11: We don't need to keep this, but it helped me weed out which things were causing N+1s, and which things weren't.

We should be able to add unit tests similar to `API::Teachers::Query` that ensure the relevant associations are loaded.

I think it's probably worth adding some comments here too that outline _where_ these associations would otherwise be lazily loaded. 

Perhaps a better abstraction presents itself, or perhaps we will be able to see a path that helps us rework our views/components/helpers more logically. But at the least it will be easier to reason about _why_ we are pre-loading something.

app/services/schools/ect_at_school_periods/query.rb
```diff
@@ -0,0 +1,46 @@
...
+                expression_of_interest
+                expression_of_interest_lead_provider
+                delivery_partner
+              ],
+              latest_training_period: %i[
+                lead_provider
+                expression_of_interest
+                expression_of_interest_lead_provider
+                delivery_partner
+              ]
```
cpjmcquillan, line 30: For example, most of this is necessary because we have a "Lead provider" row in the `Schools::ECTs::ListingCardComponent`.

The logic behind that is basically

```
if the training period has an expression of interest **only**
-> use the expression of interest lead provider name
else if the training period has an active lead provider
-> use the active lead provider lead provider name
else if there is no training period
-> use "Not available"
```

where the `training period` is defined as

```
the ect's "current" training period
or the ect's "next" training period
or the ect's "latest" training period
```

PR #3322, opened 2026-08-03
Fix semantic headings in teacher registration forms

ericaporter:
Ticket [here](https://github.com/DFE-Digital/register-ects-project-board/issues/4085) 

### Context
Some form questions in the ECT and mentor registration journeys were styled visually as headings but were not marked up with the correct semantic heading levels.

This PR changes the heading semantics identified in the accessibility review while retaining the existing visual styling and form behaviour.

- Adds semantic h2 headings to form questions in the ECT and mentor registration journeys.
- Updates the shared teacher lookup fields so “Teacher reference number (TRN)” is an h2 and dob is an h2
- Adds h2 semantics to the ECT details confirmation and previous programme choices questions.
- The shared TRN partial also applies this improvement to the appropriate body and admin ECT lookup journeys.

### Guidance to review
- No new specs added, as agreed during sizing; these are markup-only changes to existing GOV.UK form-builder components.

### DIFF
--- a/app/views/shared/_find_teacher_fields.html.erb
+++ b/app/views/shared/_find_teacher_fields.html.erb
@@ -2,7 +2,7 @@
   f.govuk_text_field(
     :trn,
     width: 10,
-    label: { text: "Teacher reference number (TRN)", size: "m" },
+    label: { text: "Teacher reference number (TRN)", size: "m", tag: "h2" },
     hint: {
       text: "
         This unique ID is usually 7 digits long, for example 4567814. It may
--- a/app/views/shared/_find_teacher_fields.html.erb
+++ b/app/views/shared/_find_teacher_fields.html.erb
@@ -20,7 +20,7 @@
   f.govuk_date_field(
     :date_of_birth,
     width: "two-thirds",
-    legend: { text: "Date of birth", size: "m", tag: "h3" },
+    legend: { text: "Date of birth", size: "m", tag: "h2" },
     hint: { text: "For example, 31 3 1980" }
   )
 %>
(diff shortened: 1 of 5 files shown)


app/views/shared/_find_teacher_fields.html.erb
```diff
@@ -20,7 +20,7 @@
   f.govuk_date_field(
     :date_of_birth,
     width: "two-thirds",
-    legend: { text: "Date of birth", size: "m", tag: "h3" },
+    legend: { text: "Date of birth", size: "m", tag: "h2" },
```
avinhurry, line 23: Just to flag in the description it says: 

> Updates the shared teacher lookup fields so “Teacher reference number (TRN)” is an h2, while retaining “Date of birth” as an h3.

but we're changing it to a h2 here. A h2 makes sense to me semantically, but just flagging incase it was unintended.

avinhurry (approved):
These look good.

Does the same need to be done for the [mentor equivalent](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/79da893f5737b257d894f40ce67eb8abf70f32d6/app/views/schools/register_mentor_wizard/_review_mentor_details.html.erb#L34)?

cpjmcquillan (approved):

app/views/shared/_find_teacher_fields.html.erb
ericaporter: Yeah I meant to update the description when I figured the tag should be updated too!

avinhurry (approved):
Thanks Erica!

JSJohal (approved):

PR #3323, opened 2026-08-03
[4261] Remove `in_partnership` from metadata

tonyheadford:
### Context

There is no need for us to have `in_partnership` recorded in metadata as it does not need to be provider specific just a general "this school has one or more partnerships this contract period" flag.

Attempt 1 was to include the `school_partnerships` in the `API::Schools::Query` and perform the query for partnerships in the serializer. Testing this locally, it doubled the average response time for the schools endpoint.

Attempt 2 was to add a transient attribute and perform the partnerships query as an `EXISTS` sub-query.
Testing locally this was twice as quick as the serializer query, so I've gone with that approach.

Approx timings based on average of 100 queries to `/api/v3/schools`, performed locally connected to prod data (for most providers, 1 provider had a slightly lower times)

| Baseline | Serializer query | EXISTS query |
| ---- | ---- | ---- |
| 650ms | 1250ms | 850ms |

An issue with adding a transient attribute is that the serializer then depends on it being present.  Therefore something is needed to prevent the specs and shared contexts becoming painful.  To skirt around that I've added `in_partnership_override_value` to the serializer options.  This value is checked in the serializer and used in preference when present.

Alternatively we could use`data[:school]&.in_partnership` in the serializer but then that is not providing a boolean and a potentially wrong value.  

I also added a flag to the initializer of the `API::School::Query` service named `include_in_partnership_flag` to allow switching on or off adding the transient attribute.

- Adds a transient attribute `in_partnership` to the schools returned from the `API::Schools::Query` query
- Modifies the school serializer to use this value
- Adds an option to the serializer to pass a value in for `in_partnership` when used outside the controller/query - e.g. some specs
- Removes `in_partnership` from the `Metadata::SchoolContractPeriods` model and associated handlers.

### Guidance to review

I use [`hurl`](https://hurl.dev) for testing the endpoint responses using a simple script like this wrapped in a shell script to iterate through the providers and provide the token as an option to `hurl`:

```
GET http://localhost:3000/api/v3/schools
Authorization: Bearer {{token}}

[Query]
filter[cohort]: 2026
page[page]: 10
[Options]
repeat: 100
```


ethax-ross (commented):
Thanks Tony - it looks like having an attribute to calculate the `in_partnership` optimally in the query and passing it through to the serializer adds a bunch of complexity, I don't think we should go ahead with this approach 😞 

Did you have a version of this PR with attempt 1 kicking around somewhere? it would be good to have a look, I wonder if there's anything we can do to improve the performance so it's less of an impact as I expect the change would be much cleaner/simpler in general?

> Attempt 1 was to include the school_partnerships in the API::Schools::Query and perform the query for partnerships in the serializer. Testing this locally, it doubled the average response time for the schools endpoint.

PR #3329, opened 2026-08-04
Fixes to API touch model

ethax-ross:
### Context

We have been asked by lead providers to provide a list of fields that, when changed, won't always result in the `updated_at` being touched/updated in the respective serializer.

As part of this we also wanted to audit all serializer fields and ensure the touch model is up to date/consistent in case we were missing any (which we were!).

- Fixes to API touch model

Include `primary_contact_email` and `secondary_contact_email` so changes from `GIAS::School` touch `contract_period_metadata` (`SchoolSerializer`) and `school_partnerships` (`SchoolPartnershipSerializer`).

Touch `Teacher` when setting `finished_on` of any `InductionPeriod` to reflect changes in `most_recent_induction_period_end_date` for `TeacherSerializer`.

Include `trs_induction_completed_date` and `trs_induction_start_date` in `Teacher` touch model to reflect changes to `induction_end_date` and `overall_induction_start_date` in `TeacherSerializer`.

Touch `Teacher` when a `SchoolPartnership` changes the `lead_provider_delivery_partnership` to reflect changes on `delivery_partner_id` in the `TeacherSerializer`.

Add comments to serializer fields that will _not_ result in the `updated_at` being touched when they change.

- Remove conditional from touch model

We have the ability to only touch a model when it's in a very specific state. We were using this to ensure we only touch the `Teacher#api_updated_at` when the first `InductionPeriod` was created, for example (as this triggers a change
to the overall start date in the API response).

The issue with doing this is it because very hard to test; the way the tests were setup we mocked the conditional before performing a change that should trigger a touch. This worked for updates but not creation (as to mock the
conditional we have to instantate the instance, so we have a chicken/egg situation).

After reviewing how this is used and talking to Nathan we've decided to pull out the conditional updates in order to simplify the touch model and accept that we may over-touch a timestamp when nothing actually changes in the API
response. We will inform lead providers of the potential for this/when it could happen.

### DIFF
--- a/app/models/induction_period.rb
+++ b/app/models/induction_period.rb
@@ -110,9 +110,8 @@ private
 
   def touch_teacher?
     first_period = saved_change_to_id? && teacher.induction_periods.count == 1
-    outcome_changed = saved_change_to_outcome?
 
-    first_period || outcome_changed
+    first_period || saved_change_to_outcome? || saved_change_to_finished_on?
   end
 
   def at_most_one_outcome_per_teacher
--- a/app/models/induction_period.rb
+++ b/app/models/induction_period.rb
@@ -17,7 +17,7 @@ class InductionPeriod < ApplicationRecord
   belongs_to :teacher
   has_many :events
 
-  touch -> { teacher }, when_changing: %i[started_on finished_on], timestamp_attribute: :api_updated_at, if: :touch_teacher?
+  touch -> { teacher }, on_event: %i[create update], when_changing: %i[started_on finished_on outcome], timestamp_attribute: :api_updated_at
 
   refresh_metadata -> { teacher },
                    on_event: %i[create destroy update],
@@ -106,12 +106,6 @@ private
 
   def teacher_distinct_period
     overlap_validation(name: "induction")
-  end
-
-  def touch_teacher?
-    first_period = saved_change_to_id? && teacher.induction_periods.count == 1
-
-    first_period || saved_change_to_outcome? || saved_change_to_finished_on?
   end
 
   def at_most_one_outcome_per_teacher
--- a/app/models/induction_period.rb
+++ b/app/models/induction_period.rb
@@ -17,7 +17,7 @@ class InductionPeriod < ApplicationRecord
   belongs_to :teacher
   has_many :events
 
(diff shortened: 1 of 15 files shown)


app/models/school_partnership.rb
```diff
@@ -41,4 +42,15 @@ class SchoolPartnership < ApplicationRecord
...
       .order("active_lead_providers.contract_period_year DESC, school_partnerships.created_at DESC")
   }
+
+  def teachers
+    # We can't use a has-many through association here because a TrainingPeriod
+    # can have a Teacher via the ECTAtSchoolPeriod or MentorAtSchoolPeriod.
+    teacher_ids = training_periods
+        .left_joins(:ect_at_school_period, :mentor_at_school_period)
+        .select("COALESCE(ect_at_school_periods.teacher_id, mentor_at_school_periods.teacher_id)")
+        .distinct
```
craigmdavidson: Is that a smell? Should we have this as a parameterised scope on Teacher?

app/models/induction_period.rb
```diff
@@ -17,7 +17,7 @@ class InductionPeriod < ApplicationRecord
   belongs_to :teacher
   has_many :events
 
-  touch -> { teacher }, when_changing: %i[started_on finished_on], timestamp_attribute: :api_updated_at, if: :touch_teacher?
+  touch -> { teacher }, on_event: %i[create update], when_changing: %i[started_on finished_on outcome], timestamp_attribute: :api_updated_at
```
craigmdavidson: Do we want `delete` here also since we can delete induction periods?

craigmdavidson (approved):
LGTM. 
The whole touch / api_updated_at stuff is pretty confusing. It would be nice not to need it for v4? 

There are far more robust ways of expressing change than this, but that is a _much_ bigger, deeper and more fundamental discussion though.

app/models/school_partnership.rb
ethax-ross: Yeah good shout 👍

app/models/induction_period.rb
ethax-ross: Ah I didn't realise we deleted induction periods, I'll add that in 👍

PR #3330, opened 2026-08-04
Change schedule api refactoring

craigmdavidson:
### Context

Resolves https://github.com/DFE-Digital/register-ects-project-board/issues/4136

To a large extent our decision to have an anaemic domain model with procedural style service objects means there isn't really much we can do here in terms of DDD elegance. Silk purses, and sows ears as they say.

This PR does make some attempts to improve the clarity:
- It collapses the 10 or 11 custom validations to a single custom validation per attribute. This allows better "at a glance reading",
- It moves validation internal guard conditions up to be conditional validations, allows better "at a glance" reading, so we only need to drill into the detail when it is relevant.
- It now uses the teachers `ect_training_periods`, and `mentor_training_periods` instead of building new queries to do this.

### Guidance to review

There are pretty significant changes to this file, as such it might be easier to read the new file, rather than attempting to decode the diff.

This is a single file refactoring, the following options to improve simplicity are structural and wider ranging and as such have been deemed out of scope:

1. Passing objects into the service than primitives;  thus avoiding primitive obsession and it's issues.  
2. Dealing with "keeping junk in the garage"(API::Concerns::Teachers::SharedAction) 
3. Moving methods to the models, as we have explicitly ruled  that out.

### DIFF
--- a/app/services/api/teachers/change_schedule.rb
+++ b/app/services/api/teachers/change_schedule.rb
@@ -6,21 +6,10 @@ module API::Teachers
     attribute :schedule_identifier
 
     validates :schedule_identifier, presence: { message: "The property '#/schedule_identifier' must be present and correspond to a valid schedule." }
-    validates :schedule_identifier, inclusion: {
-      in: Schedule.identifiers.keys,
-      message: "Enter a valid schedule identifier."
-    }, allow_blank: true
-    validate :schedule_exists
-    validate :change_to_different_schedule
-    validate :schedule_applicable_for_ect
-    validate :schedule_applicable_for_mentor
-    validate :school_partnership_exists_if_changing_contract_period
-    validate :trainee_not_completed
-    validate :lead_provider_is_currently_training_teacher
-    validate :training_period_not_withdrawn
-    validate :no_future_training_periods_exist
-    validate :can_move_to_frozen_contract_period
-    validate :schedule_does_not_invalidate_declarations
+    validates :schedule_identifier, inclusion: { in: Schedule.identifiers.keys, message: "Enter a valid schedule identifier." }, allow_blank: true
+    validate :schedule_acceptable_for_change, if: -> { errors[:schedule_identifier].empty? }
+    validate :contract_period_acceptable_for_change, if: -> { training_period && errors[:contract_period_year].empty? }
+    validate :teacher_can_change_schedule, if: -> { training_period && errors[:teacher_api_id].empty? }
 
     def change_schedule
       return false unless valid?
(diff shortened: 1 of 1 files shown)


app/services/api/teachers/change_schedule.rb
craigmdavidson, line 3: `API::Concerns::Teachers::SharedAction` - this is not good. 

It could just as usefully be named _Common_ or _Stuff_.  It tells us nothing about what it is or what it does, we just get a bunch of surprising attributes that we were not expecting. While we haven't just stuffed our crap in the loft (a superclass), we have just stuffed it all in the garage (a grab-bag concern). This could better be two or three concerns

The goal is to make this class easier to understand, addressing this would go a long way in that.
But that is a much bigger change than the time allotted so is out of scope of this PR.

app/services/api/teachers/change_schedule.rb
```diff
@@ -6,21 +6,10 @@ class ChangeSchedule
...
-    validate :trainee_not_completed
-    validate :lead_provider_is_currently_training_teacher
-    validate :training_period_not_withdrawn
-    validate :no_future_training_periods_exist
-    validate :can_move_to_frozen_contract_period
-    validate :schedule_does_not_invalidate_declarations
+    validates :schedule_identifier, inclusion: { in: Schedule.identifiers.keys, message: "Enter a valid schedule identifier." }, allow_blank: true
+    validate :validate_schedule_identifier
+    validate :validate_contract_period_year
+    validate :validate_teacher_api_id
```
craigmdavidson, line 11: What we have done here is collapse the custom validations into one per attribute. Reducing the conceptual complexity. Initially we are concerned with _"what needs to be valid?"_ if we need to look at the specifics of what makes a given attribute valid, we can look at the validation itself.

craigmdavidson: Why do we move the guard into an explicit validate pre-condition? 

So we don't need to look into the validation method itself to understand the conditions under which it fires.

app/services/api/teachers/change_schedule.rb
```diff
@@ -6,21 +6,10 @@ class ChangeSchedule
...
-    validate :schedule_applicable_for_ect
-    validate :schedule_applicable_for_mentor
-    validate :school_partnership_exists_if_changing_contract_period
-    validate :trainee_not_completed
-    validate :lead_provider_is_currently_training_teacher
-    validate :training_period_not_withdrawn
-    validate :no_future_training_periods_exist
-    validate :can_move_to_frozen_contract_period
-    validate :schedule_does_not_invalidate_declarations
+    validates :schedule_identifier, inclusion: { in: Schedule.identifiers.keys, message: "Enter a valid schedule identifier." }, allow_blank: true
```
ethax-ross: I wonder if its worth rolling the last two `validates :schedule_identifier` into the methods as well; I know we'd be foregoing the out of the box validation helpers, but it would mean we could drop all the `if: -> { errors }` conditionals 🤔

app/services/api/teachers/change_schedule.rb
ethax-ross: I agree, its not a very helpfully named or structured concern; we have a similar one for the declaration services. I've added a ticket: https://github.com/DFE-Digital/register-ects-project-board/issues/4297

ethax-ross (approved):

PR #3333, opened 2026-08-04
Fix flaky spec

cpjmcquillan:
Since we validate a statement's deadline date when updating it to ensure it
remains in the future, this test could sometimes [fail] when we update it in the
test setup.

This reworks the test setup so there is no need to update the declaration's
underlying payment statement, and this validation error is never raised.

Now, the test passes every time.

[fail]: https://github.com/DFE-Digital/register-early-career-teachers-public/actions/runs/30911561387/job/91999333801


avinhurry (approved):
Looks good.

(a rebase should fix [your failure](https://github.com/DFE-Digital/register-early-career-teachers-public/actions/runs/30929751194/job/92061098679?pr=3333))

PR #3334, opened 2026-08-04
Remove debug screenshot from spec

avinhurry:
### Summary

Remove the unnecessary screenshot capture from the mentor registration feature spec.


thenapking (approved):
Whoops.  Definitely that was me.  I will make a note to search for screenshot calls before putting up a PR.

PR #3338, opened 2026-08-05
[3300] Remove ENABLE_APPROPRIATE_BODY_RECORDS_TO_REVIEW config

tonyheadford:
### Context

The ENABLE_APPROPRIATE_BODY_RECORDS_TO_REVIEW config flag is no longer required so we should tidy up

- Remove the ENABLE_APPROPRIATE_BODY_RECORDS_TO_REVIEW env var from the terraform configs
- Remove ENABLE_APPROPRIATE_BODY_RECORDS_TO_REVIEW from `config/application.rb`
- Remove reference from view component and specs


avinhurry (approved):
Looks good.

Can we drop this one too please: https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/.env.example#L15

cpjmcquillan (approved):

PR #3339, opened 2026-08-05
[4144] GIAS::Reconciliation::Eligibility

thenapking:

Now we are moving on to automating the GIAS Imports it makes sense to pull out the conditions on what action we can take on a school into a separate service.  Also, since the importer runs on a file of all schools, including historic schools which may have closed before this service launched, and hence never had a school record created we need to add that extra guard.  Also a little tidying up of the specs.


app/services/gias/reconciliation/eligibility.rb
cpjmcquillan, line 1: I wonder whether it might be worth breaking this down further 💭

For example, 

```ruby
module GIAS::Reconciliation
  module Eligibility
    class Open
      def initialize(gias_school)
        @gias_school = gias_school
      end

      def eligible?
        closed_status? &&
          closed_on_or_before_today? &&
          school.present? &&
          !school_closed_event_recorded? &&
          successors.empty?
      end
    end
  end
```

What do you think?

thenapking: I think if we were to break it down this far, we may as well just put that in the respective services for open/close/etc.  I was planning to use this so we could determine what to do with each school in a case statement in the Reconciliation service, which is why I wanted it kept together.

cpjmcquillan (approved):

PR #3344, opened 2026-08-06
Introduce `Schools::ECTs::TeachersQuery` object

cpjmcquillan:
This adds another layer of abstraction between the `Schools::ECTsController` and
the complex query that filters teachers.

The intention is this will make things easier to test, and allow us to
explicitly preload associations that are relevant for this use case.

This query object also leans on [`preload`][preload] rather than
[`includes`][includes]. This means we use a separate query for each association,
rather than rolling them into a single, complex query with multiple joins.

The benefits of this are two-fold:

1. It means the latest training period associations are loaded even when there
is no current or next training period (as outlined in https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3308)
2. It is more performant in this use-case (as discussed in https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3287)

There is also a specific unit-test that asserts the `lead_provider_name` is
returned correctly from results when an ECT only has a finished training period
(i.e no `current_or_next`, only `latest`). This is fixed by us using `preload`
instead of `include`. See https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3308.

[preload]: https://api.rubyonrails.org/classes/ActiveRecord/QueryMethods.html#method-i-preload
[includes]: https://api.rubyonrails.org/classes/ActiveRecord/QueryMethods.html#method-i-includes

### DIFF
--- a/spec/services/schools/ects/teachers_query_spec.rb
+++ b/spec/services/schools/ects/teachers_query_spec.rb
@@ -1,0 +1,114 @@
+RSpec.describe Schools::ECTs::TeachersQuery do
+  subject(:query) { described_class.new(school:, query_string:) }
+
+  let(:school) { FactoryBot.create(:school) }
+  let(:query_string) { nil }
+
+  let(:search_double) { instance_double(Teachers::Search) }
+
+  let!(:ect_at_school_period) do
+    FactoryBot.create(:ect_at_school_period, :unfinished, school:)
+  end
+  let!(:induction_period) do
+    FactoryBot.create(
+      :induction_period,
+      :unfinished,
+      teacher: ect_at_school_period.teacher
+    )
+  end
+  let!(:training_period) do
+    FactoryBot.create(
+      :training_period,
+      :unfinished,
+      :with_expression_of_interest,
+      :with_school_partnership,
+      ect_at_school_period:
+    )
+  end
+  let(:mentor_at_school_period) do
+    FactoryBot.create(:mentor_at_school_period, :unfinished, school:)
+  end
+  let!(:mentorship_period) do
+    FactoryBot.create(
+      :mentorship_period,
+      :unfinished,
+      mentee: ect_at_school_period,
+      mentor: mentor_at_school_period
+    )
+  end
+
+  describe "#teachers" do
+    subject(:teachers) { query.teachers }
+
+    it "calls `#search` on the `Teachers::Search` service with the correct parameters" do
+      allow(Teachers::Search)
+        .to receive(:new)
+        .with(ect_at_school: school, in_progress: true, query_string:)
+        .and_return(search_double)
+      allow(search_double).to receive(:search).and_return(Teacher.all)
+
+      teachers
+
+      expect(search_double).to have_received(:search)
+    end
+
+    describe "preloading associations" do
+      subject(:result) { teachers.first }
+
+      it { expect(result.association(:ongoing_induction_period)).to be_loaded }
+      it { expect(result.ongoing_induction_period.association(:appropriate_body_period)).to be_loaded }
+      it { expect(result.association(:current_or_next_ect_at_school_period)).to be_loaded }
+      it { expect(result.current_or_next_ect_at_school_period.association(:school)).to be_loaded }
+      it { expect(result.current_or_next_ect_at_school_period.association(:school_reported_appropriate_body)).to be_loaded }
+      it { expect(result.current_or_next_ect_at_school_period.association(:current_or_next_mentorship_period)).to be_loaded }
+      it { expect(result.current_or_next_ect_at_school_period.current_or_next_mentorship_period.association(:mentor)).to be_loaded }
+      it { expect(result.current_or_next_ect_at_school_period.current_or_next_mentorship_period.mentor.association(:teacher)).to be_loaded }
+      it { expect(result.current_or_next_ect_at_school_period.association(:current_or_next_training_period)).to be_loaded }
+      it { expect(result.current_or_next_ect_at_school_period.current_or_next_training_period.association(:lead_provider)).to be_loaded }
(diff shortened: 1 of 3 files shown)


app/helpers/ect_helper.rb
rwrrll, line 1: I don't think the changes here to explicitly refer to root namespace are necessary because they're not lexically inside `module Schools`

app/services/appropriate_bodies/close_induction.rb
```diff
@@ -77,15 +77,15 @@ def finish_ect_period_today
       return unless ect_at_school_period
       return if ect_at_school_period.finished_on.present?
 
-      ECTAtSchoolPeriods::Finish.new(ect_at_school_period:, finished_on: Time.zone.today, author:, record_event: false).finish!
+      ::ECTAtSchoolPeriods::Finish.new(ect_at_school_period:, finished_on: Time.zone.today, author:, record_event: false).finish!
```
rwrrll: Same re: root namespace as it's not lexically inside `module Schools`

app/controllers/appropriate_bodies/tasks/base_details_controller.rb
```diff
@@ -13,7 +13,7 @@ def index
     private
 
       def filtered_ect_at_school_periods(query_string)
-        ECTAtSchoolPeriods::TextSearch.new(initial_ect_at_school_periods, query_string:).search
+        ::ECTAtSchoolPeriods::TextSearch.new(initial_ect_at_school_periods, query_string:).search
```
rwrrll: Same re: root namespace as it's not lexically inside `module Schools`

db/scripts/close_teacher_periods.rb
```diff
@@ -9,7 +9,7 @@ def close_teacher_periods(trns:, author_email:, finished_on: Date.current)
 
       # ect_at_school_periods + associated periods
       teacher.ect_at_school_periods.unfinished.each do |period|
-        ECTAtSchoolPeriods::Finish.new(ect_at_school_period: period, finished_on:, author:).finish!
+        ::ECTAtSchoolPeriods::Finish.new(ect_at_school_period: period, finished_on:, author:).finish!
```
rwrrll: Same re: root namespace as it's not lexically inside `module Schools`

lib/support/ect_withdrawal_history_correction.rb
```diff
@@ -35,7 +35,7 @@ def correct!
       ActiveRecord::Base.transaction do
         move_withdrawal_details!
 
-        ECTAtSchoolPeriods::Finish.new(
+        ::ECTAtSchoolPeriods::Finish.new(
```
rwrrll: Same re: root namespace as it's not lexically inside `module Schools`

app/services/schools/ect_at_school_periods/query.rb
```diff
@@ -0,0 +1,46 @@
+module Schools::ECTAtSchoolPeriods
+  class Query
```
rwrrll: It looks like this query class returns `Teacher` records, not `ECTAtSchoolPeriod` records, and it's also a layer on top of `Teachers::Query`, so it feels to me like maybe it belongs in `Teachers` service namespace.

Maybe it could be e.g. `Teachers::SchoolECTsQuery` or `Teachers::ECTsAtSchoolQuery` or something? This would also avoid the naming collision.

WDYT?

app/services/schools/ect_at_school_periods/query.rb
```diff
@@ -0,0 +1,46 @@
...
+                expression_of_interest
+                expression_of_interest_lead_provider
+                delivery_partner
+              ]
+            }
+          ]
+        )
+    end
+
+    def teachers_count
```
rwrrll: I wonder if this name is potentially a bit misleading. It excludes `query_string` because it's the "has this school registered any ECTs at all" number for the empty state, but sitting next to teachers it reads as the count of teachers. Perhaps an alternative like `total_teachers_count` or `any_teachers?` could make it a touch clearer...

spec/services/schools/ects/teachers_query_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+    FactoryBot.create(
+      :induction_period,
+      :unfinished,
+      teacher: ect_at_school_period.teacher
+    )
+  end
+  let!(:training_period) do
+    FactoryBot.create(
+      :training_period,
+      :unfinished,
```
rwrrll, line 22: By creating an `:unfinished` training period, I think this will bypass the thing that #3308 fixes, because the current training period will be present even if the query is eager loaded using `includes` and rolls up the JOIN. The 18 `association(:x)).to be_loaded` examples would all pass with `includes`. We could exercise that using a finished training period.

rwrrll (commented):
Noted a few things I spotted, in case they want a look!

app/helpers/ect_helper.rb
cpjmcquillan: Yeah that's fair enough, I just got lazy and replaced them all 😆

app/services/schools/ect_at_school_periods/query.rb
cpjmcquillan: Yeah, good point! I'll change that 👌🏼

app/services/schools/ect_at_school_periods/query.rb
cpjmcquillan: I'm not sure on this one. I chose `Schools::ECTAtSchoolPeriods` to loosely align it with its usage in `Schools::ECTsController` (which returns Teachers that are ECTs). It all feels a bit murky, though.

What about `Schools::ECTs::TeachersQuery`? 🤔

rwrrll (approved):

peteryates (approved):
I think this is a big step in the right direction and it's much better having all those `includes` out of the controller.

I do wonder if we should rethink how the record retrieval/search objects are combined eventually, as they're all individual and are easy to compose or merge.

PR #3345, opened 2026-08-06
first_pass_orphaned_fixes

peteryates:
### Context

A number of ASPs and TPs were deleted as part of ticket #4163.

We have been notified that one of the ASPs that was deleted had a TP attached which had 4 declarations. The declarations are still in the database, but the TP associated is not present.

Further investigation has found a total of 8 declarations linked to trianing_period_ids that were not present in the system.

This PR has the patch csv to create ASPs & TPs were necessary and then attach the orphaned declarations to.

### Guidance to review

* Ensure the ASPs & TPs are appearing in the database as expected
* Check there are 0 declarations linked to training periods which are not present


tonyheadford (approved):
Tested and verified on migration

PR #3346, opened 2026-08-06
[4270] Fix Change LP journey when no current/next training period

rwrrll:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/4270#issuecomment-5176404565 (second, separate issue discovered during product review)

Similar to #3308 (but in a totally different code path), if an ECT has no current/next training period, the Change LP resolver would fail as documented in @mason-emily comment in that ticket. This PR changes the resolver to consider any past training period within the same AtSchoolPeriod, rather than only withdrawn/deferred ones.

This makes it work the same way as the fix in #3308 to display a current lead provider.

### DIFF
--- a/app/services/ect_at_school_periods/change_lead_provider_resolver.rb
+++ b/app/services/ect_at_school_periods/change_lead_provider_resolver.rb
@@ -5,7 +5,7 @@ module ECTAtSchoolPeriods
     end
 
     def call
-      current_training_lead_provider || withdrawn_or_deferred_lead_provider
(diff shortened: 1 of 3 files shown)


cpjmcquillan (approved):
I wonder whether we could have an association that returned the "applicable" training period for an ECT/Mentor? (there's a better name than that I'm sure)

I think we have this "current or latest" logic scattered across the application. The scope might be a little complex, but if we can get it working it would be a pretty solid entrypoint for all of this stuff IMO.

What do you think?

PR #3347, opened 2026-08-06
Introduce product team user type main

ericaporter:
### Context
A product_team role is needed so access to the patch tool can be restricted to product-team members when it is built. This pull request adds the role, associated permissions, and a DfE persona for testing. The patch tool itself will be built separately.

- Add product_team to the PostgreSQL dfe_role_type enum
- Add the role to the Rails User model and validation
- Allow product-team users to access user management
- Add an Inspector Gadget DfE persona for testing the role
- Update factories to include product-team users
- Expose product_team? through both DfE session-user wrappers
- Include the new role in user-management options
- Add and update the relevant tests

### Guidance to review
- Confirm the database migration adds the new PostgreSQL enum value
- Confirm product_team is available when creating or editing users
- Confirm the Inspector Gadget persona can sign in as a product-team user
- Confirm both DfE session wrappers expose product_team?
- Confirm existing roles and permissions remain unchanged


thenapking (commented):
This looks fine so far.  I can change a user on the review app to a product team user.  However there is no DfE persona for the product team, so you can't test things as a member of the product team easily.  Also product team users lack the access to the User menu to be able to manage users.  I expected they would have that.  I think we could handle access to the patch tool in this PR, since that relates to the user permissions.  All we'd need to do is add a basic route to an empty form, a basic controller, and a menu item.  Then we can set the permissions correctly for the user, and that would wrap this up.

cpjmcquillan (approved):

db/schema.rb
```diff
@@ -29,7 +29,7 @@
   create_enum "declaration_payment_statuses", ["no_payment", "eligible", "payable", "paid", "voided"]
   create_enum "declaration_types", ["started", "retained-1", "retained-2", "retained-3", "retained-4", "completed", "extended-1", "extended-2", "extended-3"]
   create_enum "deferral_reasons", ["bereavement", "long_term_sickness", "parental_leave", "career_break", "other"]
-  create_enum "dfe_role_type", ["admin", "user_manager", "finance"]
+  create_enum "dfe_role_type", ["admin", "user_manager", "finance", "product_team"]
```
peteryates, line 32: Can `product_team` members do everything that the ones on the left can do, plus extra stuff?

Or are there things finance can do that product team can't and vice versa?

peteryates: Just thinking about it - maybe blazer should be a `product_team` only thing?

cpjmcquillan: Yeah I think that's the plan. Agree re-blazer. Mission Control too, probably!

avinhurry (approved):
Looks good.

Do you think it's worth adding a test here too?

https://github.com/DFE-Digital/register-early-career-teachers-public/blob/3970aa92e71380f41190c16413498a4cea35d319/spec/views/admin/users/edit.html.erb_spec.rb#L34-L38

db/schema.rb
ericaporter: For a separate PR probably?

app/assets/images/personas/inspector-gadget.png
thenapking, line 1: Love it!

thenapking (approved):
LGTM

PR #3348, opened 2026-08-06
[3260] Refine DQT importer for second round- #3332

peterdavidhamilton:
### Context

Planned bulk import of data that predates the service's launch. This PR is in response to analysis of the data after having been applied in a production-like environment. 

Refine the importer scripts:

- only curtail **_LA_** AB periods if the period extends past `2024-08-31`
- reduce induction period manipulation events by 1k as a result
- add CSV generation to capture skipped inductions for analysis
- improve resource management with more batching
- exclude all existing teachers with induction period data
- log every rejection with triage info
- script validation check

### Guidance to review

Work is now being done on the CSV files and we plan to iterate and rerun on migration until we are happy.

https://www.register-early-career-teachers.education.gov.uk/admin/blazer/queries/476-dqt-import-induction-period-changes

### DIFF
--- a/db/scripts/export_invalid_records.rb
+++ b/db/scripts/export_invalid_records.rb
@@ -1,0 +1,55 @@
+# 🫠 Wonkiness checker (use on production data but not on the production deployment)
+#
+# Example:
+#   rails runner db/scripts/export_invalid_records.rb induction_periods induction_extensions
+#
+# Run in-service model validations to help triage data
+
+if Rails.env.production?
+  $stdout.puts "Query production data using a database backup."
+  exit 0
+end
+
+if ARGV.empty?
+  $stdout.puts "No tables specified."
+  exit 0
+end
+
+batch_size = 1_000
+headers = %w[table_name record_id errors]
+date = Time.zone.today.iso8601
+filename = Rails.root.join("tmp", "invalid-records-#{date}.csv")
+results = []
+
+def validate(model, batch_size, &block)
+  model.find_in_batches(batch_size:) do |batch|
+    batch.each(&block)
+  end
+end
+
+ARGV.each do |table_name|
+  model = table_name.singularize.camelize.constantize
+
+  validate(model, batch_size) do |record|
+    next if record.valid?
+
+    errors = record.errors.full_messages.join("; ")
+
+    results << {
+      table_name: model.table_name,
+      record_id: record.id,
+      errors:,
+    }
+  end
+end
+
+if results.empty?
+  $stdout.puts "No invalid records found."
+  exit 0
+end
+
+CSV.open(filename, "w", write_headers: true, headers:) do |csv|
+  results.each { |row| csv << row.values }
+end
+
+$stdout.puts "Exported #{results.size} invalid record(s) to #{filename}"
--- a/lib/appropriate_bodies/importers/induction_period_parser.rb
+++ b/lib/appropriate_bodies/importers/induction_period_parser.rb
@@ -226,7 +226,7 @@ module AppropriateBodies::Importers
           if row.trn.nil? || row.legacy_appropriate_body_id.nil?
             log_error("cannot be imported because TRN or AB is missing",
                       trn: row.trn,
-                      dqt_id: row.legacy_appropriate_body_id)
+                      dqt_id: [row.legacy_appropriate_body_id, nil])
           else
             false
           end
@@ -235,7 +235,7 @@ module AppropriateBodies::Importers
           if row.started_on.nil?
             log_error("cannot be imported because started_on is nil",
                       trn: row.trn,
-                      dqt_id: row.legacy_appropriate_body_id)
+                      dqt_id: [row.legacy_appropriate_body_id, nil])
           else
             false
           end
@@ -244,7 +244,7 @@ module AppropriateBodies::Importers
           if row.finished_on.nil?
             log_error("cannot be imported because finished_on is nil",
                       trn: row.trn,
-                      dqt_id: row.legacy_appropriate_body_id)
+                      dqt_id: [row.legacy_appropriate_body_id, nil])
           else
             false
           end
@@ -253,7 +253,7 @@ module AppropriateBodies::Importers
           if row.started_on == Date.new(1, 1, 1)
             log_error("cannot be imported because started_on is 0001-01-01",
                       trn: row.trn,
-                      dqt_id: row.legacy_appropriate_body_id)
+                      dqt_id: [row.legacy_appropriate_body_id, nil])
           else
             false
           end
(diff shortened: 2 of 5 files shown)


tonyheadford (approved):
LGTM

db/scripts/export_invalid_records.rb
```diff
@@ -0,0 +1,55 @@
...
+  exit 0
+end
+
+batch_size = 1_000
+headers = %w[table_name record_id errors]
+date = Time.zone.today.iso8601
+filename = Rails.root.join("tmp", "invalid-records-#{date}.csv")
+results = []
+
+def validate(model, batch_size, &block)
```
ethax-ross, line 24: I had to do a double take on this method as the name is a bit misleading/it doesn't do any validation 🤔 I think you're just re-implementing `find_each(batch_size: 1_000)` 🙂

db/scripts/export_invalid_records.rb
```diff
@@ -0,0 +1,55 @@
...
+
+ARGV.each do |table_name|
+  model = table_name.singularize.camelize.constantize
+
+  validate(model, batch_size) do |record|
+    next if record.valid?
+
+    errors = record.errors.full_messages.join("; ")
+
+    results << {
```
ethax-ross, line 38: If performance is an issue you could stream it to CSV to avoid holding the records in memory; it may not be but just going off this PR comment:

> improve resource management with more batching

db/scripts/export_invalid_records.rb
```diff
@@ -0,0 +1,55 @@
...
+results = []
+
+def validate(model, batch_size, &block)
+  model.find_in_batches(batch_size:) do |batch|
+    batch.each(&block)
+  end
+end
+
+ARGV.each do |table_name|
+  model = table_name.singularize.camelize.constantize
```
ethax-ross, line 31: You could simplify this if the script takes model names instead of table names, so:

```
rails runner db/scripts/export_invalid_records.rb InductionPeriod InductionExtension
```

lib/appropriate_bodies/importers/induction_period_parser.rb
```diff
@@ -198,16 +195,15 @@ def note_event(data:, heading:, body:)
...
-                  :all_dqt_uuids,
-                  :trns_already_persisted_with_inductions
+                  :la_dqt_uuids
 
     def initialize(data_csv:, logger: nil)
       @data_csv = data_csv
       @offshore_dqt_uuids = OFFSHORE_DQT_UUIDS.to_set
-      @all_dqt_uuids = AppropriateBodyPeriod.pluck(:dqt_id).to_set
-      @trns_already_persisted_with_inductions = Teacher.where(id: UNWANTED_TEACHER_IDS).pluck(:trn).to_set
+      @la_dqt_uuids = AppropriateBodyPeriod.local_authority.pluck(:dqt_id).to_set
```
ethax-ross, line 203: `@local_authority_dqt_uuid`? 😅 I saw the abbreviation in the PR description but didn't clock what it was until I got here

lib/appropriate_bodies/importers/induction_period_parser.rb
```diff
@@ -473,14 +466,33 @@ def periods_by_trn
...
+                      finished_on: [sibling.finished_on, current.finished_on]
+                    )
                   end
 
                 end
               end
           end
 
-          keep = keep.reject { |edited_period| edited_period.length <= 1 }
+          keep = keep.reject do |edited_period|
```
ethax-ross, line 482: Is there a more descriptive variable name we can use instead of `keep`? what/why do we keep them?

lib/appropriate_bodies/importers/teacher_induction_importer.rb
```diff
@@ -23,41 +23,50 @@ class TeacherInductionImporter
...
+    # @return [Set<String>]
+    attr_reader :existing_trns_without_inductions
+    # @return [Set<String>]
+    attr_reader :importing_trns_with_inductions
     # @return [Logger]
     attr_reader :logger
 
     def initialize(teachers_csv:, induction_period_csv:, logger: nil)
+      @existing_trns = Teacher.pluck(:trn).to_set
+      @existing_trns_without_inductions = Teacher.left_joins(:induction_periods).where(induction_periods: { id: nil }).pluck(:trn).to_set
```
ethax-ross, line 68:
```
Teacher
  .where.missing(:induction_periods)
  .pluck(:trn)
  .to_set
```

spec/lib/appropriate_bodies/importers/teacher_induction_importer_spec.rb
```diff
@@ -105,18 +136,18 @@
 
       expect { induction_importer.import! }.not_to raise_error
 
-      expect(Teacher.count).to eq(589_839) # On prod where some already exist we will see 588_533 imported (1_306 delta)
+      expect(Teacher.count).to eq(589_988) # On prod where some already exist we will see 588_533 imported (1_306 delta)
```
ethax-ross, line 139: It took me a second to figure out why these are so high; so presumably we deploy this to prod, add in the CSVs and run the specs to check the data?

ethax-ross (approved):
I'm not that familiar with this aspect of the app, but from what I understand it looks good 🙂  left some minor comments/suggestions but nothing major. Feel free to take it or leave it if this is temporary code anyway 👍

lib/appropriate_bodies/importers/induction_period_parser.rb
peterdavidhamilton: This was what Pete's original scripts had, and is consistent for all the types. After a few days, I really liked the simplicity of it.

spec/lib/appropriate_bodies/importers/teacher_induction_importer_spec.rb
peterdavidhamilton: No this was intended for local runs, it only fires for a dev that has downloaded and positioned the files. It gave me a way to automate and benchmark with the caveat that the numbers WILL be different on prod where teacher records already exist.

db/scripts/export_invalid_records.rb
peterdavidhamilton: I know but there was appealing about reading it lowercase and plural

db/scripts/export_invalid_records.rb
peterdavidhamilton: Yeah I was thinking I might repackage this in-service with bells and whistles but stopped for time. 

I left it in though because I think having something we can schedule or trigger on migration would be a nice addition but its what comes after? The CSV is hard to share so I was wondering what other options we have.

Could we, for example, introduce an empty table capable of holding any "invalid record" output and let the team poke it via Blazer? CSVs are still handy and I'm sure @boonali might demand them.

But a nightly recurring job on migration that spins over the important tables and leaves you with a quick list you can skim would be interesting. With Blazer we can link out to records using the smart fields I added.

PR #3352, opened 2026-08-07
[4144] GIAS Reconciliation Service

thenapking:
### Context

This adds a service to reconcile GIAS schools after import, and cleans up the importer.  At this stage we will not append `GIAS::Reconcile` to the import job, and will run it manually for a few days in lieu of product review.  Once we're happy we will remove the TODO and link it up.

1) Add the Reconcile service.  This checks first for all replacements.  This avoids subsequent reconciliation attempts on successor schools which would lead to an error being logged claiming the school could not be reconciled.  Then we check for openings.  This means if schools are being merged into a new school, it will have been opened before we check for merges.  Now we check for merges, similarly successors are now removed from the list of URNs to reconcile to avoid mistaken error logs.  Finally we check for closures.  This means that schools which are closed after merging or replacement will have already been removed from the list to reconcile and again will prevent errors being raised mistakenly.

2) Remove the auto-create schools feature from the importer.  This is handled by `GIAS::Reconciliation::Open` service
3) Alter the rake tasks so we can run import and reconcilation separately or together as required.
4) Clean up the importer class, remove unused methods and make everything else private.
5) Fix issue with predecessor links - these were the wrong way round
6) In `GIAS::Reconciliation::Eligibility` we should check not only for explicit `Predecessor` links but also whether there's a `Successor` link where the school is being linked to.  This means if successor/predecessor links don't appear in the data feed in pairs as expected we still capture the relationship in the guard clauses.

NB As per @tonyheadford feedback we should leave the functionality to import all the GIAS schools in a fresh DB.  I've clarified the naming a little here, as the introduction of the `GIAS::Reconcile`  service implies that `import_only` means "don't reconcile", rather than "import to a fresh DB, where we don't need to check imported records against existing records".

### Guidance to review
The rake tasks will pull new CSV files and run the reconciliation, but the ENV vars must be set on the review app first.  On a blank DB the first import can take a couple hours - so I wouldn't do this.  I have been pulling daily updates and testing this locally.  The importer and reconciliation services run with in a few minutes.


lib/tasks/gias.rake
```diff
@@ -1,25 +1,70 @@
...
+    if args[:urns].blank?
+      logger.warn "No URNs provided. Usage: rake gias:reconcile[12345,67890]"
+      next
+    end
+
+    urns = args.fetch(:urns)
+      .split(",")
+      .map(&:strip)
+      .map(&:to_i)
+
```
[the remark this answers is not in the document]
thenapking: There is no school with URN 0, so this point is moot.

app/services/gias/reconciliation/eligibility.rb
```diff
@@ -9,14 +9,17 @@ def can_be_closed?
         closed_on_or_before_today? &&
         school.present? &&
         !school_closed_event_recorded? &&
-        successors.empty?
+        successors.empty? &&
+        not_a_predecessor?
```
ericaporter, line 13: Looking at the AC, it mentions it can be automated when there are no GIAS links - do we need to include all links here inc. "other"?

thenapking: Good question.  Let me have a look at the data and see if those things are ever present.  My understanding of the AC was that by no links it meant "no predecessor/successor links".

app/services/gias/reconciliation/eligibility.rb
ericaporter: I know this isn't part of this PR but I was trying to understand 1:1 merges... do you think it would be helpful to check for having only one predecessor? I think the AC asks 'the open successor to have only this closed school as its predecessor' I had a look through the other PRs to see if this check is covered elsewhere but couldnt spot it

app/services/gias/importer.rb
```diff
@@ -54,72 +76,36 @@ def parse_school_link_row(row)
...
-      # It's more efficient to skip metadata during import and refresh it
-      # all in background jobs at the end when creating a lot of schools.
-      DeclarativeUpdates.skip(:metadata) do
-        import_schools
-        import_school_links
-      end
-
-      Metadata::Handlers::School.refresh_all_metadata!(async: true)
+    def needs_reconciliation?(link, imported_link_type)
+      link.link_type != imported_link_type || link.previously_new_record? || link.link_date == Date.current
```
ericaporter, line 88: Could future dated links be excluded here unintentionally? If reconciliation doesn’t run on the exact link date will they be selected again?

app/services/gias/reconciliation/eligibility.rb
thenapking: That is a good point! There are scenarios where two schools merge into one, and we will probably need to treat that sequentially as two 1-1 merges that follow each other.

app/services/gias/importer.rb
thenapking: We're expecting this to run every day.  If we didn't include this filter, reconciliation would technically fail for schools who have a change coming in the future.  This will log an error in Sentry to be investigated, which will be a waste of time. We might need to address retries and failures though.

spec/services/gias/reconcile_spec.rb
```diff
@@ -0,0 +1,183 @@
...
+        end
+
+        it "replaces the replaced schools" do
+          expect(GIAS::Reconciliation::Replace).to receive(:replace!).with(replaced_gias_school)
+
+          service
+        end
+
+        it "does not open the successor schools of replaced schools" do
+          expect(GIAS::Reconciliation::Open).not_to receive(:open!).with(successor_gias_school_1)
```
ericaporter, line 67: Could we include the replacement successor URN in urns so this verifies that a successful replacement removes it before open?

spec/services/gias/reconcile_spec.rb
```diff
@@ -0,0 +1,183 @@
...
+        end
+
+        it "merges the merged schools" do
+          expect(GIAS::Reconciliation::Merge).to receive(:merge!).with(merged_gias_school)
+
+          service
+        end
+
+        it "does not open the successor schools of merged schools" do
+          expect(GIAS::Reconciliation::Open).not_to receive(:open!).with(successor_gias_school_2)
```
ericaporter, line 79: Could we include the successor urn in the urns so this verifies the intended behaviour?

app/models/gias/school.rb
```diff
@@ -23,11 +23,11 @@ class GIAS::School < ApplicationRecord
   has_many :contract_period_metadata, class_name: "Metadata::SchoolContractPeriod", through: :school
   has_many :gias_school_links, class_name: "GIAS::SchoolLink", foreign_key: :urn, dependent: :destroy, inverse_of: :from_gias_school
 
-  has_many :successor_links,   -> { where(link_type: GIAS::SchoolLink::SUCCESSOR_LINK_TYPES) },   class_name: "GIAS::SchoolLink", foreign_key: :urn, primary_key: :urn
+  has_many :successor_links,   -> { where(link_type: GIAS::SchoolLink::SUCCESSOR_LINK_TYPES) }, class_name: "GIAS::SchoolLink", foreign_key: :urn, primary_key: :urn
   has_many :predecessor_links, -> { where(link_type: GIAS::SchoolLink::PREDECESSOR_LINK_TYPES) }, class_name: "GIAS::SchoolLink", foreign_key: :urn, primary_key: :urn
 
   has_many :successors, class_name: "GIAS::School", through: :successor_links, source: :to_gias_school
-  has_many :predecessors, class_name: "GIAS::School", through: :predecessor_links, source: :from_gias_school
+  has_many :predecessors, class_name: "GIAS::School", through: :predecessor_links, source: :to_gias_school
```
cpjmcquillan, line 30: This is a little confusing IMO.

Are we using the `from_gias_school` association at all?

My understanding was that a `GIAS::SchoolLink` record mapped a predecessor `GIAS::School` to a successor `GIAS::School`.

Can you elaborate a little on what the issue was, and why this change fixes it? (Or maybe you have, and I've missed it!)

ericaporter (approved):
Overall, this looks good and the approach looks sound - the ordering of the reconciliation steps seems sensible! I left a few small queries/comments. Since this is quite a complex area - you might want another pair of eyes from someone familiar with the GIAS link data and the expected reconciliation behaviour...

app/models/gias/school.rb
thenapking: Yes.  This is confusing, and had confused me.  It was only when I started working with the data more that I realised I had this wrong.  A test I setup in this PR failed because it didn't link correctly to the predecessors, hence fixing this here.  A successor link does indeed connect a predecessor to a successor, and so the inverse relationship could be thought of as a predecessor link.  That's not what GIAS does.  They have two rows in the table instead, one for each direction.  As we know the data is unreliable, and sometimes we only get one link record, usually a successor link.

**How this is meant to work**
Suppose we have school A (urn: 12345) being replaced by school B (urn: 54321).  Then there will be a GIAS School link like this in the database:
```
{ urn:12345, link_urn: 54321, link_type: 'Successor' }
```
If there is a predecessor link it would be will be:
```
{ urn: 54321, link_urn: 12345, link_type: 'Predecessor' }
```

In the `GIAS::SchoolLink` model we have
```
  belongs_to :from_gias_school, class_name: "GIAS::School", foreign_key: :urn, primary_key: :urn, inverse_of: :gias_school_links
  belongs_to :to_gias_school, class_name: "GIAS::School", foreign_key: :link_urn, primary_key: :urn
```
So `from_gias_school` ties the `GIAS::School` to all its links.  The `to_gias_school` relation tells us what it is linked to, which is why it should be the source in both of those associations on `GIAS::School`.  

**Inverse successors or predecessors?**
What's confusing is that we could derive the predecessors only from the successor links by using `from_gias_school` like this
```
has_many :predecessors, class_name: "GIAS::School", through: :inverse_successor_links, source: :from_gias_school
has_many :inverse_successor_links, 
  -> { where(link_type: GIAS::SchoolLink::SUCCESSOR_LINK_TYPES) },
  class_name: "GIAS::SchoolLink",
  foreign_key: :link_urn,
  primary_key: :urn
```
I considered adding this to the model, but imo it would have only increased the confusing nature of this (and the naming of `inverse_successor_links` is mind-bending).  Since we can't guarantee predecessor/successor links will always come in pairs, I instead added this to the eligibility class:
```
    def not_a_successor?
      !GIAS::SchoolLink.where(to_gias_school: gias_school, link_type: GIAS::SchoolLink::SUCCESSOR_LINK_TYPES).exists?
    end

    def not_a_predecessor?
      !GIAS::SchoolLink.where(to_gias_school: gias_school, link_type: GIAS::SchoolLink::PREDECESSOR_LINK_TYPES).exists?
    end
```

So now, for instance, to open a school we check not only that `successors.empty?`, but that it is `not_a_predecessor?` of any other school.

**Refactor potential**
I think there's an argument for discarding predecessor links entirely and relying only on successor links.  We could consider that as a refactor, I think we'd also need to inspect the data as well.

spec/services/gias/reconcile_spec.rb
thenapking: I've realised that running `open!` before `merge!` doesn't actually achieve what I expected, viz that when two schools are amalgamating into a new school that opens on the same day, the new school will be opened first.  `open!` won't open that school because it has predecessors.  Therefore, `merge!` must ensure the successor has a school record, and we should remove that check from the eligibility.  We should then run `open!` after `merge!`.  I've added these URNs to the list and it now works.

app/services/gias/reconciliation/eligibility.rb
thenapking: I've thought about this more and I think the AC probably reflects an understanding that we'd work out how to deal with more complex merges, after getting 1-1 merges to work.  However, when several schools are merging into one, we simply merge on each predecessor.  So we don't want to add a check that there's only one predecessor.  I've added an integration test to `reconcile_spec.rb` which asserts that merge is called for each predecessor, and when you have a teacher mentoring at all schools involved (the two predecessors and a successor), their `mentor_at_school_period` is handled correctly.  I've picked an example where one merge would result in two separate periods, but the additional predecessor overlaps those and therefore causes everything to be merged into one period.

app/services/gias/reconciliation/eligibility.rb
thenapking: I checked out the other kinds of links. "Other", "Closure", "Merged" and "Expansion" were hardly used and haven't been used since before 2020.  "Children's Centre Link" and "Sixth Form Centre Link" are used, and I will find out about those separately.

app/services/gias/reconciliation/eligibility.rb
ericaporter: Ok makes sense. Might be worth leaving a brief comment on the ticket to confirm many-to-one merges are intentionally included.

app/services/gias/reconciliation/eligibility.rb
ericaporter: Ok, makes sense, thanks for checking!

PR #3353, opened 2026-08-07
[4261] (part 2) Remove `in_partnership` from metadata

tonyheadford:
### Context

This PR adopts the original approach (reworked a bit) to get the value for `in_partnership` directly in the serializer and not require it to be managed via metadata

- Add `school_partnership` includes to `API::Schools::Query`
- Check for any school partnerships in place for the given contract period year in the serializer
- Remove in_partnership from metadata models and services

### DIFF
--- a/app/models/metadata/school_contract_period.rb
+++ b/app/models/metadata/school_contract_period.rb
@@ -15,10 +15,9 @@ module Metadata
 
     validates :school, presence: true
     validates :contract_period, presence: true
-    validates :in_partnership, inclusion: { in: [true, false] }
     validates :induction_programme_choice, inclusion: { in: induction_programme_choices.keys }
     validates :school_id, uniqueness: { scope: %i[contract_period_year] }
 
-    touch -> { self }, on_event: :update, when_changing: %i[in_partnership induction_programme_choice], timestamp_attribute: :api_updated_at
+    touch -> { self }, on_event: :update, when_changing: %i[induction_programme_choice], timestamp_attribute: :api_updated_at
   end
 end
--- a/app/serializers/api/school_serializer.rb
+++ b/app/serializers/api/school_serializer.rb
@@ -12,7 +12,9 @@ class API::SchoolSerializer < Blueprinter::Base
       options[:contract_period_year].to_s
     end
     field(:in_partnership) do |data, options|
-      contract_period_metadata(school: data[:school], options:).in_partnership
+      data[:school].school_partnerships.any? do |school_partnership|
+        school_partnership.lead_provider_delivery_partnership&.active_lead_provider&.contract_period_year == options[:contract_period_year]
+      end
     end
     field(:induction_programme_choice) do |data, options|
       contract_period_metadata(school: data[:school], options:).induction_programme_choice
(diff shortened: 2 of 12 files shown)


app/models/metadata/school_contract_period.rb
```diff
@@ -15,10 +15,9 @@ class SchoolContractPeriod < Metadata::Base
 
     validates :school, presence: true
     validates :contract_period, presence: true
-    validates :in_partnership, inclusion: { in: [true, false] }
     validates :induction_programme_choice, inclusion: { in: induction_programme_choices.keys }
     validates :school_id, uniqueness: { scope: %i[contract_period_year] }
 
-    touch -> { self }, on_event: :update, when_changing: %i[in_partnership induction_programme_choice], timestamp_attribute: :api_updated_at
+    touch -> { self }, on_event: :update, when_changing: %i[induction_programme_choice], timestamp_attribute: :api_updated_at
```
ethax-ross, line 21: This change looks good assuming the performance impact isn't too bad? there is one side effect I didn't consider with this approach, though 😞 

At the moment if a school partnership is created for a school in a new contract period causing the `in_partnership` to go from `false` to `true` it will be reflected in the `updated_at` in the serializer (as the `updated_at` is derived from the metadata in the schools response; its an oddity in that it's the only endpoint that is always relative to a given `filter[cohort]`).

By making this change we're going to lose that `touch`, so now when a school gets its first partnership for a contract period the `updated_at` would not change 🤔 I think we need to run this one past @easeynathan as if we're happy with that trade-off I expect we'll need to let LPs know before it goes out.

easeynathan: When in_partnership changes from FALSE to TRUE, wouldn't induction_programme_choice also be expected to update and so the record would still be touched?

I'm not sure if it is possible for in_partnership to update without induction_programme_choice also updating, so would this be a moot point? Let me know if I've misunderstood

tonyheadford: When a new `school_partnership` lands all the `training_periods` at the school that have an EOI for the `active_lead_provider` and no current `school_partnership` get updated with the new `school_partnership`.  This in turn will trigger a `refresh_metadata` for both the school and the teacher and also a `touch` update for the teacher.

Not sure if this is what you mean or if I'm barking up the wrong tree though 😕

ethax-ross: > When in_partnership changes from FALSE to TRUE, wouldn't induction_programme_choice also be expected to update and so the record would still be touched?

@easeynathan I think you're right; we derive the `induction_programme_choice` through the TPs associated with the SPs, so when the first SP is added -- assuming it always has a school led or provider led TP at that time -- the metadata and therefore school should be touched.

Shall we deploy this with reasonable confidence the `updated_at` will still be touched based on the comments above and monitor Sentry for any refresh alerts from the nightly job?

easeynathan: > Shall we deploy this with reasonable confidence the updated_at will still be touched based on the comments above and monitor Sentry for any refresh alerts from the nightly job?

Sounds good to me.

ethax-ross (approved):

PR #3354, opened 2026-08-07
[4277] Blazer query for schools with post-closure successor links

rwrrll:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/4277

Adds an idempotent task to create the desired Blazer queries: one at a per-school level, one at a per-teacher level.

This rests entirely on the GIAS integration work that records a `school_closed` event to deterministically detect a school closure with no successor links. With no specific set of columns defined, I've built this with quite a few columns of information, including pre-empting the fact that some teachers may already be registered by the time the results of the queries are observed/actioned. It might be overkill, but I figured better to include the kitchen sink now before an automated process is designed, and trim back later once we know exactly how that process will work.

### Guidance to review

n.b. the Product Review rake task is throwaway, just to make this product-reviewable on a review app.

### DIFF
--- a/db/seeds/blazer_queries/post_closure_school_links.rb
+++ b/db/seeds/blazer_queries/post_closure_school_links.rb
@@ -1,0 +1,284 @@
+module BlazerQueries
+  # Only GIAS::Reconciliation::Close records a school_closed event, and it refuses
+  # to run while the school has any successor link. Therefore, any successor link on
+  # a school with that event arrived after we had already finished or deleted its
+  # teachers' periods.
+  class PostClosureSchoolLinks
+    SCHOOLS_QUERY_NAME = "GIAS: Schools closed without a link that gained one later"
+    TEACHERS_QUERY_NAME = "GIAS: Teachers affected by a school link added after closure"
+
+    class << self
+      def sync!
+        definitions.map do |definition|
+          query = Blazer::Query.find_or_initialize_by(name: definition[:name])
+          query.update!(
+            description: definition[:description],
+            statement: definition[:statement],
+            data_source: :main,
+            status: "active"
+          )
+          query
+        end
+      end
+
+      def definitions
+        [affected_schools, affected_teachers]
+      end
+
+    private
+
+      delegate :quote, to: "ActiveRecord::Base.connection", private: true
+
+      def affected_schools
+        {
+          name: SCHOOLS_QUERY_NAME,
+          description: "Schools that RECT closed because GIAS gave us no successor " \
+                       "link, where GIAS has since given a successor link to, and that " \
+                       "had ECTs or mentors registered when they closed. Those " \
+                       "teachers had their periods finished or deleted instead of " \
+                       "being moved to the linked school. Row per closed school & link.",
+          statement: <<~SQL.strip
+            #{common_ctes}
+            SELECT
+              l.urn AS closed_school_urn,
+              l.school_name AS closed_school_name,
+              l.closed_on,
+              l.closure_processed_at,
+              l.link_type,
+              l.link_urn,
+              l.link_school_name,
+              l.link_school_status,
+              l.link_school_registered_in_rect,
+              l.link_date AS link_established_on_in_gias,
+              l.link_first_imported_at,
+              l.link_last_changed_at,
+              l.days_between_closure_and_link,
+              counts.teachers_affected,
+              counts.ect_periods_affected,
+              counts.mentor_periods_affected,
+              counts.teachers_registered_elsewhere_since,
+              counts.teachers_registered_at_linked_school_since
+            FROM late_links l
+            INNER JOIN LATERAL (
+              SELECT
+                COUNT(DISTINCT a.teacher_id) AS teachers_affected,
+                COUNT(*) FILTER (WHERE a.role = 'ECT') AS ect_periods_affected,
+                COUNT(*) FILTER (WHERE a.role = 'Mentor') AS mentor_periods_affected,
+                COUNT(DISTINCT a.teacher_id) FILTER (
+                  WHERE #{registered_elsewhere_since_sql}
+                ) AS teachers_registered_elsewhere_since,
+                COUNT(DISTINCT a.teacher_id) FILTER (
+                  WHERE #{registered_at_linked_school_since_sql}
+                ) AS teachers_registered_at_linked_school_since
+              FROM affected a
+              WHERE a.closure_event_id = l.closure_event_id
+            ) counts ON counts.teachers_affected > 0
+            ORDER BY l.link_first_imported_at DESC, l.urn;
+          SQL
+        }
+      end
+
+      def affected_teachers
+        {
+          name: TEACHERS_QUERY_NAME,
+          description: "The teachers behind '#{SCHOOLS_QUERY_NAME}'. One row per " \
+                       "ECT or mentor period that RECT finished or deleted when we " \
+                       "closed a school GIAS has since linked to a successor. " \
+                       "'registered_elsewhere_since' and " \
+                       "'registered_at_linked_school_since' flag teachers who have " \
+                       "already been registered again (at any other school / at the " \
+                       "linked school) on or after the closure date, so they " \
+                       "probably need no further action.",
+          statement: <<~SQL.strip
+            #{common_ctes}
+            SELECT
+              l.urn AS closed_school_urn,
+              l.school_name AS closed_school_name,
+              l.closed_on,
+              l.link_type,
+              l.link_urn,
+              l.link_school_name,
+              l.link_school_status,
+              l.link_school_registered_in_rect,
+              l.days_between_closure_and_link,
+              a.role,
+              a.teacher_id,
+              t.trn,
+              #{teacher_name_sql} AS teacher_name,
+              a.what_happened,
+              a.started_on AS period_started_on,
+              a.finished_on AS period_finished_on,
+              #{registered_elsewhere_since_sql} AS registered_elsewhere_since,
+              #{registered_at_linked_school_since_sql} AS registered_at_linked_school_since
+            FROM late_links l
+            INNER JOIN affected a ON a.closure_event_id = l.closure_event_id
+            INNER JOIN teachers t ON t.id = a.teacher_id
+            ORDER BY l.link_first_imported_at DESC, l.urn, a.role, a.teacher_id;
+          SQL
+        }
+      end
+
+      def common_ctes
+        <<~SQL.strip
+          WITH #{closures_cte},
+          #{late_links_cte},
+          #{affected_cte}
+        SQL
+      end
+
+      def closures_cte
+        <<~SQL.strip
+          closures AS (
+            SELECT
+              e.id AS closure_event_id,
+              e.school_id,
+              (e.metadata ->> 'gias_school_urn')::integer AS urn,
+              e.metadata ->> 'gias_school_name' AS school_name,
+              e.created_at AS closure_processed_at
+            FROM events e
+            WHERE e.event_type = 'school_closed'
+          )
+        SQL
+      end
+
+      def late_links_cte
+        <<~SQL.strip
+          late_links AS (
+            SELECT
+              c.closure_event_id,
+              c.school_id,
+              c.urn,
+              c.school_name,
+              c.closure_processed_at,
+              gs.closed_on,
+              ls.link_type,
+              ls.link_urn,
+              ls.link_date,
+              ls.created_at AS link_first_imported_at,
+              ls.updated_at AS link_last_changed_at,
+              #{london_date_sql('ls.created_at')} - gs.closed_on AS days_between_closure_and_link,
+              lgs.name AS link_school_name,
+              lgs.status::text AS link_school_status,
+              (lrs.id IS NOT NULL) AS link_school_registered_in_rect
+            FROM closures c
+            INNER JOIN gias_schools gs ON gs.urn = c.urn
+            INNER JOIN gias_school_links ls
+              ON ls.urn = c.urn
+             AND ls.link_type IN (#{successor_link_types_sql})
+            LEFT JOIN gias_schools lgs ON lgs.urn = ls.link_urn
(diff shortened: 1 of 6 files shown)


db/seeds/blazer_queries/post_closure_school_links.rb
```diff
@@ -0,0 +1,284 @@
+module BlazerQueries
+  # Only GIAS::Reconciliation::Close records a school_closed event, and it refuses
```
thenapking, line 2: Just FYI, we didn't back fill these events, so this query might not show anything at the moment on production.

spec/seeds/blazer_queries/post_closure_school_links_spec.rb
```diff
@@ -0,0 +1,315 @@
...
+        school:,
+        happened_at: closed_on,
+        metadata: { gias_school_urn: urn, gias_school_name: gias_school.name }
+      )
+    end
+
+    school
+  end
+
+  def link_to_successor!(school, link_type: GIAS::SchoolLink::SUCCESSOR, successor_urn: nil, status: "open", registered: true)
```
thenapking, line 285: I'm not sure all of this is really necessary.  There are traits for all the link types, and the link factory will create schools (if required).  For instance, this will create two gias schools, and link them to the merged school target.
```
let(:merged_school) { FactoryBot.create(gias_school)
FactoryBot.create(:gias_school_link, :successor_merged, to_gias_school: merged_school)
FactoryBot.create(:gias_school_link, :successor_merged, to_gias_school: merged_school)
```

spec/seeds/blazer_queries/post_closure_school_links_spec.rb
```diff
@@ -0,0 +1,315 @@
...
+          .not_to raise_error
+      end
+    end
+  end
+
+  describe "the closed schools query" do
+    subject(:rows) { rows_for(described_class::SCHOOLS_QUERY_NAME) }
+
+    context "when a school we closed has since been given a successor link" do
+      let!(:school) { close_school!(urn: 100_001) }
```
thenapking, line 34: I think it's a bit confusing to call a service from `let`.  I think this should be in a before block, like it is in other examples.

spec/seeds/blazer_queries/post_closure_school_links_spec.rb
```diff
@@ -0,0 +1,315 @@
...
+          "link_school_name" => link.to_gias_school.name,
+          "link_school_status" => "open",
+          "link_school_registered_in_rect" => true,
+          "teachers_affected" => 1
+        )
+      end
+    end
+
+    context "when the closure finished some periods and deleted others" do
+      let!(:school) { close_school!(urn: 100_002) }
```
thenapking, line 55: As above

spec/seeds/blazer_queries/post_closure_school_links_spec.rb
```diff
@@ -0,0 +1,315 @@
...
+        expect(rows.fetch(0)).to include(
+          "teachers_affected" => 5,
+          "ect_periods_affected" => 3,
+          "mentor_periods_affected" => 2
+        )
+      end
+    end
+
+    context "when the link arrived three weeks after the closure" do
+      let!(:school) { close_school!(urn: 100_003) }
```
thenapking, line 75: As above

spec/seeds/blazer_queries/post_closure_school_links_spec.rb
```diff
@@ -0,0 +1,315 @@
...
+        finish_at_closure!(school)
+      end
+
+      it "reports the gap in days" do
+        expect(rows.fetch(0)).to include("days_between_closure_and_link" => 21)
+      end
+    end
+
+    context "when some of the affected teachers have been registered again" do
+      let!(:school) { close_school!(urn: 100_040) }
```
thenapking, line 89: As above

spec/seeds/blazer_queries/post_closure_school_links_spec.rb
```diff
@@ -0,0 +1,315 @@
...
+      before do
+        school = close_school!(urn: 100_012)
+        link_to_successor!(school, link_type: "Sixth Form Centre Link")
+        finish_at_closure!(school)
+      end
+
+      it { is_expected.to be_empty }
+    end
+
+    context "when we did not close the school ourselves" do
```
thenapking, line 146: Perhaps a better description is "when there is no closure event recorded".

db/seeds/blazer_queries/post_closure_school_links.rb
```diff
@@ -0,0 +1,284 @@
...
+        SQL
+      end
+
+      def registered_elsewhere_since_sql
+        <<~SQL.strip
+          EXISTS (
+                SELECT 1 FROM ect_at_school_periods x
+                WHERE x.teacher_id = a.teacher_id
+                  AND x.school_id <> a.school_id
+                  AND x.started_on >= a.closed_on
```
avinhurry, line 237: This is checking that the teacher `started_on` after the school closed, but that's not necessarily when they were registered, as we allow past, current and future start dates. It would also exclude a period that started before the closure but was still ongoing afterwards. `created_at` wouldn't necessarily be better because of migrations, console changes, etc.

Not flagging it as wrong, just checking that this is the intended definition for these. Perhaps that could be made clearer in the `affected_teachers` description though?

db/seeds/blazer_queries/post_closure_school_links.rb
```diff
@@ -0,0 +1,284 @@
...
+      def late_links_cte
+        <<~SQL.strip
+          late_links AS (
+            SELECT
+              c.closure_event_id,
+              c.school_id,
+              c.urn,
+              c.school_name,
+              c.closure_processed_at,
+              gs.closed_on,
```
avinhurry, line 153: Can `closed_on` change after reconciliation? Would it be safer to use the event's `happened_at` here, so we use the historical closure date RECT acted on rather than a potentially updated GIAS closure date?

db/seeds/blazer_queries/post_closure_school_links.rb
```diff
@@ -0,0 +1,284 @@
...
+        <<~SQL.strip
+          closures AS (
+            SELECT
+              e.id AS closure_event_id,
+              e.school_id,
+              (e.metadata ->> 'gias_school_urn')::integer AS urn,
+              e.metadata ->> 'gias_school_name' AS school_name,
+              e.created_at AS closure_processed_at
+            FROM events e
+            WHERE e.event_type = 'school_closed'
```
avinhurry, line 139: The events table has roughly 2.1 million rows in production. The plan uses a parallel sequential scan of the whole table to find ~172 `school_closed` events. That may still be acceptable for an occasional blazer query but have we checked the expected runtime and impact on production?

Would it be possible to avoid the full scan by starting from the much smaller set of qualifying school links and leaning on the existing `events.school_id` index?

lib/tasks/product_review/4277.rake
```diff
@@ -0,0 +1,118 @@
...
+
+      ECTAtSchoolPeriod.create!(teacher: randall, school: celia_mae.school, started_on: Date.new(2026, 9, 1), finished_on: nil)
+    end
+
+    require Rails.root.join("db/seeds/blazer_queries/post_closure_school_links")
+    queries = BlazerQueries::PostClosureSchoolLinks.sync!
+
+    puts
+    puts "=" * 78
+    puts "  Schools linked to a successor after we closed them (#4277)"
```
avinhurry, line 47: Professional 😎

avinhurry (commented):
I left a few comments/questions but this is looking great so far. 

FWIW, I also think including the kitchen sink is worthwhile to begin with, then we can trim back once we know is actually useful.

db/seeds/blazer_queries/post_closure_school_links.rb
thenapking: The `closed_on` attribute comes through the GIAS import, from the files.  `GIAS::Reconciliation::Close` sets the `happened_at` timestamp of the event to `closed_on`.  I set it like this because I assumed if we wanted to display this in a timeline in the future we should be setting `happened_at` to the time the event was meant to happen, not when we processed it.  Also we currently only update the `GIAS::SchoolLink` if the `link_type` attribute changes, so perhaps we want compare `event.created_at` with `gias_school.closed_on`?

spec/seeds/blazer_queries/post_closure_school_links_spec.rb
rwrrll: Good shout, I've cut this down. The factory now allocates the urn and the helper is four lines shorter.

Two bits I've kept, as the traits alone don't cover them:
- `link_school_registered_in_rect` is (lrs.id IS NOT NULL) against the schools table, and the link factory only ever creates `gias_schools`. Without the explicit `School.create`, that column is always false, and the "registered at the linked school" tests can't build their setup.
- status: "open" has to be explicit: the default factory goes through trait(:open), which is `status { %w[open proposed_to_close].sample }`, so the assertion on `link_school_status` would flake about half the time.

`from_gias_school` also has to be the closed school so that the factory doesn't invent a random one.

Cheers for spotting!

spec/seeds/blazer_queries/post_closure_school_links_spec.rb
rwrrll: Yeah fair point on three of the four, so I've moved those to locals in `before`.

Kept the `let!` here because `school` is used by `let!(:link)` and by the example, and link is used by the example, so moving it would mean threading both back out of a before.

Also a minor clarification: `close_school!` isn't a service, it's a helper in this file, `GIAS::Reconciliation::Close` is never invoked in the spec. But the name invited that reading, so I've renamed it to `closed_school!`.

spec/seeds/blazer_queries/post_closure_school_links_spec.rb
rwrrll: Yep agree, changed. It names the actual condition (record_closure: false) rather than paraphrasing it.

spec/seeds/blazer_queries/post_closure_school_links_spec.rb
thenapking: Yes, perhaps we should improve the factories.  I've run into that issue with the `open` trait sometimes returning `proposed_to_close` often.  `GIAS::School` has a `:with_school` trait, that creates the `School` and perhaps we could have a similar trait on `SchoolLink`.  Some of my specs have got a bit verbose with this sort of setup, so maybe I could do this as a refactor later in the week.

db/seeds/blazer_queries/post_closure_school_links.rb
rwrrll: Yeah as you point out, sadly we don't know the registration date, so this is the next best thing. I've renamed to make them more descriptive of what they're actually saying! And extended the criteria to deal with ongoing periods.

db/seeds/blazer_queries/post_closure_school_links.rb
rwrrll: Yeah, I think `closed_on` is still the right thing to use after reconciliation. As Tom says, it comes from GIAS, so in theory it's the authoritative closure date for the school, won't change once the school has closed, and it's the value we propagate through the closure machinery to decide when periods should end. Technically it can be changed by the importer, so we are relying on GIAS not retrospectively revising it after the school has already closed, but operationally that seems very unlikely.

Initially I was going to use `happened_at`, since it's populated from `closed_on` and immutable once written, but it felt like a proxy for the real fact. Tom's note above is what decided it for me: the value is set the way it is on an assumption about how we might want to display events in a timeline in future. That makes it owned by the events concern rather than the closure machinery, so if that assumption is ever revisited this query would break silently.

`event.created_at` doesn't work as the closure date either, it's quite reasonable for closure processing to run a day or more after the actual closure.

So for me there's no standout winner between the two; in theory they're the same but either could diverge. I've gone with `closed_on` as the least footgun-y. Based on the conversation, I’ve noted the assumption in the query description, since if it ever did break the school would silently drop out of the report rather than appear with a wrong date.

db/seeds/blazer_queries/post_closure_school_links.rb
rwrrll: Well spotted! 👀 

On driving from the links and using `index_events_on_school_id`: the snag is getting from a link to a `school_id` in the first place. The only route I can see is `schools.urn = gias_school_links.urn`, and Replace does `school.update!(urn: successor.urn)`. So once a late Successor link arrives pointing at a school we don't have registered, reconciliation repoints the school record and its urn becomes the successor's. From then on the closed school's urn matches nothing in schools, and `events.metadata ->> 'gias_school_urn'` is the only remaining record of which school the closure was for.

re: runtime, Blazer runs against a read replica so no impact on production. Both queries come in around 10s in production, so not super performant but probably OK for a rarely-used query. We could add an index but that would put write overhead on the primary to speed up this query on the replica. Worth looking at if it grinds too slow over time, but I figured probably OK without for now?

avinhurry: Ah yeah good point, I always forget Blazer uses the read replica... 

~10s sounds fine for an occasional query but we should revisit it if this ever moves into the app 👍🏽

avinhurry (approved):
Very thorough work! 

Thanks for addressing/responding to all the comments.

db/seeds/blazer_queries/post_closure_school_links.rb
cpjmcquillan: Would this be mitigated if we stamped the `School` record when processing the closure (i.e setting `closed_at`)?

I don't love the approach of using `Event` records to determine eligibility for the various processes.

thenapking: I agree about the use of Events.  This was something of a stop gap to get the GIAS processing going.  I think it's worth us trying to decide what would be a good way to do this.  It's worth noting that as I understand it closed schools will eventually get deleted from the system, and that event will disappear at that point.  I think we actually need to record this on the GIAS::School.

thenapking (approved):

PR #3356, opened 2026-08-10
Introduce `Admin::DataFixes::InlineCSV` class

cpjmcquillan:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/4252

This PORO is intended to act as some of the plumbing between a user submitting a form with some CSV as text, and eventually driving some data fix changes.

It does some basic validation and explicitly expects the correct headers to be used.

The `#parse` method intentionally returns something "truthy" when valid and "falsey" when invalid, so it can be used in controllers like this:

```ruby
if @inline_csv.parse
  # do something
else
  render :new, status: :unprocessable_content
end
```

The class isn't used anywhere yet, but will be once we start building out some user-facing slices of the admin data fixes journey.

### DIFF
--- a/spec/services/admin/data_fixes/inline_csv_spec.rb
+++ b/spec/services/admin/data_fixes/inline_csv_spec.rb
@@ -1,0 +1,112 @@
+RSpec.describe Admin::DataFixes::InlineCSV do
+  subject(:inline_csv) { described_class.new(csv_string:) }
+
+  describe "#parse" do
+    subject(:parse) { inline_csv.parse }
+
+    context "when the CSV is missing" do
+      let(:csv_string) { "" }
+
+      it { is_expected.to be_falsey }
+
+      it "validates the CSV is present" do
+        parse
+        expect(inline_csv.errors).to be_added(:csv_string, "can't be blank")
+      end
+    end
+
+    context "when the CSV is malformed" do
+      let(:csv_string) do
+        <<~ROWS
+          object_type,object_id,action,attributes
+          something,1,create,"unterminated,string
+        ROWS
+      end
+
+      it { is_expected.to be_falsey }
+
+      it "validates the CSV is formatted correctly" do
+        parse
+        expect(inline_csv.errors).to be_added(:csv_string, "is malformed")
+      end
+    end
+
+    context "when the CSV has no headers" do
+      let(:csv_string) do
+        <<~ROWS
+          something,1,create,"attribute1,value1,attribute2,value2"
+          anotherthing,2,destroy,""
+        ROWS
+      end
+
+      it { is_expected.to be_falsey }
+
+      it "validates the CSV has the correct headers" do
+        parse
+        expect(inline_csv.errors).to be_added(:csv_string, "has invalid headers")
+      end
+    end
+
+    context "when the CSV has missing headers" do
+      let(:csv_string) do
+        <<~ROWS
+          object_type,object_id,attributes
+          something,1,create,"attribute1,value1,attribute2,value2"
+          anotherthing,2,destroy,""
+        ROWS
+      end
+
+      it { is_expected.to be_falsey }
+
+      it "validates the CSV has the correct headers" do
+        parse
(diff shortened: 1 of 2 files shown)


spec/services/admin/data_fixes/inline_csv_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+        parse
+        expect(inline_csv.errors).to be_added(:csv_string, "can't be blank")
+      end
+    end
+
+    context "when the CSV is malformed" do
+      let(:csv_string) do
+        <<~ROWS
+          object_type,object_id,action,attributes
+          "something,1,create,"attribute1,value1,attribute2,value2"
```
thenapking: It took me a while to spot this unterminated string at the start, especially because all the values here are the same as the valid CSV example.  How about just leaving the last string unterminated and making it read `"unterminated,string` ?

spec/services/admin/data_fixes/inline_csv_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+      it "validates the CSV has the correct headers" do
+        parse
+        expect(inline_csv.errors).to be_added(:csv_string, "has invalid headers")
+      end
+    end
+
+    context "when the CSV has the wrong headers" do
+      let(:csv_string) do
+        <<~ROWS
+          object_type,object_id,action,wrong_header
```
thenapking, line 70: I think the code might allow us to have missing headers.  Eg:
```
object_type,attributes
```
These would come out valid, because each one is a valid header.  But it's not valid, because we're missing the `action` and `object_id`.  Since this is an internal tool, why not keep it simple and make `attributes` a required column, and the header order fixed, and just do something like:
```
raise InvalidHeaderError unless table.headers == EXPECTED_HEADERS
```

spec/services/admin/data_fixes/inline_csv_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+        <<~ROWS
+          object_type,object_id,action,attributes
+          something,1,create,"attribute1,value1,attribute2,value2"
+          another_thing,2,destroy,""
+        ROWS
+      end
+
+      it { is_expected.to be_truthy }
+
+      it "returns an array of structured row data" do
```
thenapking, line 95: I was expecting that this class, or the `Row` class, would also parse the data, so that the `object_id` would be an integer, `attributes` a hash, `object_type` would be a constant in a defined list, and so on.  

Looking at the `MigrationFixes::Processor` service, I see at the moment that it is a mixture of processing and parsing.

I think it would make sense to pull the data processing out into a separate class so that errors like unknown `action`, `object_type` and so on can be raised before processing. What do you think?

The other option I suppose is updating the title of this test to make it clear that the data is in some sense raw, and is for processing by another named class.

thenapking (commented):
Looks fine.  I think we need clear up the issue with the headers and decide where the data itself is parsed, and whether that's covered by this PR or another one.

spec/services/admin/data_fixes/inline_csv_spec.rb
cpjmcquillan: Yeah, that makes sense. I've added a test for that now 👍🏼

spec/services/admin/data_fixes/inline_csv_spec.rb
cpjmcquillan: I've updated the test to be clearer 👌🏼

spec/services/admin/data_fixes/inline_csv_spec.rb
cpjmcquillan: The plan is for the processing to be handled by https://github.com/DFE-Digital/register-ects-project-board/issues/4272.

Part of the reason we don't want to change the existing class too much is because we don't want to impact existing use.

If we want the processor to do less, or behave differently, we should circle back and rework it once this epic is done IMO.

thenapking (approved):
LGTM

PR #3359, opened 2026-08-11
Return errors from migration fix processor

ericaporter:
### Context

MigrationFixes::Processor previously raised processing errors, leaving db/scripts/migration_data_fixes.rb responsible for rescuing them.

The processor will also be used by a future UI, where processing errors need to be captured and presented to the user.

- add MigrationFixes::Result to represent successful and unsuccessful outcomes
- update the processor to capture and return errors
- expose the processed object through successful results
- update the migration script to consume and report returned errors
- preserve existing logging / processing

### Guidance to review
- the processor returns the original error without raising it
- successful results expose the processed object
- the migration script’s existing behaviour remains unchanged
- failed rows are logged without preventing later rows from being processed


db/scripts/migration_data_fixes.rb
cpjmcquillan, line 1: Should we add some tests for this script that pass before/after this refactor? 🤔

app/services/migration_fixes/result.rb
cpjmcquillan, line 1: Definitely think this is an object worth having, but does it need to be its own service?

How do you feel about defining it as a `Data` object in the `MigrationFixes::Processor` class? 💭

```ruby
Result = Data.define(:data_change, :target_object, :error) do
  def success? = error.nil?
end
```

spec/migration_fixes/real_examples/teacher_210915_spec.rb
```diff
@@ -149,7 +149,8 @@
       attrs = data_change[:attributes].stringify_keys.to_a.flatten.join(",")
       data_change[:attributes] = attrs
 
-      processor.process!(data_change:)
+      result = processor.process!(data_change:)
+      raise result.error unless result.success?
```
cpjmcquillan, line 153: I'm not sure I understand why we need to make these changes 🤔

Is it because the `Processor` class rescues errors now?

cpjmcquillan (approved):

db/scripts/migration_data_fixes.rb
ericaporter: I was in two minds about adding something to cover this - happy to add!

app/services/migration_fixes/result.rb
ericaporter: I originally made it separate since it would probably be consumed by the script and the future UI and it reduces responsibility of the processor / allows it to be tested etc. Given it currently holds three values and defines success, nesting it is simpler.

spec/migration_fixes/real_examples/teacher_210915_spec.rb
ericaporter: I was in two minds about adding this but previously process! didn’t catch errors, so any failure would immediately fail the spec. Now it catches the error and returns it in the result, so if we re-raise it here that preserves that behaviour and just makes the original problem easier to debug.

PR #3362, opened 2026-08-11
[4304] Allow adding contracts until period is frozen

tonyheadford:
### Context

Finance users should be able to add contract versions until the contract period is frozen. Currently they could only be added until the `finished_on` date of the `ContractPeriod`

Change the behaviour of contract creation/deletion to match that of editing which is to use `active_lead_provider.editable?` which in turn checks the value of the `ContractPeriod.payments_frozen_at` against `Time.zone.now`.

### Guidance to review

Signed in as a finance user, you should be able to add a new `Contract` to a lead provider for any `ContractPeriod` that is not frozen and be prevented from doing so for a frozen `ContractPeriod`.

### DIFF
--- a/spec/requests/admin/finance/active_lead_providers/contracts_spec.rb
+++ b/spec/requests/admin/finance/active_lead_providers/contracts_spec.rb
@@ -239,6 +239,52 @@ RSpec.describe "Admin finance active lead provider contracts", type: :request do
         end
       end
     end
+
+    context "when editing the current contract period" do
+      include_context "sign in as finance DfE user"
+
+      around do |example|
+        travel_to(1.day.after(contract_period.started_on)) { example.run }
+      end
+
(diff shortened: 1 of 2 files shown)


cpjmcquillan (approved):

spec/requests/admin/finance/active_lead_providers/contracts_spec.rb
```diff
@@ -179,10 +179,25 @@
           travel_to(1.day.after(contract_period.finished_on)) { example.run }
         end
 
+        it "allows creating a new contract" do
```
ethax-ross, line 182: It looks like we only have these checks for the POST/create action, do we want to include them in the other actions as well since we also restrict edit/destroy?

ethax-ross (approved):

PR #3363, opened 2026-08-11
Announce band service fee changes to screen readers

craigmdavidson:
### Context

Screen readers did not get changes to an input's value, 
so a user got no signal that the page had changed a value in response to their typing. 

With javascript enabled, the table renders exactly as before, however:
- It always derives service fee from output fee, the javascript is just an affordance.
- The screen reader announces the value onBlur if javascript present
- No Javascript, no service fee, it is left for the user to do the maths. I.e. Setting the output fee to 60%, the service fee isn't shown, but we know it must be 40%.

### DIFF
--- a/app/javascript/controllers/band_controller.js
+++ b/app/javascript/controllers/band_controller.js
@@ -1,20 +1,60 @@
 import { Controller } from "@hotwired/stimulus"
 
 export default class BandController extends Controller {
-  static targets = ["outputPercentage", "servicePercentage"]
+  static targets = ["outputPercentage", "servicePercentage", "serviceFeeStatus"]
+  static values = { announceDelay: { type: Number, default: 750 } }
 
-  // Ensure output_fee_percentage is 0..100
-  // Update service_fee_percentage with delta
+  connect() {
+    this.servicePercentageTarget.value = this.#serviceFeePercentage()
+  }
+
+  disconnect() {
+    clearTimeout(this.announceTimeout)
+  }
+
   updateServiceFee() {
-    let output = Number.parseFloat(this.outputPercentageTarget.value)
+    const serviceFee = this.#serviceFeePercentage()
 
-    if (Number.isNaN(output)) {
-      this.servicePercentageTarget.value = ""
-      return
-    }
+    this.servicePercentageTarget.value = serviceFee
+    this.#announce(serviceFee)
+  }
 
-    output = Math.max(0, Math.min(100, Math.round(output)))
-    this.outputPercentageTarget.value = output
-    this.servicePercentageTarget.value = (100 - output)
+  roundAndCapOutputFee() {
+    const outputFee = this.#roundedAndCappedOutputFee()
+
+    if (outputFee === null) return
+
+    this.outputPercentageTarget.value = outputFee
+    this.updateServiceFee()
+  }
+
+  #announce(serviceFee) {
+    clearTimeout(this.announceTimeout)
+
+    this.announceTimeout = setTimeout(() => {
+      this.serviceFeeStatusTarget.textContent = serviceFee === "" ? "" : `Service fee ${serviceFee}%`
+    }, this.announceDelayValue)
+  }
+
+  #serviceFeePercentage() {
+    const outputFee = this.#outputFee()
+
+    if (outputFee === null || outputFee < 0 || outputFee > 100) return ""
+
+    return String(100 - Math.round(outputFee))
+  }
+
+  #roundedAndCappedOutputFee() {
+    const outputFee = this.#outputFee()
+
+    if (outputFee === null) return null
+
+    return Math.max(0, Math.min(100, Math.round(outputFee)))
+  }
+
+  #outputFee() {
+    const outputFee = Number.parseFloat(this.outputPercentageTarget.value)
+
+    return Number.isNaN(outputFee) ? null : outputFee
   }
 }
--- a/app/views/admin/finance/active_lead_providers/contracts/_fields.html.erb
+++ b/app/views/admin/finance/active_lead_providers/contracts/_fields.html.erb
@@ -56,6 +56,8 @@
 
   <h3 class="govuk-heading-s">Bands</h3>
 
+  <p class="govuk-hint" id="bands-service-fee-hint">Service fee is 100% minus the output fee.</p>
+
   <%= govuk_table do |table| %>
     <% table.with_head do |head| %>
       <% head.with_row do |row| %>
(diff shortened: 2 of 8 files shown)


app/views/admin/finance/active_lead_providers/contracts/_fields.html.erb
```diff
@@ -80,20 +82,28 @@
             <% row.with_cell(text: band_term.min_declarations.to_s) %>
             <% row.with_cell(text: band_term.max_declarations.to_s) %>
             <% row.with_cell do %>
-              <%= bt.text_field :service_fee_percentage,
+              <%= tag.input type: "text",
```
craigmdavidson, line 85: This is the visual percentage change. It is now just an affordance. Since it is a text field, it _will_ be submitted, but it will be ignored...

app/views/admin/finance/active_lead_providers/contracts/_fields.html.erb
```diff
@@ -80,20 +82,28 @@
...
                     inputmode: "numeric",
                     readonly: true,
                     tabindex: -1,
                     aria: { label: "Band #{("A".ord + index).chr} service fee percentage" },
                     data: { band_target: "servicePercentage" } %>
+              <span id="band-<%= ("a".ord + index).chr %>-service-fee-status"
+                    class="govuk-visually-hidden"
+                    aria-live="polite"
+                    aria-atomic="true"
+                    data-band-target="serviceFeeStatus"></span>
```
craigmdavidson, line 96: .... and this is the screen reader version.

app/javascript/controllers/band_controller.js
```diff
@@ -1,20 +1,60 @@
...
+    this.updateServiceFee()
+  }
+
+  #announce(serviceFee) {
+    clearTimeout(this.announceTimeout)
+
+    this.announceTimeout = setTimeout(() => {
+      this.serviceFeeStatusTarget.textContent = serviceFee === "" ? "" : `Service fee ${serviceFee}%`
+    }, this.announceDelayValue)
+  }
```
craigmdavidson, line 37: So `aria-live="polite"` announces after the blur, but it will queue the changes up. 
We don't want that, we only want the latest serviceFee.

app/javascript/controllers/band_controller.js
```diff
@@ -1,20 +1,60 @@
...
+
+  #serviceFeePercentage() {
+    const output = this.#outputFee()
+
+    if (output === null || output < 0 || output > 100) return ""
+
+    return String(100 - Math.round(output))
+  }
```
craigmdavidson, line 45: We wrap the calculated value in a string for a consistent return value "".

app/javascript/controllers/band_controller.js
```diff
@@ -1,20 +1,60 @@
...
+
+  #announce(serviceFee) {
+    clearTimeout(this.announceTimeout)
+
+    this.announceTimeout = setTimeout(() => {
+      this.serviceFeeStatusTarget.textContent = serviceFee === "" ? "" : `Service fee ${serviceFee}%`
+    }, this.announceDelayValue)
+  }
+
+  #serviceFeePercentage() {
```
ethax-ross, line 39: Not a big deal, but we're a bit inconsistent with naming around serviceFee/serviceFeePercentage/ outputFee. Initially I assumed serviceFee/outputFee were the ratios but it looks like they're all percentages? A 'fee' unless clarified (feePercentage/feeRatio) also makes me think its a monetary amount.

app/javascript/controllers/band_controller.js
```diff
@@ -1,20 +1,60 @@
...
-      this.servicePercentageTarget.value = ""
-      return
-    }
+  #outputFee() {
+    const outputFee = Number.parseFloat(this.outputPercentageTarget.value)
 
-    output = Math.max(0, Math.min(100, Math.round(output)))
-    this.outputPercentageTarget.value = output
-    this.servicePercentageTarget.value = (100 - output)
+    return Number.isNaN(outputFee) ? null : outputFee
```
ethax-ross, line 58: Something I noticed when testing what I could enter a non-numerical value into the output fee, save the contract successfully and it silently results in a 0/1 ratio split. I think this behaviour is also in prod, I'll maybe raise a separate ticket for it; ideally we would show an error.

ethax-ross (approved):
Looks good and works well 👍

app/javascript/controllers/band_controller.js
craigmdavidson: Okay so the thing is in the database it is stored as `output_fee_ratio` a decimal. These are not really ratios just a decimal with a value between 0 and 1 rather than something like `3:2` or whatever, these are not intuitive the the user as such they are surfaced as percentages. I.e. allocate 70% to output fee.

craigmdavidson: Why we store then as decimals in the database 🤷 the code would be simpler if we structured these as integer based percentages, since we coerce into integers then store it as a decimal!

peteryates (approved):
This looks good to me, thanks!

Agree on the naming of the `ratio` not being the clearest.

I always understood it to be `value:1` when there's no other number, but from the discussion we're not doing that? 🤔

app/javascript/controllers/band_controller.js
ethax-ross: Yeah I agree that it could be improved; the purpose of naming them `_ratio` was to try and signify that they both represent a portion of a whole, but I agree having the column called `_ratio` containing a decimal is misleading; a ratio of 0.7 is actually a 0.7:1 ratio mathematically which isn't what we're trying to convey. I wonder if storing one of them as a percentage and inferring the other rather than having a column for it that we have to validate/keep consistent would be the way to go 🤔

craigmdavidson: > I wonder if storing one of them as a percentage and inferring the other rather than having a column for it that we have to validate/keep consistent would be the way to go

I think that would be a good idea. I think that is almost the logical extension of this change. It would also simplify things we currently validate. But is it okay if we run this one in and do that as a different PR though?

ethax-ross: Yeah definitely not one for this PR; I've raised a ticket https://github.com/DFE-Digital/register-ects-project-board/issues/4365

app/javascript/controllers/band_controller.js
ethax-ross: Ticket added: https://github.com/DFE-Digital/register-ects-project-board/issues/4366

PR #3364, opened 2026-08-11
[3544] Fix heading semantics across the app

avinhurry:
### Context

This PR fixes the heading semantic issues identified in the card, along with similar issues I found while checking the rest of the app. This keeps the heading hierarchy in the right order and makes the pages easier to understand and navigate, especially for screen reader users.

### Pages changed

- ECT and mentor lists
- Find an ECT (coverage)
- ECT Registration -> start date
- Mentor Registration -> confirmation
- Induction tutor details
- School partnerships
- Admin teacher timeline
- Lead provider selection
- Test guidance
- Time travel
- Admin induction outcome pass confirmation
- Admin induction outcome fail confirmation
- Admin teacher training history
- Admin teacher school history

### Guidance to review

Each commit covers a separate page or component, so recommend going commit by commit.

### DIFF
--- a/app/components/schools/ects/listing_card_component.html.erb
+++ b/app/components/schools/ects/listing_card_component.html.erb
@@ -1,8 +1,8 @@
 <div class="govuk-summary-card">
   <div class="govuk-summary-card__title-wrapper">
-    <h3 class="govuk-summary-card__title govuk-!-font-size-24">
+    <h2 class="govuk-summary-card__title govuk-!-font-size-24">
       <strong><%= link_to_ect(ect_at_school_period) %></strong>
-    </h3>
+    </h2>
   </div>
   <div class="govuk-summary-card__content">
     <% if withdrawn? %>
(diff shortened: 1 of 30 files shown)


cpjmcquillan (approved):
One thought I had (outside the scope of this change)

Could we automate this at all? 🤔

I wonder whether we could run a CI step that viewed each page and asserted no headings were skipped. Lighter touch than something like `axe`, because mostly we're leaning on the design system.

app/components/schools/ects/listing_card_component.html.erb
```diff
@@ -1,8 +1,8 @@
 <div class="govuk-summary-card">
   <div class="govuk-summary-card__title-wrapper">
-    <h3 class="govuk-summary-card__title govuk-!-font-size-24">
+    <h2 class="govuk-summary-card__title govuk-!-font-size-24">
       <strong><%= link_to_ect(ect_at_school_period) %></strong>
```
peteryates, line 4: I don't think we need `<strong>` inside `<h2>` here (or anywhere).

Maybe as a follow up can we snip them?

peteryates (approved):
This looks great, thanks @avinhurry 

I left a comment about snipping out `<strong>` elements from headers.

Also, I wonder if there's a way we can test this, even if it's not part of the suite. Maybe a callback that runs as part of the feature spec page-loading, that pulls out the header elements and makes sure the order is correctly hierarchical.

PR #3367, opened 2026-08-12
[4017] Merge deletes predecessor school

thenapking:
### Context

In discussion of cleaning up historic data from merged schools it became clear that there was an omission in the `GIAS::Reconciliation::Merge` service.  Once merged, the `school` record of the closed school, along with associated `school_partnerships` and `events` should be removed from the database.

This PR adds this functionality so that as changes occur we clean up the database, and so that it can be used to clean up the data from before the service launched.

Add methods to `GIAS::Reconciliation::Merge` to destroy all related partnerships and events, as well as the predecessor school itself.

Record the school merged event against the successor school

Correct mocking and test setup:
The tests had been setup so that the service which detects overlapping periods returns one `mentor_at_school_period`.  This is not possible because one period cannot overlap itself.  So the tests are now setup so that there are two overlapping periods. 
Secondly, if overlapping periods are returned, then they would be merged at the successor school.  We then test that any _remaining_ periods (which don't overlap) are then transferred to the successor.  The mocks had set the transfer service and merge service to use the same mentor period.  This is corrected, so we have setup 3 mentor periods.
Thirdly, we need to call the original services in order to allow the service to delete the school.  If there are mentors or ECTs left at the school, we cannot delete it.  Mocking the processes which merge and transfer the periods seems overkill, so making this a straightforward integration test seems to me the best path forward.  
Fourthly, now the mocking has been altered we need to make a couple of adjustments for the scenario where the successor school has to be opened.  In this case there can be no periods at the successor school and so there can be no overlapping mentor periods.

### Guidance to review
Run the merge service from the console on Monster's College and see that it is merged into the School of Excellence and deleted.

### DIFF
--- a/app/services/gias/reconciliation/eligibility.rb
+++ b/app/services/gias/reconciliation/eligibility.rb
@@ -49,10 +49,6 @@ module GIAS::Reconciliation
       closed_on <= Date.current
     end
 
-    def school_merged_event_recorded?
-      Event.where(school:, event_type: :school_merged).exists?
-    end
-
     def school_closed_event_recorded?
       Event.where(school:, event_type: :school_closed).exists?
     end
--- a/app/services/gias/reconciliation/eligibility.rb
+++ b/app/services/gias/reconciliation/eligibility.rb
@@ -36,8 +36,7 @@ module GIAS::Reconciliation
         closed_on_or_before_today? &&
         school.present? &&
         school_being_merged_or_amalgamated? &&
-        has_one_open_successor? &&
-        successor.school.present?
+        has_one_open_successor?
     end
 
   private
--- a/app/services/gias/reconciliation/eligibility.rb
+++ b/app/services/gias/reconciliation/eligibility.rb
@@ -36,8 +36,8 @@ module GIAS::Reconciliation
         closed_on_or_before_today? &&
         school.present? &&
         school_being_merged_or_amalgamated? &&
-        !school_merged_event_recorded? &&
-        has_one_open_successor?
+        has_one_open_successor? &&
+        successor.school.present?
     end
 
   private
--- a/app/services/gias/reconciliation/merge.rb
+++ b/app/services/gias/reconciliation/merge.rb
@@ -85,7 +85,7 @@ module GIAS::Reconciliation
     def successor_school = @successor_school ||= successor.school
 
     delegate :school, :closed_on, :successor, to: :gias_school
-    delegate :mentor_teachers, :ect_teachers, to: :school, prefix: false
+    delegate :mentor_teachers, to: :school, prefix: false
 
     def author = Events::SystemAuthor.new
   end
--- a/app/services/gias/reconciliation/merge.rb
+++ b/app/services/gias/reconciliation/merge.rb
@@ -1,48 +1,57 @@
 module GIAS::Reconciliation
   class Merge
-    def self.merge!(gias_school) = new(gias_school).merge!
+    def self.merge!(gias_school, refresh_metadata: true) = new(gias_school, refresh_metadata:).merge!
 
-    def initialize(gias_school)
+    def initialize(gias_school, refresh_metadata: true)
       @gias_school = gias_school
+      @refresh_metadata = refresh_metadata
     end
 
     def merge!
       return false unless eligibility.can_be_merged?
 
-      ActiveRecord::Base.transaction do
-        ensure_successor_school_exists!
+      teachers_to_refresh = affected_teachers
 
-        mentor_teachers.each do |teacher|
-          overlapping_mentor_at_school_periods(teacher).each do |periods|
-            MentorAtSchoolPeriods::Merge.call(periods:, predecessor_school:, successor_school:)
+      DeclarativeUpdates.skip(:metadata) do
+        ActiveRecord::Base.transaction do
+          ensure_successor_school_exists!
+
+          mentor_teachers.each do |teacher|
+            overlapping_mentor_at_school_periods(teacher).each do |periods|
+              MentorAtSchoolPeriods::Merge.call(periods:, predecessor_school:, successor_school:)
+            end
           end
+
+          remaining_mentor_periods = predecessor_school.mentor_at_school_periods.reload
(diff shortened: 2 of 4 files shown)


app/services/gias/reconciliation/eligibility.rb
```diff
@@ -33,7 +33,6 @@ def can_be_merged?
         closed_on_or_before_today? &&
         school.present? &&
         school_being_merged_or_amalgamated? &&
-        !school_merged_event_recorded? &&
```
cpjmcquillan, line 39: Can we remove this method if it isn't being used?

app/services/gias/reconciliation/merge.rb
```diff
@@ -49,9 +53,21 @@ def overlapping_mentor_at_school_periods(teacher)
...
+      predecessor_school.school_partnerships.destroy_all
+    end
+
+    def destroy_events!
+      predecessor_school.events.destroy_all
+    end
+
+    def destroy_predecessor_school!
+      predecessor_school.destroy!
+    end
```
cpjmcquillan: I'm not sure that these methods are helpful.

I'd argue that 

```ruby
predecessor_school.destroy!
```

is clearer than calling a method

```ruby
destroy_predecessor_school!
```

that I need to lookup.

What do you think?

cpjmcquillan: Also, are we intentionally calling `destroy!` on the school, but `destroy_all` on the partnerships and events?

If it is intentional, perhaps we could highlight _why_ we aren't calling `destroy!` on the partnerships and events too 🤔

app/services/gias/reconciliation/merge.rb
```diff
@@ -75,11 +89,31 @@ def record_school_merged_event!
...
+      Metadata::Handlers::School.new(successor_school).refresh_metadata!
+    end
+
+    def refresh_teacher_metadata!(teachers)
+      return unless refresh_metadata
+
+      teachers.each do |teacher|
+        Metadata::Handlers::Teacher.new(teacher).refresh_metadata!
+      end
+    end
```
cpjmcquillan: I wonder if this part would make sense on the `School` model 🤔

I don't think we ever considered `School` records might get destroyed when we initially scoped things out.

Would this work?

```ruby
# app/models/school.rb

refresh_metadata -> { ect_teachers }, on_event: %i[destroy]
refresh_metadata -> { mentor_teachers }, on_event: %i[destroy]
````

Comment (cpjmcquillan) [reply app/services/gias/reconciliation/merge.rb]: Although thinking about this a little more, I think this should already be covered by https://github.com/DFE-Digital/register-early-career-teachers-public/blob/main/app/models/concerns/at_school_period.rb#L18, given we are updating the `school_id` on the `ECTAtSchoolPeriod` and `MentorAtSchoolPeriod` records.

Is there a reason we need to `skip(:metadata)` and explicitly refresh teacher metadata at all? 🤔

File: app/services/gias/reconciliation/merge.rb [hunk 4]
```diff
@@ -75,11 +89,31 @@ def record_school_merged_event!
...
+      return unless refresh_metadata
+
+      @affected_teachers ||= (mentor_teachers + ect_teachers).uniq
+    end
+
+    def refresh_school_metadata!
+      return unless refresh_metadata
+
+      Metadata::Handlers::School.new(successor_school).refresh_metadata!
+    end
```
cpjmcquillan: Similarly, could this be covered by the declarations on the `SchoolPartnership`, `AtSchoolPeriod`, and `TrainingPeriod` models?

I think part of the design of the `DeclarativeUpdates` pattern is that things should happen automatically, without us worrying too much about it.

Everything runs in sync, so I think it should be safe to just delete the predecessor school metadata and rely on everything else _just working_ 🪄.

I might be missing something, though.

@ethax-ross do you have any insight here?

app/services/gias/reconciliation/merge.rb
thenapking: We're going to delete the school metadata for the predecessor school, so I thought it might be better therefore to skip metadata processing during the change and trigger it at the end.  I'm not really sure whether it is necessary to skip though, as I am not so familar with this part of the codebase. I assume we need to refresh teacher metadata explicitly if we skip, because they've moved school. 

Regarding adding ` on_event: %i[destroy]` to the school - we've moved all the teachers out of the school, and deleted all the events, school_partnerships, etc.  If adding this would remove the need to explicitly delete the metadata for the school on destroy then perhaps it is worth it.

app/services/gias/reconciliation/merge.rb
thenapking: That was only really for efficiency, which was very marginal. I've amended this to call `each(&:destroy!)` and moved all the statements back to the main `merge` call.

app/services/gias/reconciliation/merge.rb
thenapking: After discussion with @ethax-ross we think this [PR](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3391) should address the issue and then this service won't need to skip the `DeclarativeUpdates` at all.

thenapking: Following merge of the above PR this class does not skip `DeclarativeUpdates` and metadata updates are all handled as expected.

app/services/gias/reconciliation/merge.rb
```diff
@@ -81,7 +85,7 @@ def predecessor_school = school
     def successor_school = @successor_school ||= successor.school
 
     delegate :school, :closed_on, :successor, to: :gias_school
-    delegate :mentor_teachers, to: :school, prefix: false
+    delegate :mentor_teachers, :ect_teachers, to: :school, prefix: false
```
cpjmcquillan: Do we still need to add this?

cpjmcquillan (approved):

app/services/gias/reconciliation/merge.rb
thenapking: Good spot, thanks! We don't.

PR #3368, opened 2026-08-12
Remove `role="main"` from `<main>` element

peteryates:
This is no longer needed as <main>'s default role is "main".

It used to be done on GOV.UK services to provide some backwards compatibility with legacy screen readers, but now GOV.UK doesn't do it and it's been removed from the Design System documentation.


cpjmcquillan (approved):

avinhurry (approved):
~~This looks technically safe but the design system layout still includes it in [its examples](https://design-system.service.gov.uk/styles/layout/), do you have a link to the doc change you mentioned or have they just been completely removed?~~

Ignore this, I was being blind, this looks good!

PR #3374, opened 2026-08-13
Alias `dfe_sign_in_user_id` to `id` so it's picked up by dfe-analytics

peteryates:
We are currently missing user identifiers for non-DfE staff in our analytics.

Our DfE Analytics setup has called `#id` on the `current_user` object since it was introduced.

This change adds an `id`->`dfe_sign_in_user_id` alias to:

* `Sessions::Users::SchoolUser`
* `Sessions::Users::AppropriateBodyUser`

Those classes represent users who sign in via DfE Sign-in, they don't have `#id` and instead use `#dfe_sign_in_user_id`.

Aliasing to `#id` provides a quick way to test this mapping to ensure dataflows through correctly.

## Next steps

If all is successful it will probably make things more robust if we replace the alias with `#dfe_analytics_user_id` so it's clearer (without needing comments) why we're doing it.

### DIFF
--- a/app/services/sessions/users/appropriate_body_user.rb
+++ b/app/services/sessions/users/appropriate_body_user.rb
@@ -73,6 +73,12 @@ module Sessions
       # @return [Array<String>]
       alias_method :roles, :dfe_sign_in_roles
 
+      # dfe-analytics is configured to add the current_user#id to the web
+      # request log, so alias it here
+      #
+      # @return [String] the UUID returned by DfE Sign-in
+      alias_method :id, :dfe_sign_in_user_id
+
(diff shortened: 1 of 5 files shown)


avinhurry (commented):
This change undoes [this work](https://github.com/DFE-Digital/register-early-career-teachers-public/commit/d40dc32744b05a2be40ac3e36fddc6dded32be0c). 

Aliasing the DSI user ID to `id` means the raw UUID will start getting picked up by the existing Sentry/logging calls.

peteryates:
This change was replaced by #3382

PR #3376, opened 2026-08-14
Decouple serializer dependencies from query services

ethax-ross:
### Context

At the moment we define the associations to preload in the query services, which couples them to a particular serialzer and prevents them from being reused in other APIs.

We want to decouple the serializer/query and shift the responsibility of defining the dependencies into the serializer, which is where they would change.

The query can then accept any set of associations to include and work with different serializers that have different requirements.

- Define `dependencies` in serializers and update query services to accept `included_associations`.

### Guidance to review

Includes a few other minor changes such as removing unnecessary memoization, updating variable renames for clarity and reworking a shared context into a matcher.

We seem to re-query the models in some of the underlying teacher services, I think this was to avoid an N+1 in the response, but its not a good pattern and an N+1 in these endpoints isn't the end of the world so I've removed them.

I tried adding a test to ensure that only the required dependencies (and no extra) are added for each serializer, but this got complicated to test so I settled on a test that just ensures all required dependencies are specified.

This is a lot of changes, but its essentially the same pattern of changes across each model's controller/query/serializer.


app/controllers/api/v3/declarations_controller.rb
```diff
@@ -4,13 +4,13 @@ class DeclarationsController < APIController
       include API::TeacherType
 
       def index
-        conditions = {
+        query_arguments = {
```
craigmdavidson: Is there a more domain centric name than `query_arguments`? `declaration_filters` might express the domain more cleanly? 

Building this as a cohesive full filter would make it more obvious also? I.e.
```
  declaration_filters = lead_provider_filter.merge(
    contract_period_years: extract_conditions(contract_period_years, type: :integer),
    ...  
  ).compact
```

app/controllers/api/v3/declarations_controller.rb
```diff
@@ -48,18 +48,19 @@ def create_declaration_params
         params.require(:data).expect({ attributes: %i[participant_id declaration_type declaration_date course_identifier evidence_held] })
       end
 
-      def declarations_query(conditions: {})
-        API::Declarations::Query.new(**(default_query_conditions.merge(conditions)).compact)
+      def declarations_query(query_arguments: {})
```
craigmdavidson: I would consider inlining this? 
* It's not an _explaining method_, and; 
* it doesn't more clearly reveal intention? 

Instead it buries rather than surfaces the full filter? 
Building the full filter in the sunlight of the `action` makes what's happening more obvious?

app/controllers/api/v3/declarations_controller.rb
```diff
@@ -48,18 +48,19 @@ def create_declaration_params
...
+        API::Declarations::Query.new(**(base_query_arguments.merge(query_arguments)).compact)
       end
 
-      def default_query_conditions
-        @default_query_conditions ||= {
+      def base_query_arguments
```
craigmdavidson: Is burying `included_associations` here rather than as a distinct query parameter as intention revealing as it might be? 

I wonder if it is that "burying" that causes the method name to be more abstract / less intention revealing? 

The clearer name sans included_associations would be `lead_provider_filter`?

app/controllers/api/v3/declarations_controller.rb
```diff
@@ -92,7 +93,11 @@ def delivery_partner_api_ids
       end
 
       def to_json(obj)
-        API::DeclarationSerializer.render(obj, root: "data", **serializer_options)
+        serializer.render(obj, root: "data", **serializer_options)
+      end
+
+      def serializer
```
craigmdavidson, line 91: Inline?

app/controllers/api/v3/declarations_controller.rb
```diff
@@ -92,7 +93,11 @@ def delivery_partner_api_ids
       end
 
       def to_json(obj)
-        API::DeclarationSerializer.render(obj, root: "data", **serializer_options)
+        serializer.render(obj, root: "data", **serializer_options)
```
craigmdavidson: If we do the above the need for a `serializer_options` method goes away? it is just:

`API::DeclarationSerializer.render(obj, root: "data", **lead_provider_filter)`

Which tells you everything you need to know in a single line?

app/controllers/api/v3/declarations_controller.rb
```diff
@@ -4,13 +4,13 @@ class DeclarationsController < APIController
...
       def index
-        conditions = {
+        query_arguments = {
           contract_period_years: extract_conditions(contract_period_years, type: :integer),
           teacher_api_ids: extract_conditions(teacher_api_ids, type: :uuid),
           delivery_partner_api_ids: extract_conditions(delivery_partner_api_ids, type: :uuid),
           updated_since:,
         }
-        paginated_declarations = paginate(declarations_query(conditions:).declarations)
+        paginated_declarations = paginate(declarations_query(query_arguments:).declarations)
```
craigmdavidson: Filters and associations to include are different things. "Globbing" them into a single hash obfuscates rather than elucidates I think? 

Having something like:
```
paginate(
  API::Declarations::Query.new(
    **declaration_filters, 
    included_associations: API::DeclarationSerializer.dependencies
  ).declarations
)
```

would be much clearer I think? Reading the action definition would tell me everything I need to know, and sets me up for looking at the specifics if we need to?

craigmdavidson (commented):
LGTM, some thoughts on the query object interface design. 
If you disagree let me know and I'll trigger the approval.

app/controllers/api/v3/declarations_controller.rb
craigmdavidson: If we wanted to simplify without obfuscating we could have:

```
declaration_filters = lead_provider_filter
  .merge(contact_period_filter)
  ...
```

The intent would still be obvious, but a slightly better signal to noise ratio?

ethax-ross: Yeah that's fair, `base_query_arguments` doesn't reveal much

app/controllers/api/v3/declarations_controller.rb
ethax-ross: I think there's an argument for keeping `declarations_query` extracted. If we inline it we leave ourselves more open to someone not passing `lead_provider_id` to the query in a new action and/or (less importantly) not loading the associations. 

Extracting the query with filtering/preloading that is applicable to all actions feels more robust without adding much complexity. It also eludes to the fact all actions share the same query/associations, which I like 🤔

app/controllers/api/v3/declarations_controller.rb
ethax-ross: I'll look at inlining the query arguments and making it more explicit 👍

app/controllers/api/v3/declarations_controller.rb
ethax-ross: I don’t see much benefit in inlining this; keeping the serializer behind a named method gives us a consistent, single place to define the controller’s serializer, with very little indirection 🤔

app/controllers/api/v3/declarations_controller.rb
ethax-ross: Yeah agreed that's nicer 👍

app/controllers/api/v3/declarations_controller.rb
ethax-ross: Agree the combining under `base_query_arguments` isn't ideal, but not sold on inlining the query for the reasons above. I'll have a play around and see if I can find a happy middle ground 🙂

app/controllers/api/v3/declarations_controller.rb
craigmdavidson: If you want to make it a hard constraint then surface that? 
Make API::Declarations::Query take a lead_provider parameter such that it is no longer optional rather than burying it?
I.e.
`API::Declarations::Query.new(lead_provider, ...`

If you are going to extracted it it, to remain clear in needs to be a "pure" function, otherwise you are having side effects that are not obvious? As soon as you have the pure function, there is little advantage in the method, as it adds explanation nor elides noise?

craigmdavidson (approved):
LGTM

PR #3377, opened 2026-08-14
Product team users can input a data fixes CSV

cpjmcquillan:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/4251

This build out the first step of the `Admin::DataFixes` user journey.

Product team users can input rows of data fixes in CSV format via the UI.

The CSV is validated and parsed, and the proposed changes are stored in the
session.

A wizard approach has been taken, since this felt like a straightforward, but
multi-step, journey —and this is a pattern we use across the application.

This journey is currently restricted to review apps only, since it requires
`ENABLE_ADMIN_DATA_FIXES` to be enabled.

There is also no way to navigate to this page directly yet. Discussions are
ongoing to decide how we want users to access the journey.

We expect to iterate on the design and content as we build this journey out.

### DIFF
--- a/app/controllers/admin/data_fixes_controller.rb
+++ b/app/controllers/admin/data_fixes_controller.rb
@@ -1,0 +1,53 @@
+module Admin
+  class DataFixesController < AdminController
+    include WizardStoreRescuable
+
+    before_action :set_steps,
+                  :set_store,
+                  :set_wizard
+
+    before_action -> { redirect_to "/404", as: :not_found },
+                  unless: -> { wizard_class.step?(@current_step) }
+
+    before_action -> { @wizard.reset },
+                  if: -> { @current_step == :csv },
+                  unless: -> { wizard_class.step?(@previous_step) },
+                  only: :new
+
+    def new
+      render @current_step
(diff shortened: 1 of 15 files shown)


app/controllers/admin/data_fixes_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+      if @wizard.save!
+        redirect_to @wizard.next_step_path
+      else
+        render @current_step, status: :unprocessable_content
+      end
+    end
+
+  private
+
+    def authorised? = current_user&.dfe_user? && current_user.product_team?
```
thenapking, line 31: Can you have `product_team` access without being a `dfe_user`?

thenapking (approved):
LGTM.  One small query, but otherwise looks fine.

app/controllers/admin/data_fixes_controller.rb
cpjmcquillan: No, but only dfe users have the `role` enum. This adopts the same approach for authorising finance users to keep things simple.

PR #3378, opened 2026-08-14
[4198] Add trs_response status and trs_redirected_to on teachers table and remove flags

rwrrll:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/4198

Replaces the `trs_not_found` and `trs_deactivated` booleans with a single `trs_response` status plus a `trs_redirected_to` column for the destination TRN on a 308.

Also includes the [backfill assigned to #4152](https://github.com/DFE-Digital/register-ects-project-board/issues/4152), since the old columns can't be dropped until nothing reads them, and nothing can stop reading them while `trs_response` is null everywhere; status is derived from the flags and recovers destination TRNs out of `teacher_trs_merged` events.

Dropping the columns is a [separate PR](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3379) that can be merged after this one.

### Guidance to review

More moving parts than expected, so I tried to slice this into logical commits that can hopefully be reviewed independently. 

One bit worth noting is `syncable_with_trs?` treating a null `trs_response` as due a sync: the migration job in this PR will finish before the rolling deploy does, so anyone whose status changes in that window is written the old way and won't get backfilled. Leaving them syncable means they self-correct on next sync.

`analytics.yml` changes here rather than alongside the migration that drops the columns because `ignored_columns` takes them out of the model's field list, which `DfE::Analytics` validates.

n.b. [query 396](https://www.register-early-career-teachers.education.gov.uk/admin/blazer/queries/396-teachers-missing-from-trs) selects on `trs_not_found` and will break when the columns go.

Tagging @peteryates here because you had a comment on the backfill ticket!

### TODO before merging

- [ ] Notify data team of column changes

### DIFF
--- a/db/migrate/20260812163524_add_trs_response_to_teachers.rb
+++ b/db/migrate/20260812163524_add_trs_response_to_teachers.rb
@@ -1,0 +1,10 @@
+class AddTRSResponseToTeachers < ActiveRecord::Migration[8.1]
+  def change
+    create_enum :trs_responses, %w[ok not_found gone permanent_redirect]
+
+    change_table :teachers, bulk: true do |t|
+      t.enum :trs_response, enum_type: :trs_responses
+      t.string :trs_redirected_to
+    end
+  end
+end
--- a/db/migrate/20260813103015_backfill_trs_response_on_teachers.rb
+++ b/db/migrate/20260813103015_backfill_trs_response_on_teachers.rb
@@ -33,9 +33,11 @@ class BackfillTRSResponseOnTeachers < ActiveRecord::Migration[8.1]
                  substring(body from 'redirects to TRN (\\d+)') AS redirected_to
           FROM events
           WHERE event_type = 'teacher_trs_merged'
+            AND teacher_id BETWEEN #{first} AND #{last}
           ORDER BY teacher_id, created_at DESC
         ) AS merged
         WHERE merged.teacher_id = teachers.id
+          AND teachers.trs_response = 'permanent_redirect'
           AND merged.redirected_to IS NOT NULL
           AND teachers.trs_redirected_to IS NULL
           AND teachers.id BETWEEN #{first} AND #{last}
--- a/db/migrate/20260813103015_backfill_trs_response_on_teachers.rb
+++ b/db/migrate/20260813103015_backfill_trs_response_on_teachers.rb
@@ -1,0 +1,58 @@
+class BackfillTRSResponseOnTeachers < ActiveRecord::Migration[8.1]
+  disable_ddl_transaction!
+
+  BATCH_SIZE = 5_000
+
+  def up
+    each_batch do |first, last|
+      execute(<<~SQL)
+        UPDATE teachers SET trs_response = (
+          CASE
+            WHEN trs_deactivated THEN 'gone'
+            WHEN trs_not_found AND EXISTS (
+              SELECT 1 FROM events
+              WHERE events.teacher_id = teachers.id
+                AND events.event_type = 'teacher_trs_merged'
+            ) THEN 'permanent_redirect'
+            WHEN trs_not_found THEN 'not_found'
+            ELSE 'ok'
+          END
+        )::trs_responses
+        WHERE teachers.id BETWEEN #{first} AND #{last}
+          AND teachers.trs_response IS NULL
+          AND (teachers.trs_deactivated
+               OR teachers.trs_not_found
+               OR teachers.trs_data_last_refreshed_at IS NOT NULL)
+      SQL
+
+      execute(<<~SQL)
+        UPDATE teachers SET trs_redirected_to = merged.redirected_to
+        FROM (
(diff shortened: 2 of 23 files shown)


db/seeds/blazer_queries.rb
```diff
@@ -65,14 +65,19 @@ def create_query(creator:, name:, statement:, description:)
   },
   {
     name: "TRS not found",
-    statement: "SELECT id AS teacher_id, trs_not_found FROM teachers WHERE trs_not_found=true",
+    statement: "SELECT id AS teacher_id, trs_data_last_refreshed_at FROM teachers WHERE trs_response='not_found'",
```
avinhurry, line 68: Do we need to update any existing blazer queries as part of this change? Queries still using `trs_not_found` / `trs_deactivated` will break when #3379 drops those columns.

I know you've mentioned one in the description, is that the only one?

db/migrate/20260813103015_backfill_trs_response_on_teachers.rb
```diff
@@ -0,0 +1,58 @@
...
+      SQL
+
+      execute(<<~SQL)
+        UPDATE teachers SET trs_redirected_to = merged.redirected_to
+        FROM (
+          SELECT DISTINCT ON (teacher_id)
+                 teacher_id,
+                 substring(body from 'redirects to TRN (\\d+)') AS redirected_to
+          FROM events
+          WHERE event_type = 'teacher_trs_merged'
```
avinhurry: Nice to see us batching here, should we apply the current batch to the events subquery too?

```suggestion
          WHERE event_type = 'teacher_trs_merged'
            AND teacher_id BETWEEN #{first} AND #{last}
```

db/migrate/20260813103015_backfill_trs_response_on_teachers.rb
```diff
@@ -0,0 +1,58 @@
...
+          ORDER BY teacher_id, created_at DESC
+        ) AS merged
+        WHERE merged.teacher_id = teachers.id
+          AND merged.redirected_to IS NOT NULL
```
avinhurry, line 41: Optionally, since we're setting the response and redirect separately, we could add another `AND` here to ensure we only set `trs_redirected_to` for teachers we've classified as a permanent redirect.

```suggestion
          AND teachers.trs_response = 'permanent_redirect'
          AND merged.redirected_to IS NOT NULL
```

db/migrate/20260813103015_backfill_trs_response_on_teachers.rb
avinhurry, line 35: Nice to see us batching here, can we apply the current batch to the events subquery too please?

```suggestion
          WHERE event_type = 'teacher_trs_merged'
            AND teacher_id BETWEEN #{first} AND #{last}
```

avinhurry (commented):
I have one concern around applying the current batch to the events subquery, other than that i've left a couple of questions/suggestions.

Otherwise this looks good.

rwrrll:
Thanks @avinhurry -- committed both your changes! Now digging through production Blazer queries not stored in git (😬), I'll have to update these manually after deployment so I'll keep a note of what I find and do that as a manual process.

avinhurry (approved):
Looks great! Thanks @rwrrll

db/migrate/20260812163524_add_trs_response_to_teachers.rb
```diff
@@ -0,0 +1,10 @@
+class AddTRSResponseToTeachers < ActiveRecord::Migration[8.1]
+  def change
+    create_enum :trs_responses, %w[ok not_found gone permanent_redirect]
+
+    change_table :teachers, bulk: true do |t|
+      t.enum :trs_response, enum_type: :trs_responses
```
peteryates, line 6: In order to register a teacher we'll have to have had a positive result from TRS' API so maybe it'd be easier to default this to `ok`? That would have simplified the backfilling below as we only need to set the few hundred who've been deactivated or removed.

peteryates (approved):
Looks good. Left a suggestion on defaulting to `ok` which would simplify the backfill and the scope, but as the work's already done and it looks fine it's probably not worth revisiting.

spec/services/teachers/manage_spec.rb
```diff
@@ -391,4 +391,31 @@
...
+
+    context "when the teacher is not yet flagged" do
+      before { mark_teacher_as_merged }
+
+      it "records the TRN the teacher was merged into" do
+        teacher.reload
+
+        expect(teacher.trs_redirected_to).to eq(redirected_to)
+        expect(teacher).to be_trs_response_permanent_redirect
+        expect(teacher.trs_data_last_refreshed_at).to be_within(0.001.seconds).of(trs_data_last_refreshed_at)
```
cpjmcquillan: Should we `freeze_time` instead?

cpjmcquillan (approved):

PR #3380, opened 2026-08-14
[4298] add training period FK constraint to declarations

peterdavidhamilton:
### Context

DFE-Digital/register-ects-project-board/issues/4298

> We accidentally deleted a training period that belonged to a declaration.

Add the foreign key now that all the missing periods have been restored

### Guidance to review

https://www.register-early-career-teachers.education.gov.uk/admin/blazer/queries/478-missing-fk-declarations-training_period_id

### DIFF
--- a/app/models/training_period.rb
+++ b/app/models/training_period.rb
@@ -98,7 +98,7 @@ class TrainingPeriod < ApplicationRecord
   end
 
   # Callbacks
-  before_destroy :abort_destruction_when_billable_declarations_exist
+  before_destroy :destroy_destroyable_declarations
 
   # Scopes
   scope :for_ect, ->(ect_at_school_period_id) { where(ect_at_school_period_id:) }
@@ -199,11 +199,14 @@ class TrainingPeriod < ApplicationRecord
 
 private
 
-  def abort_destruction_when_billable_declarations_exist
-    return unless declarations.billable.exists?
+  def destroy_destroyable_declarations
+    if declarations.billable.exists?
+      errors.add(:base, "Cannot delete a training period with billable declarations")
+      throw(:abort)
+    end
 
-    errors.add(:base, "Cannot delete a training period with billable declarations")
-    throw(:abort)
+    declarations.destroyable.find_each(&:destroy!)
+    true
   end
 
   def only_one_at_school_period_present
--- a/spec/models/declaration_spec.rb
+++ b/spec/models/declaration_spec.rb
@@ -2,6 +2,7 @@ describe Declaration do
   describe "declarative updates" do
     def will_change_attribute(attribute_to_change:, new_value:)
       FactoryBot.create(:statement, id: new_value) if attribute_to_change.in?(%i[payment_statement_id clawback_statement_id])
+      FactoryBot.create(:training_period, id: new_value) if attribute_to_change == :training_period_id
     end
 
     describe "declarative touch target self" do
@@ -94,6 +95,28 @@ describe Declaration do
         let(:declaration_type) { :completed }
 
         it { is_expected.to have_error(:base, "Uplifts are only applicable to started declarations in an uplift enabled contract period") }
+      end
+    end
+
+    describe "training_period_id foreign key constraint" do
+      it "prevents creating a declaration that references a missing training period" do
+        declaration = FactoryBot.build(:declaration, training_period_id: TrainingPeriod.maximum(:id).to_i.next)
+
+        expect { declaration.save(validate: false) }.to raise_error(ActiveRecord::InvalidForeignKey)
+      end
+
+      it "prevents deleting a training period that has declarations" do
+        declaration = FactoryBot.create(:declaration)
+        training_period = declaration.training_period
+
+        expect { training_period.delete }.to raise_error(ActiveRecord::InvalidForeignKey)
+      end
+
+      it "allows destroying a training period that has declarations via callback" do
+        declaration = FactoryBot.create(:declaration)
+        training_period = declaration.training_period
+
+        expect { training_period.destroy! }.not_to raise_error
       end
     end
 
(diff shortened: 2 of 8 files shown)


peterdavidhamilton:
@cpjmcquillan is deleting events from the history how we should approach this? In unpicking [mentorship periods](https://www.register-early-career-teachers.education.gov.uk/admin/blazer/queries/479-missing-fk-declarations-mentorship_period_id) that also have orphaned FKs I understand the lack of events has made this harder for @boonali?

@peteryates I expected we would treat the papertrail as immutable, but I can also understand why we are purging events as we [delete periods](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/37961dea4800faafd766201dc0c2cd62d4490b7b/app/services/mentor_at_school_periods/finish.rb#L70). 

Is there room to instead nullify and update the event body to say something was deleted, so we still get the history that it happened?

spec/models/declaration_spec.rb
```diff
@@ -97,6 +98,28 @@ def will_change_attribute(attribute_to_change:, new_value:)
...
     end
 
+    describe "training_period_id foreign key constraint" do
+      it "prevents creating a declaration that references a missing training period" do
+        declaration = FactoryBot.build(:declaration, training_period_id: TrainingPeriod.maximum(:id).to_i.next)
+
+        expect { declaration.save(validate: false) }.to raise_error(ActiveRecord::InvalidForeignKey)
+      end
+
+      it "prevents deleting a training period that has declarations" do
```
ethax-ross, line 129: Feels a bit weird to test training period stuff here 🤔 would it be enough to just have `with_foreign_key(:training_period_id)` in the shoulda matcher for the association to cover all this?

app/models/training_period.rb
```diff
@@ -98,7 +98,7 @@ class TrainingPeriod < ApplicationRecord
   end
 
   # Callbacks
-  before_destroy :abort_destruction_when_billable_declarations_exist
+  before_destroy :destroy_destroyable_declarations
```
ethax-ross: `destroy_destroyable_declarations` feels a bit awkward, what about `before_destroy :destroy_non_billable_declarations` and switching the scope to `non_billable` instead of 'destroyable' which feels a bit more descriptive and generic at the same time 😅 the scope could then be inferred from the `billable` constant too:

```
scope :non_billable, -> { where.not(payment_status: BILLABLE_PAYMENT_STATUS) }
```

spec/models/training_period_spec.rb
```diff
@@ -932,6 +932,38 @@ def will_change_attribute(attribute_to_change:, new_value:)
         }
       end
     end
+
+    context "when there are both billable and non-billable declarations" do
```
ethax-ross, line 936: Do we want the happy path here as well where it only has non-billable declarations?

app/models/training_period.rb
peterdavidhamilton: You know I had something like that originally but I changed it at the last.

peteryates:
> @peteryates I expected we would treat the papertrail as immutable, but I can also understand why we are purging events as we [delete periods](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/37961dea4800faafd766201dc0c2cd62d4490b7b/app/services/mentor_at_school_periods/finish.rb#L70).

Sorry I missed this originally.

Yeah I always thought we'd leave the events in place if we deleted any referenced records.

The FKs are all `on_delete: nullify`. So if we get rid of a `mentor_at_school_period` for example, the event will still be attached to the `teacher`/`school` etc.

https://github.com/DFE-Digital/register-early-career-teachers-public/blob/85a3c7333988dfdb26c8417898dd8df83ee69d4a/db/schema.rb#L975-L995

spec/models/training_period_spec.rb
peterdavidhamilton: [Already under test](https://github.com/DFE-Digital/register-early-career-teachers-public/blob/475c6978e3b22825a286284f66f65a3ddfa9d843/spec/models/training_period_spec.rb#L907-L915)

spec/models/declaration_spec.rb
peterdavidhamilton: The matcher relies on an explict FK in the model, which because it is the default, is flagged as redundant by the linter. I also felt it warranted actually checking the db behaviour

ethax-ross: Ah makes sense 👍 maybe our own matcher would be useful for these at some point, something like:

```
foreign_keys = ActiveRecord::Base.connection.foreign_keys(table_name)

expect(foreign_keys).to include(
  an_object_having_attributes(
    column: expected_foreign_key
  )
)
```

We can think about this separately though 🙂

app/models/training_period.rb
```diff
@@ -199,11 +199,14 @@ def status
...
+  def destroy_non_billable_declarations
+    if declarations.billable.exists?
+      errors.add(:base, "Cannot delete a training period with billable declarations")
+      throw(:abort)
+    end
 
-    errors.add(:base, "Cannot delete a training period with billable declarations")
-    throw(:abort)
+    declarations.non_billable.find_each(&:destroy!)
+    true
```
cpjmcquillan, line 209: This changes the callback from a de facto validation on destroy, to a callback that **does something**.

I think it's the first of its kind across the application 🏆

Are we ok with this from a philosophical perspective? Does this set a precedent for us to use callbacks elsewhere?

I mention it because my understanding was we tended to avoid callbacks, although this hasn't been codified anywhere.

Personally I think there is a pretty compelling argument to use a callback to destroy these declarations, but I think it would be clearer if we separated the validation from the doing and used two distinct callbacks.

At the least, I think we ought to add some more context in the commit message about _why_ we decided to do it this way.

What do you think? 💭

ethax-ross: That's a good point; I like having this sort of logic in the model because it ties in to the other `dependent: :destroy` behaviour that can be on the associations defined here, but it is a bit of a step outside of how we do it elsewhere.

Is it worth taking the discussion here? https://github.com/DFE-Digital/register-ects-project-board/discussions/4375

peterdavidhamilton: https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3380/changes/9cf336d6e2ff21ecf06d2ae22e10759417e72bdb is the intent and backstory clear on the commit

PR #3381, opened 2026-08-14
[3260] DQT analysis

peterdavidhamilton:
### Context

- The justified use of mechanisms that bypass validation runs the risk of importing invalid data 
- Oversites in the schema can lead to orphaned records
- Legacy data imports have and will be shoehorned in to ECF2's new structure and need finessing afterwards
- Gaps and manipulations that naturally occur over the life of the service

A way to process and navigate invalid records within the service to aid in the triage of wonky data.

An iteration on the existing script runner that exports to CSV.

### DIFF
--- a/.github/workflows/check-validity.yml
+++ b/.github/workflows/check-validity.yml
@@ -1,0 +1,44 @@
+name: Check data validity
+
+on:
+  workflow_dispatch:
+
+permissions:
+  id-token: write
+
+jobs:
+  check-validity:
+    name: Analyse invalid records on migration
+    runs-on: ubuntu-latest
+    environment: migration
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v7
+
+      - uses: Azure/login@v3
+        with:
+          client-id: ${{ secrets.AZURE_CLIENT_ID }}
+          subscription-id: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
+          tenant-id: ${{ secrets.AZURE_TENANT_ID }}
+
+      - uses: DFE-Digital/github-actions/set-kubelogin-environment@master
+        with:
+          azure-client-id: ${{ secrets.AZURE_CLIENT_ID }}
+          azure-subscription-id: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
+          azure-tenant-id: ${{ secrets.AZURE_TENANT_ID }}
+
+      - uses: DFE-Digital/github-actions/set-kubectl@master
+
+      - name: Set AKS credentials
+        shell: bash
+        run: make ci migration get-cluster-credentials
+
+      - name: Truncate and re-run invalidity scan
+        shell: bash
+        run: |
+          kubectl -n cpd-production exec deployment/cpd-ec2-migration-web -- /bin/sh -c '
+            cd /app &&
+            DISABLE_DATABASE_ENVIRONMENT_CHECK=1 \
+            RAILS_ENV=migration \
+            /usr/local/bin/bundle exec rails runner "InvalidRecord.delete_all; CheckValidity.new.call"
+          '
--- a/db/seeds/blazer_queries.rb
+++ b/db/seeds/blazer_queries.rb
@@ -82,6 +82,33 @@ end
     name: "Active lead provider bands",
     statement: "SELECT * FROM active_lead_provider_bands ORDER BY active_lead_provider_id, allocation_order",
     description: "new data model"
+  },
+  {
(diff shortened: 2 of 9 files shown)


.github/workflows/check-validity.yml
```diff
@@ -0,0 +1,44 @@
...
+        run: make ci migration get-cluster-credentials
+
+      - name: Run invalidity scan
+        shell: bash
+        run: |
+          kubectl -n cpd-production exec deployment/cpd-ec2-migration-worker -- /bin/sh -c '
+            cd /app &&
+            DISABLE_DATABASE_ENVIRONMENT_CHECK=1 \
+            RAILS_ENV=migration \
+            /usr/local/bin/bundle exec rails runner "CheckValidity.new.call"
```
peterdavidhamilton, line 43: Another workflow might not be what we want but we could either invoke a rails task or the service in a runner

app/jobs/check_validity_job.rb
```diff
@@ -0,0 +1,9 @@
+class CheckValidityJob < ApplicationJob
+  queue_as :check_validity
```
peterdavidhamilton, line 2: I think having a queue that doesn't exist in prod is also a reasonable guard

config/recurring.yml
```diff
@@ -46,6 +46,12 @@ production:
     command: Blazer.send_failing_checks
     schedule: every day at 8am
 
+migration:
+  check_validity:
+    class: CheckValidityJob
+    queue: check_validity
+    schedule: every day at 3am
```
peterdavidhamilton: A scheduled job would rebuild the data in the queries ready for the next day

db/seeds/blazer_queries.rb
```diff
@@ -82,6 +82,33 @@ def create_query(creator:, name:, statement:, description:)
     name: "Active lead provider bands",
     statement: "SELECT * FROM framework_agreement_bands ORDER BY framework_agreement_id, allocation_order",
     description: "new data model"
+  },
+  {
+    name: "Invalid teachers",
+    description: "Fails model validations",
+    statement: <<~SQL.strip
+      SELECT record_id AS teacher_id,
```
peterdavidhamilton, line 90: `AS teacher_id` does the magic because we have configured Blazer already.

db/seeds/invalid_records.rb
```diff
@@ -0,0 +1,41 @@
+# Generate date that is invalid bypassing validations in order to test data triage
```
peterdavidhamilton, line 1: We might want this seed to be optional, but as is the teacher is clearly named to mitigate surprises

lib/tasks/check_validity.rake
```diff
@@ -0,0 +1,9 @@
+namespace :data do
```
peterdavidhamilton, line 1: Option to run on demand?

lib/check_validity.rb
```diff
@@ -0,0 +1,39 @@
...
+  class ProductionGuardError < StandardError; end
+
+  TABLES = %w[
+    teachers
+    appropriate_body_periods
+    induction_periods
+    induction_extensions
+    training_periods
+    declarations
+  ].freeze
```
peterdavidhamilton, line 18: This service is based off `db/scripts/export_invalid_records.rb` and as PoC only checks a few tables, but the team could trigger all tables or just one for a quicker result

spec/models/invalid_record_spec.rb
```diff
@@ -0,0 +1,40 @@
+RSpec.describe InvalidRecord do
```
peterdavidhamilton, line 1: Not sure the model needs any AR logic as it is an ephemeral bucket for Blazer to query against.

Blazer can use other data sources so we could even capture this data to another location that only exists in migration. That might be a segue to some other analysis tool

peterdavidhamilton (commented):
A rough idea floated to start the discussion.

peteryates:
I'm open to anything we can do to improve data quality.

My only reservation here is that we end up with a job writing data to a table nobody looks at.

If this is part of a bigger piece of work to identify bad data, fix it and push ourselves closer to 0 invalid records, I'm all for it.

Then, perhaps we can shift from reporting bad records to Sentry instead.

lib/check_validity.rb
peterdavidhamilton: Following the DQT import we have so many more teachers and inductions that a single process is not practical. 
You get usable results if you run the checker in parallel or are prepared to wait. Current testing is now only focused on induction periods.

PR #3382, opened 2026-08-14
Send user identifier to DfE Analytics

peteryates:
### Context

We don't currently send any user identifier to DfE Analytics which makes tracking a single user's journey through the site impossible.

This PR makes a few changes, it:

1. adds a `#dfe_analytics_user_id` method to all of the `Sessions::User` classes, which refer to either the `user.id` for database-backed users or the `#dfe_sign_in_user_id` for DfE Sign-in users. Users who can't be linked to either (School/AB Persona users) are `nil`
2. moves the existing fingerprint functionality from the `ApplicationController` to `Sesssions::Users::Fingerprint` and adds tests that cover it directly
3. adds a `API::AnalyticsUser` class which bridges the gap between a `LeadProvider` and a `User` by supplementing it with a `#fingerprint` method. This doesn't need to be kept secret so it just uses the `LeadProvider#id` (`#short_name` is a better fit here, see #2837)

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -24,7 +24,7 @@ class ApplicationController < ActionController::Base
     Sentry.set_user(
       email: current_user&.email,
       id: current_user.try(:id),
-      dfe_sign_in_user_fingerprint: current_user_dfe_sign_in_user_fingerprint
+      dfe_sign_in_user_fingerprint: current_user&.fingerprint
     )
   end
 
@@ -41,11 +41,6 @@ private
     super
     payload[:current_user_class] = current_user&.class&.name
     payload[:current_user_id] = current_user.try(:id)
-    payload[:current_user_dfe_sign_in_user_fingerprint] =
-      current_user_dfe_sign_in_user_fingerprint
-  end
-
-  def current_user_dfe_sign_in_user_fingerprint
-    current_user&.fingerprint
+    payload[:current_user_dfe_sign_in_user_fingerprint] = current_user&.fingerprint
   end
 end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -46,14 +46,6 @@ private
   end
 
   def current_user_dfe_sign_in_user_fingerprint
-    dfe_sign_in_user_id = current_user.try(:dfe_sign_in_user_id)
-
-    return if dfe_sign_in_user_id.blank?
-
-    OpenSSL::HMAC.hexdigest(
-      "SHA256",
-      Rails.application.secret_key_base,
-      dfe_sign_in_user_id
-    )
+    current_user&.fingerprint
   end
 end
--- a/app/services/api/analytics_user.rb
+++ b/app/services/api/analytics_user.rb
@@ -1,0 +1,14 @@
+module API
+  class AnalyticsUser
+    attr_reader :lead_provider
+
+    def initialize(lead_provider)
+      @lead_provider = lead_provider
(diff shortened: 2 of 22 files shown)


app/services/api/analytics_user.rb
```diff
@@ -0,0 +1,14 @@
...
+  class AnalyticsUser
+    attr_reader :lead_provider
+
+    def initialize(lead_provider)
+      @lead_provider = lead_provider
+    end
+
+    delegate :name, to: :lead_provider
+
+    # NOTE: would short_name make more sense here?
```
peteryates, line 11: Short names didn't actually land (#2837) but hopefully might soon.

app/controllers/application_controller.rb
```diff
@@ -46,14 +46,6 @@ def append_info_to_payload(payload)
   end
 
   def current_user_dfe_sign_in_user_fingerprint
```
peteryates: Perhaps this'd be better as `#current_user_fingerprint`?

avinhurry (approved):
Nicely done, this looks good.

One thing for the analytics team to be aware of is that OTP users won't have a fingerprint, I think that's fine and arguably preferable.

cpjmcquillan (approved):

peteryates:
> One thing for the analytics team to be aware of is that OTP users won't have a fingerprint, I think that's fine and arguably preferable.

Do you mean `OTPSchoolUser` @avinhurry ?  Yeah I left that out as I wasn't sure if it would ever make a return.

But regular OTP users (`DfEUser` and `DfEUserImpersonatingSchoolUser`) both include `Fingerprint` so _should_ work fine.

app/controllers/application_controller.rb
avinhurry: I agree now that it can fingerprint DfE staff as well as DfE sign in users.

spec/services/sessions/users/appropriate_body_persona_spec.rb
```diff
@@ -13,6 +13,10 @@
     let(:user_props) { { email:, name:, appropriate_body_period_id: } }
   end
 
+  it_behaves_like "an unfingerprintable user" do
```
ethax-ross: `an unidentifiable user`?

spec/services/sessions/users/dfe_user_impersonating_school_user_spec.rb
```diff
@@ -1,4 +1,4 @@
-describe "Sessions::Users::DfEUserImpersonatingSchoolUser" do
+describe Sessions::Users::DfEUserImpersonatingSchoolUser do # rubocop:disable RSpec/SpecFilePathFormat
```
ethax-ross, line 1: Could also add to `RSpec/SpecFilePathFormat`, though it doesn't just accept `DfE` annoyingly, you have to remap the whole spec:

```
RSpec/SpecFilePathFormat:
  CustomTransform:
    DfEUserImpersonatingSchoolUser: dfe_user_impersonating_school_user
```

ethax-ross (approved):
Looks good 👍

spec/services/sessions/users/dfe_user_impersonating_school_user_spec.rb
peteryates: Yeah, I'll do this as a follow up - it was bugging me too :sweat_smile:

peteryates:
@tbayer-dfe has confirmed this looks like it's working correctly :tada:

PR #3388, opened 2026-08-17
[4292] Fix metadata refresh bug

tonyheadford:
### Context

When the school changes on a `ECTAtSchoolPeriod`, `MentorAtSchoolPeriod` or `SchoolPartnership` the metadata handlers kick in and update the school-related metadata associated with the _new school_ not the original school.  This leads to inconsistent metadata records for the original school, which are picked up and reported in Sentry by an overnight reconciliation job.

Adds an additional `refresh_metadata` that uses the old `school_id` as the target

### DIFF
--- a/spec/models/ect_at_school_period_spec.rb
+++ b/spec/models/ect_at_school_period_spec.rb
@@ -5,6 +5,20 @@ describe ECTAtSchoolPeriod do
       let!(:target) { FactoryBot.create(:school) }
 
       it_behaves_like "a declarative metadata model", on_event: %i[create destroy update]
+
+      context "when changing :school_id", :with_metadata do
+        let!(:manager) { instance_double(Metadata::Manager, refresh_metadata!: nil) }
+        let!(:new_school) { FactoryBot.create(:school) }
+
+        before do
+          allow(Metadata::Manager).to receive(:new).and_return(manager)
+        end
+
+        it "refreshes the metadata of the previous school" do
+          instance.update!(school: new_school)
+          expect(manager).to have_received(:refresh_metadata!).with(target)
+        end
+      end
     end
 
     describe "teacher target" do
--- a/spec/models/ect_at_school_period_spec.rb
+++ b/spec/models/ect_at_school_period_spec.rb
@@ -6,18 +6,12 @@ describe ECTAtSchoolPeriod do
 
       it_behaves_like "a declarative metadata model", on_event: %i[create destroy update]
 
-      context "when changing :school_id", :with_metadata do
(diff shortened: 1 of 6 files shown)


cpjmcquillan (approved):

spec/models/ect_at_school_period_spec.rb
```diff
@@ -5,6 +5,20 @@
...
+      context "when changing :school_id", :with_metadata do
+        let!(:manager) { instance_double(Metadata::Manager, refresh_metadata!: nil) }
+        let!(:new_school) { FactoryBot.create(:school) }
+
+        before do
+          allow(Metadata::Manager).to receive(:new).and_return(manager)
+        end
+
+        it "refreshes the metadata of the previous school" do
+          instance.update!(school: new_school)
```
ethax-ross: I think this is incidentally passing because of a `refresh_metadata` we have when creating any school:

```
refresh_metadata -> { self }, on_event: %i[create]
```

If you log out what `DeclarativeUpdates` is actually doing after calling `instance.update!(school:)` it doesn't refresh the old school's metadata on the `update` hook:

```
updating old school id: 23259, new school id: **23260**
refreshing metadata of school id: 23260 ([:create, :destroy, :update] | [])
refreshing metadata of school id: 24662 ([:create, :destroy, :update] | [])
refreshing metadata of school id: **23260** ([:update] | [:school_id])
```

If you change `school_id_was` to `school_id_before_last_save` it does:

```
updating old school id: 23271, new school id: **23272**
refreshing metadata of school id: 23272 ([:create, :destroy, :update] | [])
refreshing metadata of school id: 24673 ([:create, :destroy, :update] | [])
refreshing metadata of school id: **23271** ([:update] | [:school_id])
```

I wonder if we could leverage the existing shared context for testing metadata refreshes though 🤔 I had a go but didn't quite get it all passing:

```
describe "previous school target" do
  let(:instance) { FactoryBot.create(:ect_at_school_period, :unfinished, school: target) }
  let!(:target) { FactoryBot.create(:school) }

  def will_change_attribute(attribute_to_change:, new_value:)
    DeclarativeUpdates.skip(:metadata) { FactoryBot.create(:school, id: new_value) } if attribute_to_change == :school_id
  end

  it_behaves_like "a declarative metadata model", on_event: %i[update], when_changing: %i[school_id], target_optional: false
end
```

Do you want to have a crack at it?

tonyheadford: Fantastic catch!
Yes, I'll take a look, I did try and use them for this but noticed that there were assumptions around `target` in there which is different in this case, thanks

ethax-ross (approved):

PR #3390, opened 2026-08-17
[4299] Prevent deleting TrainingPeriod with billable declarations

rwrrll:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/4299

As mentioned in the ticket, a foreign key constraint enforces this at the database layer, and the change here is to catch that in the application layer with a model error.

Simple `dependent: :restrict_with_error` doesn't fit here because we only care about _billable_ declarations, so I've used a destroy callback instead.

### DIFF
--- a/app/models/training_period.rb
+++ b/app/models/training_period.rb
@@ -97,6 +97,9 @@ class TrainingPeriod < ApplicationRecord
     validates :deferral_reason, absence: true
   end
 
+  # Callbacks
+  before_destroy :abort_destruction_when_billable_declarations_exist
+
   # Scopes
   scope :for_ect, ->(ect_at_school_period_id) { where(ect_at_school_period_id:) }
   scope :for_mentor, ->(mentor_at_school_period_id) { where(mentor_at_school_period_id:) }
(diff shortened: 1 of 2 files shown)


app/models/training_period.rb
```diff
@@ -97,6 +97,9 @@ class TrainingPeriod < ApplicationRecord
     validates :deferral_reason, absence: true
   end
 
+  # Callbacks
+  before_destroy :abort_destruction_when_billable_declarations_exist
```
thenapking, line 101: How does this interact with the various services which do destroy training periods?  I've had a look and mostly it looks like we destroy unstarted training_periods or ones with no school_partership in various cases, eg when changing lead provider via the wizard. I assume in those cases there can't be declarations, let alone billable ones.  It would be useful to have a record on this PR to confirm that this works with anything which does destroy a training period, if tests aren't really appropriate.

rwrrll: Yeah there is a spec that covers this! The spec deletes an ECTAtSchoolPeriod and checks that it doesn't succeed because this model having billable declarations prevents it.

thenapking: Yes, I get that.  What I am wondering about is service classes which destroy training periods.  If for instance the `ChangeLeadProvider` wizard failed because there was a billable declaration attached to a training period the background service tried to destroy, we'd raise an error, but the wizard would fail in a way that wouldn't be meaningful to the end user.  As it happens, I don't think this change will affect the `ChangeLeadProvider` wizard, because I don't think it would be destroying training_periods with declarations.  What I am asking is - have you checked all the places training_periods get destroyed to see whether this might cause an unexpected interaction?

cpjmcquillan: I think the idea is that we can deal with those scenarios if/when they arise, rather than pre-empting them.

The idea being that an error in Sentry is something we can action, and preferable to orphaned declarations (which is what would happen now).

cpjmcquillan (approved):

PR #3391, opened 2026-08-17
[4017] Do not update metadata on destroyed objects

thenapking:
### Context

When trying to update the `GIAS::Reconciliation::Merge` service so that it removed the old schools, I hit an issue with `DeclarativeUpdates`.  This PR resolves that issue by ensuring we check whether the object referenced by the metadata exists, before updating its metadata.

Check the target of the metadata refresh exists before updating its metadata.

We also ensure that when a school is destroyed, its metadata is cleaned up to.

We test the situation where the object and the target of the metadata update have been destroyed in the same transaction. For instance if a `SchoolPartnership` and a `School` were destroyed in the same transaction, the destruction of the `SchoolPartnership` would have previously caused a metadata update on the school.  If the `School` was also destroyed this would raise an error.

### DIFF
--- a/app/models/concerns/declarative_updates.rb
+++ b/app/models/concerns/declarative_updates.rb
@@ -11,11 +11,17 @@ module DeclarativeUpdates
       after_commit(on: on_event) do
         next if DeclarativeUpdates.skip?(:metadata)
 
+        evaluated_target = instance_exec(&target)
+        next unless evaluated_target
+        next unless evaluated_target.class.exists?(evaluated_target.id)
+
         should_touch = destroyed? || when_changing.blank? || when_changing.any? do |attr|
           saved_change_to_attribute?(attr)
         end
 
-        Metadata::Manager.new.refresh_metadata!(instance_exec(&target)) if should_touch
+        next unless should_touch
+
+        Metadata::Manager.new.refresh_metadata!(evaluated_target)
       end
     end
 
--- a/spec/support/shared_examples/declarative_updates.rb
+++ b/spec/support/shared_examples/declarative_updates.rb
@@ -348,6 +348,24 @@ RSpec.shared_examples "a declarative metadata model", :with_metadata do |when_ch
 
           expect(manager).to have_received(:refresh_metadata!).with(target)
         end
+
+        context "when the target is also destroyed in the same transaction" do
+          it "does not refresh the metadata" do
+            # In order for the target to be destroyed in the same transaction,
+            # we need to ensure that anything associated with the target that would prevent its destruction is also destroyed.
+            dependent_associations = %i[ect_at_school_periods mentor_at_school_periods school_partnerships]
+
+            ActiveRecord::Base.transaction do
+              instance.destroy!
+              dependent_associations.each do |association|
+                target.send(association).destroy_all if target.respond_to?(association)
+              end
+              target.destroy!
+            end
+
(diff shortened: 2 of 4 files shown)


app/models/concerns/declarative_updates.rb
```diff
@@ -11,11 +11,17 @@ def refresh_metadata(target, when_changing: [], on_event: %i[update])
       after_commit(on: on_event) do
         next if DeclarativeUpdates.skip?(:metadata)
 
+        evaluated_target = instance_exec(&target)
+        next unless evaluated_target
```
ethax-ross, line 15: I think these could be simplified/collapsed into:

```
next unless evaluated_target&.persisted?
```

spec/support/shared_examples/declarative_updates.rb
```diff
@@ -348,6 +348,24 @@ def generate_new_value(attribute_to_change:)
 
           expect(manager).to have_received(:refresh_metadata!).with(target)
         end
+
+        context "when the target is also destroyed in the same transaction" do
+          it "does not refresh the metadata" do
+            # In order for the target to be destroyed in the same transaction,
+            # we need to ensure that anything associated with the target that would prevent its destruction is also destroyed.
+            dependent_associations = %i[ect_at_school_periods mentor_at_school_periods school_partnerships]
```
ethax-ross: As this is a shared context I think we should avoid including particular named associations; given we're in a spec here I'd opt for a hand grenade option that should work for any future `target`:

```
ActiveRecord::Base.connection.disable_referential_integrity { target.destroy! }
```

ethax-ross (commented):
Thanks Tom, looks great just a couple of minor suggestions

app/models/concerns/declarative_updates.rb
thenapking: I tried this suggestion (along with previously trying `evaluated_target&.destroyed?`) and found that it caused the class where I am using this to error.  This existence check didn't:
```
evaluated_target.class.exists?(evaluated_target.id)
```
Having dug into that further I've found that if there are multiple in-memory references to the target within a transaction (as there can be in my service class) `evaluated_target&.persisted?` can be `true` even when the target has been destroyed.  The existence check however always catches the situation even if it is pretty verbose.  I have updated the spec to cover this, and added an explanatory note, which hopefully will mean it is not too obscure.  That test fails if we just use `evaluated_target&.persisted?` 
I also found the test failed in some circumstances unless there was a specific guard first, ie `next unless evaluated_target`.  Even though both kinds of target (schools and teachers) are required foreign keys, in the situation where we've deleted the target during the transaction it is possible `evaluated_target` is `nil`.  This line I think will evaluate `evaluated_target.id` before `evaluated_target&.class&.exists?`, so if it is `nil` will also error.

ethax-ross: Ah that makes sense, I guess `persisted?/destroyed?` are querying the state of a particular instance whereas `exists?` will actually go to the DB and check 👍

ethax-ross (approved):
LGTM! thanks Tom!

app/models/concerns/declarative_updates.rb
thenapking: Yes, that's it.  Was confusing because it worked for everything... except for when I applied it to the other branch I was working on, which is where I need to fix the issue.

PR #3393, opened 2026-08-18
Introduce `Admin::DataFixes::Changes` class

cpjmcquillan:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/4272

This PORO is intended to act as some more of the plumbing between a user submitting a form with some CSV as text, and eventually driving some data fix changes.

It takes the parsed rows from an `InlineCSV` and passes them to the `Processor` for processing. Errors are accumulated over each row. If all rows are processed successfully, then users will proceed to the next step of the wizard (where changes will be committed).

The class isn't used anywhere yet, but will be used in the `Admin::DataFixesWizard`.

### DIFF
--- a/app/services/admin/data_fixes/changes.rb
+++ b/app/services/admin/data_fixes/changes.rb
@@ -1,0 +1,31 @@
+module Admin::DataFixes
+  class Changes
+    include ActiveModel::Model
+    include ActiveModel::Attributes
+    include ActiveModel::Validations
+
+    attribute :parsed_rows
+
+    validate :all_results_successful
+
+    def process
+      return false unless valid?
+
+      results.map(&:saved_change)
+    end
+
+  private
+
+    def results = @results ||= parsed_rows.map { process_row(it) }
+
+    def process_row(row)
+      Processor.new.process!(data_change: row.with_indifferent_access)
+    end
+
+    def all_results_successful
+      results.each.with_index do |result, index|
+        errors.add(:base, "Row #{index + 1}: #{result.error}") unless result.success?
+      end
+    end
+  end
+end
--- a/spec/services/admin/data_fixes/changes_spec.rb
+++ b/spec/services/admin/data_fixes/changes_spec.rb
@@ -1,0 +1,119 @@
+RSpec.describe Admin::DataFixes::Changes do
+  subject(:changes) { described_class.new(parsed_rows:) }
+
+  describe "#process" do
+    subject(:process) { changes.process }
+
+    context "when none of the changes are valid" do
+      let!(:teacher) { FactoryBot.create(:teacher) }
+
+      let(:parsed_rows) do
+        [
+          {
+            "object_type" => "Teacher",
+            "object_id" => teacher.id.to_s,
+            "action" => "update",
+            "attributes" => "trn,,trs_first_name,New Name"
+          },
+          {
+            "object_type" => "ECTAtSchoolPeriod",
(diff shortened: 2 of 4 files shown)


spec/services/admin/data_fixes/changes_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+  subject(:changes) { described_class.new(parsed_rows:) }
+
+  describe "#process" do
+    subject(:process) { changes.process }
+
+    context "when none of the changes are valid" do
+      let!(:teacher) { FactoryBot.create(:teacher) }
+
+      let(:parsed_rows) do
+        [
```
thenapking, line 11: It feels like this would be better if we mocked the response of the `Process` class because all we care about here is whether `success?` is `true`.

app/services/admin/data_fixes/changes.rb
```diff
@@ -0,0 +1,29 @@
+module Admin::DataFixes
+  class Changes
+    include ActiveModel::Model
+    include ActiveModel::Attributes
+    include ActiveModel::Validations
+
+    attribute :parsed_rows
+
+    validate :successful_results
```
thenapking: I'm not sure about the naming here.  The results might not be "successful" in the sense that some of them have errors.  Perhaps `processed_results` would be better? What do you think?

app/services/admin/data_fixes/changes.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  private
+
+    def results = @results ||= parsed_rows.map { process_row(it) }
+
+    def process_row(row)
+      Processor.new.process!(data_change: row.with_indifferent_access)
+    end
+
+    def successful_results
```
thenapking: I'm not sure about the naming here. It includes results that weren't processed because of errors, and so aren't "successful" in that sense. Perhaps `processed_results` would be better?   What do you think?

app/services/admin/data_fixes/changes.rb
thenapking: I'm not really sure attaching validation errors to the whole set of results is the right approach.  We loose the opportunity to identify which row errored, which might help users find the rows in a long CSV file.  I think that either exposing the list of failures, or somewhere recording the row number would be useful.

thenapking (commented):
Obviously this is quite bare bones at the moment and how it develops depends on what the wizard will look like.

app/services/admin/data_fixes/changes.rb
cpjmcquillan: Yeah, the wording is a little confusing I guess. I've changed it to

```ruby
validate :all_results_successful
```

app/services/admin/data_fixes/changes.rb
cpjmcquillan: We were always returning all the errors. I've tweaked it to point back to the row number, though.

spec/services/admin/data_fixes/changes_spec.rb
cpjmcquillan: To be honest I started out in that direction, but the tests were so cumbersome to write and janky to read I binned them!

I think a key thing is whether these document the behaviour of the `Changes` class clearly enough. 

I think the tests strike a balance with testing something useful and telling a story, so I'd be inclined to keep them like this.

thenapking (approved):
LGTM

PR #3394, opened 2026-08-18
Improve DQT induction period parser rejection CSV triage output

peterdavidhamilton:
### Context

Improves the DQT induction period importer's rejection logging so analysts can reconcile every source row against what was imported and what was discarded. 

Previously several branches silently dropped source rows without writing them to the rejected CSV, so `source ≠ imported + rejected` and the triage output was incomplete.           
    
- Rely on CSV only for prod runs and reduce noise in downloaded from the pod
- Fill gaps in rejected data logging each rejected entry for a parity check
- Consistent imported vs discarded comparison

### Guidance to review

| reason | trn | dqt_id_imported | dqt_id_discarded | started_on_imported | started_on_discarded | finished_on_imported | finished_on_discarded |
|---|---|---|---|---|---|---|---|
| cannot be imported because the induction period is 1 day or shorter | 1234567 | | 9f8c2a1b-3d4e-4f5a-8b6c-7d8e9f0a1b2c | | 2012-01-01 | | 2012-01-01 |
| an overlapping induction period was discarded as fully contained within the retained period | 1234567 | 9f8c2a1b-3d4e-4f5a-8b6c-7d8e9f0a1b2c | 9f8c2a1b-3d4e-4f5a-8b6c-7d8e9f0a1b2c | 2012-06-30 | 2012-08-15 | 2013-06-30 | 2013-03-15 |
| cannot be imported because the induction period is fully absorbed into the retained period which was extended to cover it | 1234567 | 9f8c2a1b-3d4e-4f5a-8b6c-7d8e9f0a1b2c | 9f8c2a1b-3d4e-4f5a-8b6c-7d8e9f0a1b2c | 2012-01-01 | 2012-03-03 | 2012-04-04 | 2012-05-05 |
| two induction periods with different appropriate bodies where one contains the other (1 of 2) | 1234567 | | 5a1b2c3d-4e5f-4a6b-8c7d-9e0f1a2b3c4d | | 2022-02-02 | | 2022-03-03 |
| two induction periods with different appropriate bodies where one contains the other (2 of 2) | 1234567 | | 9f8c2a1b-3d4e-4f5a-8b6c-7d8e9f0a1b2c | | 2022-01-01 | | 2022-05-05 |


boonali:
Context and changes mentioned match up with what we've discussed...and to my very untrained eyes, the code looks good!

lib/appropriate_bodies/importers/induction_period_parser.rb
```diff
@@ -212,66 +223,83 @@ def rows
...
-            log_error("cannot be imported because TRN or AB is missing",
-                      trn: row.trn,
-                      dqt_id: [row.legacy_appropriate_body_id, nil])
+            log_error(
+              "cannot be imported because TRN or AB is missing",
+              trn: row.trn,
+              dqt_id: [nil, row.legacy_appropriate_body_id],
+              started_on: [nil, row.started_on],
+              finished_on: [nil, row.finished_on]
+            )
```
craigmdavidson, line 237: `log_error` will now always return false in production?, which means it won't be rejected. But the chaining represents an expectation that these failures will be rejected? E.g. we don't want to try to sort by a missing trn?

lib/appropriate_bodies/importers/induction_period_parser.rb
```diff
@@ -537,8 +628,10 @@ def genuine_data?
     def log_error(message, trn:, dqt_id: [nil, nil], started_on: [nil, nil], finished_on: [nil, nil])
       log_error_csv(message, trn:, dqt_id:, started_on:, finished_on:)
 
-      uuids = dqt_id.join(", ")
-      logger.error("#{message} trn: #{trn} dqt_id: #{uuids}")
+      if Rails.env.development? || Rails.env.test?
+        uuids = dqt_id.join(", ")
+        logger.error("#{message} trn: #{trn} dqt_id: #{uuids}")
+      end
```
craigmdavidson, line 640: This changes the return from `true` to `false` when this runs on Production, since logger.error would return true?

I'm not sure if it is a great idea to rely on logger.error as a return type anyway?

craigmdavidson (commented):
There is some pretty hairy logic to try to reason about in here?! 🤯 

I do have a query regarding behaviour change and it's impact? It might be simply that returning true in log_error addresses this?

peterdavidhamilton:
A fresh pair of eyes is appreciated @craigmdavidson thank you. So the gist is logs were very hard for analysts to consume, so at each leaf of the branch logic we now save to CSV any rejected period row that was trying to be imported. At the same time silencing the logs unless in dev/test environment, I am happy to revert that, but at this stage no one is looking at them.

lib/appropriate_bodies/importers/induction_period_parser.rb
peterdavidhamilton: Great spot! I'll add explict return values 🙇🏼

peterdavidhamilton: I felt the duplication was better than letting the logging be loadbearing

craigmdavidson (approved):
LGTM

PR #3395, opened 2026-08-18
[4356] Allow lowering band capacity but not below payment declaration count

tonyheadford:
### Context

Change the editing of band capacity to allow a lower value providing that the new value is not less than the number of payment declarations for the provider in the contract period.

Change the validation to permit lowering the capacity of a band and add additional check that the new capacity + all the previous band capacities and not less than the number of payment declarations recorded by the provider against any of the statements in any of the contracts for the contract period.

### Guidance to review

Error message:

image

### DIFF
--- a/app/models/active_lead_provider.rb
+++ b/app/models/active_lead_provider.rb
@@ -9,6 +9,7 @@ class ActiveLeadProvider < ApplicationRecord
   has_many :events
   has_many :contracts
   has_many :statements, through: :contracts
+  has_many :payment_declarations, through: :statements
   has_many :bands, -> { order(allocation_order: :asc) }, class_name: "ActiveLeadProvider::Band"
 
   # Validations
@@ -38,10 +39,6 @@ class ActiveLeadProvider < ApplicationRecord
   delegate :finished_on_before_today?, to: :contract_period
 
   def editable? = !contract_period.payments_frozen?
-
-  def payment_declarations_count
-    contracts.joins(:payment_declarations).where(payment_declarations: { clawback_status: :no_clawback }).count
-  end
 
   def bands_can_be_added_and_removed?
     contracts.none? && contract_period.started_on.future?
--- a/app/models/active_lead_provider.rb
+++ b/app/models/active_lead_provider.rb
@@ -39,6 +39,10 @@ class ActiveLeadProvider < ApplicationRecord
 
   def editable? = !contract_period.payments_frozen?
 
+  def payment_declarations_count
+    contracts.sum(&:payment_declarations_count)
+  end
+
   def bands_can_be_added_and_removed?
     contracts.none? && contract_period.started_on.future?
   end
--- a/app/models/active_lead_provider.rb
+++ b/app/models/active_lead_provider.rb
@@ -40,7 +40,7 @@ class ActiveLeadProvider < ApplicationRecord
   def editable? = !contract_period.payments_frozen?
 
   def payment_declarations_count
-    contracts.sum(&:payment_declarations_count)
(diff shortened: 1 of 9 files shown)


app/models/contract.rb
```diff
@@ -59,4 +59,8 @@ def statement_range_description
 
     "#{first.month_year} - #{last.month_year}"
   end
+
+  def payment_declarations_count
+    statements.sum { it.payment_declarations.count }
```
ethax-ross: Is it worth having a `has_many, through` to avoid potential N+1s here?:

```
has_many :payment_declarations, through: :statements
```

You could then just run `contract.payment_declaratitons.count` which should `JOIN` and `count`.

Would that also resolve the need to have those extra `reload` calls? 🤔

spec/models/active_lead_provider_spec.rb
```diff
@@ -201,4 +201,14 @@
       end
     end
   end
+
+  describe "#payment_declarations_count" do
+    let(:contract_period) { FactoryBot.create(:contract_period, :next) }
+    let!(:active_lead_provider) { FactoryBot.create(:active_lead_provider, contract_period:) }
+    let!(:declarations) { FactoryBot.create_list(:declaration, 2, :payable, active_lead_provider:) }
```
ethax-ross: Worth adding a few clawedback declarations/other contracts/statements in the `payment_declarations_count` specs to ensure they aren't/are counted?

app/models/contract.rb
tonyheadford: Haha, yes probably. I did start off with a simple count query but I couldn't find anywhere it really belonged, it was too simple for a service but because it joined several things there didn't seem anywhere "correct" for it to live.

spec/models/active_lead_provider/band_spec.rb
```diff
@@ -16,15 +16,21 @@
...
-        band.capacity = 100
-        expect(band).not_to be_valid
-        expect(band.errors.full_messages).to include("Capacity can only be increased")
+      before do
+        allow(active_lead_provider).to receive(:payment_declarations_count).and_return(1400)
+      end
+
+      it "validates that the combined capacity cannot be less than the number of payment declarations" do
+        band_c.capacity = 100
+        expect(band_c).not_to be_valid
```
ethax-ross: `to be_invalid` 😅

app/models/active_lead_provider.rb
```diff
@@ -39,6 +39,10 @@ class ActiveLeadProvider < ApplicationRecord
 
   def editable? = !contract_period.payments_frozen?
 
+  def payment_declarations_count
```
ethax-ross: I wonder if this is going to exacerbate the confusion around the meaning of `payment_declarations` as it means 'declarations that have been paid including those that were clawed back' when looking from a `Contract`/`Statement` perspective but here we're redefining that term to be declarations that were paid and _not_ clawed back.

I'd be inclined to just have this as another has_many, through as you have in `Contract` and then do `active_lead_provider.payment_declarations.count` in the `Band`, not worrying about subtracting clawbacks 🤔 I think that's what we decided on during refinement, but I helpfully haven't made that very clear in the ticket 🙃 

If we feel we do want to subtract clawbacks then I think we need to rename this to something like `unclawed_back_payment_declarations_count`. Doesn't really roll off the tongue though 😬

tonyheadford: Yes, I did think this naming felt wrong after this morning's thread

ethax-ross (approved):
Thanks Tony, looks good to go to PO review 👍 I think we should be able to drop those last two `reload` calls in the specs with how it's implemented now?

PR #3396, opened 2026-08-18
first_pass_patching_mentor_completion_info

boonali:
### Context

Currently providers are unable to self-serve if they need to declare against certain mentors. If a mentor was first registered into the 2021 or 2022 cohorts but not declared against (or has only one paid declaration), providers are allowed to train them in ECTPM (i.e. 2025 cohort or beyond) as long as they pass certain other local checks (particularly that they did not have access to learning materials for 2 or more years).

Currently, DfE has to set mentor_became_ineligible_for_funding_on and mentor_became_ineligible_for_funding_reason back to NULL before the provider can resume the participant and transfer them to the current cohort.

Contract managers have agreed that all mentors in this shape can have their fields updated so providers can self-serve.

Provider guidance may also need to be updated at the point the changes are made in prod - see linked ticket.

Patching 3,630 teacher's fields of mentor_became_ineligible_for_funding_on and mentor_became_ineligible_for_funding_reason back to NULL.

### Guidance to review

Ensure all 3,630 teachers have null entries for these fields after the patch has been ran.


tonyheadford (approved):
Tested with real data with no errors

PR #3399, opened 2026-08-19
patches_nulling_ASPs_and_TPs

boonali:
### Context
We've identified cases where schools have not told ECF1 a participant left, but during migration we set a finished_on for both at_school_period and training period.

This can cause an induction tutor to be able to re-register the participant or re-select them as a mentor.

We are patching the finished_on date of the ECT's latest ASP who have no ECF1 leaving flag and have an induction period in RECT where started_on < 28/4/2026 and finished_on is either NULL or > 28/4/2026.

If they have a null withdrawn_at & deferred_at, then their finished_on for their Training Period is also nullified.

### Guidance to review
Check that the ECT ASPs & Training Periods mentioned in the patch all have null finished_on dates.


tonyheadford (approved):
Tested on real data with no errors

PR #3402, opened 2026-08-19
[3591] Add Search and Filters to Admin School Teachers Page

avinhurry:
### Summary

Following on from #2689 where we redesigned the admin index page, this PR adds search, role, and contract period filters to the admin school teachers page.

| Before | After |
|--------|-------|
|image|image|

Updates the admin school teachers page to:

- search by name, TRN or API IDs
- filter by role and contract period
- show current ECT and mentor roles as separate rows
- display `Not available` when a role has no contract period
- order teachers by newest first

The main admin teachers search and table were extracted into shared components. Its wording now uses `Not available`, its filtering behaviour is unchanged here but will be in a follow up PR.


app/components/admin/teachers/search_component.html.erb
```diff
@@ -0,0 +1,29 @@
+<%= form_with url: form_url, method: :get, data: { turbo: false }, html: { class: "app-teachers-filters" } do |f| %>
+  <%= f.govuk_text_field(
+        :q,
+        value: filter_params[:q],
+        label: { text: "Search for teacher", size: "s" },
+        hint: { text: "Name, TRN, API participant or training record ID" },
```
ericaporter, line 6: Notice that the hint changed - just checking if dropping ID from 'API participant' is intentional?

avinhurry: Yep, @StephenC-V1 and I caught up this morning on it. It sits nicer on one line and users are internal.

ericaporter: It might be helpful to update the ticket if you haven't already?

app/services/admin/schools/teacher_search.rb
```diff
@@ -0,0 +1,86 @@
...
+        @query_string = query_string.to_s.strip
+        @role = role.to_s.presence
+        @contract_period = contract_period.to_s.presence
+      end
+
+      def rows
+        role_periods
+          .select { |_current_role, period| contract_period_matches?(period:) }
+          .map { |current_role, period| row_for(role: current_role, period:) }
+          .sort_by { |row| [-row.teacher.created_at.to_f, row.role] }
```
ericaporter, line 20: In prototype it looks like teachers are ordered by first name, I think this is in AC too - are they being ordered by created_at here?

avinhurry: Yes they are being ordered by `created_at` to match the main teachers page and we think that ordering by newest first will be more useful for schools where there are lots of teachers.

Either way, the search will mostly be used to find the teachers.

app/components/admin/teachers/search_component.rb
```diff
@@ -0,0 +1,37 @@
...
+        filter_options = ContractPeriod.order(:year).map do |contract_period|
+          year = contract_period.year.to_s
+
+          FilterOption.new(value: year, name: year)
+        end
+
+        filter_options + [
+          FilterOption.new(
+            value: Rows::CONTRACT_PERIOD_NOT_AVAILABLE,
+            name: "Not available"
```
ericaporter, line 31: The prototype is still showing 'Not applicable' which is a bit confusing - I know it should be this though

app/components/admin/teachers/table_component.rb
```diff
@@ -0,0 +1,42 @@
...
+  module Teachers
+    class TableComponent < ApplicationComponent
+      attr_reader :rows
+
+      def initialize(rows:, teacher_path:)
+        @rows = rows
+        @teacher_path = teacher_path
+      end
+
+      def teacher_path_for(teacher_row)
```
ericaporter: You could make this a private method but also fine if you don't

app/services/admin/teachers/rows.rb
```diff
@@ -25,7 +25,7 @@ def role_name
 
         def contract_period_name
           return if contract_period.blank?
-          return "Not applicable" if contract_period == Rows::CONTRACT_PERIOD_NOT_APPLICABLE
+          return "Not available" if contract_period == Rows::CONTRACT_PERIOD_NOT_AVAILABLE
```
ericaporter, line 29: The AC (2d) refers to this as "Not applicable" (as does prototype) but I'm pretty sure 'Not available' is actually right - can you just confirm! 😆

avinhurry: 😆  your confusion is valid!

I decided to go with "Not available" for any current ECT/Mentor role that doesn't have a contract period, rather than limiting it to school-led ECTs as there are valid records where a contract period isn't available so "Not available" reflects the data more accurately (and also the filters).

app/components/admin/teachers/table_component.rb
avinhurry: yeah good shout. Updated here https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3402/changes/5916f8a822c76994daa5bca04b4f504fcc2355b6

spec/requests/admin/schools_spec.rb
```diff
@@ -105,11 +105,30 @@
...
+          )
+        end
+
+        it "keeps the teachers navigation item current when filtered" do
+          get admin_school_teachers_path(school.urn, role: "mentor")
+
+          expect(Capybara.string(response.body)).to have_css(
+            "a[aria-current='page']",
+            text: "Teachers"
+          )
```
ericaporter, line 123: You could also verify that the role filter has been applied as a nice to have

avinhurry: I intentionally kept this spec focused on the navigation behaviour. The role filtering behaviour is already covered in the service spec.

spec/services/admin/schools/teacher_search_spec.rb
```diff
@@ -0,0 +1,236 @@
...
+        expect(rows.map(&:teacher)).not_to include(role_only_teacher, contract_period_only_teacher)
+      end
+    end
+
+    context "when filtering by the not available contract period option" do
+      let(:contract_period) { Admin::Teachers::Rows::CONTRACT_PERIOD_NOT_AVAILABLE }
+      let!(:school_led_teacher) { FactoryBot.create(:teacher, trs_first_name: "Goku") }
+      let!(:provider_led_teacher) { FactoryBot.create(:teacher, trs_first_name: "Vegeta") }
+      let!(:untrained_mentor) { FactoryBot.create(:teacher, trs_first_name: "Piccolo") }
+      let!(:untrained_ect) { FactoryBot.create(:teacher, trs_first_name: "Trunks") }
```
ericaporter: I'm just wondering about including untrained_ect here - I get what you are testing but I don't think they would realistically show up in this list (unless they would through migrated data?). Maybe a school_led_teacher is enough

app/services/admin/schools/teacher_search.rb
ericaporter: Ok yeah that makes sense. Created_at being the most recently created in RECT rather than most recently started at the school - in most cases it's probably the same? Maybe just add a note to the ticket.

ericaporter (approved):
Overall looks great! Just a few small queries where prototype, AC and implementation look slightly diff.

spec/services/admin/schools/teacher_search_spec.rb
avinhurry: Good point. I checked prod and there are currently no ECTs in that state,[ so i've removed `untrained_ect`](https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3402/changes/c344cd30b0924a72d45006b2744496c8d04164e2). I've kept `untrained_mentor` though because there are 5862 current mentor periods without training periods.

PR #3403, opened 2026-08-19
Framework agreement events

craigmdavidson:
### Context

Part of https://github.com/DFE-Digital/register-ects-project-board/issues/4334

Following on from rename in the code, this renames the created and deleted events from ActiveLeadProvider to FrameworkAgreement

### Guidance to review

Simple PR.


app/models/event.rb
ethax-ross, line 3: Can these be dropped as well?

ethax-ross (approved):

cpjmcquillan (approved):

PR #2968, opened 2026-05-26
Prefix page titles with Error:

ericaporter:
### Context
When validation errors are shown during change journeys for ECTs, mentors and SITs, the page title was not consistently prefixed with Error: as required by the GOV.UK Design System. Some wizards had it correctly, others were missing it.

### Changes proposed in this pull request

Added error: @wizard.current_step.errors.present? to the page_data call in four change wizard edit views that were missing it:

schools/ects/change_email_address_wizard/edit
schools/ects/change_lead_provider_wizard/edit
schools/ects/change_mentor_wizard/edit
schools/mentors/change_email_address_wizard/edit

and

schools/mentors/change_name_wizard/edit
schools/ects/change_name_wizard/edit
schools/mentors/change_lead_provider_wizard/edit
schools/induction_tutor/update_induction_tutor_wizard/edit
schools/ects/change_mentor_wizard/lead_provider
schools/ects/change_training_programme_wizard/lead_provider

Also added validation-failure scenarios to the corresponding feature specs asserting that the page title starts with Error: when the form is re-rendered with errors.

### Guidance to review
Can check the four edited views on the review app by submitting each form without valid input and inspecting the browser tab title.

Issue #1763, opened 2025-11-20
Add error title prefix to adjustment forms


peteryates:
The `page_data` in `app/views/admin/finance/adjustments/new.html.erb` and `app/views/admin/finance/adjustments/edit.html.erb` doesn't include an `error` argument even though there's a form that can display errors on the page.

It's needed to prefix the page title with `Error: ` when the forms in an error state.

PR #3405, opened 2026-08-19
Prefix invalid form page titles with Error

ericaporter:
### Context
Issue #1763 identified that the new and edit adjustment forms did not pass their validation state to page_data. As a result, the browser page title was not prefixed with Error: when validation failed.

Issue #2968 also relates to this where the same fix was applied to the wizard change journey files.

This PR includes the fixes for new and edit adjustment forms as well as fixing other similar examples in the admin area.

- Pass each form object’s error state to page_data for the affected forms
- Prefix affected page titles with Error: when validation fails
- Add the missing error summary to the school support-query form

### Guidance to review
A check to see if any are still missing would be useful

### DIFF
--- a/app/controllers/admin/finance/adjustments_controller.rb
+++ b/app/controllers/admin/finance/adjustments_controller.rb
@@ -30,7 +30,7 @@ module Admin::Finance
           Events::Record.record_statement_adjustment_updated_event!(author: current_user, statement_adjustment: @adjustment)
           redirect_to statement_path, alert: "Adjustment changed"
         else
-          render :new, status: :unprocessable_content
+          render :edit, status: :unprocessable_content
         end
       end
     end
--- a/app/controllers/admin/finance/adjustments_controller.rb
+++ b/app/controllers/admin/finance/adjustments_controller.rb
@@ -30,7 +30,7 @@ module Admin::Finance
           Events::Record.record_statement_adjustment_updated_event!(author: current_user, statement_adjustment: @adjustment)
           redirect_to statement_path, alert: "Adjustment changed"
         else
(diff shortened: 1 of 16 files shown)


cpjmcquillan (approved):
I wonder if it's worth having a specific helper for page data for pages with forms? 🤔

I haven't tried it but I wonder if this could work.

```ruby
def form_page_data(model:, **args)
  page_data(error: model.errors.any?, **args)
end
```

It might improve the ergonomics a little and make it easier for people to remember to use it.

We could even lint for any templates that contain `<%= form_with` without using `<% form_page_data`.

What do you think?

avinhurry (approved):

peteryates (approved):

ericaporter:
> I wonder if it's worth having a specific helper for page data for pages with forms? 🤔
> 
> I haven't tried it but I wonder if this could work.
> 
> ```ruby
> def form_page_data(model:, **args)
>   page_data(error: model.errors.any?, **args)
> end
> ```
> 
> It might improve the ergonomics a little and make it easier for people to remember to use it.
> 
> We could even lint for any templates that contain `<%= form_with` without using `<% form_page_data`.
> 
> What do you think?

@cpjmcquillan Yeah, I think a form_page_data helper could make this a bit easier to use consistently.

I wonder if checking for govuk_error_summary rather than every form_with might be safer, since some forms (search/filter forms etc.) won't have a validation error state.

Maybe we could keep this PR focused and create a follow-up ticket for introducing the helper?

JSJohal (approved):
I've had a quick check with the review app to see if the 'Error' prefix is passing through in the titles - it looks to be on all the pages I've tested in registering and editing details.

PR #3406, opened 2026-08-20
Product team users can preview data fixes

cpjmcquillan:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/4254

#### Product team users can preview data fixes
Before, in https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3377 users could input a CSV of data fixes which were parsed and
presented to the user.

But we didn't do anything with those changes.

This builds on https://github.com/DFE-Digital/register-early-career-teachers-public/pull/3393 and processes the proposed changes in a preview step. Any
errors are presented to the user. If there are no errors, we display the
"changes".

This is executed in a transaction and rolled back, so no changes are persisted
yet.

Both the parsed rows and the processed changes are displayed to users in the
most rudimentary way possible. We intend to iterate on this soon!

In order to display *all* the errors when processing the CSV, we need to pass a
custom `ErrorPresenter` to the form builder. This ensures each error message is
displayed, rather than only the first.

#### Pre-populate `csv_string` from parsed rows in `CSVStep`
When a user navigates "back" from the `/preview` step to the `/csv` step after
previously entering a CSV, we want to pre-populate the form input with what they
submitted.

This generates a CSV from the parsed rows in the session store if the
`csv_string` is missing.

As part of this change, the step unit tests have been reworked to stub the
behaviour of the `Admin::DataFixes::InlineCSV` class. This is intended to keep
the tests focussed on the behvaiour of the step, rather than the underlying PORO
(which is already extensively tested).

### DIFF
--- a/app/controllers/admin/data_fixes_controller.rb
+++ b/app/controllers/admin/data_fixes_controller.rb
@@ -14,6 +14,10 @@ module Admin
                   unless: -> { wizard_class.step?(@previous_step) },
                   only: :new
 
+    around_action :wrap_in_transaction,
+                  if: -> { @current_step == :preview },
+                  only: :create
+
     def new
       render @current_step
     end
@@ -49,5 +53,13 @@ module Admin
     end
 
     def wizard_class = DataFixesWizard::Wizard
+
+    def wrap_in_transaction
+      ActiveRecord::Base.transaction do
+        yield
+      ensure
+        raise ActiveRecord::Rollback
+      end
+    end
   end
 end
--- a/app/wizards/admin/data_fixes_wizard/preview_step.rb
+++ b/app/wizards/admin/data_fixes_wizard/preview_step.rb
@@ -1,6 +1,33 @@
 module Admin::DataFixesWizard
   class PreviewStep < Step
     def previous_step = :csv
-    def next_step = nil
+    def next_step = :verify
+
+    def save!
+      processed_changes = changes.process
+      store.processed_changes = processed_changes if processed_changes
+    end
+
+    delegate :errors, to: :changes
+    delegate :parsed_rows, to: :store
+    def error_presenter = ErrorSummaryPresenter
+
+  private
+
+    def changes
+      @changes ||= Admin::DataFixes::Changes.new(parsed_rows: store.parsed_rows)
+    end
+
+    class ErrorSummaryPresenter
+      def initialize(error_messages)
+        @error_messages = error_messages
+      end
+
+      def formatted_error_messages
+        @error_messages.flat_map do |attribute, messages|
+          messages.map { |message| [attribute, message] }
+        end
+      end
+    end
   end
 end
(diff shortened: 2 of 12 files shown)


app/controllers/admin/data_fixes_controller.rb
```diff
@@ -49,5 +53,13 @@ def set_wizard
     end
 
     def wizard_class = DataFixesWizard::Wizard
+
+    def wrap_in_transaction
+      ActiveRecord::Base.transaction do
+        yield
+      ensure
```
ericaporter: Is the ensure here to guarantee the preview is always rolled back? It might mean an unexpected exception from yield gets replaced by ActiveRecord::Rollback, which is then handled by the transaction and hard to debug.

app/wizards/admin/data_fixes_wizard/preview_step.rb
```diff
@@ -1,6 +1,33 @@
 module Admin::DataFixesWizard
   class PreviewStep < Step
     def previous_step = :csv
-    def next_step = nil
+    def next_step = :verify
+
+    def save!
+      processed_changes = changes.process
```
ericaporter, line 7: What is the strategy for stale processed_changes? Is that to be dealt with in a follow up PR?

ericaporter (approved):
Looks good! Left a few questions, but nothing blocking

app/wizards/admin/data_fixes_wizard/preview_step.rb
cpjmcquillan: What do you mean exactly by stale `processed_changes`?

app/controllers/admin/data_fixes_controller.rb
cpjmcquillan: Yeah, that's the intention. 

We always want the changes to get rolled back in the preview, so we're using `ensure`.

You're right though that means unexpected exceptions get swallowed and the response is a `200 OK`.

I'm not sure how to avoid that though 🤔. Any thoughts?

ericaporter: I think it might be ok to just remove the ensure and have
```
  yield
  raise ActiveRecord::Rollback
end
```
so if yield is a success path - the execution reaches `raise ActiveRecord::Rollback` and Rails catches the exception. If yield raises another exception that never reaches `raise ActiveRecord::Rollback` then the transaction block is escaped and the exception continues upwards. I think that is the desired behaviour?

app/wizards/admin/data_fixes_wizard/preview_step.rb
ericaporter: Ah, I meant if a user previews a set of changes, so store.processed_changes gets populated, then goes back and submits a different CSV (which fails during preview) - perhaps unlikely, but would the previous processed_changes still remain in the store?

cpjmcquillan: Ok, I get what you mean!

Yeah, I guess that's something we're not really handling. I'll update the steps to set the stored value to `nil` if the changes cannot be parsed/processed 👌🏼

ericaporter: Update looks good 👍

PR #3408, opened 2026-08-20
Framework Agreement Table rename

craigmdavidson:
### Context

Part of https://github.com/DFE-Digital/register-ects-project-board/issues/4334

Following on from rename in the code, this renames the tables and columns from ActiveLeadProvider to FrameworkAgreement

### Guidance to review

* This change includes both database table renames and the code impact of these. This means that deployment will require a short quiet window.

### DIFF
--- a/app/components/admin/select_lead_providers_form_component.rb
+++ b/app/components/admin/select_lead_providers_form_component.rb
@@ -33,7 +33,7 @@ module Admin
     end
 
     def currently_selected_ids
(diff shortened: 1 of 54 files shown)


ethax-ross (approved):
Looks good, can we reflect the change on the [data schema](https://github.com/DFE-Digital/register-ects-project-board/wiki/Data-schema) diagram too?

There are still a few short name references to active lead provider (`alp`, `active_lp`) kicking around, are we going to mop those up after this?

PR #3409, opened 2026-08-20
[4318] Exclude upcoming mentors from the list of "eligible" mentors

rwrrll:
### Context

https://github.com/DFE-Digital/register-ects-project-board/issues/4318

`Schools::AssignMentor` now closes every current or upcoming mentorship period, rather than just the first one, so that changing an ECT's mentor overwrites a future mentorship if it exists, instead of failing with an overlap validation error. I put it in `AssignMentor` rather than the change mentor wizard because all the assignment paths (change mentor, register new mentor, assign existing mentor) go through it and hit the same bug.

A mentorship that's already started gets truncated to the day before the new one begins (not extended). One that hasn't started yet is deleted outright, which gives the desired "only ever one future mentor" behaviour.

n.b. The removed future mentor keeps their `TrainingPeriod`, which hangs off their own at-school period rather than the mentorship and can already carry declarations. This matches what already happens to an outgoing current mentor, so I think this is fine, just mentioning in case it's not what we want. If it's wrong, then we could change just in this case, or maybe tidying those up belongs in its own ticket covering both cases.

### Guidance to review

I'm a bit on the fence about the feature test, it is a bug that occurs as part of a wizard journey, but OTOH it's an edge case. Figured I'd add it so that we can see it, happy to revisit if there's a strong opinion against it.

### DIFF
--- a/app/services/schools/assign_mentor.rb
+++ b/app/services/schools/assign_mentor.rb
@@ -11,7 +11,7 @@ module Schools
 
     def assign!
       ActiveRecord::Base.transaction do
-        close_current_mentorship!
+        close_current_and_upcoming_mentorships!
         @mentorship_period = assign_new_mentor!
         record_events!
       end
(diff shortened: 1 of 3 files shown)


cpjmcquillan (approved):
FWIW, I don't think the feature test is essential.

It might be worth adding some more unit tests though that cover the nuances between the different dates when closing/finishing mentorship periods.

In particular what happens (i.e closing behaviour) for each combination of 

- "started" mentorship?
- earliest possible start **after** mentorship started on?
- mentorship finished on **before** earliest possible start?

It can be a bit annoying, but I think it's helpful to highlight the complexity of the mentor assignment behaviour.

What do you think?

rwrrll:
@cpjmcquillan cool beans, in that case 🪓  feature test gone!

Yeah I agree worth having the coverage. All these scenarios were actually already covered by existing unit tests, but the coverage was incidental and could easily get changed or removed in future inadvertently. So I did a bit of shuffling to make this coverage more explicit, see what you think.

